PR #213, opened 2015-01-05
Docs: update pkg-config to use vips8 C++ API

lovell:
I've started experimenting with the new and significantly more pleasant C++ API :)

### DIFF
--- a/doc/reference/using-cpp.xml
+++ b/doc/reference/using-cpp.xml
@@ -26,7 +26,7 @@
 
 <programlisting language="C++">
 /* compile with:
- *      g++ -g -Wall try.cc `pkg-config vips-cc --cflags --libs`
+ *      g++ -g -Wall try.cc `pkg-config vips-cpp --cflags --libs`
  */
 
 #include &lt;vips/vips8&gt;


jcupitt:
Off-topic: git master C++ (will be 7.42.2) has a new feature, you can set enums with strings. You can use:

```C++
a.embed(10, 10, 100, 100, VImage::option()->set("extend", "copy"));
```

as well as the old:

```C++
a.embed(10, 10, 100, 100, VImage::option()->set("extend", VIPS_EXTEND_COPY));
```

If you have any good ideas, please say, this is the time to break API, it'll all be frozen once 8.0 is out. The whole ::set() thing is slightly ugly, but I can't think if anything better that's not horribly complicated :-(

lovell:
Thanks John. Given this is a brand new binding, would you consider moving to C++11? It could mean client code that currently looks like:
```cpp
VImage rows;
VImage cols = image.profile(rows);
```
could be made more consistent with the use of tuples:
```cpp
VImage cols, rows;
tie(cols, rows) = image.profile();
```
It's a bit more Python-esque too, which I know you <3.

The only (minor) problem I've run into so far was with the `scale()` method of VImage, which seems to overload both `vips_image_get_scale` and `vips_scale`. I had to use `image.scale(VImage::option())` to coerce the compiler into accessing the latter.

jcupitt:
Yes, I kept wanting to use C++11 features :-( I was trying to avoid it, though I can't remember exactly why.

Good point on the `scale()` thing, I'll have a look.

lovell:
Exploring the `profile()` method further, using sample code like:

```cpp
#define VIPS_DEBUG_VERBOSE
#include <vips/vips8>
using namespace vips;
...
VImage image = VImage::new_from_file("/path/to/test.jpg");
VImage rows;
VImage cols = image.profile(rows);
```
aborts with the output:
```
VObject constructor, obj = (nil), steal = 1
VObject assignment
   reffing 0x7f06f4541010
   unreffing (nil)
VObject destructor
   unreffing 0x7f06f4541010
VObject constructor, obj = (nil), steal = 1
**
ERROR:/usr/local/include/vips/VImage8.h:101:vips::VObject::VObject(const vips::VObject&): assertion failed: (VIPS_IS_OBJECT( a.vobject ))
Aborted (core dumped)
```
Am I doing something wrong? I notice the VObject constructor [allows a NULL init](https://github.com/jcupitt/libvips/blob/master/cplusplus/include/vips/VImage8.h#L70); should the copy constructor do so also?

jcupitt:
You're right, output `VImage` parameters are messed up. `VImage::profile` is declared as:

```C++
VImage VImage::profile( VImage rows , VOption *options )
    throw( VError )
```

There should be a `&` there, I think.

jcupitt:
All required output parameters were broken (though that's only `profile` and `project`), argh! This now works:

```C++
        VImage rows;
        VImage cols = in.profile( &rows );
```

I went for `&` on the caller for consistency with `->set()` output params. You have to have the `&` on the caller with `->set()` so it can tell the difference between input and output params. I rather like it anyway, it makes it clear to people using the function which params are modified.

jcupitt:
I removed `VImage::scale()` as well, ouch.

lovell:
Both these changes work perfectly, thank you John.

PR #224, opened 2015-01-22
Ensure resize operation has the required number of temp images

lovell:
Hi John,

The tilecache created during the resize operation uses the [7th](https://github.com/jcupitt/libvips/blob/master/libvips/resample/resize.c#L160) temporary image but only 6 are allocated.

This fix prevents a memory leak and an occasional segfault.

(PR is for the master branch as I can't find a 7.42 branch for bug fixes.)


jcupitt:
Argh! Well spotted.

PR #228, opened 2015-01-27
Improve performance of openslide2vips argb2rgba()

bgilbert:
Almost all pixels emitted by OpenSlide are either fully transparent or fully opaque.  Add a special case for opaque pixels to avoid the multiplication/division, and restructure both special cases to compile to a single shift + OR + bswap.  On my system, this produces a 1.95x speedup in user time for `vips openslideload CMU-1.svs out.v` and a 1.28x speedup for `vips openslideload CMU-1.mrxs --level 2 out.v`.


jcupitt:
Hey, nice speedup.

PR #241, opened 2015-02-15
Add ImageMagick buffer support

mcuelenaere:
This commits adds buffer support for the ImageMagick backend, using the Blob
API.


mcuelenaere:
This fixes https://github.com/jcupitt/libvips/issues/238

jcupitt:
Looks great, thank you @mcuelenaere 

It'll be in libvips-8.0.0, due on 1st April.

PR #262, opened 2015-04-13
Expose JPEG extension parameters. Support trellis quantisation, deringing and scan optimisation.

lovell:
These features are available now in mozjpeg and support for the extension parameter API will be added to the next release of libjpeg-turbo.

This PR adds `./configure` time detection by looking for the new `J_BOOLEAN_PARAM` enum type of `jpeglib.h`, which will also be present in libjpeg-turbo >= 1.5.

It also uses a runtime guard via the new `jpeg_c_x_param_supported` methods. I believe libjpeg-turbo is hoping to support overshoot deringing first, then add other features later as they are optimised to use intrinsics.

If JPEG extension parameters are not available, which will be for the majority of users right now until libjpeg-turbo updates roll out, the new features that use them silently have no effect. Would you prefer these options to fail if unavailable at runtime?

Here are some examples of the file size reductions one can expect using these new features:

When running the following (existing `optimize_coding` option included here for comparison):
```sh
curl -O https://raw.githubusercontent.com/lovell/sharp/master/test/fixtures/2569067123_aca715a2ee_o.jpg
vipsthumbnail --size=256 --linear --format=tn_%s.jpg 2569067123_aca715a2ee_o.jpg
vipsthumbnail --size=256 --linear --format=tn_oc_%s.jpg[optimize_coding] 2569067123_aca715a2ee_o.jpg
vipsthumbnail --size=256 --linear --format=tn_tq_%s.jpg[trellis_quant] 2569067123_aca715a2ee_o.jpg
vipsthumbnail --size=256 --linear --format=tn_os_%s.jpg[interlace,optimize_scans] 2569067123_aca715a2ee_o.jpg
vipsthumbnail --size=256 --linear --format=tn_ostq_%s.jpg[interlace,optimize_scans,trellis_quant] 2569067123_aca715a2ee_o.jpg
```
With mozjpeg:
```
10103 tn_2569067123_aca715a2ee_o.jpg
 9731 tn_oc_2569067123_aca715a2ee_o.jpg
 8866 tn_tq_2569067123_aca715a2ee_o.jpg
 9608 tn_os_2569067123_aca715a2ee_o.jpg
 8847 tn_ostq_2569067123_aca715a2ee_o.jpg
```
Without mozjpeg:
```
10103 tn_2569067123_aca715a2ee_o.jpg
10103 tn_tq_2569067123_aca715a2ee_o.jpg
 9731 tn_oc_2569067123_aca715a2ee_o.jpg
 9813 tn_os_2569067123_aca715a2ee_o.jpg
 9813 tn_ostq_2569067123_aca715a2ee_o.jpg
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -696,6 +696,22 @@ FIND_JPEG(
    with_jpeg=no
   ])
 
+# JPEG extension parameters available in libjpeg-turbo >=1.5.0, mozjpeg >=3.0
+if test x"$with_jpeg" = "xyes"; then
+  AC_MSG_CHECKING([for JPEG extension parameters])
+  AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
+    #include <stdio.h>
+    #include <jpeglib.h>
+  ]], [[
+    J_BOOLEAN_PARAM test;
+  ]])], [
+    AC_MSG_RESULT([yes])
+    AC_DEFINE([HAVE_JPEG_EXT_PARAMS],1,[libjpeg has extension parameters])
+  ], [
+    AC_MSG_RESULT([no])
+  ])
+fi
+
 # libexif
 AC_ARG_WITH([libexif], 
   AS_HELP_STRING([--without-libexif], [build without libexif (default: test)]))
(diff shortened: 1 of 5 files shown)


jcupitt:
Looks great Lovell, thank you very much for this patch.

Perhaps it could print a warning if the libjpeg does not support the feature? tiffsave warns if you use an unsupported feature, for example. `vips_warn()` is the official thing to use, I expect you know about it. 

http://www.vips.ecs.soton.ac.uk/supported/current/doc/html/libvips/libvips-error.html#vips-warn

lovell:
Thanks John. I've updated the `jpeg-extension-parameters` branch to use `vips_warn()` where any attempt is made to use the new options but the version of libjpeg does not support them.

I've rebased against upstream/master to include 3f9aded but it looks like Travis might be having problems with it at the moment.

jcupitt:
Lovely! NIce patch Lovell.

PR #264, opened 2015-04-15
Ensure expected signature in ExceptionInfo before destroy

lovell:
As `read_free` can be called twice since fc5a4a9, this change is required for GraphicsMagick compatibility. It avoids failing the [assert](http://sourceforge.net/p/graphicsmagick/code/ci/default/tree/magick/error.c#l359) in `DestroyExceptionInfo` on the second occasion.


jcupitt:
That's great, thanks Lovell. You'd think magick would spot this for you :(

PR #265, opened 2015-04-16
Modified vips_foreign_load_start()

aferrero2707:
Added metadata copy from load->out to load->real in vips_foreign_load_start(), so that custom metadata fields added after the image has been opened do not get destroyed when reading of the pixels is started.

The modification does not introduce, as far as I can tell, any bad side effect. However I might be missing "the big picture" ;-)

### DIFF
--- a/libvips/foreign/foreign.c
+++ b/libvips/foreign/foreign.c
@@ -816,6 +816,14 @@ vips_foreign_load_start( VipsImage *out, void *a, void *b )
 		if( !vips_foreign_load_iscompat( load->real, out ) )
 			return( NULL );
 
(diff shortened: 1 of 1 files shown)


felixbuenemann:
@jcupitt The issue that motivated this PR is closed, so probably the PR can be closed as well?

aferrero2707:
I've been able to "circumvent" this issue by first running the loaded image through vips_copy() and then adding the metadata to the vips_copy() output...

If no one else is affected by this, I'm perfectly fine with the above workaround, and this pull request can be safely ignored.

jcupitt:
OK, I'll close. Thanks @aferrero2707

PR #297, opened 2015-05-18
Fix .gir file dependencies

bgilbert:
Ensure `introspect` is built before `g-ir-scanner` tries to run it.  This should fix MacPorts' Buildbot builds on OS X 10.8 (which run with `make -j8`).


jcupitt:
Oh, well done, that's a good thing to fix. Thank you Benjamin.

PR #335, opened 2015-10-04
acinlude.m4: python3 compatible print

PolarNick239:
Python3 and python2 compatible print call in configure script.


jcupitt:
Looks good, thank you @PolarNick239

PR #359, opened 2015-12-21
Use 254 pixels as default tile size

gasi:
I remember @benvanik, a friend and colleague, who worked on our DZI renderer (Seadragon) told us that, I believe, it’s better for GPUs to have power of two sized tiles, so 254px + 2 x 1px tile overlap will give you 256 pixels.

> To keep my tiles nicely sized, I shrink in instead of expand out – that means my tiles are really 254×254 with a 1px border, so the images are 256x256px with 254×254 of useful imagery. If you wanted larger tiles, you’d go 510×510 with 1px border making 512×512 tiles.
> — Source: http://www.noxa.org/blog/2009/11/29/megatextures-in-webgl-2/

P.S. @benvanik: Please correct me if I am wrong or fill us in on more of the details :smile:

### DIFF
--- a/libvips/foreign/dzsave.c
+++ b/libvips/foreign/dzsave.c
@@ -116,9 +116,9 @@
 
 	Overlap handling
 
-   For deepzoom, tile-size == 256 and overlap == 1 means that edge tiles are 
-   257 x 257 (though less at the bottom right) and non-edge tiles are 258 x 
-   258. Tiles are positioned across the image in tile-size steps. This means 
+   For deepzoom, tile-size == 254 and overlap == 1 means that edge tiles are 
+   255 x 255 (though less at the bottom right) and non-edge tiles are 256 x 
+   256. Tiles are positioned across the image in tile-size steps. This means 
    (confusingly) that two adjoining tiles will have two pixels in common.
 
    This has caused bugs in the past. 
@@ -1938,7 +1938,7 @@ vips_foreign_save_dz_class_init( VipsForeignSaveDzClass *class )
 		_( "Tile size in pixels" ),
 		VIPS_ARGUMENT_OPTIONAL_INPUT,
 		G_STRUCT_OFFSET( VipsForeignSaveDz, tile_size ),
-		1, 8192, 256 );
+		1, 8192, 254 );
 
 	VIPS_ARG_ENUM( class, "depth", 13, 
 		_( "Depth" ), 
@@ -1999,14 +1999,14 @@ vips_foreign_save_dz_class_init( VipsForeignSaveDzClass *class )
 		_( "Tile width in pixels" ),
 		VIPS_ARGUMENT_OPTIONAL_INPUT | VIPS_ARGUMENT_DEPRECATED,
 		G_STRUCT_OFFSET( VipsForeignSaveDz, tile_size ),
-		1, 8192, 256 );
+		1, 8192, 254 );
 
 	VIPS_ARG_INT( class, "tile_height", 12, 
 		_( "Tile height" ), 
 		_( "Tile height in pixels" ),
 		VIPS_ARGUMENT_OPTIONAL_INPUT | VIPS_ARGUMENT_DEPRECATED,
 		G_STRUCT_OFFSET( VipsForeignSaveDz, tile_size ),
-		1, 8192, 256 );
+		1, 8192, 254 );
 
 }
 
@@ -2016,7 +2016,7 @@ vips_foreign_save_dz_init( VipsForeignSaveDz *dz )
 	VIPS_SETSTR( dz->suffix, ".jpeg" );
 	dz->layout = VIPS_FOREIGN_DZ_LAYOUT_DZ; 
 	dz->overlap = 1;
-	dz->tile_size = 256;
+	dz->tile_size = 254;
 	dz->tile_count = 0;
 	dz->depth = VIPS_FOREIGN_DZ_DEPTH_ONEPIXEL; 
 	dz->angle = VIPS_ANGLE_D0; 


jcupitt:
Hi @gasi, I agree the defaults are a bit crazy. 258 pixels across is bad for JPEG too: you get a 1/4 8x8 blocks down the right and across the bottom. If I were king I think tiles would always be 256 x 256 and overlaps would be generated by reducing the image step instead.

However ... the defaults are set to match the deepzoom default, which I think is for 258 x 258 tiles and a fixed 256 x 256 step. For example, with the [openslide deepzoom writer](https://github.com/openslide/openslide-python/tree/master/examples/deepzoom) I see:

```
$ python deepzoom_tile.py /data/john/pics/wtc.tif
Tiling slide: wrote 1877/1877 tiles
$ vipsheader wtc_files/13/1_1.jpeg 
wtc_files/13/1_1.jpeg: 258x258 uchar, 3 bands, srgb, jpegload
```

benvanik:
@gasi still right, on older hardware (think crappy Android tablets and Intel integrated chips). Most drivers will pad the texture out for you during upload instead of preventing you from using non-power of two textures at all, so that's one additional memcpy you'd be incurring. Yuck.

The later versions of DeepZoom should work just fine with 256x256 total tiles (so 254 + 1border), as it switched to using newer code that optimized for them (and for ones <256 would decode them right into buffers of 256 to avoid memcpys). If your image sets are small and rarely have interior tiles I guess it doesn't matter too much, but in most deep pyramids interior tiles dominate and having them be optimal is important.

gasi:
> If I were king I think tiles would always be 256 x 256 and overlaps would be generated by reducing the image step instead.

What do you mean by ‘reducing the image step’?

> However ... the defaults are set to match the deepzoom default

Hmm, as for DeepZoom default, since it was originally developed at Microsoft, I think _Deep Zoom Composer_ would be the closest reference and it seems to generate 256x256 pixels tiles (total?): https://msdn.microsoft.com/en-us/library/dd409068.aspx Do you think changing the defaults would cause interoperability issuess? Assuming other libraries read the tile size from the XML manifest, I would think that wouldn’t be a problem.

Either way, @benvanik was one of the developers most familiar with the whole Deep Zoom system, so I trust his judgement.

Thanks for chiming in, @benvanik :smile:

jcupitt:
DeepZoom was developed by some company whose name I forget, then they were bought out by MS. I think that page means tiles are 256 x 256 unique pixels, but they have small borders in addition to that, so 258 x 258 for a non-edge tile. Or that's how I interpret it. I can't actually test composer myself.

You're right, changing the defaults could break some code. If someone is invoking `dzsave` and not specifying the values for tile size and overlap, this patch could change their output and potentially cause trouble.

benvanik:
lol I'm the author of the DeepZoom implementation that shipped in Silverlight :)
Silverlight 2.0 was the only version that suffered from weirdness around tile sizes. All versions after use code that is optimized to have final image sizes be powers of two. For that, you definitely want 254 with a 1 overlap, and not 256 with a 1 overlap (as the final image becomes 258, and is padded to 512, using 4x as much memory).

gasi:
@jcupitt Yes, DeepZoom was developed by [Seadragon](https://en.wikipedia.org/wiki/Seadragon_Software) and @benvanik, @aseemk, and I, among others, worked on it after Microsoft acquired it as part of [Microsoft Live Labs](https://en.wikipedia.org/wiki/Microsoft_Live_Labs). See @benvanik demo Seadragon on the iPhone: https://channel9.msdn.com/Blogs/LarryLarsen/First-Look-Seadragon-Mobile :smile: 

I understand your concern about backwards compatibility. How about including these new defaults in the upcoming VIPS 8.0 release? Given that it’s a major version, it should allow for breaking changes. Either way, it’s not a huge issue as one can override the defaults, but it’d be good to have people fall into the _pit of success_ when using VIPS. Maybe we can document the 254 recommendation at least.

BTW, a couple of us ex-Seadragon folks, i.e. @iangilman, @aseemk, @nein09, @kenperkins, and I are working on http://zoom.it, an online service for generating and hosting DZIs, and are using VIPS for the image pyramid generation. **Example:** http://zoom.it/bQTEg :smile: 

Happy holidays to you! :christmas_tree:

jcupitt:
Oh, how funny, well done for making seadragon, and zoom.it seems interesting. I have a little history on this as well, I made the original IIP back in 1997. That used pyramidal jpeg tiff with no overlaps ... much simpler. 

This might make you laugh:

http://jcupitt.github.io/iipmooviewer/

It's a deepzoom viewer (a modified iipmooviewer) with relighting. Click on the terrible lightbulb icon (first item, second row), then click-drag on the image. You can set the position of the light source! It's using webgl to draw the tiles. 

vips8.0 came out in May 2015, the next release is 8.2, due real soon now. 

I wonder if I could ask you for an opinion, @bgilbert ? Benjamin wrote (I think) the openslide deep zoom creator. Do you know what the generally accepted defaults are for dzi and what their origin is?

gasi:
@jcupitt Wow, such a small world and your viewer with the lighting effect is incredible!

I wrote [OpenZoom](https://github.com/openzoom/sdk/) and had an adapter for IIP: https://github.com/openzoom/sdk/blob/9cc5a61330be1448b8e2eb93645d2ca7d4e15dc4/src/org/openzoom/flash/descriptors/iip/IIPImageDescriptor.as OpenZoom was then used by Ruven to build IIPZoom: http://iipimage.sourceforge.net/2010/02/introducing-iipzoom-the-new-iipimage-flash-viewer/

If it helps, this is a Python DZI generator I wrote that should follow the formats / defaults set by the official [DeepZoomTools.dll](http://blogs.msdn.com/b/expression/archive/2008/11/26/hello-deepzoomtools-dll-deep-zoom-image-tile-generation-made-easy.aspx?PageIndex=2): https://github.com/openzoom/deepzoom.py

iangilman:
Hey, nice little reunion we're having here on this ticket! I was a member of that company Microsoft bought (we were called Sand Codex and then Seadragon), and now I manage OpenSeadragon. 

I agree VIPS should move to 254 tile size (with 1 pixel overlap) by default. I know there are a lot of DZI creation tools, and many of them have different defaults, but VIPS should be leading the way to the configuration we know to be the best. 

As for backward compatibility, I wouldn't be that concerned...whatever settings were used in the past were written out to the .dzi file, so even if someone has some of the old 256 tile size files mixed in with newer 254 tile size files, it's all there in the DZI and whatever viewer they're using will pick that up.

jcupitt:
OK, I concede! Thank you all very much, this was an interesting discussion. I'll add a note to the docs on this as well.

libvips/foreign/dzsave.c
```diff
@@ -1999,14 +1999,14 @@ vips_foreign_save_dz_class_init( VipsForeignSaveDzClass *class )
...
-		1, 8192, 256 );
+		1, 8192, 254 );
 
 	VIPS_ARG_INT( class, "tile_height", 12, 
 		_( "Tile height" ), 
 		_( "Tile height in pixels" ),
 		VIPS_ARGUMENT_OPTIONAL_INPUT | VIPS_ARGUMENT_DEPRECATED,
 		G_STRUCT_OFFSET( VipsForeignSaveDz, tile_size ),
-		1, 8192, 256 );
+		1, 8192, 254 );
```
gasi: @jcupitt For these two above here, are they legacy settings? I ask this because of this comment `/* How annoying. We stupidly had these in earlier versions. */` If so and they are used for non-Microsoft DeepZoom images and have no overlap, I should not have changed them to 254 but rather keep them at 256. The goal is to have the majority of tiles be 256x256px and none of them exceed it (or nearest power of two).

On a related note, we should keep in mind that [_Deep Zoom_](https://en.wikipedia.org/wiki/Deep_Zoom) is the product name of the original Microsoft release and generally, the umbrella term would probably better be _image pyramid_ or similar. I am worried that it might be confusing for users to save a _Deep Zoom_ image with _Zoomify_ layout. It might be better to say _image pyramid_ with _Zoomify_ / _Google_ / _Deep Zoom_ layout. Also, as far as I know, only _Deep Zoom_ (DZI) supports tile overlap, which was originally added avoid seams between images due to numerical rounding errors. I don’t think Google Maps or Zoomify pyramids are valid with a non-zero tile overlap value. Is that something we could enforce? Probably too late, but it’d be clearer to name the feature `vips pyramid` or similar and then distinguish between _DZI_ (Deep Zoom Image), Zoomify, etc. Also, consider prefixing the `overlap` parameter with `tile-*` to clarify its meaning :smile: 

Thanks again for making image pyramid generating wicked fast :christmas_tree:

libvips/foreign/dzsave.c
```diff
@@ -116,9 +116,9 @@
 
 	Overlap handling
 
-   For deepzoom, tile-size == 256 and overlap == 1 means that edge tiles are 
-   257 x 257 (though less at the bottom right) and non-edge tiles are 258 x 
-   258. Tiles are positioned across the image in tile-size steps. This means 
+   For deepzoom, tile-size == 254 and overlap == 1 means that edge tiles are 
+   255 x 255 (though less at the bottom right) and non-edge tiles are 256 x 
+   256. Tiles are positioned across the image in tile-size steps. This means 
    (confusingly) that two adjoining tiles will have two pixels in common.
```
gasi: With the explanation of the purpose of tile overlap, hopefully the `(confusingly)` is not necessary anymore. If not, let me know and I’d be happy to expand based on my knowledge :eyes:

gasi: If you’d like to learn more, I wrote some articles a couple of years back:

- http://web.archive.org/web/20150429205932/http://www.gasi.ch/blog/inside-deep-zoom-1/
- http://web.archive.org/web/20150429205932/http://www.gasi.ch/blog/inside-deep-zoom-2/
- http://web.archive.org/web/20150429205932/http://www.gasi.ch/blog/inside-deep-zoom-3/

libvips/foreign/dzsave.c
jcupitt: They are just there for compatibility with very old versions of `dzsave`: they add invisible `tile_width` and `tile_height` options which are synonyms for `tile_size`. I think it's correct to change their defaults to match. 

I'm not too worried about the name. It's a deepzoom writer which can optionally make zoomify or maps pyramids as well. It needs to be called `xxxsave` to be consistent with the other operations in that class.

Overlap defaults to 0 for for maps and zoomify, so I think it's pretty safe. You'd have to specifically ask for a zoomify with a 1-pixel overlap, for example, and perhaps someone might want it.

libvips/foreign/dzsave.c
jcupitt: Hehe, well, I find it confusing. DZI's overlap behaves more like border or margin, I think.

I'll have a look at your articles, thanks for the links.

iangilman: @gasi Thanks for the reminder about those posts! I've added them now to my zooming resource page: 

https://github.com/iangilman/zooming

PR #360, opened 2015-12-28
fixing typo, should be Ifdef

ericcorriel:
I think it has to be ifdef–no? Couldn't get it to compile otherwise (XCode 7)


jcupitt:
Ooop, thanks! That's been there a long time.

I guess no one noticed because `VIPS_ENABLE_DEPRECATED` defaults to `(1)`, which would also pass. Strange you had a compile error. Did you run `configure`?

PR #363, opened 2016-01-11
Enable compilation of vips8 C++ bindings with MSVC

lovell:
Hi John,

C++ name mangling means the `libvips-cpp-42.dll` file created via mingw can't be used with Node on Windows as MSVC is the only supported compiler on that platform.

This means I need to build the vips8 C++ bindings as a DLL using MSVC, but to do this I had to make a couple of syntactic changes and conditionally add MS-only `__declspec` modifiers to the public API.

Cheers,
Lovell

### DIFF
--- a/cplusplus/VImage.cpp
+++ b/cplusplus/VImage.cpp
@@ -189,7 +189,8 @@ VOption::set( const char *name, std::vector<double> value )
 	pair->input = true;
 
 	g_value_init( &pair->value, VIPS_TYPE_ARRAY_DOUBLE );
-	vips_value_set_array_double( &pair->value, NULL, value.size() ); 
+	vips_value_set_array_double( &pair->value, NULL,
+		static_cast< int >( value.size() ) );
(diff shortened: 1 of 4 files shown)


jcupitt:
I wondered if there was some gcc switch would would make name mangling match the MSVC convention, but it seems that would not be sufficient. There are a large range of other issues relating to exception handling, new/delete, call sequencing and so on which also make C++ MSVC / gcc linking impossible.

http://www.mingw.org/wiki/Interoperability_of_Libraries_Created_by_Different_Compiler_Brands

Your patch looks good, thanks!

lovell:
Great, thanks for merging.

lovell:
https://github.com/lovell/libvips-cpp-dll is the quickly-thrown-together build config for this.

jcupitt:
Nice!

lovell:
Binaries at https://github.com/lovell/libvips-cpp-dll/releases

PR #366, opened 2016-01-14
Improve portability of VError class

lovell:
Hi John, 

This change to the internals of the VError class switches to a more portable `const char*` C reference from a `std::string` without altering its public API.

The requirement for this stems from MSVC's habit of changing its std library class ABI with every release. For example. a `std::string` allocated in a DLL compiled with MSVC 2013 can't be freed from an executable compiled with MSVC 2014.

This was causing segfaults on Windows within the `~VError()` destructor.

I was able to work around a similar problem on Linux with gcc <= 5.0 vs gcc >= 5.1 by setting the `-D_GLIBCXX_USE_CXX11_ABI=0` flag. Sadly there's no equivalent in MSVC land.

Cheers,
Lovell

### DIFF
--- a/cplusplus/include/vips/VError8.h
+++ b/cplusplus/include/vips/VError8.h
@@ -40,15 +40,14 @@
 VIPS_NAMESPACE_START
 
 class VIPS_CPLUSPLUS_API VError : public std::exception {
-	std::string _what;
+	const char *_what;
 
 public:
-	VError( std::string what ) : _what( what ) {}
-	VError() : _what( vips_error_buffer() ) {}
+	VError( std::string what ) { _what = what.c_str(); }
+	VError() { _what = vips_error_buffer(); }
 	virtual ~VError() throw() {}
 
-	// Extract string
-	virtual const char *what() const throw() { return _what.c_str(); }
+	virtual const char *what() const throw() { return _what; }
 	void ostream_print( std::ostream & ) const;
 };
 


lovell:
The more I think about this, the more I'm tending towards having sharp ship with the source code to build `libvips-cpp.dll` at install time and completely avoid this problem. It might be `std::string` changes at fault here, but many other methods return members of the std library that could easily succumb to this problem in the future.

Please don't merge this until I've tried that approach first, sorry.

jcupitt:
Are C++ objects moving between the plugin and V8? Or is the interface all C? I'm vague how the details work on this :-( 

If C++ objects are being passed into V8 then I agree, I think there's no way to distribute a binary. But if the interface is plain C, can't we fix it somehow?

lovell:
> "If you have multiple binaries loaded into the same process, and they pass C++ Standard Library objects to each other, those binaries must be built with the same major version [of MSVC]"

http://stackoverflow.com/a/15951016/4625124

Instances of `VError` containing instances of `std::string` are passed from the `libvips-cpp.dll` shared library to the `sharp.node` shared library within the `node.exe` process.

From what I can tell, it looks like an exception's destructor runs in the context of the catcher of that exception.

jcupitt:
So if you built sharp.node and libvips-cpp.dll with the same compiler, it'd be fine, is that right? Can't you just ship a single binary containing both bits?

I'm probably being dumb :(

lovell:
Sadly the V8 ABI changes even more frequently than the MSVC std library ABI. https://github.com/lovell/sharp/issues/186 is tracking the work to provide pre-built binaries of `sharp.node`.

I've managed to get Windows to build the vips8 C++ bindings DLL at install time, the least bad solution right now, so please ignore this PR and sorry for any time wasted.

I've had to include the [C++ sources for vips8](https://github.com/lovell/sharp/tree/mind/src/libvips/cplusplus) to do this. They all contain comments with licensing details - are you aware of any further copyright/licence requirements?

jcupitt:
I did the C++ binding, so it's OK for me to hereby declare this part of libvips to be apache 2.0 and therefore able to be included in sharp. 

Perhaps the C++ binding should be broken out into a separate project with a separate license, it would prevent this strangeness. 

Are you missing the libvips-cplusplus headers? I can't see them, I'm being dumb.

lovell:
The header files are included with the pre-built binaries, which for Windows are based on the `vips-dev-w64-*.zip` you provide. There's a [temporary workaround](https://github.com/lovell/sharp/blob/mind/packaging/win/Dockerfile#L18-L21) until 8.2.2 is out. This approach is a little messy and needs documenting, but it works.

sharp uses libvips under the terms of LGPL3. The C++ bindings are compiled to a shared library (DLL), the use of which I believe is compatible with Apache 2.0. However if you're OK with either changing or dual licensing this part then that would make things more explicit (and allow for possible future static linking).

PR #369, opened 2016-01-21
Allow compiler to use vector unit with vips_shrinkh for ~25% boost

lovell:
Hi John,

Based on `callgrind` testing, these changes seem to grant the compiler (gcc 5.3 in my tests) use of the vector unit with `vips_shrinkh`.

When compiled with -O3 this allows `vips_shrinkh_gen` to run with ~25% fewer instructions, so the `vips_shrink` operation as a whole runs up to about 10% faster.

I'm seeing the performance of a more typical shrink+affine reduction improved by ~5%.

Below are the `callgrind` results of running a 2x shrink operation on a WebP image 100x.

Current code, using -O2, `vips_shrinkh_gen` and `vips_shrinkv_gen` are about the same.
```
6,191,513,100  ???:0x000000000002ca10 [/usr/lib/x86_64-linux-gnu/libwebp.so.5.0.1]
5,729,485,600  ???:0x000000000000d140 [/usr/lib/x86_64-linux-gnu/libwebp.so.5.0.1]
2,634,604,404  resample/shrinkh.c:vips_shrinkh_gen [/usr/local/lib/libvips.so.42.3.1]
2,619,589,004  resample/shrinkv.c:vips_shrinkv_gen [/usr/local/lib/libvips.so.42.3.1]
2,150,232,000  ???:0x0000000000021380 [/usr/lib/x86_64-linux-gnu/libwebp.so.5.0.1]
```
Current code, using -O3, `vips_shrinkv_gen` uses half the instructions of `vips_shrinkh_gen`.
```
6,191,513,100  ???:0x000000000002ca10 [/usr/lib/x86_64-linux-gnu/libwebp.so.5.0.1]
5,729,485,600  ???:0x000000000000d140 [/usr/lib/x86_64-linux-gnu/libwebp.so.5.0.1]
2,911,713,804  ???:vips_shrinkh_gen [/usr/local/lib/libvips.so.42.3.1]
2,150,232,000  ???:0x0000000000021380 [/usr/lib/x86_64-linux-gnu/libwebp.so.5.0.1]
2,091,408,000  ???:0x00000000000211e0 [/usr/lib/x86_64-linux-gnu/libwebp.so.5.0.1]
1,806,064,000  ???:0x0000000000020c60 [/usr/lib/x86_64-linux-gnu/libwebp.so.5.0.1]
1,772,999,400  ???:0x0000000000037f10 [/usr/lib/x86_64-linux-gnu/libwebp.so.5.0.1]
1,723,104,000  ???:VP8YuvToRgb32 [/usr/lib/x86_64-linux-gnu/libwebp.so.5.0.1]
1,554,228,000  ???:0x00000000000216f0 [/usr/lib/x86_64-linux-gnu/libwebp.so.5.0.1]
1,489,110,804  ???:vips_shrinkv_gen [/usr/local/lib/libvips.so.42.3.1]
1,405,190,245  ???:0x0000000000033600 [/usr/lib/x86_64-linux-gnu/libwebp.so.5.0.1]
```
With this patch, using -O3, `vips_shrinkh_gen` uses ~25% fewer instructions.
```
6,191,513,100  ???:0x000000000002ca10 [/usr/lib/x86_64-linux-gnu/libwebp.so.5.0.1]
5,729,485,600  ???:0x000000000000d140 [/usr/lib/x86_64-linux-gnu/libwebp.so.5.0.1]
2,150,232,000  ???:0x0000000000021380 [/usr/lib/x86_64-linux-gnu/libwebp.so.5.0.1]
2,140,833,200  ???:vips_shrinkh_gen [/usr/local/lib/libvips.so.42.3.1]
2,091,408,000  ???:0x00000000000211e0 [/usr/lib/x86_64-linux-gnu/libwebp.so.5.0.1]
1,806,064,000  ???:0x0000000000020c60 [/usr/lib/x86_64-linux-gnu/libwebp.so.5.0.1]
1,772,999,400  ???:0x0000000000037f10 [/usr/lib/x86_64-linux-gnu/libwebp.so.5.0.1]
1,723,104,000  ???:VP8YuvToRgb32 [/usr/lib/x86_64-linux-gnu/libwebp.so.5.0.1]
1,554,228,000  ???:0x00000000000216f0 [/usr/lib/x86_64-linux-gnu/libwebp.so.5.0.1]
1,489,110,804  ???:vips_shrinkv_gen [/usr/local/lib/libvips.so.42.3.1]
1,405,190,245  ???:0x0000000000033600 [/usr/lib/x86_64-linux-gnu/libwebp.so.5.0.1]
```
Cheers,
Lovell

### DIFF
--- a/libvips/resample/shrinkh.c
+++ b/libvips/resample/shrinkh.c
@@ -111,18 +111,15 @@ vips_shrinkh_start( VipsImage *out, void *a, void *b )
 	TYPE * restrict q = (TYPE *) out; \
 	\
 	for( x = 0; x < width; x++ ) { \
-		for( b = 0; b < bands; b++ ) \
+		for( b = 0; b < bands; b++ ) { \
 			sum[b] = 0; \
-		\
-		for( b = 0; b < bands; b++ ) \
 			for( x1 = b; x1 < ne; x1 += bands ) \
 				sum[b] += p[x1]; \
-		p += ne; \
-		\
-		for( b = 0; b < bands; b++ ) \
 			q[b] = (sum[b] + shrink->xshrink / 2) / \
 				shrink->xshrink; \
-		q += b; \
+		} \
+		p += ne; \
+		q += bands; \
 	} \
 }
 
@@ -134,17 +131,14 @@ vips_shrinkh_start( VipsImage *out, void *a, void *b )
 	TYPE * restrict q = (TYPE *) out; \
 	\
 	for( x = 0; x < width; x++ ) { \
-		for( b = 0; b < bands; b++ ) \
+		for( b = 0; b < bands; b++ ) { \
 			sum[b] = 0.0; \
-		\
-		for( b = 0; b < bands; b++ ) \
 			for( x1 = b; x1 < ne; x1 += bands ) \
 				sum[b] += p[x1]; \
-		p += ne; \
-		\
-		for( b = 0; b < bands; b++ ) \
 			q[b] = sum[b] / shrink->xshrink; \
-		q += b; \
+		} \
+		p += ne; \
+		q += bands; \
 	} \
 } 
 
@@ -227,7 +221,7 @@ vips_shrinkh_gen( VipsRegion *or, void *vseq,
 
 		s.left = r->left * shrink->xshrink;
 		s.top = r->top + y;
-		s.width = ceil( r->width * shrink->xshrink );
+		s.width = r->width * shrink->xshrink;
 		s.height = 1;
 #ifdef DEBUG
 		printf( "shrinkh_gen: requesting line %d\n", s.top ); 


jcupitt:
You wouldn't think collapsing those b loops would make so much difference. Thanks! It's also clear with your reorganisation that we can get rid of the `sum` intermediate and just use a local, that should help a bit more. 

Have you tried the gcc vector trace flags?

```
        $ FLAGS="-O2 -msse4.2 -ffast-math"
        $ FLAGS="$FLAGS -ftree-vectorize -fdump-tree-vect-details"
        $ CFLAGS="$FLAGS" CXXFLAGS="$FLAGS" \
                ./configure --prefix=/home/john/vips
```

Now gcc will write a copy of each source file with lines marked up to show which loops were vectorised, and which were not and why they were not.

jcupitt:
Ooop, meant to add, I worked through vips a year or two ago fixing all the important loops and making them vectorisable, but I've not done it since. Any code from the last year or so probably ought to be checked.

jcupitt:
I removed the sum buffer and it gets another 5% or so for me. Fastest of 10 runs before:

```
$ time vips shrinkh wtc.tif x.tif 10 
real	0m0.324s
user	0m0.740s
sys	0m0.188s
```

And after:

```
$ time vips shrinkh wtc.tif x.tif 10 
real	0m0.313s
user	0m0.748s
sys	0m0.140s
```

It saves 50 lines of code too, which is very nice.

jcupitt:
I tried turning on the vector analysis output. Here's the inner loop we want to vectorise:

```
                        sum = 0; 
                        for( x1 = b; x1 < ne; x1 += bands )
                                sum += p[x1]; 
                        q[b] = (sum + shrink->xshrink / 2) /
                                shrink->xshrink;
```

It complains that it can't determine the number of times to loop. It's the `x1 += bands`: it can't tell when the loop will finish. If you rewrite as:

```
                        sum = 0; 
                        for( x1 = b; x1 < ne; x1 += 3 )
                                sum += p[x1]; 
                        q[b] = (sum + shrink->xshrink / 2) /
                                shrink->xshrink;
```

Now it can vectorise, and outputs a lot of stuff about the code it's generated. Of course the code won't work for anything but RGB now :( 

I'm not sure what to do: the problem is the structure of the data. We need to sum every Nth element, and there's just no neat way to do that as a general vector loop.

One solution would be to generate special cases for 1, 3 and 4 bands. We could have the current code as a fallback, and pick a hardwired path for the three most common cases. That would be very easy.

jcupitt:
... and with special paths for 1, 3 and 4 bands. Another 5% speed up! I now see:

```
$ time vips shrinkh wtc.tif x.tif 10 
real	0m0.297s
user	0m0.692s
sys	0m0.140s
```

lovell:
Nice work on those extra boosts John. I've not used `dump-tree-vect-details` before so thank you for that tip!

jcupitt:
I moved the inner loop to `VIPS_UNROLL` for another 5%.

```
$ time vips shrinkh wtc.tif x.tif 10 
real	0m0.288s
user	0m0.600s
sys	0m0.188s
```

I can't think of anything else, except maybe transposing the pixels between v and h shrink.

PR #372, opened 2016-01-22
Small perf improvements to vips_affine + bicubic

lovell:
Hi John,

There were a few copies of the `FAST_PSEUDO_FLOOR` macro, so these all now reference the same version and, if available, will use the ~7% faster `__builtin_floor` function.

`vips_afffine_gen` already uses `vips_interpolate_get_method` so the `interpolate` instance isn't required by the `interpolate` method. I've run the sharp unit tests over this to ensure all the interpolators still work.

It looks like register starvation was starting to affect `vips_interpolate_bicubic_interpolate` so there's now one less temp variable in there.

The result is ~10% less time spent in `vips_affine_gen` and ~5% less time spent in `vips_interpolate_bicubic_interpolate`.

I see an overall improvement to `vips_affine` of ~3%.

Cheers,
Lovell

Before:

```
4,444,599,308  ???:decode_mcu [/usr/local/lib/libjpeg.so.8.1.2]
1,532,000,000  ???:vips_interpolate_bicubic_interpolate(_VipsInterpolate*, void*, _VipsRegion*, double, double) [/usr/local/lib/libvips.so.42.3.1]
1,141,811,867  ???:conv_gen'2 [/usr/local/lib/libvips.so.42.3.1]
1,133,844,404  ???:conv_gen [/usr/local/lib/libvips.so.42.3.1]
  735,118,204  ???:decompress_onepass [/usr/local/lib/libjpeg.so.8.1.2]
  526,680,000  ???:jsimd_idct_2x2 [/usr/local/lib/libjpeg.so.8.1.2]
  289,870,000  ???:jpeg_fill_bit_buffer [/usr/local/lib/libjpeg.so.8.1.2]
  245,142,470  /build/buildd/eglibc-2.19/string/../sysdeps/x86_64/memset.S:memset [/lib/x86_64-linux-gnu/libc-2.19.so]
  155,446,012  ???:vips_affine_gen [/usr/local/lib/libvips.so.42.3.1]
  139,199,712  ???:g_hash_table_lookup [/lib/x86_64-linux-gnu/libglib-2.0.so.0.4002.0]
```
After:
```
4,444,599,308  ???:decode_mcu [/usr/local/lib/libjpeg.so.8.1.2]
1,460,000,000  ???:vips_interpolate_bicubic_interpolate(_VipsInterpolate*, void*, _VipsRegion*, double, double) [/usr/local/lib/libvips.so.42.3.1]
1,141,811,631  ???:conv_gen'2 [/usr/local/lib/libvips.so.42.3.1]
1,133,844,404  ???:conv_gen [/usr/local/lib/libvips.so.42.3.1]
  735,118,204  ???:decompress_onepass [/usr/local/lib/libjpeg.so.8.1.2]
  526,680,000  ???:jsimd_idct_2x2 [/usr/local/lib/libjpeg.so.8.1.2]
  289,870,000  ???:jpeg_fill_bit_buffer [/usr/local/lib/libjpeg.so.8.1.2]
  245,125,190  /build/buildd/eglibc-2.19/string/../sysdeps/x86_64/memset.S:memset [/lib/x86_64-linux-gnu/libc-2.19.so]
  139,206,416  ???:vips_affine_gen [/usr/local/lib/libvips.so.42.3.1]
  139,196,155  ???:g_hash_table_lookup [/lib/x86_64-linux-gnu/libglib-2.0.so.0.4002.0]
```

### DIFF
--- a/libvips/resample/affine.c
+++ b/libvips/resample/affine.c
@@ -155,29 +155,6 @@ typedef VipsResampleClass VipsAffineClass;
 
 G_DEFINE_TYPE( VipsAffine, vips_affine, VIPS_TYPE_RESAMPLE );
 
-/*
- * FAST_PSEUDO_FLOOR is a floor and floorf replacement which has been
- * found to be faster on several linux boxes than the library
- * version. It returns the floor of its argument unless the argument
- * is a negative integer, in which case it returns one less than the
- * floor. For example:
- *
- * FAST_PSEUDO_FLOOR(0.5) = 0
- *
- * FAST_PSEUDO_FLOOR(0.) = 0
- *
- * FAST_PSEUDO_FLOOR(-.5) = -1
- *
- * as expected, but
- *
- * FAST_PSEUDO_FLOOR(-1.) = -2
- *
- * The locations of the discontinuities of FAST_PSEUDO_FLOOR are the
- * same as floor and floorf; it is just that at negative integers the
- * function is discontinuous on the right instead of the left.
- */
-#define FAST_PSEUDO_FLOOR(x) ( (int)(x) - ( (x) < 0. ) )
-
 /* We have five (!!) coordinate systems. Working forward through them, these
  * are:
  *
@@ -374,8 +351,7 @@ vips_affine_gen( VipsRegion *or, void *seq, void *a, void *b, gboolean *stop )
 					(int) iy - window_offset + 
 						window_size - 1 ) );
 
-				interpolate( affine->interpolate, 
-					q, ir, ix, iy );
+				interpolate( NULL, q, ir, ix, iy );
 			}
 			else {
 				for( z = 0; z < ps; z++ ) 
--- a/libvips/resample/bicubic.cpp
+++ b/libvips/resample/bicubic.cpp
@@ -153,19 +153,17 @@ bicubic_unsigned_int_tab( void *pout, const VipsPel *pin,
 		const T qua_thr = in[l3_plus_b2];
 		const T qua_fou = in[l3_plus_b3];
 
-		int bicubic = bicubic_unsigned_int<T>(
+		out[z] = bicubic_unsigned_int<T>(
 			uno_one, uno_two, uno_thr, uno_fou,
 			dos_one, dos_two, dos_thr, dos_fou,
 			tre_one, tre_two, tre_thr, tre_fou,
 			qua_one, qua_two, qua_thr, qua_fou,
 			cx, cy );
 
-		if( bicubic < 0 )
-			bicubic = 0;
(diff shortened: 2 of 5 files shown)


jcupitt:
Interesting, I wonder why the number of calls to `vips_interpolate_bicubic_interpolate` has fallen? Your very useful change to floor I guess. 

I've merged, but I'll undo a couple of these, sorry :-( I want the interpolate parameter in there in case a (future) interpolator takes parameters. We have had parameterised interpolators before.

jcupitt:
You're right, all the floor stuff was a mess. I've made a `VIPS_FLOOR` / `VIPS_CEIL` pair which use the fast builtins, and they are now used everywhere. No more `PSEUDO_FLOOR`, phew.

jcupitt:
Sorry, I had to reverse your change to bicubic as well.

Bicubic is not strictly interpolatory: if you have points A B C D and you're trying to calculate a value for a point x between B and C, x will not always be between B and C in value, it can be greater or less than both of them. For example, if there's a peak between B and C (A -> B is rising, C -> D is falling) then x will be greater than both B and C.

So ... we need to capture the result of bicubic in a type larger than the type we are working on and then clip the range down. Otherwise we'll get nasty overflows or wrapping around on edges. 

Although no current interpolators use the interpolate param, having it
there makes it possible for other (perhaps user-defined) interpolators
to be parameterised.

For example, you could perhaps write:

```
vips affine in.v out.v "2 0 0 2" --interpolate sharp[2]
```

and `sharp` could maybe do a bicubic and sharpen in a single operation, with the parameter controlling the degree of sharpening.

jcupitt:
You probably saw, but there's quite a large test suite now, and it tests the interpolators, though it wouldn't have spotted the bicubic overflows unfortunately. You can run the whole lot:

```
john@kiwi:~/GIT/libvips/test (master)$ ./test_all.py 
.......................................................................................................................................................................
----------------------------------------------------------------------
Ran 167 tests in 48.052s

OK
memory: high-water mark 114.37 MB
```

Or a particular one:

```
john@kiwi:~/GIT/libvips/test (master)$ ./test_resample.py TestResample.test_affine
.
----------------------------------------------------------------------
Ran 1 test in 2.398s

OK
memory: high-water mark 11.89 MB
```

lovell:
Thanks for merging, and sorry I didn't spot the bicubic overflow thing. The ability to blend two or more tasks into a single interpolation operation could be very powerful, I agree.

The callgrind stats show the number of CPU instructions spent within each function rather than number of times that function is called. I've been using them as a quick, semi-automated way of proving a given small change is worth investigating further.

PR #374, opened 2016-01-25
Improve convolution performance by 20-25%

lovell:
Hi John,

Replacing the "Duff's device" with a for-loop seems to allow gcc to optimise this code, including permitting the use of vectorisation. It reduces the CPU instruction count of `conv_gen` by about 30%, which leads to a performance improvement to the (non-3x3) convolution operation as a whole by somewhere between 20 to 25%.

Hopefully I've not made any glaring errors in my maths this time!

The following tests should demonstrate the fairly dramatic drop in instruction count for a Gaussian blur with sigma=10 and liborc disabled. (I've included libjpeg's `decode_mcu` function in the output as a "control" to ensure the same test was run for each scenario.)

Cheers,
Lovell

precision=integer
```
-- master -O2

4.415s

18,956,604,353  libvips/convolution/im_conv.c:conv_gen'2 [/usr/local/lib/libvips.so.42.3.1]
15,287,450,934  libvips/convolution/im_conv.c:conv_gen [/usr/local/lib/libvips.so.42.3.1]
   592,029,158  ???:decode_mcu [/usr/local/lib/libjpeg.so.8.1.2]

-- master -O3

4.432s

18,956,603,353  ???:conv_gen'2 [/usr/local/lib/libvips.so.42.3.1]
15,287,457,226  ???:conv_gen [/usr/local/lib/libvips.so.42.3.1]
   592,029,158  ???:decode_mcu [/usr/local/lib/libjpeg.so.8.1.2]

-- conv-allow-vectorise -O2

4.190s

13,974,480,980  libvips/convolution/im_conv.c:conv_gen'2 [/usr/local/lib/libvips.so.42.3.1]
11,269,633,462  libvips/convolution/im_conv.c:conv_gen [/usr/local/lib/libvips.so.42.3.1]
   592,029,158  ???:decode_mcu [/usr/local/lib/libjpeg.so.8.1.2]

-- conv-allow-vectorise -O3

3.395s

11,362,912,900  ???:conv_gen'2 [/usr/local/lib/libvips.so.42.3.1]
 9,163,516,484  ???:conv_gen [/usr/local/lib/libvips.so.42.3.1]
   592,029,158  ???:decode_mcu [/usr/local/lib/libjpeg.so.8.1.2]
```
precision=float
```
-- master -O2

4.525s

18,305,707,960  libvips/convolution/im_conv_f.c:conv_gen'2 [/usr/local/lib/libvips.so.42.3.1]
13,264,259,684  libvips/convolution/im_conv_f.c:conv_gen [/usr/local/lib/libvips.so.42.3.1]
   592,029,158  ???:decode_mcu [/usr/local/lib/libjpeg.so.8.1.2]

-- master -O3

4.439s

20,164,137,120  ???:conv_gen'2 [/usr/local/lib/libvips.so.42.3.1]
13,264,371,184  ???:conv_gen [/usr/local/lib/libvips.so.42.3.1]
   592,029,158  ???:decode_mcu [/usr/local/lib/libjpeg.so.8.1.2]

-- conv-allow-vectorise -O2

4.210s

17,550,623,384  libvips/convolution/im_conv_f.c:conv_gen'2 [/usr/local/lib/libvips.so.42.3.1]
12,655,099,920  libvips/convolution/im_conv_f.c:conv_gen [/usr/local/lib/libvips.so.42.3.1]
   592,029,158  ???:decode_mcu [/usr/local/lib/libjpeg.so.8.1.2]

-- conv-allow-vectorise -O3

3.926s

17,891,636,579  ???:conv_gen'2 [/usr/local/lib/libvips.so.42.3.1]
11,593,399,095  ???:conv_gen [/usr/local/lib/libvips.so.42.3.1]
   592,029,158  ???:decode_mcu [/usr/local/lib/libjpeg.so.8.1.2]
```


jcupitt:
Wow Lovell, that looks fantastic. I've not reworked im_conv() for years :-( certainly not since gcc auto-vec became useful. Duff was a good thing back then, but it's obviously not now. Brilliant!

jcupitt:
I ran the vector analyzer and it's not actually being vectorised. It doesn't seem to like the `p[i][x]` in the inner loop -- we'd need to find another way to code this.

The error is:

```
im_conv.c:773:3: note: bad data references.
```

I made a branch here to mess about:

https://github.com/jcupitt/libvips/tree/convolution-tuning

The uchar loop for `im_conv.c` is there three times: as duff, as not-duff, and as expanded not-duff (useful for checking loop line numbers with the vec tree dump).

I do see a very nice speedup with your change though, I guess gcc is able to do something clever even without vectorising:

First make a 4x4 sharpen mask (so we avoid the special 3x3 path):

```
$ cat >sharp.mat
4 4  
-1 -1 -1 -1
-1 3 3 -1
-1 3 3 -1
-1 -1 -1 -1
```

Then `conv` as was:

```
$ time vips conv wtc.tif x.tif sharp.mat --vips-novector
real	0m4.974s
user	0m15.992s
sys	0m0.376s
```

Your duff-free conv:

```
$ time vips conv wtc.tif x.tif sharp.mat --vips-novector
real	0m3.414s
user	0m10.092s
sys	0m0.396s
```

And finally Orc:

```
$ time vips conv wtc.tif x.tif sharp.mat 
real	0m1.368s
user	0m2.268s
sys	0m0.384s
```

Which is rather sobering.

lovell:
Thanks for merging John.

My attempt to break the inner loop apart, from:
```c
for ( i = 0; i < nnz; i++ )
  sum += t[i] * p[i][x];
```
to:
```c
int t2[nnz];
for ( i = 0; i < nnz; i++ )
  t2[i] = p[i][x];
for ( i = 0; i < nnz; i++ )
  sum += t[i] * t2[i];
```
did vectorise but with worse performance :(

Are the under/overflow counts in the `VIPS_CLIP_` macros used by anything? If not, a possible improvement might be using the compiler's built-in min and max functions for pixel value clipping (with the aim to avoid branching).

jcupitt:
I tried:

```C
                for ( i = 0; i < nnz; i++ ) {
                        sum += t[i] * *p[i];
                        p[i] += 1; 
                }
```

But it's hardly different and also didn't vectorise :( 

At the moment, conv tries to squeeze out zero mask elements. If we abandon that we could remove a level of indirection in the accesses to pixels and make it vectorisable (I think). I'll see if I can make a test.

Yes, it looks like GCC has __builtin_fmin()/_fmax(), but nothing for int values. We should probably make a VIPS_FMIN()/_FMAX() and use it where we can. 

I think g++ has a min/max for int with the built in `a <? b`.

jcupitt:
Yes, swapping the

```C
IM_CLIP_UCHAR( sum, seq );
```

for 

```C
sum = VIPS_CLIP( 0, sum, 255 );
```

Saves about 10%:

```
$ time vips conv wtc.tif x.tif sharp.mat --vips-novector
real	0m3.137s
user	0m9.008s
sys	0m0.440s
```

`__builtin_fmaxf` is actually slower, since it's having to go to float and back:

```C
sum = __builtin_fminf( 255.0f, __builtin_fmaxf( 0.0f, sum ) );
```

```
$ time vips conv wtc.tif x.tif sharp.mat --vips-novector
real	0m2.761s
user	0m9.472s
sys	0m0.404s
```

lovell:
To use fast integer-based min/max clipping on all compilers (not just gcc), we'd need to use SSE2 intrinsics, available on all 64-bit Intel CPUs.

This would look something like:
```c
#if defined(__x86_64__)
#  include <emmintrin.h>
#  define VIPS_CLIP_INT( A, V, B ) _mm_extract_epi16( _mm_min_epi16( _mm_max_epi16( _mm_set1_epi16( V ), _mm_set1_epi16( A ) ), _mm_set1_epi16( B ) ), 0 )
#else
#  define VIPS_CLIP_INT( A, V, B ) VIPS_CLIP( A, V, B )
#endif
```
The real power here is the ability to clip 8 integers at once (with `_mm_set_epi16` rather than `_mm_set1_epi16`), but that's a world of complexity I'd be happy to ignore by hiding behind liborc :)

jcupitt:
I tried adding `VIPS_ICLIP()` with that scrap of sse, but it's no faster than `?:`, unfortunately.

I timed the 3x3 path and it still seems usefully quicker.

jcupitt:
Shall we stop here? The easy stuff seems to be done. 

I'll make 8.2.2 with all these changes.

lovell:
I agree, release at will, thanks John!

PR #380, opened 2016-02-09
bootstrap.sh: Better doc pruning / support acdir dirlists

felixbuenemann:
* Clean doc dir with find instead of moving files to keep to temp dir (old code could break on ctrl+c)
* Limited support for automake ACDIR dirlists (require for mac homebrew, no wildcard handling)

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -13,47 +13,43 @@ rm -f swig/vipsCC/*.cxx
 rm -f swig/vipsCC/VImage.h
 rm -f swig/vipsCC/VImage.py python/vipsCC/VError.py python/vipsCC/VMask.py python/vipsCC/Display.py
 rm -f benchmark/temp*
-( mkdir poop ; \
-	mv doc/libvips-docs.xml.in poop ; \
-	mv doc/Makefile.am poop ; \
-	mv doc/images poop ; \
-	mv doc/*.xml poop ; \
-	mv doc/*.py poop ; \
-	rm -rf doc/* ; \
-	mv poop/* doc ; \
-	rmdir poop \
-)
+find doc -depth \( \
+      -path doc/libvips-docs.xml.in \
+   -o -path doc/Makefile.am \
+   -o -path 'doc/images/*' \
+   -o -name '*.xml' ! -name libvips-docs.xml ! -path 'doc/xml/*' \
+   -o -name '*.py' \
+\) -prune -or \( \
+      -type f \
+   -o -type d -empty \
+\) -delete
 
-# glib-gettextize asks us to copy these files to m4 if they aren't there
-# I don't have $ACDIR/isc-posix.m4, how mysterious
 ACDIR=`aclocal --print-ac-dir`
-
-# OS X with brew sets ACDIR to
-# /usr/local/Cellar/automake/1.13.1/share/aclocal, the staging area, which is
-# totally wrong argh
-if [ ! -d $ACDIR ]; then
-	ACDIR=/usr/local/share/aclocal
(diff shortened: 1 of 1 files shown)


bootstrap.sh
```diff
@@ -13,47 +13,41 @@ rm -f swig/vipsCC/*.cxx
...
 mkdir -p m4
-cp $ACDIR/codeset.m4 m4
-cp $ACDIR/gettext.m4 m4
-cp $ACDIR/glibc21.m4 m4
-cp $ACDIR/iconv.m4 m4
-cp $ACDIR/lcmessage.m4 m4
-cp $ACDIR/progtest.m4 m4
-cp $ACDIR/introspection.m4 m4
+# glib-gettextize asks us to copy these files to m4 if they aren't there:
+files="codeset gettext glibc21 iconv isc-posix lcmessage progtest introspection"
```
felixbuenemann: I have none of these files except introspection.m4 in /usr/local/share/aclocal and doc generation seems to work fine.
Is this actually required on linux?

felixbuenemann:
Seems I need to adapt the `find` command for GNU find (works fine on BSD find):

```
find: warning: you have specified the -depth option after a non-option argument (, but options are not positional (-depth affects tests specified before it as well as those specified after it).  Please specify options before other arguments.
```

felixbuenemann:
Adjusted to work with both GNU and BSD find.

bootstrap.sh
```diff
@@ -13,47 +13,43 @@ rm -f swig/vipsCC/*.cxx
...
-	mv doc/*.py poop ; \
-	rm -rf doc/* ; \
-	mv poop/* doc ; \
-	rmdir poop \
-)
+find doc -depth \( \
+      -path doc/libvips-docs.xml.in \
+   -o -path doc/Makefile.am \
+   -o -path 'doc/images/*' \
+   -o -name '*.xml' ! -name libvips-docs.xml ! -path 'doc/xml/*' \
```
felixbuenemann: The generated libvips-docs.xml was previously preserved, because it matched `*.xml`.

jcupitt:
I had a good idea: how about using git to remove all files not under source control from `doc/`? That would be much simpler and safer than a find command or that craziness with `mv` I had. 

Googling seems to suggest something like:

```
git clean -dfX doc
```

should do it. It seems `libvips-docs.xml` is under source control, it should probably be added to `.gitignore`.

felixbuenemann:
That was my initial idea, I thought you intentionally avoided git so it also works with a tarball.

jcupitt:
bootstrap predates git, vips was SVN back when it was made (!!).

You only need bootstrap when building from source control. Tarballs have configure all set up and ready to go. I think it's OK to depend on git within bootstrap.

felixbuenemann:
The gtk-doc.make should probably also be removed from git, it's auto-generated by `gtkdocize`.

felixbuenemann:
Oh and I have fond memories of the times of svn and slandering into the kitchen to brew a cup of coffee and have a chat while waiting for `svn checkout` to complete… ;-)

jcupitt:
I've tried removing `gtk-doc.make` before, but gtkdocize on OS X wouldn't run without that file, so you couldn't then build from git on OS X. Perhaps that's been fixed.

felixbuenemann:
Seems to be fixed, it gets generated on my machine (OS X 10.11) with `gtkdocize (gtk-doc) 1.24`.

jcupitt:
Merged and I'll have a go with git clean in a separate commit. Thanks!

jcupitt:
Actually, you're right, let's not add another dependency. find it is.

PR #402, opened 2016-03-09
Performance: only build LAB to sRGB look-up tables if needed

lovell:
This change makes `vips_LabQ2sRGB_line` follow the pattern of the pixel-based conversion functions such as `vips_col_scRGB2sRGB_8` by only building the LUTs when first used.

More importantly, this change reduces the start-up time of the `vips`, `vipsthumbnail` etc. command line tools by ~20ms. For trivial operations on small images, where a LAB to sRGB conversion is not used, I see a ~3x speed-up.


jcupitt:
Nice! 

I'm telling myself it's an oversight, there's a lock on table create ready to move it to first use, but the call was stupidly left in class create.

lovell:
Thanks for merging!

PR #412, opened 2016-04-05
Remove = in #define

charmander:
Fixes a compiler warning and not much else.


jcupitt:
Ooop! Thank you very much for this fix @charmander.

PR #421, opened 2016-04-14
Fix required poppler-glib version in AC_MSG_WARN()

bgilbert:
The pkg-config test requires >= 0.30.0, but the warning message asks for >= 0.40.0.


jcupitt:
Ooops again, thank you!

PR #424, opened 2016-04-15
Allow MSVC to link CPP overloads introduced in da43a40

lovell:
Testing v8.3.0 with Windows, as promised :)


jcupitt:
Ooop, I keep forgetting about the declspec stuff :( Thanks!

PR #430, opened 2016-04-23
Use WebP Advanced Encoding API and make it tunable

felixbuenemann:
* Switch to WebP Advanced Encoding API to allow access to `WebPConfig`
* Add `preset` enumerable to tune WebP lossy compression for certain types of images
* Add `smart_subsample` switch that greatly reduces color bleeding from 4:2:0 subsampling
* Add `near_lossless` switch that allows preprocessing in lossless mode using `Q` option
* Add `alphaQ` option similar to `Q` that allows to reduce alpha palette in lossy mode

### DIFF
--- a/libvips/foreign/foreign.c
+++ b/libvips/foreign/foreign.c
@@ -2458,6 +2458,8 @@ vips_webpload_buffer( void *buf, size_t len, VipsImage **out, ... )
  * Optional arguments:
  *
  * @Q: quality factor
+ * @lossless: enables lossless compression
+ * @preset: #VipsForeignWebpPreset choose lossy compression preset
  *
  * See also: vips_webpload(), vips_image_write_to_file().
  *
@@ -2486,6 +2488,8 @@ vips_webpsave( VipsImage *in, const char *filename, ... )
  * Optional arguments:
  *
  * @Q: JPEG quality factor
+ * @lossless: enables lossless compression
+ * @preset: #VipsForeignWebpPreset choose lossy compression preset
  *
  * See also: vips_webpsave().
  *
--- a/libvips/foreign/foreign.c
+++ b/libvips/foreign/foreign.c
@@ -2497,6 +2497,12 @@ vips_webpsave( VipsImage *in, const char *filename, ... )
  * @near_lossless: use preprocessing in lossless mode (controlled by Q)
  * @alpha_q: set alpha quality in lossless mode
  *
+ * As vips_webpsave(), but save to a memory buffer.
+ *
+ * The address of the buffer is returned in @obuf, the length of the buffer in
+ * @olen. You are responsible for freeing the buffer with g_free() when you
+ * are done with it. The buffer is freed for you on error.
+ *
  * See also: vips_webpsave().
  *
  * Returns: 0 on success, -1 on error.
--- a/libvips/foreign/foreign.c
+++ b/libvips/foreign/foreign.c
@@ -2460,6 +2460,7 @@ vips_webpload_buffer( void *buf, size_t len, VipsImage **out, ... )
  * @Q: quality factor
  * @lossless: enables lossless compression
  * @preset: #VipsForeignWebpPreset choose lossy compression preset
+ * @smart_subsample: enables high quality chroma subsampling
  *
  * See also: vips_webpload(), vips_image_write_to_file().
  *
@@ -2490,6 +2491,7 @@ vips_webpsave( VipsImage *in, const char *filename, ... )
  * @Q: JPEG quality factor
  * @lossless: enables lossless compression
  * @preset: #VipsForeignWebpPreset choose lossy compression preset
+ * @smart_subsample: enables high quality chroma subsampling
  *
  * See also: vips_webpsave().
  *
--- a/libvips/foreign/foreign.c
+++ b/libvips/foreign/foreign.c
@@ -2462,6 +2462,7 @@ vips_webpload_buffer( void *buf, size_t len, VipsImage **out, ... )
  * @preset: #VipsForeignWebpPreset choose lossy compression preset
  * @smart_subsample: enables high quality chroma subsampling
  * @near_lossless: use preprocessing in lossless mode (controlled by Q)
+ * @alpha_q: set alpha quality in lossless mode
  *
  * See also: vips_webpload(), vips_image_write_to_file().
  *
@@ -2494,6 +2495,7 @@ vips_webpsave( VipsImage *in, const char *filename, ... )
  * @preset: #VipsForeignWebpPreset choose lossy compression preset
  * @smart_subsample: enables high quality chroma subsampling
  * @near_lossless: use preprocessing in lossless mode (controlled by Q)
+ * @alpha_q: set alpha quality in lossless mode
  *
  * See also: vips_webpsave().
  *
--- a/libvips/foreign/foreign.c
+++ b/libvips/foreign/foreign.c
@@ -2461,6 +2461,7 @@ vips_webpload_buffer( void *buf, size_t len, VipsImage **out, ... )
  * @lossless: enables lossless compression
  * @preset: #VipsForeignWebpPreset choose lossy compression preset
  * @smart_subsample: enables high quality chroma subsampling
+ * @near_lossless: use preprocessing in lossless mode (controlled by Q)
  *
  * See also: vips_webpload(), vips_image_write_to_file().
  *
@@ -2492,6 +2493,7 @@ vips_webpsave( VipsImage *in, const char *filename, ... )
  * @lossless: enables lossless compression
  * @preset: #VipsForeignWebpPreset choose lossy compression preset
  * @smart_subsample: enables high quality chroma subsampling
+ * @near_lossless: use preprocessing in lossless mode (controlled by Q)
  *
  * See also: vips_webpsave().
  *
--- a/libvips/foreign/vips2webp.c
+++ b/libvips/foreign/vips2webp.c
@@ -85,7 +85,7 @@ get_preset( VipsForeignWebpPreset preset )
 static int
 write_webp( WebPPicture *pic, VipsImage *in,
 	int Q, gboolean lossless, VipsForeignWebpPreset preset,
-	gboolean smart_subsample )
+	gboolean smart_subsample, gboolean near_lossless )
 {
 	VipsImage *memory;
 	WebPConfig config;
@@ -97,9 +97,11 @@ write_webp( WebPPicture *pic, VipsImage *in,
 		return( -1 );
 	}
 
-	config.lossless = lossless;
+	config.lossless = lossless || near_lossless;
 	if( smart_subsample )
 		config.preprocessing |= 4;
+	if( near_lossless )
+		config.near_lossless = Q;
 
 	if( !WebPValidateConfig(&config) ) {
 		vips_error( "vips2webp",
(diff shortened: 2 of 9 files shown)


felixbuenemann:
Here's a comparison of PNG24 (123k), WebP (15k) and WebP with smart downsamling (16k; LTR):
compare webp subsampling

Please click image for full size. The default WebP in the center looks mushy especially in the pocket area.

felixbuenemann:
It seems `config.near_lossless` is not present in the webp version on Travis. I should add a version check.

felixbuenemann:
It seems even before my changes the webpload already depended on libwebp-0.1.3, so I've added that as a requirement to configure.

@jcupitt Should be good for review now…

libvips/foreign/vips2webp.c
```diff
@@ -82,6 +83,73 @@ get_preset( VipsForeignWebpPreset preset )
 	return( -1 );
 }
 
+#if WEBP_ENCODER_ABI_VERSION < 0x0100
+typedef struct {
+	uint8_t *mem;
+	size_t size;
+	size_t max_size;
+	uint32_t pad[1];
```
jcupitt: Do we need `pad`? I don't see another reference to it.

libvips/foreign/vips2webp.c
```diff
@@ -82,6 +83,73 @@ get_preset( VipsForeignWebpPreset preset )
...
+	writer->max_size = 0;
+}
+
+static int WebPMemoryWrite( const uint8_t *data, size_t data_size,
+	const WebPPicture *picture ) {
+	WebPMemoryWriter* const w = (WebPMemoryWriter*) picture->custom_ptr;
+	size_t next_size;
+
+	if (w == NULL) {
+		return 1;
```
jcupitt: Should this be 0? It's probably an error.

libvips/foreign/vips2webp.c
```diff
@@ -82,6 +83,73 @@ get_preset( VipsForeignWebpPreset preset )
...
+	next_size = w->size + data_size;
+
+	if( next_size > w->max_size ) {
+		uint8_t *new_mem;
+		size_t next_max_size = w->max_size * 2;
+
+		if( next_max_size < next_size )
+			next_max_size = next_size;
+		if( next_max_size < 8192 )
+			next_max_size = 8192;
```
jcupitt: You could do:

```
const size_t next_max_size = 
    VIPS_MIN( 8192, VIPS_MIN( next_size, w->max_size * 2 ) ); 
```  

It gets rid of a non-const local, always a good idea.

libvips/foreign/vips2webp.c
```diff
@@ -82,6 +83,73 @@ get_preset( VipsForeignWebpPreset preset )
...
+
+		new_mem = (uint8_t*) malloc( next_max_size );
```
jcupitt: We need to use `g_malloc()` and `g_free()`. 

GLib has it's own malloc family, since some platforms (eg. Windows) have an extremely slow default malloc implementation. Sadly this means you have to be very careful which bits of memory you `free()` and which you `g_free()`. To avoid confusion, vips is supposed to use `g_free()` everywhere. 

In this case, you want `g_try_malloc()`, the one which can fail. 

This was probably a bug in the previous `vips_webpsave_buffer()`.

libvips/foreign/foreign.c
```diff
@@ -2458,6 +2458,11 @@ vips_webpload_buffer( void *buf, size_t len, VipsImage **out, ... )
  * Optional arguments:
  *
  * @Q: quality factor
+ * @lossless: enables lossless compression
+ * @preset: #VipsForeignWebpPreset choose lossy compression preset
+ * @smart_subsample: enables high quality chroma subsampling
+ * @near_lossless: use preprocessing in lossless mode (controlled by Q)
+ * @alphaQ: set alpha quality in lossless mode
  *
  * See also: vips_webpload(), vips_image_write_to_file().
```
jcupitt: Could we add a note about the need to `g_free()` thre result? See `vips_pngsave_buffer()`.

jcupitt:
Hi @felixbuenemann, thanks for this, the improvement is very obvious. Nice!

Reading it through it looks good, I just noticed a few small things:

* vips uses K&R layout, so there should be a newline between the function's return type and the function name, and between the `)` at the end of the function prototype and the first `{`, eg. `WebPMemoryWrite()` is missing them, and there are a few others. It also brackets the results of `return()`, to make it match `if()`.

* The `WebP` prefix should be reserved for functions defined in the webp library or debugging becomes very confusing. `WebPMemoryWrite()` should be something like `VipsWebPMemoryWrite()`.

* Actually, vips uses Gnu capitalization, so `this_is_a_function()`, `ThisIsAType` and `THIS_IS_A_MACRO`. Memory write probably be `vips_webp_memory_write()`.

I've added a few lines notes too.

* Use `g_malloc()` etc.

* Docs should note caller's need to `g_free()` the result.

* Small logic cleanup.

libvips/foreign/vips2webp.c
felixbuenemann: The whole code in this block is copied from libjpeg 0.1.99, it has only been changed to use the same formatting style as the rest of the file. This is done because 0.1.3 only has a private memory writer with a different data struct.

libvips/foreign/vips2webp.c
felixbuenemann: This is copied from libwebp-0.1.99, but is essentially the same in 0.5.0. I chose the older version because it does not use libwebp memory allocation functions that are missing in 0.1.3.

libvips/foreign/vips2webp.c
felixbuenemann: Do you mind replacing the default memory writer from libwebp with a vips specific one? Then we could use that for all supported libwebp versions. You should look at 0.5.0 for reference, it has a newer version that uses uint64_t internally instead of size_t.

libvips/foreign/foreign.c
felixbuenemann: Sure.

felixbuenemann:
As mentioned above the WebP memory writer code is simply backported from 0.1.99 for 0.1.3, which is why the original naming was used. There is nothing stopping us from using our own version with different naming conventions snd memory allocators. I just don't know enough about the underlying logic, but I can try to update it with your help.

It should also be possible to use a custom file writer, that flushes out to a file. I'm not sure that's much of a gain, because a lot more memory is allocated for uncompressed image data.

Additionally it should be possible to just point to an existing buffer in the ARGB case using pic.argb and pic.argb_stride in pixels, skipping the import function, as long as the underlying buffer uses packed argb 32-bit uint instead of uchar. Not sure if vips can provide that.

My C and low level programming skills are flaky at best, so I just wrote the code in write file and buffer to allow switching out the writer and using different implementations for buffer and file. I'm sure it will be much easier for you to optimize that code.

libvips/foreign/vips2webp.c
jcupitt: Sure, let's have our own. vips2jpeg.c has quite a nice one at line 1204: it builds a linked list of 10kb memory blocks, then on end of write does a single malloc of the correct size and copies the list to that.

jcupitt:
vips uses RGBA bytes, so we'd need to swap, I think.

libvips/foreign/vips2webp.c
felixbuenemann: Btw. the current version of this code uses uint64_t and later downcasts to size_t. It has a comment that this is safe due to WebPSafeMalloc. Would this be safe with g_malloc as well?

See: https://github.com/webmproject/libwebp/blob/master/src/enc/picture.c#L208

felixbuenemann:
Don't take my word on byte order in the packed version. I will dig up the code from webp.

libvips/foreign/vips2webp.c
jcupitt: Yes, `size_t` is fine for us. webp does not support large images anyway, so it doesn't matter .

felixbuenemann:
The various import functions for RGB use VP8PackARGB or VP8PackRGB, so you're probably right and conversion is always required.

felixbuenemann:
I'll adress the smaller issues and update the writer to use proper naming conventions and memory allocators and then you can later swap it out with a better implementation.

libvips/foreign/vips2webp.c
felixbuenemann: It looks strange indeed. My guess is that internally the memory writer is used in some places where they just want to ignore, similar to the DummyWriter, if the pointer is not set. We can certainly change it for our implementation.

libvips/foreign/vips2webp.c
felixbuenemann: This needs to be max not min:

```c
const size_t next_max_size =
	VIPS_MAX( 8192, VIPS_MAX( next_size, w->max_size * 2 ) );
```

felixbuenemann:
@jcupitt I've addressed your feedback.

What do you think about the `alphaQ` option naming? I wanted it to be in line with `Q`, but it could get confusing if we later add `alpha_filtering` or `alpha_compression`. The opton is named `alpha_quality` in libwebp and the switch for `cwebp` is `-alpha_q`.

Should I switch from `alphaQ` to `alpha_q`?

felixbuenemann:
Btw. I should look into adding tests for writing WebP, I noticed that `make check` worked locally, even when saving webp always failed with an error.

felixbuenemann:
@jcupitt I've added two web format tests for `make check` using lossy and lossless compression.

I also discovered that there is a more comprehensive python test suite:

* Is there a good reason those tests are *not* run on Travis CI?
* How do you tell the python tests to use the libvips and gir from the source dir instead of the system?

jcupitt:
Yes, it'd be great if the py test suite could be run on Travis. Last time I tried, the linux on Travis didn't have a working gobject-introspection, so all that would have to be built as well. 

There's another problem: GOI sadly has no way to give a path for the overrides file, so Vips.py must be copied to /usr before it'll work. You can point it at the typelib by setting `GI_TYPELIB_PATH`.

jcupitt:
Re. `alphaQ`, I agree it looks a little odd. `alpha_q` would probably be better.

felixbuenemann:
I'll rename the alphaQ parameter and look into getting the python tests running in ci. If the packages on precise are too old we could use `dist: trusty` or even better configure matrix builds on os x, precise and trusty.

felixbuenemann:
OK, renamed alpha_q and straightened up the commits.

Do you want me to squash into a single commit before merging?

felixbuenemann:
@jcupitt Is there anything left to do here?

jcupitt:
Ooop! Sorry Felix, I've been distracted.

Thank you very much for adding this feature!

felixbuenemann:
No problem. Have a nice weekend!

jcupitt:
I tried to add some more doc comments on `vips_webpsave()`, not sure how helpful they are.

jcupitt:
I forgot to say, your work is in master and will be in supported in 8.4. It seemed like too large a change to add in a micro release, I hope that's OK.

felixbuenemann:
Fine with me.

jcupitt:
I revised the doc comment again, does it sound OK?

https://github.com/jcupitt/libvips/blob/master/libvips/foreign/foreign.c#L2467

felixbuenemann:
Sounds fine.

PR #434, opened 2016-05-07
Add mozjpeg quant_table option to jpegsave

felixbuenemann:
This allows to choose from 8 predefined quantization tables when libvips is compiled against mozjpeg.

Tuning the quantization table helps to improve file size in the higher quality range and avoid artefacts in the lower range.

One example would be to use table 3 at quality 75, where it consistently compresses better than any of the other tables, but to use table 2 for a "retina" image compressed at qualities below 50, where it performs better at keeping color accurate and avoiding banding in flat areas and gradients.

I have also added some advice on which table to choose, based on limited personal observations, which is probably much better than blindly having to choose a numerical index ;-)


jcupitt:
Nice! Thank you!

PR #436, opened 2016-05-08
Fix configure check for giflib4

### LINKED ISSUE #435: Configure check for giflib4 broken
The configure check for giflib4 was broken in commit d11a7960b7595636f59b1c802c5dca3f735a7410  while adding giflib5 support (see #407).

Both giflib 4.x and 5.x have `EGifSetGifVersion` but with different signatures:

```c
// giflib4
void EGifSetGifVersion(const char *Version);
// giflib5
void EGifSetGifVersion(GifFileType *GifFile, const bool gif89);
```

The configure check tries to compile `EGifSetGifVersion(0,0)` which fails for giflib4 (tested with 4.2.3):

```
conftest.c:121:21: error: too many arguments to function call, expected single argument 'Version', have 2 arguments
EGifSetGifVersion(0,0)
~~~~~~~~~~~~~~~~~   ^
/usr/local/include/gif_lib.h:118:1: note: 'EGifSetGifVersion' declared here
void EGifSetGifVersion(const char *Version);
^
1 error generated.
```

felixbuenemann:
This uses a combination of the old giflib4 and the new giflib5 checks.

Fixes #435.

### DIFF
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -631,14 +631,26 @@ if test "$GIFLIB_LIBS" = ""; then
   INCLUDES="$GIFLIB_INCLUDES $INCLUDES"
 
   # Try the standard search path first
-  AC_TRY_LINK([#include <gif_lib.h>],[EGifSetGifVersion(0,0)], [
+  AC_TRY_LINK([#include <gif_lib.h>],[
+    #if defined(GIFLIB_MAJOR) && GIFLIB_MAJOR > 4
+    EGifSetGifVersion(0,0);
+    #else
+    GifLastError();
+    #endif
+  ], [
     GIFLIB_LIBS="-lgif"
   ], [
     # giflib is not in the standard search path, try $prefix
 
     LIBS="-L${prefix}/lib $LIBS"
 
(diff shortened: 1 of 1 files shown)


acinclude.m4
```diff
@@ -631,14 +631,26 @@ if test "$GIFLIB_LIBS" = ""; then
...
   # Try the standard search path first
-  AC_TRY_LINK([#include <gif_lib.h>],[EGifSetGifVersion(0,0)], [
+  AC_TRY_LINK([#include <gif_lib.h>],[
+    #if defined(GIFLIB_MAJOR) && GIFLIB_MAJOR > 4
+    EGifSetGifVersion(0,0);
+    #else
+    GifLastError();
+    #endif
+  ], [
     GIFLIB_LIBS="-lgif"
```
jcupitt: How about calling `DGifSlurp(0)` instead? This should compile on giflib4 and giflib5, and we only link, we don't execute, so it doesn't matter that it won't run.

jcupitt:
I'll change to using slurp in a sec.

PR #438, opened 2016-05-08
Run python tests on travis and setup build matrix

felixbuenemann:
This is still a work in progress, but I've succeeded to get the python tests running on Travis CI.

There are still some TODOs and things I'd like to discuss:

- [x] Add a Makefile target to run the python tests, eg. `make test`
- [x] Decide wether `make check` should run in addition to python tests and in what order
- [x] Test if python tests can run on precise as well
- [x] Decide on the test matrix: run only on trusty or precise and trusty
- [x] Check minimum required libpoppler-glib version (probably another issue / PR)
- [x] Ensure we are using the system python (currently hardcoded into .travis.yml)
- [x] Do we need to run `make install` at all?
- [x] Check nothing broke by moving Vips.py (gi override) to a new directory


felixbuenemann:
@jcupitt Could you help me with adding the make target to the generated `test/Makefile`?

I'm not familiar with automake, so I don't know where to add it.

I basically need to do this:

```sh
# inside test directory
# on Linux
GI_TYPELIB_PATH=../libvips LD_LIBRARY_PATH=../libvips/.libs python -m unittest -v test_all
# on MacOS X
GI_TYPELIB_PATH=../libvips DYLD_LIBRARY_PATH=../libvips/.libs python -m unittest -v test_all
```

Additionally I need to ensure I'm using /usr/bin/python on Travis CI and not the one found in the path, because it does not contain the gi / pygobject packages installed through `apt-get`.

felixbuenemann:
There's probably a bit of explanation needed for the changes in b1ba4cb60f6279ec87b5d3334119d2bbedcb4362:

In order to load gi overrides from a custom path, we need to add the following code to `__init__.py` of the gi and gi/overrides directories in that path to extend the search path with pkgutil:

```python
from pkgutil import extend_path
__path__ = extend_path(__path__, __name__)
```

This is the reason why I moved the file `python/Vips.py` to `python/packages/gi/overrides/Vips.py` so I can extend the PYTHONPATH with the `python/packages` directory and load the overrides from there.
I'm currently extending the python sys path from inside test_all.py, so I don't need to look up the existing paths, which would be required when setting it through the `PYTHONPATH` env var (the default overrides directory needs to come first in the sys path or it won't work).

jcupitt:
Hey, nice! I'll try this out today.

felixbuenemann:
Btw. I tried to get python vips8 running on precise, but had no luck. It would require both a newer glib and a newer pygobject. So my plan would be to only run `make check` on precise and run the python tests on trusty. I should be able to do that through the build matrix settings.

jcupitt:
Also, I wanted to thank you for doing all this work fixing up vips, it's great to have all this stuff sorted out.

lovell:
Quick thought: the use of a tool like Docker might help here. Defining and building a Linux container with the correct dependencies required by the tests would also make it easier to run them locally. Happy to help with this.

felixbuenemann:
Yes, we could prepare docker containers with the packages preinstalled and use them to run the tests.
I'd have to check how much time is spent installing packages and see if it outweighs the cost of pulling down the docker image.

Running the tests locally could also be made easier by providing a shell script that installs the required packages on let's say Mac OS X Homebrew, Ubuntu/Debian and Redhat-based distributions.

@lovell The packages on Trusty are fine, but it's still interesting to see if vips still works with the old package versions on Precise, which is still used on many servers. If you install newer packages on precise, you're not really testing that.

lovell:
Perhaps the Dockerfile(s) could live within libvips, a bit like with the [Windows binary builder](https://github.com/jcupitt/build-win64).

For example, if libvips had a `test/containers/precise/Dockerfile` file that defined a `ubuntu:precise` based image with all test dependencies, then a new `./run-tests.sh` that looks a bit like the following could both build and run the container, e.g.:

```sh
docker build -t libvips-test-precise test/containers/precise
docker run -v $PWD:/libvips libvips-test-precise sh -c "cd /libvips; make clean tests"
```

This would then allow us (and Travis) to attempt to compile+test libvips against the various dependencies/versions as provided by the more popular Linux-based OS flavours/versions.

jcupitt:
Oh clever, I like the `__init__.py` trick. I'll see if I can get it working here.

felixbuenemann:
I've got a working multi OS Travis CI config in my [travis-multi-env](https://github.com/felixbuenemann/libvips/tree/travis-multi-env) branch.

It currently runs:

* precise: make check
* trusty: make check, python tests
* el capitan: make check, python tests

I nearly fell asleep waiting for the OS X tests to complete tough (~25 min vs. 6 min for ubuntu). So the workers are pretty slow. If we decide to add OS X to the build matrix, we should probably make it an allowed failure and add `fast_finish: true` to the matrix to speed up build feedback in pull requests.

OS X with Homebrew is an interesting CI target, because we're pretty much guaranteed to build against the latest package versions. It also means package versions are constantly changing, so another reason to mark it as allowed failure.

There's also some tricks that could speed up the osx build, like caching /usr/local/Cellar between builds, so only new versions need to be installed. I'll also try enabling ccache for all builds.

jcupitt:
I had a look at my `/usr/lib/python2.7/dist-packages/gi/overrides/__init__.py` and there seems to be stuff in there already for this. It has this near the top:

```python
# support overrides in different directories than our gi module
from pkgutil import extend_path
__path__ = extend_path(__path__, __name__)
```

I think that means we could just add the `Vips.py` area to `PYTHONPATH` and it would load, and we wouldn't need our own `__init__.py`. I'll experiment.

felixbuenemann:
You need it in both places.

jcupitt:
It seems to be working for my without the `__init__.py`, just setting `PYTHONPATH`. If I install git master vips to `/home/john/vips`, I see:

```
john@mm-jcupitt5 ~ $ ls /usr/lib/python2.7/dist-packages/gi/overrides/Vips.py
ls: cannot access '/usr/lib/python2.7/dist-packages/gi/overrides/Vips.py': No such file or directory
john@mm-jcupitt5 ~ $ ls ~/vips/lib/python2.7/site-packages
gi  vipsCC
john@mm-jcupitt5 ~ $ ls ~/vips/lib/python2.7/site-packages/gi
overrides
john@mm-jcupitt5 ~ $ ls ~/vips/lib/python2.7/site-packages/gi/overrides/
Vips.py  Vips.pyc  Vips.pyo
john@mm-jcupitt5 ~ $ echo $PYTHONPATH 
/home/john/vips/lib/python2.7/site-packages
john@mm-jcupitt5 ~ $ python
Python 2.7.11+ (default, Apr 17 2016, 14:00:29) 
[GCC 5.3.1 20160413] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from gi.repository import Vips
__main__:1: PyGIWarning: Vips was imported without specifying a version first. Use gi.require_version('Vips', '8.0') before import to ensure that the right version gets loaded.
>>> a = Vips.Image.new_from_file("/data/john/pics/k2.jpg")
>>> 
```

I have no `Vips.py` in the main gi area, just the one in `~/vips`, no `__init__.py` in `~/vips` (though there is one in `/usr/lib/python2.7/dist-packages/gi/overrides/` and it seems to be working! 

I don't know when they added the extra stuff in their `__init__.py`, I'm certain this used to not work. This is Ubuntu 16.04 if that's a factor.

jcupitt:
Ooops, oh drat, I had `Vips.pyc` still in `/usr/lib/python2.7/dist-packages/gi/overrides`, that's why it was working. I'll investigate some more.

felixbuenemann:
:-) I had the same situation when I did `brew unlink vips` which left the `Vips.pyc` in place.

jcupitt:
This makes "make check" run the py tests:

https://github.com/jcupitt/libvips/commit/ffe27469ac5b58c5ec66fd5d7f332ab41287aa07

it seems to work. magick is failing for some reason. giflib is failing on 16.04 because of the uninit bug.

felixbuenemann:
@jcupitt I think it would be better if there where two different make targets eg. `make check` to run the cli tests and `make test` to run the python tests. This makes it easier to run the wanted tests for example in packaging workflows.

My idea would be that `make check` does some very rough testing through the vips cli, while the python tests do the in depth testing, checking if all the options and operations are working as expected.

We could then slim down `make check` a bit, so it only contains a few fast tests.

felixbuenemann:
Oh, I must've misread the commit logs. You made the python tests an additional part of make check and kept the existing tests.

felixbuenemann:
> it seems to work. magick is failing for some reason. giflib is failing on 16.04 because of the uninit bug.

The giflib and magick errors seem to be reladed. I had the same failures on xenial, so I created a giflib backport package of 5.1.4 and now both gif and magick tests are working fine.

You can add [ppa:louisinternet/giflib](https://launchpad.net/~louisinternet/+archive/ubuntu/giflib) and `apt-get upgrade libgif7 libgif-dev` to see if it fixes your problems as well. Alternatively you can just [grab the two .deb files](https://launchpad.net/~louisinternet/+archive/ubuntu/giflib/+build/9721529).

felixbuenemann:
@jcupitt I've rebased and merged your changes a couple of days ago.

**What direction would you prefer for the build matrix?**

We could go with containers as suggested by @lovell which allows testing against arbitrary linux distributions or with the VMs provided by Travis CI which currently means Precise, Trusty and various OS X / Xcode combinations (or a mix of both).

Build performance with ccache is pretty good for linux. OS X is sluggish. We could speed up both by making the gtk-doc requirement in bootstrap.sh optional through an env var (it causes a libxml2 compile from source for libxml2 on os x).

jcupitt:
That's great @felixbuenemann !

I don't really have an opinion. Using the Travis VMs seems to be a bit simpler, so I suppose I'd just do that unless there's a pressing need to test on other systems. 

The CLI tests on "make check" are very slow and are mostly useless, they date from before all the py tests were written. I think it would be a good idea to get rid of almost all of them, if we can. I was holding back because getting the py tests working on Travis was so annoying, but now (thank goodness) you've fixed that, we could drop them. 

Maybe we should do this in two phases: get the python tests running everywhere first, then do a second round where we cull the CLI tests and perhaps add a few more py ones.

felixbuenemann:
Getting the python tests running on Precise won't work unless you backport a LOT of gnome libraries at which point you're no longer testing if libvips works on precise out of the box. I've just build all the dependencies of libvips/pyvips8 for use on aws lambda with python this week.

First I've hacked my way around getting gobject-introspection and pygobject3 built against the Amazon Linux 2015.05 Glib 2.36.3 with 2.35.9 being the lowest workable version, by using older gobject-instrospection and pygobject3 versions and patching their configure.ac.

That worked out OK and then I thought: *Well I could build my own glib 2.43 and then I can use all the latests versions of all the important libs.* After that crazy idea I spent many many hours compiling vendored copies of all the intertwined gnome libraries and making sure they didn't link against any unwanted system libraries.

Given that Precise ships with glib 2.32.4 you're pretty much out of luck, unless you want to spent time making libvips and the overrides compatible with old gobject-introspection versions.

felixbuenemann:
@jcupitt We need to find a better way to check wether to install the `__init__.py` stubs to extend the pygobject overrides path.

Currently the files are installed uncodnitionally, but it should only happen, if the target path is *not* the main overrides path used by pygobject3, otherwise we overwrite the pygobject3 `__init__.py` files which contain a lot more than the stub code. The install should probably also check if the files already exist.

This also leads to conflicts in homebrew with the brewed python where both paths are the same:

```sh
$ brew switch vips DEV
Error: Could not symlink lib/python2.7/site-packages/gi/__init__.py
Target /usr/local/lib/python2.7/site-packages/gi/__init__.py
is a symlink belonging to pygobject3. You can unlink it:
  brew unlink pygobject3

To force the link and overwrite all conflicting files:
  brew link --overwrite vips
```

We can find the primary overrides path using python:

```sh
python -c 'import gi;print(gi._overridesdir)'
/usr/local/lib/python2.7/site-packages/gi/overrides
```

I also noted another issue with installation of the python files:

On Linux systems with multiarch support, files where installed in the wrong directory, eg. `$PREFIX/lib/python2.7/site-packages` instead of `$PREFIX/lib64/python2.7/site-packages`.

jcupitt:
Here's a patch to only install `__init__.py` if it's not there already.

https://github.com/jcupitt/libvips/commit/15fe0aea8af9a823ffaedd51a4fa0aad1bf2989d

Which is ugly and fragile, but probably less dangerous :(

jcupitt:
I also removed the `build_pylinks` stuff, it smells a bit fishy to me. For example, does `all-local` actually execute?

felixbuenemann:
I've thought about it a bit more, and I think we shouldn't install the `__init__.py` files at all. They're fine to be able to run the test suite from inside the source directory, but installing them seems very error prone.

Even if we install it, the user likely needs to set `GI_TYPELIB_PATH` and `PYTHON_PATH` with the proper order to use them.

What we could do is warn the user at the end of configure, if the prefix gi path differs from the selected python's main gi path.

jcupitt:
I added something to master to try to print some py install hints at the end of configure:

https://github.com/jcupitt/libvips/commit/8b592585d7fe2bfb87512e4f2166a4954a7aa38c

What do you think?

Yes, the `__init__.py` stuff makes me uncomfortable. I wish there was a better way to install this stuff.

felixbuenemann:
@jcupitt Your current approach breaks if both pygobject and pyvips8 are installed to a prefix that differs from the default python packages path:

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: No module named gi
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: No module named gi
The vips Python overrides file will install to
/var/task/lib/python2.7/site-packages/overrides/Vips.py, but your
system gi area seems to be .
You may need to copy this file, for example:

	cp /var/task/lib/python2.7/site-packages/overrides/Vips.* /overrides
```

There are two issues here:

1. We need to append the prefix pyexecdir when checking for gi, because it might also be installed to the same prefix.
2. The pyexec prefix must use the architecture dependent path (like lib64), vipsCC has .so files and the Vips.py override depends on pygobject, which in turn ships with .so files and thus goes in the architecture dependent path.

This can be fixed by replacing `sysconfig.get_python_lib(0, 0, prefix='$am_py_prefix')` with `sysconfig.get_python_lib(1, 0, prefix='$am_py_prefix')` ([see distutils docs]((https://docs.python.org/2.7/distutils/apiref.html?highlight=get_python_lib#distutils.sysconfig.get_python_lib)).

The python code should slow only be executed inside the `enable_pyvips8` conditional to avoid error when no python is available.

We should also probably do a small check for homebrew environment, because in this case the vips pyecec path witll be /usr/local/Cellar/... but it's symlinked to /usr/local/lib, so we're fine and don't need to display the warning.

I'll send a PR to fix this up.

felixbuenemann:
@jcupitt Btw. I gave you collaborator access to my fork, if you need to add anything to this branch.

felixbuenemann:
Meh, the python detection code comes from automake and it hardcodes the platform specific path to `lib` instead of using the `lib64` it got back from python:

http://git.savannah.gnu.org/cgit/automake.git/tree/m4/python.m4#n203

This leads to configure using the wrong exec prefix. If I comment out the above mentioned line, I get the proper exec prefix.

felixbuenemann:
@jcupitt I looked at pygobject, and they have a vendored `m4/python.m4` without the above mentioned problems. Do you have anything against adding that to vips as well?

felixbuenemann:
@jcupitt Do you plan on adding compatibility with older gobject-intropection / pygobject versions or should we go with just testing through the vips CLI on Precise and CLI + Python on Trusty / OS X?

jcupitt:
Let's do the CLI on precise and CLI + py on trusty. I don't think there's much point spending effort supporting very old versions.

felixbuenemann:
@jcupitt Could you post a diff of what needs to be changed in this tree so that the __init__.py files are no longer installed? Or is is as simple as reverting the "add Makefile.am to install new py init" commit?

jcupitt:
Hi @felixbuenemann, you just need to remove the `install-data-hook` rule from `libvips/python/packages/gi/Makefile.am` and from `gi/overrides/Makefile.am`.

jcupitt:
Here's a patch that does that, tested and seems to work OK. 

https://github.com/jcupitt/libvips/commit/d2dcaa1bc440ae4a6e54134d0706b2e3d73e1c79

felixbuenemann:
@jcupitt Thanks, sorry for iterating so slowly on this. I've been pretty busy lately.

felixbuenemann:
@jcupitt OK, configured the matrix build and made the build output a lot less verbose. Should be good to merge from my side.

The Mac OS X build uses some hacks to avoid recompiling libxml2 with python support on each build, but is still very slow, due to the sluggish OS X workers.

I have marked OS X builds as allowed failure and enabled fast feedback mode, so PR feedback is given once the must have builds are finished.

We could really speed up CI builds if the gtk-doc dependency would be optional, because it pulls in a large amount of big dependencies like docbook.

Would that be hard to implement? (It's currently required by `bootstrap.sh`)

jcupitt:
Sorry, I think it's very hard to build the configure system without gtk-doc. It becomes standalone when you `make dist`, but it is necessary for bootstrap.

felixbuenemann:
Yeah, I had a quick go at making it optional 2 days ago and came to a similar conclusion.
Maybe it's possible to pull down a standalone version of `gtkdocize` without all the other stuff we only need when actually generating docs.

felixbuenemann:
Oh, for Mac OS X where this has the biggest imapct I can just use `brew install --ignore-dependencies gtk-doc`. I'll update the `.travis.yml` to use that and remove the libmxl2 caching hacks.

felixbuenemann:
OK, the OS X build now installs gtk-doc without dependencies, which is fine, because we only need three files to run `./bootstrap.sh` (`gtkdocize`, `gtk-doc.make` and `gtk-doc.m4`).

felixbuenemann:
@jcupitt Let me know if you have questions or want something changed. This is good from my side.

jcupitt:
Great! Let's turn it on. It's going to be very useful to have the test suite running on every commit, this is a good improvement.

PR #439, opened 2016-05-08
Fix poppler version requirement, add loader check

felixbuenemann:
This fixes the poppler-glib version requirement to version 0.16, which is the lowest version that compiles and works properly with pdfload.

Also added a check for pdfload and enabled testing on Travis CI.


felixbuenemann:
Oops, there's an error in test_formats.sh, it's not comparing the PNG, will fix.

felixbuenemann:
Compile is failing because pdfload.c is referencing `vips__cairo2rgba()` which is defined in `svgload.c`. I didn't see that problem on my local precise VM, because I also lowered the librsvg requirement.

I'll look into moving that code into a shared file.

felixbuenemann:
@jcupitt I've added a commit that moves `vips__cairo2rgba()` into a shared `cairo.c` file. Feel free to ignore that commit and do it differently on master. Just ping me and I'll rebase to get rid of the change.

felixbuenemann:
Meh, seems the antialiasing differs between poppler versions, so the fine lines in the reschart cause too much difference. I'll try with the simpler cmyk.pdf.

felixbuenemann:
OK, I'm now using a minimal blank page pdf for testing, which reduces file size and generates the same image in all supported poppler versions (unsurprisingly ;-).

Sorry for the noise.

jcupitt:
Looks good, thanks @felixbuenemann!

I might move `vips__cairo2rgba()` back into `svload.c`, just outside the `HAVE_RSVG` ifdef, but i can do that on master, so I've just merged the whole thing.

jcupitt:
Actually, you're right, let's have `cairo.c`. Thanks again.

PR #441, opened 2016-05-08
Fix threading and thumbnail tests without make install

felixbuenemann:
Both scripts were not using the $vips variables required to use vips from the source directory instead of from the system.


jcupitt:
Ooops! Thank you @felixbuenemann

PR #442, opened 2016-05-09
Test system webp on Travis CI and fix alpha_q default value

felixbuenemann:
This switches from the current backport ppa to testing against the system provides libwebp-0.1.3 on Travis CI. As a consequence the lossless test in `test_formats.sh` was dropped, because it's not supported on that old version.

I've also fixed the default value for webpsave `alpha_q` to be 100 instead of 0. I had assumed that would be already handled by defining the `VIPS_ARG_INT`, but that doesn't seem to be the case.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: cpp
 before_install:
-  - sudo add-apt-repository ppa:lyrasis/precise-backports -y
   - sudo apt-get update -qq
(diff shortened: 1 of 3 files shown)


jcupitt:
Yes, I get confused by that, the default value is only there for UIs, you need something in init too.

jcupitt:
Looked over, seems great. Thanks!

PR #443, opened 2016-05-09
Remove jpegsave default quant_table debug print

felixbuenemann:
This was only ut in for debugging which quant_table is used in the `JCP_FASTEST` vs. `JCP_MAX_COMPRESSION` profile and should've never been committed.

This is a fixup for #434.


jcupitt:
Ooop, done.

felixbuenemann:
Thanks, John.

Issue #334, opened 2015-10-03
Compilation error in vimagemodule (PyBuffer_FromMemory, char** to wchar_t**)


PolarNick239:
Compilation error on Ubuntu 14.04, Python 3.4 (virtualenv).
May be I am doing something wrong? Can you help please?

```
vimagemodule.cxx: In function 'PyObject* _wrap_VImage_tobuffer(PyObject*, PyObject*)':
vimagemodule.cxx:31987:70: error: 'PyBuffer_FromMemory' was not declared in this scope
     resultobj = PyBuffer_FromMemory ((&result)->data, (&result)->size);
                                                                      ^
vimagemodule.cxx: In function 'PyObject* PyInit_vimagemodule()':
vimagemodule.cxx:33684:42: error: cannot convert 'char**' to 'wchar_t**' for argument '2' to 'void PySys_SetArgv(int, wchar_t**)'
     PySys_SetArgv (args->argc, args->argv);
```

May be configure output is important:
```
* build options
native win32:                           no
native OS X:                            no
open files in binary mode:              no
enable debug:                           no
build deprecated components:            yes
build docs with gtkdoc:                 no
gobject introspection:                  yes
build vips7 Python binding:             yes
install vips8 Python overrides:         yes
  (requires pygobject-3.12.0 or later)

* optional dependencies
use fftw3 for FFT:                      yes
Magick package:                         
file import with libMagick:             no
accelerate loops with orc:              no
  (requires orc-0.4.11 or later)
ICC profile support with lcms:          no
file import with OpenEXR:               yes
file import with OpenSlide:             no
  (requires openslide-3.3.0 or later)
file import with matio:                 no
file import with cfitsio:               no
file import/export with libwebp:        no
text rendering with pangoft2:           yes
file import/export with libpng:         yes (pkg-config libpng >= 1.2.9)
  (requires libpng-1.2.9 or later)
file import/export with libtiff:        yes (pkg-config libtiff-4)
file import/export with libjpeg:        yes
image pyramid export:                   yes
  (requires libgsf-1 1.14.27 or later)
use libexif to load/save JPEG metadata: yes

```

jcupitt:
Hi, sorry for the delay, I missed this issue.

That error is in the old SWIG-generated vips7 Python binding. I'm not sure if it works with python3, I guess not. 

I would disable the vips7 Python binding (`--without-python`), you'll still get the much better and not deprecated vips8 Python binding, which is tested and works with python3. 

[...]

PR #448, opened 2016-05-18
Fix Python and GI path detection problems

felixbuenemann:
* Vendor the pygobject `m4/python.m4` to fix wrong multiarch exec paths
* Fix several problems with gi path warnings
* Disable vips7 SWIG bindings for python3 (see #334)
* Warn if the gi module could not be loaded
* Fix `VIPS_LIBDIR` being `"NONE"` without explicit `--prefix`

For details see #438 and the individual commit messages.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,7 +48,8 @@ libtool
 libvips/include/vips/version.h
 fred
 ltmain.sh
-m4/
+m4/*
+!m4/python.m4
 missing
 mkinstalldirs
 po/Makefile.in.in
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -5,7 +5,7 @@
 # a bunch of cleaning up ... make certain everything will be regenerated
 rm -f Makefile Makefile.in aclocal.m4 
 rm -rf autom4te.cache
-rm -f m4/*
+find m4 ! -name python.m4 -type f -name '*.m4' -delete
 rm -f config.* configure depcomp
 rm -f install-sh intltool-* libtool ltmain.sh missing mkinstalldirs
 rm -f stamp-* vipsCC-7.19.pc vips-7.19.spec vips-7.19.pc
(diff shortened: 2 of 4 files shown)


felixbuenemann:
Tested on Amazon Linux 2015.09, Precise, Trusty, Xenial, OS X 10.11 w/Homebrew.

Btw. warning the user that we know there is a problem still seems a bit clunky to me. I'll explore the idea of instead installing symlinks to the two `__init__.py` from the system gi path to the prefix path. That should be a lot less error prone that the stub approach.

jcupitt:
That all looks great, nice job @felixbuenemann

jcupitt:
I found a small problem: `acinclude.m4` had a version of `AM_CHECK_PYTHON_HEADERS` as well, I don't know which version was being used. Anyway, it was confusing.

I've broken the vips `acinclude.m4` up into a series of separate tests (`tiff.m4` etc) and put them, with your `python.m4` (and not including the extra headers test) into a new directory called `m4-safe`. bootstrap now removes all of m4/ and copies everything fresh of m4-safe. 

It should be a little simpler to maintain anyway.

felixbuenemann:
I like the idea of splitting up a include.m4. 

Do we really need another directory to achieve that? We could just delete the hard coded list of m4 files from the m4 directory and blacklist them in gitignore…

jcupitt:
The separate directory is a bit cleaner, I think. We can drop more .m4 files in there and we won't need to change .gitignore, or that `find` in bootstrap.

felixbuenemann:
That's why I was suggesting to switch from whitelisting to blacklisting. I think the list of copied m4 files has been pretty static.

felixbuenemann:
Btw. I think we don't need to copy those files mentioned by glib-gettextize at all. They are already in the aclocal search paths, so they will automatically be picked up by aclocal. We would only need to copy them, if they were outside the ACDIRS.

So I think we can save ourselves a lot of throuble by just ignoring the hint from `glib-gettextize` and maintain a single `m4` directory with custom macros.

This works fine on my mac, I'll have to verify on a few different linux versions to check if they behave differently.

jcupitt:
I agree, let's try that. I've removed m4-safe on master.

felixbuenemann:
Btw. I usually test such stuff on a feature branch, push to ci, see if everything works, fixup, rebase and repeat. Keeps the git history clean of all those oops and argh commits ;-)

jcupitt:
Hehe yes, I should have done that. Overconfidence :(

PR #451, opened 2016-05-23
Add buffer write support for HDR Radiance

unusual-thoughts:
As `dzsave` only supports buffer output, I added `radsave_buffer` functionality and split the class into file and buffer subclasses as in `pngsave.c`


jcupitt:
That's a great looking patch, thank you very much @unusual-thoughts !

I suppose there's one more thing: the test suite could exercise rad buffer write as well. It would go in here:

https://github.com/jcupitt/libvips/blob/master/test/test_foreign.py#L451

You could write the buf to a temp file and try reloading it I guess.

PR #455, opened 2016-05-26
Use printf instead of non-portable echo -n in bootstrap.sh

felixbuenemann:
Not all versions of echo support `-n` so using it is not portable. One example where this does not work is bash in sh mode on Mac OS X. Replaced with `printf` which the BSD manpage suggest to use for portability.


jcupitt:
Ooop, I'd forgotten, thanks Felix.

PR #456, opened 2016-05-26
Support older librsvg2 / Add Travis checks for svgload / gifload

felixbuenemann:
* Lower librsvg version requirement to 2.34
* Test svgload in make check and on Travis CI
* Test gifload in make check and on Travis CI


felixbuenemann:
Btw. it's possible that older versions of librsvg2 would be compatible, but because 2.34 is the first version to properly support gtk3 it was the first version I could compile on my mac with otherwise up-to-date versions. Given that both Precise and Wheezy ship with 2.36 this should not be an issue.

jcupitt:
All looks good, nice work!

Issue #352, opened 2015-11-22
Skip fully black or white tiles when creating Deep Zoom pyramids?


perog:
I have lots of large mirax (mrxs) slides (120000x260000pixels on level 0) with a lot of white areas between the samples on the slide. Is it possible to skip these tiles when creating the stack and by that lower the time needed to create these stacks as well as the total stack size that becomes shrunk by 80% (2.6GB to 0.5GB)? I use the stacks with openseadragon and missing tiles are handled there instead.

The vips command I run is 'vips dzsave example.mrxs[autocrop] example' and it takes over 30 minutes to create the stack. Is it expected to take such long time for these sizes? The server I run on is a quite new multicore linux server, not entirely sure about the specs though... (40 *
Intel(R) Xeon(R) CPU E5-2650 v3 @ 2.30GHz according to 'less /proc/cpuinfo') Could you advice me if there is some speed optimiziation that I can do?

jcupitt:
It could skip blank tiles, but I don't think it would speed it up much. I believe google maps layout allows blank tiles, but I'm not sure if zoomify or deepzoom permit this. You're right, it would at least save some disk space. 

120k x 260k is large, I'd expect it to take a while. With 256x256 tiles, that's over 600,000 separate jpeg files, ouch. Even if it takes 30 minutes, that still means it's generating 350 jpeg files every second. 

You could turn the tile size up, that might speed it up a bit. An SSD would help too, probably. You could also experiment with zip output, it can be faster if your disk is very slow. 

Another option is to convert on the fly. So leave your mrxs files on disc, and serve tiles from them dynamically as clients request them. For small numbers of users, it's not too bad. I made a small on-the-fly server:

https://github.com/jcupitt/tilesrv

jcupitt:
I had a quick hack:

https://github.com/jcupitt/libvips/tree/no-blank-tiles

It seems to work. In google maps mode, it now skips tiles which are equal to the background colour. I had some code in there earlier to do this but removed it as it was a lot slower. This version seems to be fast enough that you can probably leave it on all the time. Before:

```
$ time vips dzsave CMU-1.svs x --layout google --background "243 243 243 255"
real	1m1.940s
user	2m15.004s
sys	0m37.092s
```

After:

```
$ time vips dzsave CMU-1.svs x --layout google --background "243 243 243 255"
strip_work: skipping blank tile 176 x 67
strip_work: skipping blank tile 21 x 112
real	1m3.503s
user	2m16.012s
[...]

PR #461, opened 2016-06-03
Vips dzsave zip compression support

felixbuenemann:
This adds support for enabling ZIP compression and setting the deflate compression level when using dzsave with the zip container type.

The general consensus is, that there's no use in compressing already compressed image types like JPEG. However this is not true if the file itself is not well compressible by JPEG, as is the case with (mostly) blank tiles.

In my tests I am saving between 15-25% on the total ZIP size depending on the JPEG compression level used and how many solid areas the source image contains. Also see my comments in #352.

Example using the [Blue Marble Next Generation](http://visibleearth.nasa.gov/view.php?id=73909) image, which contains some nearly solid areas like oceans and Antarctica:

```sh
for i in `seq 0 9`; do
  echo -n "$i "
  time vips dzsave world-86400x43200.tif world-86400x43200-dz900z${i}.zip \
    --suffix '.jpg[Q=75,quant_table=3]' --layout dz --strip --background 0 \
    --tile-size 900 --depth onetile --compression $i
done

0 vips dzsave world-86400x43200.tif   dz     107,88s user 33,57s system 199% cpu 1:10,75 total
1 vips dzsave world-86400x43200.tif   dz     114,88s user 37,65s system 192% cpu 1:19,16 total
2 vips dzsave world-86400x43200.tif   dz     111,73s user 36,37s system 189% cpu 1:18,02 total
3 vips dzsave world-86400x43200.tif   dz     105,16s user 34,60s system 190% cpu 1:13,32 total
4 vips dzsave world-86400x43200.tif   dz     104,94s user 34,05s system 188% cpu 1:13,74 total
5 vips dzsave world-86400x43200.tif   dz     105,67s user 34,55s system 187% cpu 1:14,95 total
6 vips dzsave world-86400x43200.tif   dz     104,77s user 34,38s system 188% cpu 1:13,87 total
7 vips dzsave world-86400x43200.tif   dz     106,27s user 34,49s system 187% cpu 1:15,13 total
8 vips dzsave world-86400x43200.tif   dz     105,94s user 34,56s system 186% cpu 1:15,39 total
9 vips dzsave world-86400x43200.tif   dz     105,75s user 34,07s system 187% cpu 1:14,76 total

ls -hal world-86400x43200-dz900z?.zip
-rw-r--r--  1 felix  staff   237M  3 Jun 23:33 world-86400x43200-dz900z0.zip
-rw-r--r--  1 felix  staff   205M  3 Jun 23:34 world-86400x43200-dz900z1.zip
-rw-r--r--  1 felix  staff   204M  3 Jun 23:36 world-86400x43200-dz900z2.zip
-rw-r--r--  1 felix  staff   204M  3 Jun 23:37 world-86400x43200-dz900z3.zip
-rw-r--r--  1 felix  staff   203M  3 Jun 23:38 world-86400x43200-dz900z4.zip
-rw-r--r--  1 felix  staff   203M  3 Jun 23:39 world-86400x43200-dz900z5.zip
-rw-r--r--  1 felix  staff   202M  3 Jun 23:40 world-86400x43200-dz900z6.zip
-rw-r--r--  1 felix  staff   202M  3 Jun 23:42 world-86400x43200-dz900z7.zip
-rw-r--r--  1 felix  staff   202M  3 Jun 23:43 world-86400x43200-dz900z8.zip
-rw-r--r--  1 felix  staff   202M  3 Jun 23:44 world-86400x43200-dz900z9.zip
```

As we can see there is not much to gain from raising compression level above deflate level 1, but there is also nearly no overhead for enabling compression at any level.

I also tested with my normal images which are only 3000x3000 to 5400x5400 but contain a lot of solid areas and I'm seeing around 20% savings.

**ToDo:**

- [x] Add python test that checks compression is working

### DIFF
--- a/libvips/foreign/dzsave.c
+++ b/libvips/foreign/dzsave.c
@@ -147,6 +147,14 @@
 #include <vips/vips.h>
 #include <vips/internal.h>
 
+#define VIPS_ZIP_STORE 0
+#define VIPS_ZIP_DEFLATE 8
+#define VIPS_ZIP_DEFAULT_COMPRESSION -1
+
+/* libgsf before 1.14.31 did not support deflate-level.
+ */
+#define vips_gsf_has_zip64 vips_gsf_has_deflate_level
+
 /* Track this during property save.
  */
 typedef struct _WriteInfo { 
@@ -326,6 +334,10 @@ typedef struct _VipsGsfDirectory {
 	 */
         GsfOutput *container;
 
+	/* Set deflate compression level for zip container.
+	 */
+	gint compression_level;
+
 } VipsGsfDirectory; 
 
 /* Close all dirs, non-NULL on error.
@@ -383,7 +395,7 @@ vips_gsf_tree_free( VipsGsfDirectory *tree )
 /* Make a new tree root.
  */
 static VipsGsfDirectory *
-vips_gsf_tree_new( GsfOutput *out, gboolean is_zip )
+vips_gsf_tree_new( GsfOutput *out, gboolean is_zip, gint compression_level )
 {
 	VipsGsfDirectory *tree = g_new( VipsGsfDirectory, 1 );
 
@@ -393,6 +405,7 @@ vips_gsf_tree_new( GsfOutput *out, gboolean is_zip )
 	tree->out = out;
 	tree->is_zip = is_zip;
 	tree->container = NULL;
+	tree->compression_level = compression_level;
 
 	return( tree ); 
 }
@@ -430,12 +443,13 @@ vips_gsf_dir_new( VipsGsfDirectory *parent, const char *name )
 	dir->children = NULL;
 	dir->is_zip = parent->is_zip;
 	dir->container = NULL;
+	dir->compression_level = parent->compression_level;
 
 	if( dir->is_zip ) 
 		dir->out = gsf_outfile_new_child_full( 
 			(GsfOutfile *) parent->out, 
 			name, TRUE,
-			"compression-level", 0, 
+			"compression-level", VIPS_ZIP_STORE,
 			NULL );
 	else
 		dir->out = gsf_outfile_new_child( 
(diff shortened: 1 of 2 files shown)


felixbuenemann:
@jcupitt I still need to add tests, but I wanted to get your feedback / review first.

felixbuenemann:
**My use case for this feature:**

I am storing zoom images as ZIP containers on AWS S3 blob storage, which is much faster, more cost effective and easier to manage than uploading thousands of small files (each upload operation is billed).

Before I upload the file to S3 I extract the byte ranges for each tile from the ZIPs central directory using a few lines of Python and store them as a separate file.

I can then request each image using a HTTP Range request either directly from the Browser or through an intermediate reverse proxy server. The reverse proxy can also fetch the complete zip file and serve all subsequent range request directly from its cache, which is great if edge caching is used (CDNs).

This already works great with uncompressed ZIPs, but because all browsers already support DEFLATE compression as content encoding, it also works nicely with compressed ZIP files.

There are additional gains when using a sparse layout, like google. Because I know which tiles exists from the extracted metadata, I can make sure that only a single HTTP request is made for all missing tiles.

Combining all these techniques with the flexible computing power provided by AWS Lambda makes it viable to generate (and serve) zoom images on the fly as long as they are reasonably sized.

Vips is such a great tool, without it this simply wouldn't be viable. Thanks, John!

felixbuenemann:
OK, added a simple test to ensure compressed zip is smaller than uncompressed.

jcupitt:
Hi Felix, that's interesting, I'm surprised the savings are so large. What if you turn on the `optimize-coding` option to jpeg save? Does that make zip encoding less useful?

I hadn't thought of sending the zipped jpg files out directly, I'd assumed you'd have to decompress for transmission, that's nice. 

This flag will have to default off, since some people are using a zip container for DZ and assume it's uncompressed. Maybe call it `compression` and default to zero?

jcupitt:
Oooop, misread your patch, that's exactly what you've done.

felixbuenemann:
Thanks for mentioning `optimize_coding`, I usually leave it disabled for large images, but it has almost no overhead for the zoom image case.

I tested again with the above image and the savings gained by deflate go down from ~14% to ~5%.

However on my actual images I am still seeing ~25% savings even on top of `no_subsample`, `overshoot_deringing` and `trellis_quant`. Combining all those options and deflate compression I'm seeing up to 55% savings over the unoptimized and stored jpeg.

Interestingly adding `interlace` and `optimize_scans` which adds quite some cpu overhead, leads to smaller files before deflate compression, but larger files after deflating. So in this case deflate is far more cpu efficient in bringing down the size.

jcupitt:
This is great, nice job @felixbuenemann!

PR #462, opened 2016-06-04
[WIP] More accurate dzsave zip size estimation

felixbuenemann:
This improves the checks on when to abort dzsave due to lack of ZIP64
support in libgsf. ZIP64 is required for archives/files larger than 4GB
and if more than 65535 files need to be stored, which is the more likely
limit to hit when saving zoom images.

* Add check to abort when exceeding the 64k file limit
* Modify size check to account for per file and directory overhead

The previous estimation was ways off, subtracting only 100k for overhead
when the actual overhead would be several megabytes for file counts near
the limit.

Estimation Basis:

Per file overhead: 30B fixed size fields + filename + extras
Central directory overhead: 46B fixed size fields + filename + extras
End of central directory: 22B fixed size fields

ZIP64 specific header sizes are not estimated, because the checks are
only done if zip64 support is lacking.

In order to estimate the variable space for filenames we need to track
the total length of all filenames and double that, because the filename
is stored in both the local file header and the central directory.

Because we don't know how big the blank tile and metadata files will be,
we subtract 64k from the max size and also leave room for up to 3 more
files (blank tile, metadata, vips properties).

**ToDo:**

- [ ] Install an old libgsf version (pre 1.14.31) and verify the checks are working properly for various layouts and both file count and file size limit constraints.
- [ ] Make sure the 16KB headroom is enough with very large tiles and with nearly 64k tiles.

### DIFF
--- a/libvips/foreign/dzsave.c
+++ b/libvips/foreign/dzsave.c
@@ -326,6 +326,14 @@ typedef struct _VipsGsfDirectory {
 	 */
         GsfOutput *container;
 
+	/* Track number of files for ZIP64 limit check.
+	 */
+	unsigned int file_count;
+
+	/* Track length of file names for ZIP64 limit check.
+	 */
+	unsigned int filename_lengths;
+
 } VipsGsfDirectory; 
 
 /* Close all dirs, non-NULL on error.
@@ -393,6 +401,8 @@ vips_gsf_tree_new( GsfOutput *out, gboolean no_compression )
 	tree->out = out;
 	tree->no_compression = no_compression;
 	tree->container = NULL;
+	tree->file_count = 0;
+	tree->filename_lengths = 0;
 
 	return( tree ); 
 }
@@ -465,13 +475,18 @@ vips_gsf_path( VipsGsfDirectory *tree, const char *name, ... )
 	char *dir_name;
 	GsfOutput *obj;
 
+	tree->file_count += 1;
+	tree->filename_lengths += strlen(tree->out->name) + 1 + strlen(name);
+
 	dir = tree; 
 	va_start( ap, name );
-	while( (dir_name = va_arg( ap, char * )) ) 
+	while( (dir_name = va_arg( ap, char * )) ) {
+		tree->filename_lengths += strlen(dir_name) + 1;
 		if( (child = vips_gsf_child_by_name( dir, dir_name )) )
 			dir = child;
 		else 
 			dir = vips_gsf_dir_new( dir, dir_name );
+	}
 	va_end( ap );
 
 	if( dir->no_compression )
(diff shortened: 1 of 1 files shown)


jcupitt:
That looks like a very nice improvement, good job!

jcupitt:
Oh honestly, sorry, I merged without thinking. More coffee.

libvips/foreign/dzsave.c
```diff
@@ -326,6 +326,14 @@ typedef struct _VipsGsfDirectory {
 	 */
         GsfOutput *container;
 
+	/* Track number of files for ZIP64 limit check.
+	 */
+	unsigned int file_count;
```
jcupitt: I think I'd just use an int. Mixed uint/int arithmetic is very error-prone, and there's no need for the extra range a uint would give you.

libvips/foreign/dzsave.c
```diff
@@ -326,6 +326,14 @@ typedef struct _VipsGsfDirectory {
...
+
+	/* Track length of file names for ZIP64 limit check.
+	 */
+	unsigned int filename_lengths;
```
jcupitt: `strlen()` is a `size_t`, so this probably should be as well.

libvips/foreign/dzsave.c
felixbuenemann: Actually the ZIP64 format uses uint64 for the number of entries in thr ZIP64 EOCD, so this should probably be size_t as well.

felixbuenemann:
@jcupitt Can you re-open this PR or do I have to create a new one?

felixbuenemann:
Btw. there's also a bug in libgsf's ZIP64 support causing the generated ZIP files to be corrupt due to a wrong size header in the ZIP64 EOCD. This should probably be fixed in 1.14.38, see GNOME/libgsf#1 for details. So should you get any reports about problems with generated ZIP archives that's probably why.

felixbuenemann:
I did some testing with libgsf 1.14.30, the file count check seems to work ok, but the file size check is not working as expected, because it happens after writing the tile instead of before. If you have large tiles, you'll exceed the 4GB barrier making the ZIP invalid.

I'll look into re-arranging the code so the check happens when we know how big the tile will be, but before we write it to the container.

felixbuenemann:
@jcupitt The current way of checking the size of the zip is flawed in that it doesn't account at compression, which will be a problem if #461 gets merged.

Do you know a way to get a reference to the GsfOutputStdio that's backing the GsfOutputfileZip which in turn backs the `dz->tree->container`?

I tried that yesterday but couldn't find out how to get at the "sink" that according to the gsf docs should be present on the zip output and using `gsf_output_size()` or `gsf_output_tell()` on the zip output itself did not give meaningful results.

felixbuenemann:
@jcupitt Can you re-open this PR or is that not possible?

jcupitt:
I'm very sorry Felix, I don't think I can :(  Or github isn't offering me a button anyway.

felixbuenemann:
NP, I'll just create a new PR and reference this one once I have time to work on this again.

jcupitt:
Sorry for being an idiot :(

felixbuenemann:
It's no big deal and I'm pretty sure youre not an idiot 😄

jcupitt:
Hi again, I've applied this patch by hand, so you can delete your branch, if you like.

Off-topic, but if you're interested in webp still, git master libvips can now load and save webp files with EXIF, XMP and ICC metadata.

felixbuenemann:
Thanks John!

> Off-topic, but if you're interested in webp still, git master libvips can now load and save webp files with EXIF, XMP and ICC metadata.

I tend to convert to sRGB and then strip all metadata including profiles before output to WebP/JPEG/whatever, so I've no particular use for that feature, but it's nice to know it's there :-)

Issue #460, opened 2016-06-03
Dzsave properties xml wrong path


felixbuenemann:
This is written of the top of my head, so not 100% accurate:

The path to the vips dzsave properties file inside zip is duplicated, eg. foo.zip/foo/foo/vips_properties.xml with expected path foo.zip/foo/vips_properties.xml

felixbuenemann:
Hmm, I guess this depends a bit on the layout used. For example with the deep zoom layout, the file ends up at `test1dz500p/test1dz500p_files/vips-properties.xml` which makes sense in order to avoid clashing with other deep zoom images in the same directory when not using zip.

But for the google layout it ends up at `test1dz500pgm/test1dz500pgm/vips-properties.xml` with other files like `test1dz500pgm/0/0/0.jpg` or `test1dz500pgm/blank.png` in which case the duplicate folder name is unexpected.

felixbuenemann:
@jcupitt This is honestly not a big issue, so feel free to close if you don't feel like changing it,

felixbuenemann:
Fixed in PR #464.

jcupitt:
OK, merged!

PR #464, opened 2016-06-05
Fix dzsave vips-properties path for non-dz layout

felixbuenemann:
This fixes a duplicate root path for the vips-properties.xml, when either the zoomify or goolgle layouts are used.

Example: `foo/foo/image-properties.xml` becomes `foo/image-properties.xml`

This resolves issue #460.


felixbuenemann:
@jcupitt Any reason this hasn't been merged?

jcupitt:
Sorry, I was worried about zip vs. non-zip output and dz vs. gm. Merged!

PR #465, opened 2016-06-05
dzsave: Remove toplevel dir from zip container

felixbuenemann:
This simplifies the code quite a bit and gets rid of deepzoom hacks.
Also removed redundant code setting no compression on zip directories.

@jcupitt Do you have any objections to dropping the top-level directory in ZIP archives? It would save quite same space due to shorter filenames and also makes it easier/more memory effizient to work with the zip files programatically. If you have to load the directory into memory those extra bytes add up quickly.

It seems this was added originally to support ZIP containers, but I can write ZIP just fine without a parent directory, by simply passing in `zip` as the root of the `dz->tree` and removing `dz->tree->container`. 

This also allows to get rid of some FIXMEs to change the directory structure for the dz layout.

### DIFF
--- a/libvips/foreign/dzsave.c
+++ b/libvips/foreign/dzsave.c
@@ -321,11 +321,6 @@ typedef struct _VipsGsfDirectory {
 	 */
 	gboolean no_compression;
 
-	/* The root node holds the enclosing zip file or FS root ... finish
-	 * this on cleanup.
-	 */
-        GsfOutput *container;
-
 } VipsGsfDirectory; 
 
 /* Close all dirs, non-NULL on error.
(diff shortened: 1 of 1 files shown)


jcupitt:
I think you can't write to the filesystem without a parent directory :( so for dzsave with an fs container, you need to move things around after the write. It's annoying.

felixbuenemann:
Hmm, looks like it's failing on trusty which has libgsf 1.4.27, so this might be a limitation of older libgsf versions. It certainly works without a parent directory in newer versions.

I'll try on 1.4.27 locally.

kleisauke:
It might be worth investigating https://github.com/richgel999/miniz (MIT licensed) as possible replacement for libgsf. miniz is a self-contained library for reading/writing zipfiles that unlike other zip libraries also includes libz compatible compress/decompress support. It is a single header and a single C file without any other dependencies.

jcupitt:
Yes, it looks nice. It's not part of ubuntu, so we'd need to bundle the sources, which would be annoying for maintenance. 

libgsf supports filesystem and zip output, so we'd need a little wrapper. But we have something very like that already, so perhaps it's not a problem.

jcupitt:
I think this is too old now. I'll close to keep things tidy.

PR #468, opened 2016-06-07
Re-enable Travis CI builds on all branches

felixbuenemann:
This re-enables travis ci to trigger for all branches, so it can be used to check the build on forks without having to modify `.travis.yml`.

The reason given for the change was the failing build badge, so this changes the build badge to monitor the master branch instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,3 @@
-branches:
-  only:
-    - master
-    - stable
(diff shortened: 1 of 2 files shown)


jcupitt:
Good plan! I googled but managed to not find out how to get the badge to just show master.

Issue #473, opened 2016-06-23
svgload does not detect svgz files


felixbuenemann:
Currently svgload is not used for `.svgz` files, which is the file extension used for gzip-compressed svg files.

Most major svg programs can generate `.svgz` in addition to `.svg`, so it would be nice if this format was supported by svgload as well. I'm pretty sure, this is only an issue with the format detection code, because `rsvg-convert` from librsvg 2.4 handles `.svgz` files just fine.

I'm not sure what's the best solution to the problem, because the current detection code bypasses librsvg and does some plain text sniffing, which would mean it would have to also detect gzip, uncompress a couple of bytes and then search on the result…

If you also have magickload enabled, it will eventually load the svgz, but magickload is slow and its svg support is worse then librsvg's.

jcupitt:
I suppose we could add a `svgzload` operation which just looked at the filename suffix. If there's no file sniffer, it'll fall back to just testing the filename suffix:

https://github.com/jcupitt/libvips/blob/master/libvips/foreign/foreign.c#L471

It wouldn't work for memory buffers. Perhaps users could just call `svgzload_buffer` by hand if they wanted that.

felixbuenemann:
Hmm, don't you think it's a bit confusing to have two different loaders, that end up calling the exact same code after the detection stage?

jcupitt:
Well, that's true. The alternative is to look for the gz magic number and if we see it, decompress the first few hundred bytes. I've not looked into libz, I don't know how easy that is.

felixbuenemann:
I looked at the [zlib manual](http://www.zlib.net/manual.html) and I think you could use the `uncompress` function with a fixed size destination buffer, that's big enough to hold as much uncompressed xml as is required to run the detection code. The call to `uncompress` should return `Z_BUF_ERROR`, but you should still be able to read the output buffer.

felixbuenemann:
OK, it's not actually that easy. I tried it with a small sample program and it returned a stream error, so I guess uncompress is for raw deflate streams. I'll try to come up with a short example using the advanced zlib api.

felixbuenemann:
Here's a small sample program that decodes up to `OSIZE` bytes from the input buffer:

```c
#include <stdio.h>
#include <zlib.h>

[...]

PR #476, opened 2016-06-27
Add svgload svgz detection

felixbuenemann:
* Add support for .svgz files and buffers to svgload
* Add configure check for zlib with inflateInit2 which replaced old unused `FIND_ZIP` code

This fixes issue #473.


libvips/foreign/svgload.c
```diff
@@ -390,15 +456,10 @@ vips_foreign_load_svg_is_a_buffer( const void *buf, size_t len )
...
+			return( 0 );
 
 	for( i = 0; i < 200 && i < len - 5; i++ ) {
-		char txt[5];
-
-		/* 5, since we include the \0 at the end.
-		 */
-		vips_strncpy( txt, buf + i, 5 );
-		if( strcasecmp( txt, "<svg" ) == 0 )
+		if( strncmp( (const char *) str + i, "<svg", 4 ) == 0 )
```
felixbuenemann: @jcupitt Is this change ok?

I removed the extra copy, because I only replace `str` in the SVGZ block, but not `buf`.

I switched to `strncmp` so `<SVG` also gets detected, because tag names are case insensitive.

Is casting a pointer in a loop a bad idea or will the compiler be smart about that?

felixbuenemann:
@jcupitt I left the configure changes in a separate commit, in case you don't like them.

felixbuenemann:
Oh, and I tried to add `.svg.gz` as a file extension, but it was not detected, so I only left `.svgz` in there. I didn't investigate further, but I guess duplicate extensions aren't currently supported.

So for now `.svg.gz` will still be handled by magickload…

libvips/foreign/svgload.c
felixbuenemann: Oh, and do we need to add a configure check for `strncmp`?

libvips/foreign/svgload.c
```diff
@@ -364,13 +377,66 @@ typedef VipsForeignLoadSvgClass VipsForeignLoadSvgBufferClass;
...
+		zs.zalloc = (alloc_func) vips_zalloc;
+		zs.zfree = (free_func) vips_zfree;
+		zs.opaque = Z_NULL;
+		zs.next_in = str;
+		zs.avail_in = len;
+
+		if( inflateInit2(&zs, 15 | 32) != Z_OK ) {
+			vips_error( "svgload",
+				"%s", _( "Zlib init failed" ) );
+			return( -1 );
```
felixbuenemann: The -1 return code isn't currently handled by the caller of `vips_foreign_load_svg_is_a_buffer`.

felixbuenemann: @jcupitt Should I simply log the error and `return 0` here?

felixbuenemann: Oh, other is_a_buffer functions also return -1, so it should be OK.

jcupitt:
Oh dear @felixbuenemann sorry I've been very distracted. 

I'll read this over and commit. Thanks!

felixbuenemann:
No problem, John. Let me know, if you have any questions.

libvips/foreign/svgload.c
jcupitt: I think this should be strncasecmp(), shouldn't it? The old code did the copy because I didn't want to test for strncasecmp(), but maybe we can just assume all strings.h have this now.

jcupitt: Yes, case in a loop is fine, it has no runtime cost.

libvips/foreign/svgload.c
```diff
@@ -315,6 +325,9 @@ vips_foreign_load_svg_file_header( VipsForeignLoad *load )
 
 static const char *vips_foreign_svg_suffs[] = {
 	".svg",
+#if LIBRSVG_CHECK_FEATURE(SVGZ)
```
jcupitt: I think this should be `LIBRSVG_CHECK_FEATURE(SVGZ) && defined(HAVE_ZIP)` to match the other ifs. I'll make a #define for this test to be safe.

libvips/foreign/svgload.c
jcupitt: Oh hang on, the n has no function anyway with a fixed second string. Let's just use strcasecmp() and no copy.

libvips/foreign/svgload.c
```diff
@@ -364,13 +377,66 @@ typedef VipsForeignLoadSvgClass VipsForeignLoadSvgBufferClass;
...
+				vips_error( "svgload",
+					"%s", _( "Zlib inflate failed" ) );
+				return( -1 );
+			}
+			opos = sizeof(obuf) - zs.avail_out;
+		} while( opos < sizeof(obuf) && zs.avail_out == 0 );
+
+		inflateEnd(&zs);
+
+		str = obuf;
```
jcupitt: This leaves str pointing into an array that's out of scope. It's probably OK, but makes me uncomfortable, I'll move obuf to the outer scope.

libvips/foreign/svgload.c
felixbuenemann: Do you mean `strcmp()`? Because matching should be case insensitive, `<SVG` should also be matched.

jcupitt: Actually `strcasecmp()` is the case-insensitive one, the names are backwards, thanks to history.

libvips/foreign/svgload.c
felixbuenemann: No, it'll work without zlib, librsvg uses libgsf internally and matching on file extension does not go through the buffer detection code (unless I'm confused ;-).

libvips/foreign/svgload.c
felixbuenemann: Argh, thanks for catching that!

libvips/foreign/svgload.c
jcupitt: Ah good point!

libvips/foreign/svgload.c
```diff
@@ -364,13 +377,66 @@ typedef VipsForeignLoadSvgClass VipsForeignLoadSvgBufferClass;
...
+		do {
+			zs.avail_out = sizeof(obuf) - opos;
+			zs.next_out = obuf + opos;
+			if( inflate(&zs, Z_NO_FLUSH) < Z_OK ) {
+				vips_error( "svgload",
+					"%s", _( "Zlib inflate failed" ) );
+				return( -1 );
+			}
+			opos = sizeof(obuf) - zs.avail_out;
+		} while( opos < sizeof(obuf) && zs.avail_out == 0 );
```
jcupitt: Is this test right? avail_out will be zero exactly when opos == sizeof(obuf), so they are testing the same thing. 

Should it be `while( output_not_full && input_available )`? which would be `while( opos < sizeof(obuf) && zs.avail_in > 0 )`, I think.

jcupitt:
OK, all in master! Thank you Felix, this is great. I'll have a look at the `.svg.gz` problem, you'd think it would work.

felixbuenemann:
> I'll have a look at the .svg.gz problem, you'd think it would work.
I didn't look into the code, but my wild guess would be that it only looks at the last dot when extracting the file extension.

libvips/foreign/svgload.c
felixbuenemann: If I remember correctly, I came to the same conclusion at some point and tried to remove the second condition and it broke some cases so I put it back. The code is supposed to be able to handle the case where the input produces less output than the size of the output buffer.

felixbuenemann:
@jcupitt I think we need to add another test, that actually exercises `vips_svgload_is_a_buffer`. The test I added manually specifies the loader using `buffer_loader("svgload_buffer", …)`, so it only shows that librsvg can handle svgz, but not that the detection code works…

felixbuenemann:
Please disregard my last comment, `buffer_loader` is fine because it uses `new_from_buffer()` after the explicit load:

```python
    def buffer_loader(self, loader, test_file, validate):
        with open(test_file, 'rb') as f:
            buf = f.read()

        im = Vips.call(loader, buf)
        validate(self, im)
        im = Vips.Image.new_from_buffer(buf, "")
        validate(self, im)
```

I also manually tested an svgz that was smaller than the output buffer and if worked fine with the current master. The smallest valid svg I could generate was 28 bytes uncompressed and 45 bytes compressed: `<svg width="1" height="1"/>`.

felixbuenemann:
Thanks for adding the final missing pieces, John!

PR #489, opened 2016-07-20
Round WebP shrink-on-load dimensions

lovell:
Hi John, this was discovered by @asilvas in lovell/sharp#508

(Casting to a float should provide enough accuracy here as WebP uses only 14-bits for dimensions.)

### DIFF
--- a/libvips/foreign/webp2vips.c
+++ b/libvips/foreign/webp2vips.c
@@ -186,8 +186,8 @@ read_new( const char *filename, const void *data, size_t length, int shrink )
 
 	read->config.options.use_threads = 1;
 
-	read->width = read->config.input.width / read->shrink;
-	read->height = read->config.input.height / read->shrink;
+	read->width = VIPS_RINT( (float) read->config.input.width / read->shrink );
+	read->height = VIPS_RINT( (float) read->config.input.height / read->shrink );
 
 	if( read->width == 0 ||
 		read->height == 0 ) {


jcupitt:
This change makes me uncomfortable, @lovell. vips does round-down rather than round-to-nearest everywhere else, I'm not sure I understand why it has to be different in this one case.

Can you show a test case where the current behaviour is wrong? I see some sharp code in the original issue:

```javascript
     .ignoreAspectRatio()
     .resize(233, 131)
     .extract({ top: 1, left: 0, width: 233, height: 130 })
```

That makes it look like `.resize()` is not making an image of the size it's being asked to make.

lovell:
Thanks John, no worries. libjpeg(-turbo) appears to round-to-nearest for its shrink-on-load feature but in this instance it's probably best I modify sharp to recalculate dimensions again after any shrink-on-load, be it JPEG or WebP.

jcupitt:
Yes, that's what `vipsthumbnail` does. You do need to be careful about aspect ratio if you recalc the shrinks, of course, argh.

jcupitt:
Oh you're right, jpeg shrink-on-load does round to nearest, I'd never noticed. 

I don't think resize can make any assumptions about the rounding behaviour of shrink-on-load; other formats might do it differently, or even other jpeg decoders.

jcupitt:
Oh crap, sorry, the touchpad on this laptop is too sensitive and I accidentally posted too early and also reopened. Argh.

PR #490, opened 2016-07-23
Fix truncation if target_width/target_height

ioquatix:
Prefer to use round as scale factors might carry some error. Possible fix for https://github.com/jcupitt/ruby-vips/issues/82

Completely untested - just did it in the github editor. Not sure if this is a systematic issue that might need to be addressed elsewhere in the code base.

### DIFF
--- a/libvips/resample/resize.c
+++ b/libvips/resample/resize.c
@@ -171,11 +171,11 @@ vips_resize_build( VipsObject *object )
 
 	/* The image size we are aiming for.
 	 */
-	target_width = in->Xsize * resize->scale;
+	target_width = round(resize->scale * in->Xsize);
 	if( vips_object_argument_isset( object, "vscale" ) ) 
-		target_height = in->Ysize * resize->vscale;
+		target_height = round(resize->vscale * in->Ysize);
 	else
-		target_height = in->Ysize * resize->scale;
+		target_height = round(resize->scale * in->Ysize);
 
 	int_hshrink = vips_resize_int_shrink( resize, resize->scale );
 	if( vips_object_argument_isset( object, "vscale" ) ) 


jcupitt:
This makes me uncomfortable. vips always rounds down everywhere, otherwise you get fractional pixels at edges. 

I think a better fix for jcupitt/ruby-vips#82 would be to add 0.1 to your target size before calculating the resize factor. This is what `vipsthumbnail` does:

https://github.com/jcupitt/libvips/blob/master/tools/vipsthumbnail.c#L188

ioquatix:
If you add 0.1 to the size, doesn't that skew the resize to be off by 0.1, since effectively you'd be resizing to x + 0.1?

ioquatix:
Adding 0.1 to a function to get it to do the right thing seems like a hack to work around buggy behaviour in resize (incidentally, it's close do what I implemented as a temporary fix). 

If you are accepting floating point arguments, you have to assume they are +/- some error. Integer truncation definitely seems like the wrong approach here.

It doesn't make sense for every piece client code to work around the integer truncation behaviour of the rounding code. The resize function is leaking it's implementation details into client code.

Having a resize function that requires a bias to work correctly just seems like a bad design. It's almost enough to put me off using this library entirely.

If you aren't willing to use floating point numbers correctly, with respect to their inherent errors, I'd suggest you make the resize function take absolute integrals for width and height and internally compute the scale factor that suits. That way, at least you are not leaking (IMHO) buggy behaviour.

jcupitt:
Ah perhaps you're right. I'll have a think.

jcupitt:
I made an enhancement issue to try this out.

ioquatix:
Thanks for taking the time to think about this issue more.

felixbuenemann:
I've also been bitten by this in cases were I was shrinking to an evenly divisable ration of the original and the addition of the 0.1 lead to selecting bad shrink-on-load factors.

I ended up with some code like this when calculating the shrink factor:

```python
def calculate_shrink(im, out_w, out_h, crop=False):
    """Calculate shrink factor based on in/out sizes."""
    horizontal = float(im.width) / out_w
    vertical = float(im.height) / out_h
    # +0.1 to avoid rounding below target size in resize
    if int(im.width / horizontal) != out_w:
        horizontal = float(im.width) / (out_w + 0.1)
    if int(im.height / vertical) != out_h:
        vertical = float(im.height) / (out_h + 0.1)

    if horizontal < vertical:
        return horizontal if crop else vertical
    else:
        return vertical if crop else horizontal

```

It would be great if the client code didn't need such workarounds to get the desired size.

Another idea to solving this would be to add width and height parameters in pixels to the resize functions instead of having to calculate a float shrink factor:

```python
# current:
# scale by factor:
im.resize(scale)
# proposed:
# scale by target width:
im.resize(width=w)
# scale by target height:
im.resize(height=h)
# scale by width and height (distort):
im.resize(width=w, height=h)
```

Vips would then ensure the new image matches the given width, height or both exactly and if the second dimension is missing it can calculate it by rounding.

This api would also feel a lot more natural to me.

ioquatix:
The proposed API by @felixbuenemann seems really good. However, I'd still argue that we need to use round when scaling by a floating point value. However, this would allow us to work around the issue correctly.

ioquatix:
Have we made any progress on this issue?

jcupitt:
#491 is supposed to be tracking this. I've not done anything on it yet, I've been working on convolution and ImageMagick7 support. 

I've added that issue to 8.4, so it should some attention soon.

felixbuenemann:
@ioquatix The changes from #491 have been merged to master. If everything is OK for you I think this  PR could be closed.

jcupitt:
I'll close, I think an equivalent to this is now in stable.

PR #518, opened 2016-09-07
Better support for exif resolution

UlysseM:
Pull request for the issue mentioned here https://github.com/jcupitt/libvips/issues/516

### DIFF
--- a/libvips/foreign/jpeg2vips.c
+++ b/libvips/foreign/jpeg2vips.c
@@ -542,6 +542,34 @@ get_entry_double( ExifData *ed, int ifd, ExifTag tag, double *out )
 	return( vips_exif_get_double( ed, entry, 0, out ) );
 }
 
+static void
+set_entry_double( ExifData *ed, int ifd, ExifTag tag, double in)
+{
+	ExifEntry *entry;
+
+	if( (entry = exif_content_get_entry( ed->ifd[ifd], tag )) ) {
+		/* Let's remove the current value */
+		exif_content_remove_entry( ed->ifd[ifd], entry );
+		entry = NULL;
+	}
+
+	if ( !( entry = exif_entry_new() ) )
+		return;
+
+	entry->tag = tag;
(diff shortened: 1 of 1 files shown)


jcupitt:
Looks very reasonable. 

Rather than falling back to jfif, I suppose the alternative would be to set yres to match xres, or xres to match yres, if one is missing. Do you think the jfif solution is better?

UlysseM:
I think that, if an exif only has one value, it should be considered invalid altogether.
At this point, falling back to a value we know is valid (with both x and y) is better than assuming the pixel are square and copying the resolution over.

jcupitt:
OK, I agree, I'll merge. Thank you very much for this nice improvement!

jcupitt:
I think I found a small problem: it's not setting the res unit if that's missing. I see:

```
$ vipsheader -a bad-exif-res.jpg
vips warning: VipsJpeg: error reading resolution
bad-exif-res.jpg: 1024x683 uchar, 3 bands, srgb, jpegload
...
xres: 13.779528
yres: 13.779528
...
exif-ifd0-XResolution: 13779/1000 (13.779, Rational, 1 components, 8 bytes)
exif-ifd0-YResolution: 13779/1000 (13.779, Rational, 1 components, 8 bytes)
```

It's just set the vips res, which is in pixels/mm. I changed it to use the exif res writer from jpeg save, I now see:

```
$ vipsheader -a bad-exif-res.jpg
vips warning: VipsJpeg: error reading resolution
bad-exif-res.jpg: 1024x683 uchar, 3 bands, srgb, jpegload
...
xres: 13.779528
yres: 13.779528
...
exif-ifd0-YResolution: 350/1 (350, Rational, 1 components, 8 bytes)
exif-ifd0-ResolutionUnit: 2 (Inch, Short, 1 components, 2 bytes)
exif-ifd0-XResolution: 350000/1000 (350.000, Rational, 1 components, 8 bytes)
```

UlysseM:
Looks good, thanks

PR #519, opened 2016-09-07
Allow for output directory names containg a dot

unusual-thoughts:
Don't strip the extension at the end until you know you are saving to a file


jcupitt:
I though of a small problem: you need to always remove the suffix, not just if you use it to set the container type. I added remove-.dz as well.

Anyway, thank you very much!

PR #540, opened 2016-10-15
cplusplus - move assignment operators outside class, enables DLL exporting

lovell:
Hi John, this PR updates the assignment operator overloads introduced in 28efdf1 to work with MSVC by moving them "outside" the class and therefore enable DLL exporting.

Without this, MSVC generates the following rather cryptic error:
```
error C2487: '+=': member of dll interface class may not be declared with dll interface
```

There's also a bit of an MSVC annoyance in that when returning by reference, the __declspec bit has to go before the type, so what I ended up doing was ensuring a consistent modifier order across all exported functions, which looks a bit neater.

### DIFF
--- a/cplusplus/VImage.cpp
+++ b/cplusplus/VImage.cpp
@@ -770,21 +770,21 @@ operator+( VImage a, std::vector<double> b )
 }
 
 VImage & 
-VImage::operator+=( const VImage b )
+operator+=( VImage a, const VImage b )
 {
-	return( *this = *this + b ); 
+	return( a = a + b ); 
 }
 
 VImage & 
-VImage::operator+=( const double b )
+operator+=( VImage a, const double b )
 {
-	return( *this = *this + b ); 
+	return( a = a + b ); 
 }
 
 VImage & 
-VImage::operator+=( std::vector<double> b )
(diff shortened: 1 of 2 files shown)


jcupitt:
I don't understand why they have to be outside the class, you'd think being `friend' would be enough. I find C++ very mysterious :( 

Anyway, it looks good. Do you think this is too large a change for a minor release? I'm inclined to put it in 8.4.3 rather than waiting for 8.5 since it seems to be important for sharp.

lovell:
Thanks for merging John.

sharp is happily using patched [VImage8.h](https://github.com/lovell/sharp/blob/quill/packaging/build/win.sh#L17) and [VImage.cpp](https://github.com/lovell/sharp/blob/quill/src/libvips/cplusplus/VImage.cpp#L773) files so this can wait for libvips 8.5.

A possible alternative to selecting individual functions for export might be to export the whole VImage class itself. This would include private functions, which feels a little dangerous and could lead to accidental coupling between lipvips-cpp.dll and anything depending upon it.

jcupitt:
OK, 8.5 then. Good news about sharp working with 8.4 now.

PR #543, opened 2016-10-19
Increase SVG buffer check length by 50 bytes

lovell:
Hi John, was the 200 value an otherwise arbitrary limit? Increasing it to 250 will provide support for "real world" SVG files such as the following unnecessary preamble exported by Adobe Illustrator:

```svg
<?xml version="1.0" encoding="utf-8"?>
<!-- Generator: Adobe Illustrator 18.1.1, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<svg ...
```
https://raw.githubusercontent.com/sandstorm-io/sandstorm/master/icons/key.svg

### DIFF
--- a/libvips/foreign/svgload.c
+++ b/libvips/foreign/svgload.c
@@ -482,7 +482,7 @@ vips_foreign_load_svg_is_a_buffer( const void *buf, size_t len )
 		if( !isascii( str[i] ) )
 			return( FALSE );
 
-	for( i = 0; i < 200 && i < len - 5; i++ ) 
+	for( i = 0; i < 250 && i < len - 5; i++ )
 		if( g_ascii_strncasecmp( str + i, "<svg", 4 ) == 0 )
 			return( TRUE );
 


jcupitt:
Yes, I thought 200 bytes was plenty for any kind of DOCTYPE line. I was wrong!

lovell:
Adobe Illustrator is to SVG as Microsoft Word is to HTML.

jcupitt:
Actually, if you trim that example down do the character after the `svg`, it's 239 bytes. That makes 250 seems a little marginal. Shall we make it 300?

lovell:
Go for it!

Should anyone* really want to determine the largest "real world" value, there are almost 10 million SVG files on GitHub that contain a comment before the svg element.

https://github.com/search?l=SVG&q=%27%3C%3Fxml+version%3D%221.0%22+encoding%3D%22utf-8%22%3F%3E+%3C%21--%27&ref=searchresults&type=Code&utf8=%E2%9C%93

\* not me

jcupitt:
OK, bumped to 300!

thomas15k:
In this example, a file generated by adobe ilustrator, <svg appears after 714 bytes

jcupitt:
Oh good grief. Let's bump it to 1000 then.

PR #668, opened 2017-05-29
Silence 'missing sentinel' warnings in GCC v6+

lovell:
GCC v6+ is starting to warn about the use of `NULL` as a sentinel value in C++ sources (as `NULL` is considered a zero value rather than a null pointer).

```
reduceh.cpp: In function 'int vips_reduceh_build(VipsObject*)':
reduceh.cpp:509:8: warning: missing sentinel in function call [-Wformat=]
   NULL ) )
        ^
reduceh.cpp:514:41: warning: missing sentinel in function call [-Wformat=]
   VIPS_DEMAND_STYLE_THINSTRIP, in, NULL ) )
                                         ^
reducev.cpp:790:40: warning: missing sentinel in function call [-Wformat=]
   VIPS_DEMAND_STYLE_FATSTRIP, in, NULL ) )
                                        ^
reducev.cpp: In function 'int vips_reducev_build(VipsObject*)':
reducev.cpp:873:8: warning: missing sentinel in function call [-Wformat=]
   NULL ) )
        ^
reducev.cpp:897:9: warning: missing sentinel in function call [-Wformat=]
    NULL ) )
         ^
```

### DIFF
--- a/libvips/resample/reduceh.cpp
+++ b/libvips/resample/reduceh.cpp
@@ -506,12 +506,12 @@ vips_reduceh_build( VipsObject *object )
 		reduceh->n_point / 2 - 1, 0, 
 		width, in->Ysize,
 		"extend", VIPS_EXTEND_COPY,
-		NULL ) )
+		(void *) NULL ) )
 		return( -1 );
 	in = t[1];
 
 	if( vips_image_pipelinev( resample->out, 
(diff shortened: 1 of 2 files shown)


jcupitt:
NULL is simply 0? That seems crazy.

http://en.cppreference.com/w/cpp/types/NULL

Daft C++! Thank you for this patch Lovell.

PR #701, opened 2017-08-01
Autofit text in the requested dimensions

gargsms:
This PR implements auto fitting of text in requested dimensions.

Two new flags were added to `vips text` to achieve this. `text` now supports `--height gint` and `--gravity VIPS_GRAVITY` to determine the position, and font size of the text automatically.

To autofit text, one needs to use `vips text` like:
```
vips text out.png "input text here" --font "Font Name" --width XXX --height XXX
```

The following table summarizes how different flags work to create a text output image. Font Size takes priority over others, and no autofit logic is applied if the user has specified a font size. The older behavior of `vips text` is preserved for user specified font size.

Font Size | Width | Height | Will text autofit?
--|--|--|--
✗ | ✗ | ✗ | ✗
✗ | ✓ | ✗ | ✗ 
✗ | ✗ | ✓ | ✗
✗ | ✓ | ✓ | ✓

As evident, `height` by itself does not have any effect on the autofit.

To make the process efficient, binary search is used to figure out the font size that has least deviation to requested dimensions. Incremental changes are made after a loosely fitting value of font size is obtained.

This is not a perfect working version of the implementation, it fails horribly for a single letter being autofit, however, it works just fine for most practical purposes. I have tested this thoroughly for edge cases, and have been using it in production for quite some time now. There is, however, quite some scope for improvement.

@jcupitt Comments welcomed. I have tried to maintain full backward compatibility, so any application using `vips text` the way it was before should not be harmed by this.

### DIFF
--- a/libvips/create/text.c
+++ b/libvips/create/text.c
@@ -108,6 +108,12 @@ static PangoFontMap *vips_text_fontmap = NULL;
  */
 static GMutex *vips_text_lock = NULL; 
 
+/* The maximum deviation to tolerate for autofitting the text
+ *
+ * TODO: Should the user be able to change this?
+ */
+static int const MAX_TOLERANCE = 10;
+
 static void
 vips_text_dispose( GObject *gobject )
 {
@@ -232,8 +238,7 @@ append_to_flist( FontSizeList *flist, FontSizeList *nflist )
 
 static PangoRectangle
 fit_to_bounds( VipsText *text,
-	int tolerance, char *name, int size, PangoRectangle rect,
-	FontSizeList *flist, bool coarse )
+	char *name, int size, PangoRectangle rect, FontSizeList *flist, bool coarse )
 {
 	int buf_size = strlen( name ) + digits_in_num( size ) + 2;
 	int deviation;
@@ -261,7 +266,7 @@ fit_to_bounds( VipsText *text,
 	text->layout = text_layout_new( text->context,
 		text->text, buf, text->width, text->spacing, text->align );
 
-	pango_layout_get_extents( text->layout, &rect, NULL );
+	pango_layout_get_extents( text->layout, NULL, &rect );
 
 	deviation = determine_deviation( text->width, text->height, rect );
 
@@ -275,7 +280,7 @@ fit_to_bounds( VipsText *text,
 	// smallest deviation and then fit in small adjustments
 	if( search_flist( flist, size ) ) {
 		if( coarse ) {
-			return fit_to_bounds( text, tolerance, name, size, rect,
+			return fit_to_bounds( text, name, size, rect,
 				least_deviation_flist( flist ), false );
 		} else {
 			// We cannot do better than this because we will
@@ -284,8 +289,8 @@ fit_to_bounds( VipsText *text,
 		}
 	}
 
-	if( deviation > tolerance )  {
-		return fit_to_bounds( text, tolerance, name, size, rect, flist, coarse );
+	if( deviation > MAX_TOLERANCE )  {
+		return fit_to_bounds( text, name, size, rect, flist, coarse );
 	} else {
 		return rect;
 	}
@@ -300,7 +305,6 @@ vips_text_build( VipsObject *object )
 	FontSizeList *flist = (FontSizeList *) malloc( sizeof( FontSizeList ) );
 
 	PangoRectangle logical_rect;
-	PangoRectangle ink_rect;
 	int left;
 	int top;
 	int width;
@@ -310,9 +314,6 @@ vips_text_build( VipsObject *object )
 	int font_size = 0;
 	char *last;
 	bool is_font_size_provided = true;
-
-	// TODO: Should user be allowed to change this?
-	const int TOLERANCE = 100; // 10% * 10% for width, height
 
 	if( VIPS_OBJECT_CLASS( vips_text_parent_class )->build( object ) )
 		return( -1 );
@@ -335,7 +336,7 @@ vips_text_build( VipsObject *object )
 	if( font_size ) {
 		strncat( font_name, text->font, last - text->font );
 	} else {
-		// Font was more than 1 word. "Fira Code" would have last
+		// Font was more than 1 MAX_word. "Fira Code" would have last
 		// pointing to "Code", leading atol to output 0
 		// Fix font_name back to the original in this case
 		strcpy( font_name, text->font );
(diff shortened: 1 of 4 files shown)


gargsms:
There are a few shortcomings that should be rectified before this PR should be accepted.

- [ ] Text of a few characters cannot be fit reliably. `vips text out.png "I" --width 100 --height 100 --font "Sans"`
- [ ] Font name has to be specified with the text command always. `vips text out.png "some text" --width 100 --height 100` would default to `sans 12`
- [ ] The output image is 1-bit, so no color information can be written to the image

I will keep updating this comment as I further realize the shortcomings. :slightly_smiling_face:

jcupitt:
8.6 is due in September, it'd be nice to get this in before then. I've added it to the 8.6 milestone.

gargsms:
Ah, perfect. I will try to wrap this up by then. I had a few questions I will put here by the end of the day.

jcupitt:
There are a few small style things I'd like to change, but the overall PR looks great.

I don't feel I can nag you to do the style stuff, so how about merging and I'll do a few changes after on master. Does that sounds OK? 

Thanks you for making this nice improvement!

jcupitt:
I had an idea: how about adjusting DPI rather than the font name? It seems like a more natural way to size text, and you'd avoid all the string manipulation.

I would do a binary search, so:

* render at 72 dpi (the default)
* too large? halve the dpi; too small? double the dpi
* render again, until we have a dpi either side of the correct size
* now find the midpoint between the too-large and too-small dpi and try that
* keep finding midpoints until large - small is zero

You could make it converge a bit more quickly by assuming rendered size changes linearly with dpi and picking not the mid-point, but the proportion.

gargsms:
Changing DPI did not occur to me, really - and with autofit, it doesn't have an affect on the rendered output either.I have a few improvements to do on this regarding gravity and positioning. I should be done in a day or two max.String manipulation was a hassle, yes. What are the style changes, exactly? I can confirm to the code convention if that is what you mean.  

    On Sunday, 17 September 2017 1:44 PM, John Cupitt <notifications@github.com> wrote:
 

 I had an idea: how about adjusting DPI rather than the font name? It seems like a more natural way to size text, and you'd avoid all the string manipulation.I would do a binary search, so:   
   - render at 72 dpi (the default)
   - too large? halve the dpi; too small? double the dpi
   - render again, until we have a dpi either side of the correct size
   - now find the midpoint between the too-large and too-small dpi and try that
   - keep finding midpoints until large - small is zero
You could make it converge a bit more quickly by assuming rendered size changes linearly with dpi and picking not the mid-point, but the proportion.—

jcupitt:
I think the PR should use glib more, so gboolean not stdbool, GSList not the intrusive linked list, C++-style `//` comments should be avoided, enums need a `LAST` element and a gtk-doc comment, and the formatting should be in line with the rest of libvips. 

I can make these changes, I realize they are not really functional, and that getting the behaviour right is the important thing.

My dpi thought was that if the user specifies width and height but not dpi, dpi could be chosen to fit the text to the box.

jcupitt:
What do you think about changing dpi rather than font size? Would that meet your needs for this feature?

gargsms:
@jcupitt I have addressed some code conventions, removed `stdbool`. I tried to use GSList in place of the self-implemented Linked List, but gave up in between. I couldn't find exactly where the PR did not match the formatting, so I haven't fixed any of that.

Functionally, this works perfect at the moment. DPI would have no effect for text that is to be autofit, but would work just as expected if the font size is specified, or the autofit conditions are not met.

I, on purpose, did not change DPI, because I would have needed to rewrite most of the code. This works as intended. I can do that if it is required.

jcupitt:
I like this feature, but I don't like all the string stuff with font size :( it feels wrong. I don't want to ask you to rewrite it, I'm sure you have better and more interesting things to do. 

How about we merge as is, then I'll switch it to change dpi instead? The effect would be the same (fit text to box), but the API would be slightly different, and you'd need to adapt upstream code a bit.

gargsms:
I have no issues with you changing it. How much of a rewrite is that?

gargsms:
Anyhow, I have another requirement on top of this. I would want to know the computed font size. ImageMagick exposes this in the metadata object for the image. Should we do the same here?
Also, since we support Pango Markup, is there a possibility to make the text a RGB image in place of 1-bit image? That should bring us on feature parity with ImageMagick.

jcupitt:
It shouldn't be much work to make it change DPI rather than font size, let me merge to a branch and try.

Yes, we could expose computed dpi / points as an output param.

The output image is 8-bit, so you can do anti-aliased blends into an image, but you're right, it's not colour. 

You can add colour in the blend, I guess you know, eg.

```ruby
image = Vips::Image.text("hello world").ifthenelse(image, [255, 0, 0], blend: true)
```

To write `hello world` in red text on top of an image, but of course you can't change colour during the text.

gargsms:
I have been using that to write the color to text, but it limits the possibilities with Pango Markup.
We can work on that accordingly once this is good enough. Thanks for the quick update.

gargsms:
Tried changing the DPI settings on the latest vips master. It works as expected. I will try to work that out as you suggested.

    On Sunday, 17 September 2017 9:17 PM, Sukant Garg <garg_sms@yahoo.in> wrote:
 

 Changing DPI did not occur to me, really - and with autofit, it doesn't have an affect on the rendered output either.I have a few improvements to do on this regarding gravity and positioning. I should be done in a day or two max.String manipulation was a hassle, yes. What are the style changes, exactly? I can confirm to the code convention if that is what you mean.  

    On Sunday, 17 September 2017 1:44 PM, John Cupitt <notifications@github.com> wrote:
 

 I had an idea: how about adjusting DPI rather than the font name? It seems like a more natural way to size text, and you'd avoid all the string manipulation.I would do a binary search, so:   
   - render at 72 dpi (the default)
   - too large? halve the dpi; too small? double the dpi
   - render again, until we have a dpi either side of the correct size
   - now find the midpoint between the too-large and too-small dpi and try that
   - keep finding midpoints until large - small is zero
You could make it converge a bit more quickly by assuming rendered size changes linearly with dpi and picking not the mid-point, but the proportion.—

jcupitt:
Oh dear, I had a go at changing the dpi:

https://github.com/jcupitt/libvips/tree/gargsms-text-autofit

It seems to work. I've not added gravity yet though.

gargsms:
Wow. This is so much shorter. I will check if this works for my requirements.
It would have been better if I discussed that with you before implementing. Anyhow, if this works for me, I can implement gravity to this.

jcupitt:
How about having gravity as a separate operation? You'd use it something like:

```ruby
mask = Vips::Image.text("hello", width: 100, height: 100).gravity(100, 100, direction: :north)
```

Now `text` expands to fit just within 100 x 100, then gravity expands to exactly 100 x 100, centred on the top edge. You could then combine `gravity` with other operations, like resize, perhaps. 

`gravity` is `embed` with x, y chosen for you, so it'd be simple to subclass that.

gargsms:
That makes sense, actually.

gargsms:
I tried your implementation. It works fine, but uses more iterations to fit the text, on an average, than the earlier implementation. I haven't yet tried to understand the code much.
I would make my app use your changes once they are in master. For now, I am using mine. Thanks.

jcupitt:
Yes, I didn't try to get the iterations down, I'll see if I can improve that. I'll add a `gravity` operator too.

gargsms:
Should I close this PR?

jcupitt:
I added a `gravity` operator too, so this works:

```ruby
mask = Vips::Image.text("hello", width: 100, height: 100).gravity(:north, 100, 100)
```

Would you mind testing it? It you say it's OK, I'll merge to master and we can close. 

Supporting multi-coloured text is a separate issue I think.

gargsms:
This seems to be work just fine. You should merge it and release.
Yes, working with coloured text is a different PR, or issue in itself.
Also, it is not necessary to credit me on this. Most of my code was useless anyway. :confused: 

Closing this.

gargsms:
@jcupitt There is one probable shortcoming to adjusting the DPI to find an exact match. The computed font size of the rendered text may not be an integer in that case. So if we want to expose that as a property of the image, we would need to floor that first and then render.
I implemented that here: https://github.com/gargsms/libvips/commit/6adbd9ef7ff4507bc0cff37279bea985fa99781d

jcupitt:
I'd forgotten about exposing the detected scale. Would it be enough to tag the dpi parameter as read-write? You can calculate font size from that.

jcupitt:
I tried adding a new output param, I now see:

```
$ irb
irb(main):001:0> require 'vips'
=> true
irb(main):002:0> x = Vips::Image.text "hello", width: 100, height:100, autofit_dpi: true
=> [#<Image 98x48 uchar, 1 bands, b-w>, {"autofit_dpi"=>248}]
```

jcupitt:
OK, merged to master! 8.6 soon, hopefully.

PR #706, opened 2017-08-08
Add C header to GIR file

astavale:
This patch adds the `--c-include` option to the `g-ir-scanner` options in the build system. It enables the main C header, `vips/vips.h`, to be identified in the GIR. This is needed by bindings generated for Vala.


jcupitt:
Looks good, thank you @astavale !

Issue #688, opened 2017-07-08
Resolution issue when rasterizing svg files through svgload


llwy:
Hello,

I found a strange behaviour in the svgload command: I never seem to get the requested rasterizing resolution.

To illustrate the issue, I created 3 svg files of 10mm, 25.4mm and 50mm width (Files sent to the mailing list - I could not find a way to add files to the bug report.).

if I run the bellow command on the 25.4mm file (1 inch) requesting 8000dpi resolution, I get a file which is 10000 pixels wide (where it should be 8000 since we are talking about a 1 inch file):
vips.exe extract_band F254mm.svg[dpi=8000] F254mm.tiff[compression=ccittfax4,squash,miniswhite,resunit=VIPS_FOREIGN_TIFF_RESUNIT_INCH] 2

If I run the same command requesting a 2000dpi resolution rasterization, I obtain a file which is 2500 pixel wide instead of the theoretical 2000 pixels.

For both cases, the ratio between the actual and theoretical number of pixels is exactly 1.25.

Now if I do the same experiment with the 10mm wide image, I get 3888 pixels instead of 3150 at 8000dpi and 972 pixels instead of 787 at 2000dpi. The ratio between the actual and theoretical number of pixels is 1.23444

Finally, the same experiment performed with the 50mm wide image, I get 19666 pixels instead of 15748 at 8000dpi and 4916 pixels instead of 3937 at 2000dpi. The ratio between the actual and theoretical number of pixels is 1.24879

What is strange is that for each input image size, I get a different ratio between the requested resolution and the obtained resolution. The difference between the ratio is big enough between those files that I cannot use one ratio to correct all resolution requests.

As an alternative test, I tried to use the "scale" parameter instead of the DPI parameter. The documentation says that the default resolution is 72dpi so I tried using the scale parameter set to 72/8000 for instance but still did not get the right amount of pixels. 
As a final test, I rasterized the 25.4mm file without any scale or dpi parameter and I got a 90 pixels wide tiff file instead of a 72dpi file.

This looks like a bug to me but maybe I am missing something?

Does anyone have an idea of what is hapening here?

Thank you in advance,

Fosk

jcupitt:
Hello, it looks like librsvg defaults to 90 DPI, but vips thinks it defaults to 72 DPI. I'll see if I can fix it.

jcupitt:
It seems to be working better now. With this patch I see:

```
[...]

PR #724, opened 2017-09-03
Correct scaling of SVG input without width and height attributes

lovell:
Hi John,

Since the changes for #688, made available in libvips v8.5.7+, SVG images without `width` and `height` attributes can no longer be scaled using the `dpi` and `scale` parameters of `vips_svgload`.

For example, the following commands all produce the same 620x472 output:

```sh
$ curl -O https://upload.wikimedia.org/wikipedia/commons/b/bd/Test.svg
$ vips copy Test.svg Test.png
$ vips copy Test.svg[dpi=150] Test-dpi-150.png
$ vips copy Test.svg[scale=2] Test-scale-2.png
$ vipsheader *.png
Test-dpi-150.png: 620x472 uchar, 4 bands, srgb, pngload
Test.png: 620x472 uchar, 4 bands, srgb, pngload
Test-scale-2.png: 620x472 uchar, 4 bands, srgb, pngload
```

This PR is a proposed solution for this problem.

When `dpi`/`scale` are set to non-default values, two pairs of `rsvg_handle_set_dpi` / `rsvg_handle_get_dimensions` calls are used to detect the most appropriate scaling method.

The first call checks the dimensions librsvg will render for the default 72DPI, the second call checks the dimensions for the desired dpi/scale. If the same values are returned for both calls, the SVG file did not specify a width and height, so any DPI value passed to `rsvg_handle_set_dpi` is ignored (I believe this behaviour is correct according to the SVG spec). Instead, this PR uses cairo to scale the image to the desired DPI/scale and restore the previous functionality.

With this change applied, dimensions are scaled as expected:

```sh
$ vipsheader *.png
Test-dpi-150.png: 1291x983 uchar, 4 bands, srgb, pngload
Test.png: 620x472 uchar, 4 bands, srgb, pngload
Test-scale-2.png: 1240x944 uchar, 4 bands, srgb, pngload
```

In addition, the `sharp` test suite also passes cleanly with this change where one of its SVG rendering tests currently fails when using libvips v8.5.7+.

### DIFF
--- a/libvips/foreign/svgload.c
+++ b/libvips/foreign/svgload.c
@@ -89,6 +89,10 @@ typedef struct _VipsForeignLoadSvg {
 	 */
 	double scale;
 
+	/* Scale using cairo when SVG has no width and height attributes.
+	 */
+	double cairo_scale;
+
 	RsvgHandle *page;
 
 } VipsForeignLoadSvg;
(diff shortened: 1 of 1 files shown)


lovell:
@jcupitt If you're getting v8.6.0 ready then this is probably something to consider, be it via this PR or an alternative method, as I'm aware SVG rendering is a well-loved feature of sharp+libvips.

jcupitt:
I agree, sorry, I've been distracted.

kleisauke:
I've tested this pull request and it works as intended. I would like to see this shipped in v8.6.0.

jcupitt:
I've finally got off my arse and read your patch @lovell, it looks great. Nice job!

PR #736, opened 2017-09-14
cplusplus: offset coords are int rather than double

lovell:
Hi John, I spotted this small API gotcha in the C++ bindings, mostly harmless but worth fixing even though it's a breaking change.


jcupitt:
Oooops, looks like a copy-paste error from xres/yres. Nice job!

jcupitt:
It's an ABI break, but almost everything about C++ breaks ABI all the time, so it's hopefully not an issue.

Issue #735, opened 2017-09-13
vips_thumbnail: SVG/PDF input broken (v8.6.0-alpha*)


kleisauke:
Currently there is a regression (from the `VIPS_SIZE_FORCE` mode) in libvips (8.6.0-alpha*) that prevents SVG/PDF input being processed.

Test images:
[SVG](https://raw.githubusercontent.com/lovell/sharp/master/test/fixtures/check.svg)
[PDF](https://t0.nl/pdf-sample.pdf)

SVG test:
```bash
vipsthumbnail check.svg \
    -o tn_%s.png
```

Output:
```bash
svgload: no property named `shrink'
```

PDF test:
```bash
vipsthumbnail pdf-sample.pdf \
    -o tn_%s.png
```

Output:
```bash
pdfload: no property named `shrink'
```

After taking a quick look in the source code, it seems that the [`vips_thumbnail_file_open`](https://github.com/jcupitt/libvips/blob/e346fd639e190c310e3671dd43f8c60553318235/libvips/resample/thumbnail.c#L697) function has a `shrink` parameter with the type `int`. I think it should be the type `double` (because the type was changed [here](https://github.com/jcupitt/libvips/commit/260180197d234c7a200c735350c29cc367184c88#diff-91d7263682820594ab7f1b7c450778c1R252)). [This check](https://github.com/jcupitt/libvips/blob/e346fd639e190c310e3671dd43f8c60553318235/libvips/resample/thumbnail.c#L701) needs also be changed to `if( shrink != 1.0 )`.

Maybe `pyvips` might need more `thumbnail` tests to prevent this in the future.

PR #737, opened 2017-09-14
vips_thumbnail: Fix SVG/PDF input (#735)

kleisauke:
Fix regression from https://github.com/jcupitt/libvips/commit/260180197d234c7a200c735350c29cc367184c88#diff-91d7263682820594ab7f1b7c450778c1R272.

Closes https://github.com/jcupitt/libvips/issues/735.


jcupitt:
Ah, mustn't set shrink _and_ scale, I see. Nice!

PR #739, opened 2017-09-17
Remove additional install of pkg-config and automake from macOS CI

astavale:
It look as though the Travis build for macOS is failing because brew errors
out when it finds pkg-config and automake have already been installed.


jcupitt:
Sounds good, let's try it.

PR #743, opened 2017-09-19
Update GObject introspection annotations for /libvips/arithmetic

astavale:
PR adds `(out)` for out parameters. Most functions take a single Image input and those have been marked as `(method)`. Functions that take two Image inputs, e.g. add, remain as functions, rather than methods. There are also a few `(array length=n)` additions.

### DIFF
--- a/libvips/arithmetic/measure.c
+++ b/libvips/arithmetic/measure.c
@@ -253,9 +253,9 @@ vips_measure_init( VipsMeasure *measure )
 }
 
 /**
- * vips_measure:
+ * vips_measure: (method)
  * @in: image to measure
- * @out: array of measurements
+ * @out: (out): array of measurements
  * @h: patches across chart
  * @v: patches down chart
  * @...: %NULL-terminated list of optional named arguments
--- a/libvips/arithmetic/abs.c
+++ b/libvips/arithmetic/abs.c
@@ -253,9 +253,9 @@ vips_abs_init( VipsAbs *abs )
 }
 
 /** 
- * vips_abs:
+ * vips_abs: (method)
  * @in: input #VipsImage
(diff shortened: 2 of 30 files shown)


libvips/arithmetic/measure.c
```diff
@@ -253,9 +253,9 @@ vips_measure_init( VipsMeasure *measure )
 }
 
 /**
- * vips_measure:
+ * vips_measure: (method)
  * @in: image to measure
- * @out: array of measurements
+ * @out: (out)(array): array of measurements
```
jcupitt: Is (array) correct? This is an output image whose pixels are the array elements, if you see what I mean.

astavale:
Hmm, I made the change and did a `git push --force` after squashing the change. May be there is a clearer way I could have made the change?

jcupitt:
Hello, that's strange, I think you just push to the branch. Maybe the squash confused it? 

I'll merge anyway and patch master.

jcupitt:
Huh, it's fine after merge. Maybe try without squash if this ever comes up again.

astavale:
You usually seem pretty quick to respond and I submitted the change within an hour of your comment so I wondered if you had missed it :)

GitHub will remove/hide the original commit when a forced push is done by the submitter. That's why there is a "Show outdated" button below your review comment. The squash (actually fixup) means there is just one commit for this change. As it was a simple change it made sense to me to keep the PR tidy. This is how other projects I've sent PRs for have worked. I can do a single line commit after review comments if that is better for you.

jcupitt:
Yes, I think a simple push would let github see it as a response to the review. github auto-squash commits on a PR, you don't normally need to do them yourself. 

Or that's my understanding :(

Issue #741, opened 2017-09-17
Improve GObject instrospection annotations


astavale:
#740 is a small test of updating the annotations for a function in libvips. This is to improve the Vala bindings to libvips, but that PR needs to be checked to find that it doesn't upset any other language bindings.

Given that any G-I annotation changes will be across many files it is probably best to outline what impact is acceptable:

### Out parameters 
Out parameters are used extensively in libvips. By adding `(out)` to the docblock the GIR (GObject Introspection Repository) is generated with the relevant metadata. This change is minimal and will hopefully have no impact on the other bindings such as Python.

An example annotation is:
`@pixel_average: (out): output pixel average`

### Methods
An example function is `avg`. Currently this takes an `Image` as a parameter, but really this is an instance parameter. By adding the `(method)` annotation the GIR is generated with this additional metadata. 

This does significantly change the API because a function is no longer static, so no longer requiring an explicit instance argument to be passed. For example `Vips.avg( my_image, out pixel_avg )` becomes `my_image.avg( out pixel_avg )`.

An example annotation is:
```
vips_avg: (method)
@in: input #VipsImage
```

### Variadic arguments
While the GIR will contain the details of variadic arguments, the GIR will add `introspectable="0"` to the metadata because GI does not support variadic arguments at the current time.

For Vala bindings this means some small additional work by adding Vala specific metadata when generating the VAPI. For other bindings the function may not be included because of the `introspectable="0"` attribute. This is probably a question of API design and how much benefit variadic arguments bring.

### Use of GError
This also relates to #707 and the comments made in the related lovell/sharp issue.

The use of GError enables `try...catch` blocks in Vala (`try...except` in Python). At the moment a lot of functions in libvips return an `int` to signal if the function has worked. This means an out parameter needs to be used to get the value. The API would be easier to use from higher level language bindings if the function returned the value and raised a GError if the function failed. GError also allows for additional details to be returned, including text.

Clearly any switch to GError is a major breaking change in API for libvips.

### Is GIR worth it?
GIR brings a large number of possible bindings, see [Languages Binding with GObject-Introspection](https://wiki.gnome.org/Projects/GObjectIntrospection/Users). Although they are in various states of usefulness.

I understand the [Haskell generator](https://wiki.haskell.org/GObjectIntrospection) is very good. Both the [Go generator](https://github.com/linuxdeepin/go-gir-generator) and [Ruby generator](https://github.com/mvz/gir_ffi) are active, having commits for 2017.

[...]

PR #748, opened 2017-09-26
Update GObject annotations for libvips/conversion directory

astavale:
Further work on #741 

Two queries came up when looking through `libvips/conversion`:

 1. Should `vips_bandjoin_const()` in `bandjoin.c` be `**in` instead of `*in` ? The docs state it is an array.

 2. The annotation for `vips_bandrank( VipsImage **in, VipsImage **out, int n, ... )` in `bandrank.c` is ` @in: (array length=n): array of input images`. Should this include `(tansfer none)`, similar to `vips_arrayjoin()` ?

### DIFF
--- a/libvips/conversion/bandjoin.c
+++ b/libvips/conversion/bandjoin.c
@@ -428,8 +428,8 @@ vips_bandjoin_constv( VipsImage *in, VipsImage **out,
 }
 
 /**
- * vips_bandjoin_const:
- * @in: (array length=n) (transfer none): array of input images
+ * vips_bandjoin_const: (method)
+ * @in: input image
  * @out: (out): output image
  * @c: (array length=n): array of constants to append
  * @n: number of constants
--- a/libvips/conversion/bandjoin.c
+++ b/libvips/conversion/bandjoin.c
@@ -204,7 +204,7 @@ vips_bandjoinv( VipsImage **in, VipsImage **out, int n, va_list ap )
 /**
  * vips_bandjoin:
  * @in: (array length=n) (transfer none): array of input images
- * @out: output image
+ * @out: (out): output image
  * @n: number of input images
  * @...: %NULL-terminated list of optional named arguments
  *
@@ -243,7 +243,7 @@ vips_bandjoin( VipsImage **in, VipsImage **out, int n, ... )
  * vips_bandjoin2:
  * @in1: first input image
  * @in2: second input image
- * @out: output image
+ * @out: (out): output image
  * @...: %NULL-terminated list of optional named arguments
  *
  * Join a pair of images together, bandwise. See vips_bandjoin().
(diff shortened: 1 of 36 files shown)


jcupitt:
Hi again, 

1. Nice find, this is a copy-paste error in the docs, it should just be a single image.

2. Sure, add a transfer-none, though I think that's the default.

jcupitt:
It's great you are adding this extra markup, I think it will improve the docs, but I'm worried you won't get what you want. I don't think adding this markup will give you a working Vala binding :( 

As I said, I think you will need to use the layer below with the (new / set / build / get) machinery.

jcupitt:
This set of changes look fine.

libvips/conversion/bandjoin.c
```diff
@@ -428,9 +428,9 @@ vips_bandjoin_constv( VipsImage *in, VipsImage **out,
 }
 
 /**
- * vips_bandjoin_const:
- * @in: (array length=n) (transfer none): array of input images
- * @out: output image
+ * vips_bandjoin_const: (method)
+ * @in: array of input images
```
jcupitt: Could you fix the comment line for @in as well? Something like "input image"

astavale: My apologies, I had done that, but I had re-based against master that caused a problem and had to make the changes again.

astavale:
The Vala binding is working, but could be more elegant. Vala has a tool, `vapigen`, that uses the GIR file to generate a VAPI file. The VAPI provides translation details to the Vala compiler so the right C code is generated for the bound library. If you've tried this yourself then the base class ObjectClass needed to be adjusted to inherit from GLib.ObjectClass. Also ArgumentTable needed to inherit from GLib.HashTable. I'll publish what I've done so far in the coming weeks.

There are a couple of advantages to using the GIR:

 * The VAPI is generated automatically with the `vapigen` tool, otherwise a VAPI needs to be written manually
 * The documentation for the VAPI is generated automatically. The Vala community have a process where GIRs are held in a public repository, https://github.com/nemequ/vala-girs/tree/master/gir-1.0, and then the documentation is re-generated periodically and appears on the Vala website https://valadoc.org/ I was hoping libvips would fit in to that process

I understand you wish to release in September. So I'm updating the annotations to hopefully fit in with that. This gives a workable binding for people to try out. Given time constraints I'm looking to update libvips/iofuncs and libvips/resample. That should then cover the most common use cases. I may consider a few more directories if I get the time.

Long term, I'm thinking there should probably be a GObject Introspection binding. While Vala can handle varargs, PyGObject doesn't. See - https://wiki.gnome.org/Projects/PyGObject/IntrospectionPorting#Non-introspectable_functions.2BAC8-methods . PyGObject does however allow properties to be set at construct time, which is like keyword args. See - https://wiki.gnome.org/Projects/PyGObject/IntrospectionPorting#Constructors Vala allows for an Object( ... ) keyword, but the only examples I'm finding are in the constructor itself. So I need to investigate that further. I'm mentioning Python here because I understand the typelib file is only generated from a GIR.

jcupitt:
Oh OK, perhaps it's not an issue. I was concerned that you would be unable to use the optional arguments from Vala, but if it can call varargs functions, perhaps it's going to work. 

There's a Python binding based on pygobject here, you probably saw:

https://github.com/jcupitt/libvips/tree/master/python

But it's rather unsatisfactory for various reasons, so I'm planning to deprecate it and push people towards a plain FFI binding instead.

PR #749, opened 2017-09-26
Update GObject introspection annotations for libvips/foreign directory

astavale:
More for #741 

Apart from the `(method)` and `(out)` annotations, I've added some annotations for buffers. For example `vips_magickload_buffer`:

```
@buf: (array length=len) (element-type guint8): memory area to load
@len: (type gsize): size of memory area
```

`(element-type guint8)` has been used for other functions already, so it should be fine. g-ir-compiler ignored `@len` being set to type `size_t`, it used `gulong` instead. So I used `gsize` instead. From https://developer.gnome.org/glib/stable/glib-Basic-Types.html#gsize it states `gsize` is "An unsigned integer type of the result of the sizeof operator, corresponding to the size_t type defined in C99"

### DIFF
--- a/libvips/foreign/analyzeload.c
+++ b/libvips/foreign/analyzeload.c
@@ -149,7 +149,7 @@ vips_foreign_load_analyze_init( VipsForeignLoadAnalyze *analyze )
 /**
  * vips_analyzeload:
  * @filename: file to load
- * @out: decompressed image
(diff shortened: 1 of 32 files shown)


jcupitt:
Huh the size_t thing is weird, you'd think it would spot that. 

Otherwise it all looks fine!

PR #752, opened 2017-09-30
Update GObject introspection annotations for libvips/resample directory

astavale:
#741 

`resample.h` had to be added to `headers` in `configure.ac` for the functions to appear in the GIR. I assumed that `resample.h` was part of the public VIPS8 API.

There are also a couple of C++ files: reducev.cpp and reduceh.cpp. These look as though they have public functions, but aren't part of `instrospection_sources` in `configure.ac`. I've not made any changes to the build system for these two files.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -113,6 +113,7 @@ headers="\
 	morphology.h \
 	type.h \
 	rect.h \
+	resample.h \
 	memory.h \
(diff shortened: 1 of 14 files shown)


jcupitt:
You're right, those cpp files should go in too.

PR #754, opened 2017-09-30
Update GObject introspection annotations for libvips/convolution dire…

astavale:
…ctory

#741


jcupitt:
Your head must be spinning!

PR #760, opened 2017-09-30
Update GObject introspection annotations for libvips/mosaicing directory

astavale:
#741 

`mosaicing.h` had to be added to `headers` in `configure.ac`


jcupitt:
You are on fire!

PR #765, opened 2017-10-03
Faster and more accurate integral nearest-neighour upsampling

lovell:
Hi John, you'll be pleased to learn that sharp will soon be moving to use `vips_resize` (instead of 4x separate `vips_shrink*` and `vips_reduce*` operations in duplicate code).

One feature of sharp missing from libvips is this faster, more accurate integral nearest-neighbour upsampling path using `vips_zoom`.

See https://github.com/lovell/sharp/issues/752 for the background.


jcupitt:
Nice!

PR #766, opened 2017-10-08
Add ldconfig step to building instructions

lovell:
...as I noticed this step was missing when helping to investigate https://github.com/lovell/sharp/issues/980


jcupitt:
It's not always necessary, but I guess it's harmless to add. Thanks!

PR #810, opened 2017-11-19
Smartcrop: clamp sigma to avoid out of range warnings

lovell:
Hi John, this prevents large reductions from generating `GLib-GObject-WARNING **: value "0.000453" of type 'gdouble' is invalid or out of range for property 'sigma' of type 'gdouble' gaussblur: parameter sigma not set`.

```sh
$ vips smartcrop 10000x10000.v out.v 1 1
```


jcupitt:
Looks good!

PR #816, opened 2017-11-24
Colour: ~20% sRGB to scRGB speedup for 3/4 channel images

lovell:
Hi John, this PR moves some of the logic in `vips_sRGB2scRGB_gen` so the `BandFmt` conditional is outside the outer for-loop. It then adds optimised paths to the inner for-loop for common 3 (RGB) and 4 (RGBA) band input images. This reduces the instruction count per pixel by 5, including the removal of 2 branches.

Before:
```sh
$ time vips colourspace 10000x10000.v out.v scrgb
real    0m9.478s
user    0m1.340s
sys     0m1.240s
```
```
3,900,000,014  ???:vips_col_sRGB2scRGB_8 [/usr/local/lib/libvips.so.42.8.0]
2,301,250,004  ???:vips_sRGB2scRGB_gen [/usr/local/lib/libvips.so.42.8.0]
1,200,211,873  /build/glibc-bfm8X4/glibc-2.23/string/../sysdeps/x86_64/multiarch/memcpy-avx-unaligned.S:__memcpy_avx_unaligned [/lib/x86_64-linux-gnu/libc-2.23.so]
```

After:
```sh    
$ time vips colourspace 10000x10000.v out.v scrgb
real    0m9.071s
user    0m1.048s
sys     0m1.024s
```
```
3,900,000,014  ???:vips_col_sRGB2scRGB_8 [/usr/local/lib/libvips.so.42.8.0]
1,801,200,004  ???:vips_sRGB2scRGB_gen [/usr/local/lib/libvips.so.42.8.0]
1,200,211,873  /build/glibc-bfm8X4/glibc-2.23/string/../sysdeps/x86_64/multiarch/memcpy-avx-unaligned.S:__memcpy_avx_unaligned [/lib/x86_64-linux-gnu/libc-2.23.so]
```

### DIFF
--- a/libvips/colour/sRGB2scRGB.c
+++ b/libvips/colour/sRGB2scRGB.c
@@ -73,6 +73,24 @@ typedef VipsOperationClass VipssRGB2scRGBClass;
 
 G_DEFINE_TYPE( VipssRGB2scRGB, vips_sRGB2scRGB, VIPS_TYPE_OPERATION );
 
+/* Convert an 8-bit pixel.
+ */
+static void
+vips_sRGB2scRGB_pel_rgb_8( float * restrict q,
+	VipsPel * restrict p )
+{
+	const int r = p[0];
+	const int g = p[1];
+	const int b = p[2];
+
+	float R, G, B;
+	vips_col_sRGB2scRGB_8( r, g, b, &R, &G, &B );
+
+	q[0] = R;
+	q[1] = G;
+	q[2] = B;
+}
+
 /* Convert a buffer of 8-bit pixels.
  */
 static void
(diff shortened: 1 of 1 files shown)


libvips/colour/sRGB2scRGB.c
```diff
@@ -113,27 +152,30 @@ vips_sRGB2scRGB_line_16( float * restrict q, unsigned short * restrict p,
...
+	if( extra_bands == 0 ) {
+		for( i = 0; i < width; i++ ) {
+			vips_sRGB2scRGB_pel_rgb_16(q, p);
+			p += 3;
+			q += 3;
+		}
+	} else if( extra_bands == 1 ) {
+		for( i = 0; i < width; i++ ) {
+			vips_sRGB2scRGB_pel_rgb_16(q, p);
+			q[3] = p[3];
```
jcupitt: I think this needs a /256 to match the loop case

libvips/colour/sRGB2scRGB.c
```diff
@@ -81,28 +99,49 @@ vips_sRGB2scRGB_line_8( float * restrict q, VipsPel * restrict p,
...
-		p += 3;
-
-		vips_col_sRGB2scRGB_8( r, g, b, &R, &G, &B );
+	if( extra_bands == 0 ) {
+		for( i = 0; i < width; i++ ) {
+			vips_sRGB2scRGB_pel_rgb_8(q, p);
+			p += 3;
+			q += 3;
+		}
+	} else if( extra_bands == 1 ) {
```
jcupitt: nitpick: vips usually has the "else" on a newline, K&R style

jcupitt:
Looks good!

Off-topic: I did some performance work comparing against pillow-simd 4.3 recently. libvips (in single-threaded mode) and pillow-simd are generally close, but pillow-simd does win on downsize by about 30 - 40%.

I think one of the things they do well is using the RGBX mode of libjpeg-turbo. In this mode, libjpeg outputs 4-byte pixels, with the fourth byte always 255. This makes pixels a power of two in size and means pixels are always four-byte aligned -- they can therefore vectorize quite a bit more.

I think we'd need to add `VIPS_INTERPRETATION_RGBX`, have a `rgbx` option on `jpegload`, and make `jpegsave` spot RGBX and reenable for write.

Anyway, it would be interesting to experiment with that for 8.7.

jcupitt:
This branch adds `reducevt`, a version of reducev that transposes the image as it writes. So you run it twice: the first time with the vertical shrink factor to make a vertically-shrunk and then transposed image, the second time with the horizontal shrink factor to shrink the other axis (which is now vertical) and transpose again on write to get the image the right way up again. It's very confusing :(

Anyway, because you use the same operation for both axes, and that operation is vertical, you can vectorise the whole of the reduce phase. Sadly, because RGB pixels are not aligned, it ends up slower than what we have now, because of the scattered three-byte write for the transpose.

With RGBX I think it would have a chance of pulling ahead! Worth a try anyway.

jcupitt:
Ooop, forgot the actual link:

https://github.com/jcupitt/libvips/tree/add-reducevt

jcupitt:
Another obvious optimisation is to have a special path for u8 and u16 data. vips_col_sRGB2scRGB() has the CLIP in there, which is slow and obviously unnecessary if we know the range of the source.

lovell:
Thanks for merging John - the 3+1 channel approach will definitely help auto-vectorisation.

jcupitt:
I chopped it about a bit to move the LUT lookup inline and remove the clipping, it's a bit faster again now.

PR #824, opened 2017-12-06
Allow compilation of debug build

lovell:
Hi John, this small PR removes compilation errors from the following:
```sh
CFLAGS="-DDEBUG" CXXFLAGS="-DDEBUG" ./configure && make
```


jcupitt:
Good!

PR #832, opened 2017-12-16
Use ChangeLog date instead of build date

bmwiedemann:
in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good.

This date call works with GNU date and BSD date.

Without this patch, every build of the 'vips' openSUSE package would differ
in version.h like this:

```diff
-#define VIPS_VERSION_STRING    "8.5.9-Thu Dec 14 22:40:35 UTC 2017"
+#define VIPS_VERSION_STRING    "8.5.9-Sun Jan 20 11:56:52 UTC 2019"
```

and differ in the resulting /usr/lib64/libvips.so.42.7.8


jcupitt:
Oh, nice!

Issue #620, opened 2017-03-14
'gifsave'


brian-armstrong-discord:
Hi!

I was wondering if you have plans to use giflib's save/writer to create a gifsaver in libvips. I've had a look through and it seems like it may be possible, but I don't know if there are any major underlying conceptual blockers that would inherently prevent this from being possible. I tried to read through some of gifload.c and pngsave.c to get some idea of what might be involved but I'm not sure I entirely understand still how libvips works :)

jcupitt:
Hi, thanks for suggesting this improvement. 

gifsave is quite hard to do well, unfortunately. 

libvips treats all images as simple C-style arrays. GIF images are all indexed, so each pixel is a single byte which holds an index into a colour palette, and each frame just holds the set of differences to the previous frame. libvips has to expand each frame out into a complete RGB image and it loses all the GIF structure. 

For gifsave, you'd need to analyze the image and recreate the structure that was thrown away. You'd need to do some clustering to find the best 256 colours in each frame, and you'd need to analyze inter-frame differences to find the minimum set of rectangular changes. giflib just provides the basic write facilities, it doesn't include any of this analysis. 

There are probably libraries which implement this analysis and which libvips could just call out to, but I've not investigated them. We'd need to experiment with a couple and see which was best.

jcupitt:
A simple solution might be to add image write via ImageMagick / GraphicsMagick. We have a load - via - libMagick, we'd just need a matching save.

We wrote a loader for the new IM7 API last summer, that would be a good starting point:

https://github.com/jcupitt/libvips/blob/master/libvips/foreign/magick7load.c

lovell:
A few other options for possible experimentation:

1. giflib v5+ re-added an undocumented [quantize.c](https://sourceforge.net/p/giflib/code/ci/master/tree/lib/quantize.c) feature, although we should heed the "doesn't really belong in the core library" comment.

2. I used [exoquant](https://github.com/exoticorn/exoquant) to good effect in the original work for [attention](https://github.com/lovell/attention) and can recommend it.

3. The quantisation feature of pngquant is available as the GPLv3 [libimagequant](https://github.com/ImageOptim/libimagequant)

Detaching the quanitisation step from the write step could allow for things like saving indexed PNG in addition to GIF.

brian-armstrong-discord:
[...]

PR #833, opened 2017-12-17
Add support for writing files with ImageMagick

dlemstra:
This PR adds support for writing files with ImageMagick as requested in #620. Today was the first day I took a dive into the code of libvips so I am pretty sure that this needs a proper review. I did not add anything to the C++ api and I probably missed some other stuff. 

I can add additional changes when requested. But also feel free to accept this PR in it's current form and make the changes yourself @jcupitt.


jcupitt:
Hi Dirk, that's great! Your code looks very nice. I'll try to have a proper look this week. I can't today, unfortunately. I had some quick thoughts:

I'd put all the code in `magicksave.c`, I think, the way `magick7load.c` does it. The separate `magick2vips.c` file is there for vips7 compatibility, but you don't need that. 

You might need separate savers for magick6 and magick7 -- I don't know how large the API differences are. The magick6 one ought to work with graphicsmagick too, but it would need testing. 

We will need configure flags to enable magick save separately from magick load, but that can be added later, of course. For security reasons, many users will not want to allow load via imagemagick, but will want imagemagick save.

dlemstra:
I wrote the code in a way that the api differences are in methods that have a different implementation for im6 and im7. I could add the new defines for saving and use the old value. You could then move those to configure, don't really have experience with that. How do `MAGICKSAVE` and `MAGICK7SAVE` sound? I will move all the code to `magicksave.c`. Not sure what you mean by vips7 compatibility though.

dlemstra:
I decided to use a slightly different approach. I have kept both `HAVE_MAGICK` and `HAVE_MAGICK7` which can be used to determine if we are building with IM6 or with IM7. And two new defines called `HAVE_MAGICKLOAD` and `HAVE_MAGICKSAVE` have been added. These can be used to toggle the save and load methods separately. I have now added those two new defines in various source files but this should be moved to a configure option. Could you handle that part @jcupitt? You can find the spots where I have added it with the following search text: `Should be removed and added as a configure option`.

jcupitt:
Hello again, I built your branch, it works well! I noticed one small thing: one-band images (monochrome) are saved as RGB, but with the image data in R, and GB zero, so they come out red.

How would you like to move forward on this? I'm happy to merge to master and do further improvements as a set of small issues. It would widen the group of testers, which is probably the most useful thing to do.

jcupitt:
OK, I've started tweaking a bit and merging to master. I'll try to get it done tonight. 

Thank you very much for contributing this nice thing, Dirk! Let's address any remaining things in issues.

dlemstra:
I can take a look at the issue tonight. @kleisauke send me info on how to test it. Did you try this with IM6 or IM7? Feel free to merge this so you can do the configure stuff and I can address the grayscale problem in a separate pull request.

dlemstra:
@jcupitt Just pushed a fix for the grayscale issue.

jcupitt:
OK, I've merged to master! This is great, thank you again Dirk.

I made a couple of large-ish changes:

1. There was a separate `Write` struct to manage resources during a write operation. I've folded this into the main `VipsForeignSaveMagick` class, and made the operations on `Write` members of that class. This makes things a bit simpler. I moved some more stuff into the base magicksave class too. 

2. There's going to need to be some code sharing between read and write via libMagick, especially for library initialization, so I split your little wrapper functions out into a separate file, `magick.c`.

Plus some very small style things, and I've bumped the version to 8.7.

I'll make an issue with a TODO list of the things that still need resolving.

jcupitt:
TODO list here:

https://github.com/jcupitt/libvips/issues/839

Please add any more problems you think of.

PR #842, opened 2018-01-02
Perf: ~15% improvement in conversion to greyscale

lovell:
Hi John, this slight reordering of the logic in `vips_col_scRGB2BW` reduces the number of costly `isinf` and `isnan` calls from 6 to 2 per pixel.

Before:
```
436,545,000  libvips/colour/LabQ2sRGB.c:vips_col_scRGB2BW [/usr/local/lib/libvips.so.42.8.1]
```
After:
```
339,535,000  libvips/colour/LabQ2sRGB.c:vips_col_scRGB2BW [/usr/local/lib/libvips.so.42.8.1]
```
I see a ~15% performance improvement as a result.

(This PR targets the `8.6` branch as the latest *magick changes on the `master` branch are not compiling cleanly on my machine and I've yet to investigate why.)

### DIFF
--- a/libvips/colour/LabQ2sRGB.c
+++ b/libvips/colour/LabQ2sRGB.c
@@ -350,23 +350,18 @@ vips_col_scRGB2BW( int range, int *lut, float R, float G, float B,
 	int Yi;
 	float v;
 
-	/* RGB can be Nan, Inf etc. Throw those values out, they will break
-	 * our clipping.
-	 *
-	 * Don't use isnormal(), it is false for 0.0 and for subnormal
-	 * numbers. 
+	/* The usual ratio. We do this in linear space before we gamma.
 	 */
-	if( VIPS_ISNAN( R ) || VIPS_ISINF( R ) ||
-		VIPS_ISNAN( G ) || VIPS_ISINF( G ) ||
-		VIPS_ISNAN( B ) || VIPS_ISINF( B ) ) {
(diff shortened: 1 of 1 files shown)


jcupitt:
Oh, clever! I didn't realize isnan etc. were slow - you'd think it would be a simple operation. 

I'll do an 8.6.1 today with this stuff in.

lovell:
Thanks John, looking forward to it.

jcupitt:
I came across some macros in ITK to do `isnan` etc. as a series of AND and OR bit tests. They might give you a few more % on this operation. I'll dig out a link.

jcupitt:
Look at this lovely thing I'm having to work on:

https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/ThirdParty/VNL/src/vxl/core/vnl/vnl_math.cxx#L124

```c++
// Assume IEEE floating point number representation
bool isnan( float x) {return bMe(&x,0x7f800000L,sz_f)&&bMp(&x,0x007fffffL,sz_f);}
bool isnan(double x) {return bMe(&x,0x7ff00000L,sz_d)&&(bMp(&x,0x000fffffL,sz_d)||bMp(&x,0xffffffffL,1-sz_d));}
```

Would something like that help benchmarks, do you think?

lovell:
There's only one way to find out :)

A possible baseline for this might be to specify `-ffast-math` as this will "optimise" away the existing use of `__builtin_isnan` as always returning false.

PR #864, opened 2018-01-25
composite.cpp: silence 'missing sentinel' warnings in GCC v6+

lovell:
Hi John, this quick fix uses the same approach as #668.

```
composite.cpp: In function 'int vips_composite_base_build(VipsObject*)':
composite.cpp:983:39: warning: missing sentinel in function call [-Wformat=]
    if( vips_addalpha( in[i], &x, NULL ) )
                                       ^
composite.cpp:1035:39: warning: missing sentinel in function call [-Wformat=]
    composite->compositing_space, NULL ) )
                                       ^
```

### DIFF
--- a/libvips/conversion/composite.cpp
+++ b/libvips/conversion/composite.cpp
@@ -980,7 +980,7 @@ vips_composite_base_build( VipsObject *object )
 		if( !vips_image_hasalpha( in[i] ) ) {
 			VipsImage *x;
 
-			if( vips_addalpha( in[i], &x, NULL ) )
+			if( vips_addalpha( in[i], &x, (void *) NULL ) )
 				return( -1 );
 			g_object_unref( in[i] );
 			in[i] = x;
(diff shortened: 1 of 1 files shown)


jcupitt:
Nice!

What flags are you using to see these warnings? I usually build with `g++ -Wall` and I don't seem to get them here with gcc7.

lovell:
Thanks John, the warning suggests `-Wformat` is involved.

Interestingly when specifying `-Wall` on Alpine/gcc6 this is set to `0x1`:
```
gcc version 6.4.0 (Alpine 6.4.0) 
$ g++ -Wall -Q --help=warnings | grep format=
  -Wformat=                   		0x1
```
but on Ubuntu/gcc7 it is `1`:
```
gcc version 7.2.0 (Ubuntu 7.2.0-8ubuntu3) 
$ g++ -Wall -Q --help=warnings | grep format=
  -Wformat=                   		1
```
The former reports this warning but the latter does not. Confused.

jcupitt:
I guess there's something messed up in alpine's libc. Are they not using glibc, perhaps? Not our problem anyway, phew.

PR #871, opened 2018-02-03
Make size_t to int cast explicit to prevent compiler warning

lovell:
Silences this warning seen when compiling with MSVC.

> cplusplus\include\vips\VImage8.h(477): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data


jcupitt:
Huh I wonder why gcc didn't warm about that. Useful!

PR #905, opened 2018-03-19
Add is_a support to SVG file class loader

lovell:
Hi John, I noticed the other day that there is currently no `is_a` check in the SVG file loader.

```sh
$ file what-is-it
what-is-it: SVG Scalable Vector Graphics image

$ head -3 what-is-it
<?xml version="1.0" encoding="UTF-8"?>
<svg version="1.2" baseProfile="tiny" xml:id="svg-root" width="100%" height="100%"
  viewBox="0 0 480 360" xmlns="http://www.w3.org/2000/svg"

$ vipsheader what-is-it
vipsheader: VipsForeignLoad: "what-is-it" is not a known file format
```

This PR adds a new internal method so that the above now works:

```sh
$ vipsheader what-is-it
what-is-it: 480x360 uchar, 4 bands, srgb, svgload
```

Rather than massively re-order functions in the `svgload.c` source file, I've declared the existing `vips_foreign_load_svg_is_a_buffer` function before using it in the proposed `vips_foreign_load_svg_is_a` function as it currently appears after the `vips_foreign_load_svg_file_class_init` function.

I'm happy to update this PR to re-order functions and avoid the declaration, but that would be a bigger change and I wanted to get the gist of the proposal understood first.

### DIFF
--- a/libvips/foreign/svgload.c
+++ b/libvips/foreign/svgload.c
@@ -350,82 +350,6 @@ typedef VipsForeignLoadSvgClass VipsForeignLoadSvgFileClass;
 G_DEFINE_TYPE( VipsForeignLoadSvgFile, vips_foreign_load_svg_file, 
 	vips_foreign_load_svg_get_type() );
 
-static int
-vips_foreign_load_svg_file_header( VipsForeignLoad *load )
-{
-	VipsForeignLoadSvg *svg = (VipsForeignLoadSvg *) load;
-	VipsForeignLoadSvgFile *file = (VipsForeignLoadSvgFile *) load;
-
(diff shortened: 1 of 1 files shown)


jcupitt:
Oh huh that's an odd oversight.

Yes, I'd move the `is_a` function above the use of it in file. You could remove the _buffer suffix too, since it's no longer really tied to the buffer loader.

lovell:
Thanks John, I'll re-order the functions and update this PR.

lovell:
Re-ordered and updated, thank you.

jcupitt:
Great! And nice to see you last night.

PR #906, opened 2018-03-20
Formatting fix on --vips-cache-trace

janko-m:


jcupitt:
Hello Janko, this `.xml` file is generated from the matching `.md` file:

https://github.com/jcupitt/libvips/blob/master/doc/How-it-opens-files.md

But libvips master has had this section removed -- it didn't seem very useful :(

janko-m:
Oh, yeah, I was confused that the `.md` file is missing the section. Ok, closing this then 😃

PR #913, opened 2018-03-25
Remove mention of deprecated radius parameter in vips_sharpen()

janko-m:
I noticed that the `radius` parameter has been deprecated, and it's not listed in the parameter list anymore, so I also removed its mention from the documentation.


jcupitt:
Nice!

PR #914, opened 2018-03-26
Add "(out)" to "out" parameter in vips_addalpha()

janko-m:
I don't know if this is used for function introspection (and in this case if it's related to https://github.com/jcupitt/ruby-vips/issues/168), but I saw all other functions having `(out)` specified like this.


jcupitt:
Sure, why not. I added `(method)` too.

The language bindings no longer use gobject introspection, so this change won't make `addalpha` appear.

PR #915, opened 2018-03-27
WIP: Adds a region shrink method

harukizaemon:
Addresses issue https://github.com/jcupitt/libvips/issues/859 to add a mechanism for preserving the domain of input pixel values when performing `dzsave` and `tiffsave`.

The solution allows us to build image viewers with tools that can manipulate individual phases of very large (multi-GiB) segmented images via deep zoom and pyramid TIFF representations.

* Adds a `VipsRegionShrink` enum with values `VIPS_REGION_SHRINK_MEAN`, `VIPS_REGION_SHRINK_MEDIAN`, and `VIPS_REGION_SHRINK_MODE`

* Adds `VipsRegionShrink` as parameter `method` to `vips_region_shrink()`

* Existing calls to `vips_region_shrink()` default to `VIPS_REGION_SHRINK_MEAN` to effect the current behaviour

The new `VIPS_REGION_SHRINK_MEDIAN` and `VIPS_REGION_SHRINK_MODE` methods are implemented so as to perform well, and such that we can make the following guarantees:

* *ONLY* available for non-complex uncoded images pixel types
* *ALWAYS* draws from the input values
* *NEVER* interpolates
* *NOT* stable with respect to the ordered set of input values
* *IS* stable with respect to the initial arrangement of input values

For example, given every possible permutation of the input pixels values `1, 3, 7, 11`, we would generate the following output values:

```
[1, 3, 7, 11] => 3
[1, 3, 11, 7] => 3
[1, 7, 3, 11] => 7
[1, 7, 11, 3] => 7
[1, 11, 3, 7] => 7
[1, 11, 7, 3] => 7
[3, 1, 7, 11] => 3
[3, 1, 11, 7] => 3
[3, 7, 1, 11] => 7
[3, 7, 11, 1] => 7
[3, 11, 1, 7] => 7
[3, 11, 7, 1] => 7
[7, 1, 3, 11] => 7
[7, 1, 11, 3] => 7
[7, 3, 1, 11] => 7
[7, 3, 11, 1] => 7
[7, 11, 1, 3] => 3
[7, 11, 3, 1] => 3
[11, 1, 3, 7] => 7
[11, 1, 7, 3] => 7
[11, 3, 1, 7] => 7
[11, 3, 7, 1] => 7
[11, 7, 1, 3] => 3
[11, 7, 3, 1] => 3
```

As you can see, there is a slight bias towards the upper median value.


harukizaemon:
Still to do:

- [ ] Make options available for `dzsave` and `tiffsave` via C/C++ APIs
- [ ] Make options available for `dzsave` and `tiffsave` via vips command line
- [ ] Make options available for `dzsave` and `tiffsave` via Ruby/Python APIs

harukizaemon:
Ugh, I've just noticed that, due to editor settings, end-of-lines and whitespace changes have also occurred. Apparently GitHub supports adding `?w=1` to the URL to ignore them. Alternatively, I can redo if that would be more acceptable.

jcupitt:
minmax is an interesting solution! I hadn't thought of that. 

Yes, libvips style is hard tabs set to 8 spaces and a strict 80 char line (so K&R).

Could you switch to `/*`? `//` comments are C99, which is not supported everywhere, sadly. 

The language bindings should all be automatic -- you just need to add an optional param to `dzsave` and `tiffsave`.

jcupitt:
Argh the whitespace changes do make your patch difficult to read, but `?w=1` makes it OK. All looks good otherwise. 

The test suite is part of pyvips, you probably saw. I'd run that to verify that `dzsave` etc. still work.

```
john@kiwi:~/GIT/pyvips (master)$ pytest tests/test_foreign.py 
============================== test session starts ==============================
platform linux2 -- Python 2.7.13[pypy-5.8.0-final], pytest-3.2.1, py-1.4.34, pluggy-0.4.0
rootdir: /home/john/GIT/pyvips, inifile: tox.ini
collected 19 items                                                               

tests/test_foreign.py ...................

=========================== 19 passed in 8.19 seconds ===========================
```

harukizaemon:
I’ll rebase and re-apply the patches to fix the white space and other stylistic issues, I’ll run the tests and check those too, and I’ll add the optional params.

harukizaemon:
I intuited the minmax but had to a) write a test to very and then b) rationalise that I was happy with the consequences :)

harukizaemon:
@jcupitt OK, I've fixed the whitespace. Sorry about that!

It's probably because it's late but I'm not sure how best to expose the methods as optional params and then plumb them all the way through.

AFAICT, `vips_region_shrink()` is a fair way down. For example: `vips_foreign_save_dz_class_init()` -> `vips_foreign_save_dz_build()` -> `pyramid_strip()` -> `strip_arrived()` -> `strip_shrink()` -> `vips_region_shrink()`.

Do you have a suggestion for how _you_ would do it? Again, I'm more than happy to do the leg work here.

Cheers,

Simon

harukizaemon:
I ran the tests locally and got a bunch of skipped tests:

```
❯ pytest tests/test_foreign.py 
==================================================================== test session starts =====================================================================
platform darwin -- Python 3.6.4, pytest-3.5.0, py-1.5.3, pluggy-0.6.0
rootdir: jcupitt/pyvips, inifile: tox.ini
collected 19 items                                                                                                                                           

tests/test_foreign.py ....s..s.ss........                                                                                                              [100%]

============================================================ 15 passed, 4 skipped in 6.99 seconds ============================================================
vips_image_dispose: temp-5 float2rad computed 200%
vips_image_dispose: temp-16 relational_const computed 800%
memory: high-water mark 60.23 MB
```

jcupitt:
The skipped tests will be for things like webp load and save, which perhaps your libvips is missing. You can run a particular test with:

```
$ pytest tests/test_foreign.py::TestForeign::test_dzsave
```

To add an optional param:

1. Add a class member to `struct _VipsForeignSaveDz` just after `compression`.
2. Give it a default value in `vips_foreign_save_dz_init()`. 
3. Declare it as an optional param in `vips_foreign_save_dz_class_init()` -- copy-paste one of the other enum args as a model. Your enum will be automatically wrapped as `VIPS_TYPE_REGION_SHRINK`.
4. In `strip_shrink()` access the class member as `layer->dz->region_shrink`, or whatever you called it.

And that should be all you need to do, I think. 

The tiff one is harder because you need to pass the value down through a plain C layer left over from vips7, but the idea is the same.

jcupitt:
A friend suggested this very clever hack:

```c++
#include <stdio.h>

struct int4 {
    int v[4];
};

int mode4(int4 input)
{
    int b0 = (input.v[0] == input.v[1]) | (input.v[0] == input.v[2]) | (input.v[0] == input.v[3]);
    int b1 = (input.v[1] == input.v[0]) | (input.v[1] == input.v[2]) | (input.v[1] == input.v[3]);
    int index = ((~b0) & 0x1) + (~(b0 ^ b1) & 0x1);
    return input.v[index];
}

int main()
{
    printf("%i\n", mode4({1, 1, 2, 3})); // 1
    printf("%i\n", mode4({1, 2, 1, 3})); // 1
    printf("%i\n", mode4({1, 2, 3, 1})); // 1
    printf("%i\n", mode4({2, 1, 1, 3})); // 1
    printf("%i\n", mode4({2, 1, 3, 1})); // 1
    printf("%i\n", mode4({2, 3, 1, 1})); // 1
    printf("%i\n", mode4({1, 1, 1, 2})); // 1
    printf("%i\n", mode4({1, 1, 2, 1})); // 1
    printf("%i\n", mode4({1, 2, 1, 1})); // 1
    printf("%i\n", mode4({2, 1, 1, 1})); // 1
    printf("%i\n", mode4({1, 1, 2, 2})); // 1
    printf("%i\n", mode4({1, 2, 1, 2})); // 2
    printf("%i\n", mode4({1, 2, 3, 4})); // 3
    return 0;
}
```

A fast mode4. It might be worth considering.

harukizaemon:
Oh wow, nice. Yes. I like it! Will have a go now.

harukizaemon:
Here are some sample tiles from a segmented 4070x3052 SEM:

Using mean:
segmented tile using mean

Using median:
segmented tile using median

jcupitt:
Nice! Do you have any thoughts about how median and mode compare for speed or quality?

jcupitt:
Re. enum, I don't like the name either, but I'm unsure what else to call it.

harukizaemon:
>Do you have any thoughts about how median and mode compare for speed

I ran all three against that same 12MiB (12,422,086) image, the run times were pretty much the same for all three.

>or quality?

For the 8-bit segmented images that are really the driver of this, both median and mode give pretty good results. I have a slight bias towards the median but I'm not sure why. I think perhaps because the median feels more "stable" but that is again nothing more than an unqualified bias :)

Of course both lead to a slight loss of apparent features—given the loss of gradient you get with the mean—but again, given the desired use with deep zoom images, there's always going to be some kind of feature loss no matter what you do. We are also using PNG tiles to avoid further feature loss and importantly artefact creation you get with JPEG.

I don't know how broad the appeal of this is likely to be but assuming you feel it is something you could live with, it's definitely something we will want to use specifically in a tool that allows the viewer to switch the 8 or so phases of a multi-GiB segmented images on and off but I don't think there's a lot of value for more general purpose use such as with colour, JPEG, or even  images or even 16 bpp/float grayscale images.

The real _value_ for us is that in order to do this after the fact requires all sorts of shenanigans with histograms to try and refit the interpolated values back to one of the original inputs either as another step either server side, or on the client in real-time which we didn't even think would produce a particularly good result anyway.

harukizaemon:
>Re. enum, I don't like the name either, but I'm unsure what else to call it.

Same.

Related, I'm also _slightly_ concerned with the names. Whilst I am happy to defend them as being technically correct (in the case median), I also felt the need to add in the comments with the specific guarantees and constraints.

I can imagine someone seeing "median" and assuming it will give you the average of the upper and lower two values. I can also imagine someone presuming the results are stable with respect to the strict ordering of the values rather than the arrangement of inputs.

But again, I can't think of a better name. It's designed to achieve some specific outcomes and it does that and is named as appropriately as I can come up with.

harukizaemon:
Oh another observation is that whilst `MODE` seems to do a pretty good job of preserving features cf. `MEAN` it also creates an apparent "shift". So if you take the images above and quickly switch between them, you can hopefully see what I mean (no pun intended).

jcupitt:
Elsewhere in libvips, median means putting all the values in order and taking the centre one, so I think that name at least is fine. 

It's not public API that has to be supported until 8.7 comes out, so I'm happy to leave the names for now. We should experiment with other images and see how they work. 

I work with brain MRI images where voxels are given identifiers, so this might be useful for me too.

jcupitt:
Ooof, finally merging this. Thank you again for your hard work, @harukizaemon ! I'll add something to the changelog crediting you (of course).

Let's use it a bit in master before 8.7 goes out.

Issue #922, opened 2018-03-30
Use O_TMPFILE for temporary files


Alexander--:
`O_TMPFILE` is a Linux-specific flag for open(), that allows creation of already-unlinked temporary files, that don't need to be explicitly removed via `unlink`. It was added in Linux 3.11 (~5 years old).

Alternatively, you can accomplish the same with `memfd_create`, added Linux 3.17 (~4 years old). The created file is hosted in tmpfs and, unlike with `O_TMPFILE`, you don't need to specify a parent directory to use it.

jcupitt:
Yes, we could have a configure test and use it here:

https://github.com/jcupitt/libvips/blob/master/libvips/iofuncs/vips.c#L184

libvips temp files can be very large, so tmpfs would probably not work.

Alexander--:
@jcupitt tmpfs is backed by swap, so it should not be critical (besides, vips already tries to create it's files in /tmp). But having option to create temp files in specific filesystem is nice too.

jcupitt:
It seems most glibc do not have support yet. On Ubuntu 17.10 (for example) I see:

```
vips.c:203:12: error: ‘O_TMPFILE’ undeclared (first use in this function); did you mean ‘__O_TMPFILE’?
   flags |= O_TMPFILE;
```

The `__O_TMPFILE` is the kernel bit, but I'm reluctant to use that until it's blessed by glibc. Let's leave this open as an enhancement and revisit when it's more widely supported. 

(unless I've misunderstood something!)

Alexander--:
@jcupitt You forgot to define `_GNU_SOURCE` before including `fcntl.h`. See `man 2 open`:

````
The O_DIRECT, O_NOATIME, O_PATH, and O_TMPFILE flags are Linux-specific. One must define _GNU_SOURCE to obtain their definitions.
````

jcupitt:
OK, this is in master now, I'll close.

Issue #921, opened 2018-03-30
Too many open() calls for same file


Alexander--:
I have been investigating possible alternatives to ImageMagick and discovered your library. Thanks for creating such amazing software!

But it looks your converter program is a little less mature than ImageMagick's `convert`. When I traced it's system calls with strace (in preparation to sandboxing it with seccomp), I noticed, that it's single invocation results in 7 similar open() calls for the same image file:

````
[pid  4966] openat(AT_FDCWD, "/home/user/5366", O_RDONLY) = 3
[pid  4966] close(3)                    = 0
[pid  4966] openat(AT_FDCWD, "/home/user/5366", O_RDONLY) = 3
[pid  4966] close(3)                    = 0
[pid  4966] openat(AT_FDCWD, "/home/user/5366", O_RDONLY) = 3
[pid  4966] close(3)                    = 0
[pid  4966] openat(AT_FDCWD, "/home/user/5366", O_RDONLY) = 3
[pid  4966] close(3)                    = 0
[pid  4966] openat(AT_FDCWD, "/home/user/5366", O_RDONLY) = 3
[pid  4966] close(3)                    = 0
[pid  4966] openat(AT_FDCWD, "/home/user/5366", O_RDONLY) = 3
[pid  4966] close(3)                    = 0
[pid  4966] openat(AT_FDCWD, "./output.jpeg", O_WRONLY|O_CREAT|O_TRUNC, 0666) = 3
strace: Process 4967 attached
strace: Process 4968 attached
strace: Process 4969 attached
[pid  4969] openat(AT_FDCWD, "/home/user/5366", O_RDONLY) = 4
````

Counting preceding `access` call (why are you doing it btw?) and associated `close` calls, `vips` is using over 14 system calls to open a single image file. Is this expected behavior?

I am running `vips` like this:

````
LANG=C vips --vips-concurrency=1 resize /home/user/5366 ./output.jpeg 1
````

It is a self-build version from Arch Linux AUR. Output of `vips --version`:

````
vips-8.6.1-Wed Jan  3 08:34:04 UTC 2018
````

Alexander--:
Passing `--vips-cache-max-files=2` does not change the observed behavior

jcupitt:
Hello @Alexander--,

It's because of the way the `VipsForeignLoadFile` class `is_a` method works:

https://github.com/jcupitt/libvips/blob/master/libvips/include/vips/foreign.h#L172

[...]

PR #930, opened 2018-04-06
Several minor improvements

Alexander--:
This PR resolves some issues, discussed in #921 and #922 

Namely, it implements file descriptors as source and destination and allows using `O_TMPFILE` for creating temporary files. The implementation is ugly (as expected from C code I wrote) and does not really accounts for user's best interests. In particular, there probably should be a fallback for case when filesystem does not support `O_TMPFILE`. The robustness of syntax, used for implementing file descriptor source, is also of some concern.

This PR was tested only with single thread (I don't actually need threads for my purposes). Feel free to ignore it — I don't really care about upstreaming these changes, although that'd be nice from maintenance standpoint.

### DIFF
--- a/libvips/foreign/foreign.c
+++ b/libvips/foreign/foreign.c
@@ -523,12 +523,6 @@ vips_foreign_find_load( const char *name )
 
 	vips__filename_split8( name, filename, option_string );
 
-	if( !vips_existsf( "%s", filename ) ) {
-		vips_error( "VipsForeignLoad", 
-			_( "file \"%s\" not found" ), name );
-		return( NULL );
-	}
-
 	if( !(load_class = (VipsForeignLoadClass *) vips_foreign_map( 
 		"VipsForeignLoad",
 		(VipsSListMap2Fn) vips_foreign_find_load_sub, 
--- a/libvips/include/vips/vips.h
+++ b/libvips/include/vips/vips.h
@@ -82,6 +82,13 @@
 extern "C" {
 #endif /*__cplusplus*/
 
+#define VIPS_EINTR_RETRY(expression) \
+  (__extension__                                                              \
+    ({ long int __result;                                                     \
+       do __result = (long int) (expression);                                 \
+       while (__result == -1L && errno == EINTR);                             \
+       __result; }))
+
 #include <glib.h>
 #include <glib/gstdio.h>
 #include <gmodule.h>
--- a/libvips/iofuncs/image.c
+++ b/libvips/iofuncs/image.c
@@ -43,6 +43,8 @@
 
  */
 
+#define _GNU_SOURCE
+
 /*
 #define VIPS_DEBUG
  */
@@ -58,6 +60,7 @@
 #ifdef HAVE_UNISTD_H
 #include <unistd.h>
 #endif /*HAVE_UNISTD_H*/
+#include <fcntl.h>
 #include <ctype.h>
 
 #include <vips/vips.h>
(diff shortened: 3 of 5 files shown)


jcupitt:
This looks nice @Alexander--, I'll try to read it over this weekend. Thanks!

jcupitt:
Hello again, I've been working through this.

I've committed a variant of your `O_TMPFILE` change to master. I think it should work on all platforms, though I've not actually tested it on Windows or macOS yet. I don't think `vips_image_new_temp_file()` needs to change: it'll end up calling `vips__open_image_write(TRUE)` because of the `delete_on_close` flag, and `vips__temp_name()` already checks `TMPDIR`.

I'd like to keep that `access` check -- it improves the error message for the very common "no such file" error. If you remove it, then you see:

```
$ vipsheader z.jpg
vipsheader: vips__file_open_read: unable to open file "z.jpg" for reading
unix error: No such file or directory
VipsForeignLoad: "z.jpg" is not a known file format
```

which is pretty ugly.

Your hack for allowing file descriptors is clever, but has a few problems :( I'll try to think about this. I'll leave this PR open as a reminder.

Alexander--:
@jcupitt note, that you can pass `--author` argument to command-line `git commit` or set `GIT_AUTHOR_` and `GIT_COMMITTER_` environment variables for duration of commit to credit a contributor, when re-implementing their changes on your own (github has [another way to do that](https://help.github.com/articles/creating-a-commit-with-multiple-authors/), but I am not sure, how well it is supported by Git itself). The created commit will include their contact information and will look like rebased/merged changes when viewed in github interface.

> I don't think vips_image_new_temp_file() needs to change

This might not be obvious from the PR, but the primary motivation behind these changes isn't implementing O_TMPFILE and file descriptors arguments, — that's just means to the end. I had to achieve full seccomp-sandboxing of vips, which naturally required removing all filesystem access.

The changes in `vips_image_new_temp_file` were required to remove a need for opening useless temporary file in /tmp.

Removal of access() was necessary to get rid of `access()` system call. Alternatively, it might be possible to keep it, but make it ignore some error codes except `ENOENT` (so it still display pretty error message when the file does not exist, but does not break program when it fails for other reasons).

File descriptor arguments allow to forbid `open()` syscall.

> Your hack for allowing file descriptors is clever, but has a few problems

I, for one, think that it's perfect :) Either way, you should have a look at `VIPS_EINTR_RETRY` — performing `read()`  without retrying on `EINTR` is a bad idea on Linux.

jcupitt:
I didn't know about `--author`, thanks! I'll use that next commit.

> The changes in vips_image_new_temp_file were required to remove a need for 
> opening useless temporary file in /tmp.

Ah you need to get rid of the use of `g_mkstemp()` I guess? I can see that would be useful for you, I'll have a look. 

Ignoring errors except `ENOENT` sounds reasonable, I'll try.

I didn't know about `EINTR`, thanks. Reading `g_mkstemp() and  `g_open()`, for example, has stuff around this. I'll have a look.

Alexander--:
@jcupitt 

> Ignoring errors except ENOENT sounds reasonable, I'll try.

Actually, that might be a bad idea. libvips file abstraction layer really needs a separate flag ("virtual"?, "ephemeral"?) to deal with O_TMPFILE and similar file descriptors. Once it is in place, the logic, that tries to deal with filenames, can be adjusted to check for it.

> I didn't know about EINTR, thanks.

Linux `EINTR` is just another form of the same issue, that results in short reads/writes. There are many situations when a call to `read()` or `open()` may block (for example, when reading from socket, opening filesystem pipe or a file, or trying to open a file, locked with file lease). If blocking `read()` gets interrupted by signal before consuming at least single byte, you receive `EINTR`. `fwrite`/`fread` hide those interruptions, but on a flip-side you can't interrupt them with signals (except for killing entire process).

jcupitt:
libvips master can now load and save most popular format from descriptors, so I'll close. There's a note about it here:

https://libvips.github.io/libvips/2019/11/29/True-streaming-for-libvips.html

PR #934, opened 2018-04-09
Adding x,y arguments to vips_composite

medakk:
This references issue https://github.com/jcupitt/libvips/issues/892

I've added optional x, y coordinates to the `vips_composite` functions. 
This uses `vips_embed` to resize all the images to the size of the target image, placed at the specified x, y coordinate. Example usage:
```
vips composite2 in0.png in1.png out.png over --x "400" --y "20"
```
in0
in1
out

### DIFF
--- a/libvips/conversion/composite.cpp
+++ b/libvips/conversion/composite.cpp
@@ -933,21 +933,13 @@ vips_composite_base_gen( VipsRegion *output_region,
 
 	VIPS_GATE_START( "vips_composite_base_gen: work" );
 
-	// printf("r: %d %d\n", r->width, r->height);
-	// printf("seq0: %d %d\n", seq->ir[0]->valid.width, seq->ir[0]->valid.height);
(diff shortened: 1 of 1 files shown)


jcupitt:
Oh dear, I'm sorry I sat on this for so long @medakk. 

Thank you very much for adding this useful thing! I'll add you to the ChangeLog and this should be in 8.7.

clcaalu:
Hi! Useful feature. Is there a way I can use it on c++? I can only see composite, not composite2. Thanks!

jcupitt:
`composite2` was missing from the C++ binding in 8.6, unfortunately. 8.7 has it added, though that's still not out officially.

You can add it yourself very easily, it's just defined as:

```c++
VImage VImage::composite2( VImage overlay , VipsBlendMode mode , VOption *options ) const
{
    VImage out;

    call( "composite2" ,
        (options ? options : VImage::option()) ->
            set( "base", *this ) ->
            set( "overlay", overlay ) ->
            set( "out", &out ) ->
            set( "mode", mode ) );

    return( out );
}
```

PR #959, opened 2018-05-03
Fix wrong value for VIPS_FOREIGN_PNG_FILTER_ALL

felixbuenemann:
The value is derived by oring all the filter flags:

    0x08 | 0x10 | 0x20 | 0x40 | 0x80 == 0xf8

The value was also checked against PNG_ALL_FILTERS in png.h of libpng 1.6.34.


jcupitt:
Ooop! Nice one.

PR #960, opened 2018-05-06
Fixed deadlock in vips_cache_drop_all if vips__cache_dump is true

lhecker:
_From the commit message:_

Both `vips_cache_drop_all` and `vips_cache_print` acquire `vips_cache_lock` on entry, but `vips_cache_drop_all` calls `vips_cache_print` internally if `vips__cache_dump` is true, leading to a deadlock situation.
This commit solves the issue using one further level of indirection and a new `vips_cache_print_nolock` function.

---

I tried adhering to the existing code style, but I'm not sure if I got this right as I only started working with libvips last week.

### DIFF
--- a/libvips/iofuncs/cache.c
+++ b/libvips/iofuncs/cache.c
@@ -480,6 +480,16 @@ vips_cache_print_fn( void *value, void *a, void *b )
 	return( NULL );
 }
 
+void
+vips_cache_print_nolock( void )
+{
+	if( vips_cache_table ) {
+		printf( "Operation cache:\n" );
+		vips_hash_table_map( vips_cache_table,
+			vips_cache_print_fn, NULL, NULL );
+	}
+}
+
 /**
  * vips_cache_print:
  *
@@ -490,11 +500,7 @@ vips_cache_print( void )
 {
 	g_mutex_lock( vips_cache_lock );
 
-	if( vips_cache_table ) {
-		printf( "Operation cache:\n" );
-		vips_hash_table_map( vips_cache_table, 
-			vips_cache_print_fn, NULL, NULL ); 
-	}
+	vips_cache_print_nolock();
 
 	g_mutex_unlock( vips_cache_lock );
 }
@@ -662,7 +668,7 @@ vips_cache_drop_all( void )
 		VipsOperation *operation;
 
 		if( vips__cache_dump )
-			vips_cache_print();
+			vips_cache_print_nolock();
 
 		/* We can't modify the hash in the callback from
 		 * g_hash_table_foreach() and friends. Repeatedly drop the


jcupitt:
Oh, nice catch @lhecker ! Thank you for the PR.

Yes, the lib vips layout standard is a little odd :(

lhecker:
@jcupitt Thanks for the quick merge! And: No problem. :slightly_smiling_face:
If you consider the age of the project (and thus the cruft it _must_'ve accumulated) it's actually very very well written.

I'm planning to replace [imaging](https://github.com/disintegration/imaging) with libvips in a certain proprietary microservice and I'm seeing a 5-8x speedup already. I'll thus of course make sure to provide you with more PRs in the future if any other issue pops up.

<details>
<summary>Some offtopic thoughts...</summary>

And in fact there are a couple things I can think of from the experience I could gather in the last 1-2 weeks. For instance:
* One of the biggest "bugs" right now is probably `vips_error_buffer`. Especially since running `strlen` on the returned buffer is not thread safe. I know that this is a hard-ish problem though and coming up with a good solution is probably quite complicated.
  At least the race condition could be solved by having a new `vips_error_buffer` function though, which returns the pointer and length simultaneously.
  Furthermore if one could bind the error logging to a `VipsImage` instance: That would be great. Or maybe in combination with thread-local logging which gets shared into the worker-thread-pool somehow? Or... well... thousands of more possibilities which are just as hard.
* I need to get the size of an image before "thumbnailing" it, but running `vips_image_new_from_buffer` on a JPEG image seems to be a rather expensive operation. Due to that doing another JPEG decode later on with `vips_thumbnail_buffer` is no faster than a `vips_thumbnail_image`, despite shrink-on-load's advantage.
  It would be quite nice, if libvips' would support some kind of lazy-decode which only extracts the size or other metadata etc. here...

Well as I said, working with libvips was very pleasant and I'll make sure to redeem myself if anything pops up. :slightly_smiling_face: I just sadly can't solve the above issues, since I fear my understanding of and my spare time for the library are both too low for now. :disappointed:
</details>

jcupitt:
Oh nice. I'm glad you think the code looks OK. I rewrote the whole thing in 2010 and threw out as much cruft as I could, while keeping source compatibility with vips7.

On your off-topic issues:

1. I don't know much about Go, but of course you're right, you must handle memory lifetime for `new_from_buffer`.
2. Yes, the error buffer is rather ugly. The best solution would probably be to immediately `strdup` the result of `vips_error_buffer()`, though there's still a chance of a race. Perhaps we need some API to return a `strdup()`d buffer from within the lock. libvips9 ought to have a new error mechanism. 
3. `new_from_buffer()` should be cheap, it just reads the header, it does not decode any pixels. Do you have a benchmark which shows poor performance?

jcupitt:
Sorry, I replied without thinking. You're right, a relocating GC would break `new_from_memory` very badly. There's `vips_image_new_from_memory_copy()`, but a better solution would be some kind of streaming API.

I tried:

```python
#!/usr/bin/env python
  
import sys
import time
import pyvips

start = time.time()
with open(sys.argv[1], "rb") as f:
    buf = f.read()
print "buffer read", time.time() - start, "seconds"

start = time.time()
x = pyvips.Image.new_from_buffer(buf, "")
print "new_from_buffer", time.time() - start, "seconds"
print "width =", x.width

start = time.time()
y = pyvips.Image.thumbnail_buffer(buf, 128)
y.write_to_file(sys.argv[2])
print "thumbnail_buffer", time.time() - start, "seconds"

start = time.time()
y = pyvips.Image.thumbnail_image(x, 128)
y.write_to_file(sys.argv[2])
print "thumbnail_image", time.time() - start, "seconds"
```

And for a 1.5k x 2k JPG I see:

```
$ ./thumb.py ~/pics/k2.jpg x.jpg
buffer read 0.000164031982422 seconds
new_from_buffer 0.000643014907837 seconds
width = 1450
thumbnail_buffer 0.0127930641174 seconds
thumbnail_image 0.0788509845734 seconds
```

jcupitt:
Sorry, final post. I would not try to call the C API from Go. It makes heavy use of varargs, in an attempt to be convenient, but of course that makes calling it from other languages fragile and inflexible.

I would call the layer below, as the Python / node / Ruby / C++ / PHP / Lua / etc. APIs do. The Python one is probably the simplest one to borrow from:

https://github.com/jcupitt/pyvips/

First wrap GValue, the GLib generic value type, then GObject, the GLib base object type (you just need to manage object lifetime), then VipsObject on top of that combined with GValue will let you get and set properties of all libvips types with introspection. VipsImage is just a VipsObject with some convenience methods. Finally wrap VipsOperation and you can call any operation in libvips via run-time introspection.

It's not a lot of code, just a few 100 lines, and once it's done it's all self-updating. As new operations, or new options, are added to libvips they should be available immediately.

https://github.com/jcupitt/pyvips/blob/master/pyvips/gobject.py

Then VipsObject, the libvips base class. 

https://github.com/jcupitt/pyvips/blob/master/pyvips/gobject.py

lhecker:
@jcupitt I'm sorry: I was mistaken. ☹️ `vips_image_new_from_buffer` is indeed quite fast. I'm sorry I wasted your time testing this. 😕 <sup>1</sup>

In return I finally figured out why my custom libjpeg-turbo is not correctly being picked up by `configure` even though I set `PKG_CONFIG_PATH`! I opened https://github.com/jcupitt/libvips/pull/963 for this.

But yeah, I've already switched to using `vips_call` yesterday, since I noticed something like `VipsBlob` exists internally anyways already, making my life managing memory buffers in Go significantly easier. (I didn't know it exists, coming from just the C API documentation.)

I believe [govips](https://github.com/davidbyttow/govips) is indeed the way forward, as it uses a form of introspection already (code generation similar to how libvipscc works). I was more comfortable writing my own C-wrapper though, as it can be written very easily and is easier to reason about. I now have a Go-wrapper for VipsBlob though, which should be possible to be integrated into govips, fixing its unsafety when facing a relocating GC, making it the first "safe" Go wrapper for libvips.

---

<sup>1</sup>
What I noticed though is that if I link with libjpeg-turbo and generate >= 3 thumbnails, it's faster to use `thumbnail_image`.
I general libjpeg-turbo seems to be so fast nowadays that the difference between `thumbnail_(buffer|image)` is actually relatively small and I feel like the recommendation to always use the former might be revised even when dealing with JPEGs.

For instance the output of your script using [this](https://upload.wikimedia.org/wikipedia/commons/7/7d/Barack_Obama_with_artistic_gymnastic_McKayla_Maroney_2.jpg) image:
```
buffer read 0.002251148223876953 seconds
new_from_buffer 0.010608911514282227 seconds
width = 2496
thumbnail_buffer 0.05549216270446777 seconds
thumbnail_image 0.10920906066894531 seconds
```

And when I run the `thumbnail_(buffer|image)`/`write_to_file` lines 4x in a row:
```
buffer read 0.0022971630096435547 seconds
new_from_buffer 0.010755062103271484 seconds
width = 2496
thumbnail_buffer 0.20664477348327637 seconds
thumbnail_image 0.16148996353149414 seconds
```

lhecker:
@jcupitt Yeah that's actually exactly what I meant: The one-time decode of a large JPEG is nowadays relatively fast and can be faster than shrink-on-load if you reuse the decompressed image for thumbnailing only a few times already. In fact this "reuse"-trick seems to be only viable with libjpeg-turbo IIRC (due to its raw speed).

BTW Your communicativeness with the users of vips is staggering - in pretty much every issue I've seen before and now in mine too. Thanks for being such an awesome person! 🙂

jcupitt:
Ah OK. Yes ... it depends. I think most people now are not generating derivatives in advance, but resizing on the fly. Computation has become cheaper quicker than storage, plus there is just a huge range of devices to size for, so on-the-fly is now cheaper, fast enough, and gives users a better experience.

*blush* I'm only communicative at the moment because I'm stuck at home, don't be too impressed!

PR #963, opened 2018-05-07
Fixed configure not picking up JPEG_CFLAGS

lhecker:
Without `$JPEG_CFLAGS` in there, the output of `pkgconfig --cflags libjpeg` is missing when compiling. This means that libjpeg(-turbo) at non-standard installation paths will not get picked up properly.

Merging this PR will also fix an issue on macOS using homebrew:
Many packages in homebrew depend on the `jpeg` package and you might've installed it already for whatever reason. Now if you intend to compile vips with `jpeg-turbo` you'll end up with 2 installed jpeg libraries.

You can configure vips to use jpeg-turbo using:
```
PKG_CONFIG_PATH="/usr/local/opt/jpeg-turbo/lib/pkgconfig" ./configure
```

But until now this will cause it to be linked with `jpeg-turbo` (v8), but include `jpeg`'s (v9) headers. The result is a runtime panic about unmatched versions between the include header and the linked version. (The workaround was to specify the CFLAGS manually.)


jcupitt:
Oh duh, that's a dumb error. Thanks!

Issue #732, opened 2017-09-08
vips fails with very high resolution SVGs


brewin:
Running `vips dzsave` on the following SVG files, the first one works as expected but the second one results in blank tiles.

```svg
<svg height="5000" width="5000">
    <circle cx="2500" cy="2500" r="2000" stroke="black" stroke-width="50" fill="red" />
</svg> 
```

```svg
<svg height="50000" width="50000">
    <circle cx="25000" cy="25000" r="20000" stroke="black" stroke-width="50" fill="red" />
</svg> 
```

lovell:
There are similar reports at https://github.com/lovell/sharp/issues/715

jcupitt:
Hello, the SVG loader works like this:

1. On initial load, the SVG file is parsed into memory. Files which include embedded JPG images probably decode the entire JPG to a memory buffer at this point (though I'm not certain). libvips gets the metadata (width, height, etc.) from the parsed SVG. 

2. To handle pixel requests, libvips uses a linecache, with tiles as wide as the image and 128 pixels high. I think there will usually be 3 or 4 tiles in the cache. Each tile is filled by one call to svg render, so it's relying on librsvg to efficiently cull bits of the SVG which don't touch the tile (it seems pretty good at this). If embedded JPG images are decoded at this stage, you can expect appalling performance, since the whole image would be decoded for each tile.

It sounds like the 50,000 x 128 pixel tiles are triggering some librsvg limit. Perhaps we should split horizontally as well, perhaps into 10,000 pixel chunks? It would be a simple change.

jcupitt:
I made a slightly smaller test-case. This works fine:

```
$ cat >circle30.svg
<svg height="30000" width="30000">
    <circle cx="15000" cy="15000" r="15000" stroke="black" stroke-width="50" fill
="red" />
</svg> 
^D
$ vips copy circle30.svg x.v
```

```

This just writes a black image:

```
$ cat >circle40.svg
<svg height="40000" width="40000">
    <circle cx="20000" cy="20000" r="20000" stroke="black" stroke-width="50" fill
="red" />
</svg> 
^D
$ vips copy circle40.svg x.v
```

jcupitt:
Oh hehe, of course, 32767 works fine, 32768 fails. I'll see if limiting the tile size helps.

jcupitt:
Yes, seems to have fixed it, I can now render a 40k pixel circle successfully. Thanks for reporting this!

[...]

PR #969, opened 2018-05-09
Reduce SVG re-renders with taller cache tiles

lovell:
Hi John, this was originally reported in https://github.com/lovell/sharp/issues/1222

The example [image.svg](https://raw.githubusercontent.com/nikwen/sharp-demo/master/image.svg) makes heavy use of Gaussian blur filters and takes 3.5s to render via `rsvg-convert`.

```sh
$ time rsvg-convert image.svg >output.png

real    0m3.351s
user    0m3.320s
sys     0m0.032s
```

A similar operation using the `vips` command line (latest master) takes ~10x longer.

```sh
$ time vips copy image.svg output.png

real    0m38.264s
user    0m37.548s
sys     0m0.304s
```

The SVG loader currently [defines a cache](https://github.com/jcupitt/libvips/blob/5176b4a17ec9ca5151f1b8b503b2f237f3c83b27/libvips/foreign/svgload.c#L387) to typically hold three 30000x16 tiles, but this doesn't appear to be enough to prevent multiple renders being requested from librsvg.

The proposed change in this PR alters the cache to hold vertical 32767px wide strips (see #732).

This is more in keeping with the way the PDF loader [cache is defined](https://github.com/jcupitt/libvips/blob/749f4a902a2b65a8880ee56ae5ce3fd858f28ef5/libvips/foreign/pdfload.c#L431) to avoid re-renders.

After applying this change I see:

```sh
$ time vips copy image.svg output.png

real    0m3.345s
user    0m3.316s
sys     0m0.032s
```

So comparable with, if not very slightly faster than, `rsvg-convert`.

### DIFF
--- a/libvips/foreign/svgload.c
+++ b/libvips/foreign/svgload.c
@@ -67,6 +67,10 @@
 #include <cairo.h>
 #include <librsvg/rsvg.h>
 
+/* The maximum pixel width librsvg is able to render.
+ */
+#define RSVG_MAX_WIDTH 32767
+
 /* Old librsvg versions don't include librsvg-features.h by default.
  * Newer versions deprecate direct inclusion.
  */
@@ -359,13 +363,7 @@ vips_foreign_load_svg_load( VipsForeignLoad *load )
 
 	int tile_width;
 	int tile_height;
-	int n_lines;
 	int max_tiles;
-
-	/* Use this to pick a tile height for our strip cache.
-	 */
-	vips_get_tile_size( load->real,
-		&tile_width, &tile_height, &n_lines );
 
 	/* Read to this image, then cache to out, see below.
 	 */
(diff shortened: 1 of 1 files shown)


lovell:
Afterthought: this code should also clip `tile_height` to a sensible limit to prevent memory ballooning, perhaps ~2000px == ~256MB?

jcupitt:
Yes, this will need tens of GB for large SVG images, ouch! I'll see if I can tune it a bit.

jcupitt:
Yes, 2000 seems to work well. I'll trim the number of tile rows too.

lovell:
Brilliant, thanks John!

nikwen:
Thank you for reviewing and merging this PR, @jcupitt. :)

jcupitt:
Oh no problem, thanks for pointing out the issue. We were using rsvg with very thin strips and it obviously does not cull drawing well in this situation.

PR #989, opened 2018-06-01
Ensure WebP can add metadata when compiled with libwebpmux

lovell:
Hi John, when libvips is compiled with support for libwebpmux, writing WebP images with `strip=FALSE` can currently fail silently in `vips_webp_add_metadata`.

This was discovered whilst investigating https://github.com/lovell/sharp/issues/1242, which this change fixes.


jcupitt:
Oh lols, nice find!

I'll add this to 8.6 as well.

PR #1002, opened 2018-06-13
Fix missing vips_addalpha() docs

janko-m:
Docs for [`vips_addalpha()`](http://jcupitt.github.io/libvips/API/current/libvips-conversion.html#vips-addalpha) are missing from the website. I don't know if this PR fixes that, but that's the only difference I saw from the other function documentations.


jcupitt:
Oooop! Thanks janko!

PR #1005, opened 2018-06-13
Run pyvips test suite on Travis CI

### LINKED ISSUE #742: run pyvips test suite in travis on commit
libvips no longer has the Python test suite built in, so Travis is currently running rather a small set of tests on each commit.

After building libvips, Travis should clone pyvips and run the test-suite in that.

felixbuenemann:
* Download pyvips version specified by PYVIPS_VERSION environment variable and run its test suite after the C based tests
* Fix the macOS build and use the latest Xcode/macOS versions
* Disable homebrew updates for more predictable builds
* Sync dependencies between linux and osx targets

Fixes #742.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 language: cpp
 
 before_script:
+  - $PYTHON -m pip download --no-deps https://github.com/jcupitt/pyvips/archive/$PYVIPS_VERSION.tar.gz
+  - tar xf $PYVIPS_VERSION.tar.gz
+  - $PYTHON -m pip install --user --upgrade pyvips-$PYVIPS_VERSION/[test]
   - ./autogen.sh
     --disable-dependency-tracking
     --with-jpeg-includes=$JPEG/include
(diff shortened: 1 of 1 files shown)


felixbuenemann:
@jcupitt Looks like pytest segfaulted for the macOS build, just restart it – never did that on my builds:

https://travis-ci.org/felixbuenemann/libvips/jobs/391968602

jcupitt:
This is great! I've been meaning to do this for ages, but been too lazy.

felixbuenemann:
@jcupitt Could you retry the build on Travis? That should get rid of the macOS failure due to the segfault on macOS.

If those crashes repeat in the future that's worth looking into, but for now I wouldn't worry about it, since I had several green builds on macOS without crashes.

felixbuenemann:
Btw. I think it would be better if the bulk of the pyvips tests that are there to test libvips and not the binding itself would live in the libvips repository.

That would allow the pyvips repository to be much slimmer and it would make it easier to add python based tests at the same time as adding features to vips without having to make a PR in a different repo.

The test suite could then do a `pip install pyvips` and then run the test suite inside the libvips tree.

It would also make it easier to test pyvips itself against multiple libvips versions, since it wouldn't have to include tests for file formats that have only been added in newer vips versions.

jcupitt:
Let's just merge then: that'll kick travis.

I think I put the tests in pyvips because I didn't want to make a circular dependency between libvips and pyvips. But you're right, now that we are running pyvips on libvips commit, the tests could move.

felixbuenemann:
I could reproduce the crash locally on my mac, it happens in `TestConversion::test_cache`:

```
============================================================================================= test session starts ==============================================================================================
platform darwin -- Python 2.7.15, pytest-3.6.1, py-1.5.3, pluggy-0.6.0 -- /usr/local/opt/python@2/bin/python2.7
cachedir: pyvips-master/.pytest_cache
rootdir: /Users/felix/dev/misc/libvips/pyvips-master, inifile: tox.ini
plugins: flake8-1.0.1
collected 37 items

pyvips-master/tests/test_conversion.py::TestConversion::test_band_and PASSED                                                                                                                             [  2%]
pyvips-master/tests/test_conversion.py::TestConversion::test_band_or PASSED                                                                                                                              [  5%]
pyvips-master/tests/test_conversion.py::TestConversion::test_band_eor PASSED                                                                                                                             [  8%]
pyvips-master/tests/test_conversion.py::TestConversion::test_bandjoin PASSED                                                                                                                             [ 10%]
pyvips-master/tests/test_conversion.py::TestConversion::test_bandjoin_const PASSED                                                                                                                       [ 13%]
pyvips-master/tests/test_conversion.py::TestConversion::test_bandmean PASSED                                                                                                                             [ 16%]
pyvips-master/tests/test_conversion.py::TestConversion::test_bandrank PASSED                                                                                                                             [ 18%]
pyvips-master/tests/test_conversion.py::TestConversion::test_cache [1]    18572 segmentation fault  DYLD_LIBRARY_PATH=$PWD/libvips/.libs python2 -m pytest -v
```

I'll see if I can reproduce it with LLDB attached.

felixbuenemann:
If I run that test in a for loop I can reproduce the crash repeatedly, took my about 50 runs on the last try. I'll open a new issue for that.

felixbuenemann:
@jcupitt I opened jcupitt/pyvips#51 to track the crashes.

jcupitt:
I moved the pyvips tests back into libvips. I've started a new set of tests for pyvips that should just check the binding, but I've not got very far.

felixbuenemann:
That's great, thanks for the update!

PR #1009, opened 2018-06-18
8-Bit PNG Quantization Support

### LINKED ISSUE #734: 8-bit PNG Support
It would be great if libvips had support for saving 8-bit palette based PNG with alpha transparency which are a great choice for saving small images with transparency on the web.

More info: [pngquant](https://pngquant.org)

From my limited understanding this would require the following changes:

* Add a good quantizer algorithm to reduce from RGBA to 8-bit indexed-color (eg. [libimagequant](https://github.com/ImageOptim/libimagequant))
* Add support for writing to PNG8 format with the proper tRNS-Chunk

Note that there are two ways to store transparency in 8-bit PNG:

* Mark a single color as transparent
* Store a transparency value for each color in the index (so eg. if you have 256 colors in the index, you have 256 transparency values in the same order that specify the transparency for the color)

I think only the latter is useful and makes sense to implement.

Here's some info on the tRNS chunk from libpng:

* [png_get_tRNS](http://refspecs.linuxbase.org/LSB_3.1.0/LSB-Desktop-generic/LSB-Desktop-generic/libpng12.png.get.trns.1.html)

There is similar discussion in #620 regarding gifsave support.

One approach to make it simpler to integrate could be to move the quantization logic into a custom saver like png8save which accepts RGBA data and then does the quantization internally.

felixbuenemann:
This adds support for saving 8-Bit one band palette based PNG images with palette based alpha channel (often called PNG8+Alpha).

The image is first converted to sRGBA and then quantized using [libimagequant](https://pngquant.org/lib/) controlled by the `colors`, `Q` and `dither` params.

* By default quantization is disabled and it must be enabled by specifying the `palette` flag.
* The `colours` parameter specifies the maximum number of colors (2-256).
* The `Q` parameter controls how lossy the quantization should be, similar to JPEG.
* The `dither` parameter controls how much Floyd-Steinberg dithering is used.

All of the code that deals with quantization is in the method `quantise_image` which writes back two `VipsImage` objects: a 1-band grayscale image where the pixel values correspond to the palette index and a 4-band sRGBA image of up to 256x1 px which contains the palette.

This structure should make is easy to share the quantization code with other writers or swap in a different quantization algorithm in the future.

Fixes #734.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1393,7 +1393,7 @@ support webp metadata:			$with_libwebpmux
 text rendering with pangoft2: 		$with_pangoft2
 file import/export with libpng: 	$with_png
   (requires libpng-1.2.9 or later)
-support 8bpp PNG quantization:		$with_imagequant
+support 8bpp PNG quantisation:		$with_imagequant
 file import/export with libtiff:	$with_tiff
 file import/export with giflib:		$with_giflib
 file import/export with libjpeg:	$with_jpeg
--- a/configure.ac
+++ b/configure.ac
@@ -1142,6 +1142,24 @@ if test x"$with_png" != "xno"; then
   )
 fi
 
+# look for libimagequant with pkg-config (only if libpng is enabled)
+AC_ARG_WITH([imagequant],
+  AS_HELP_STRING([--without-imagequant], [build without imagequant (default: test)]))
+
+if test x"$with_imagequant" != "xno" && test x"$with_png" != "xno"; then
+  PKG_CHECK_MODULES(IMAGEQUANT, imagequant,
+    [AC_DEFINE(HAVE_IMAGEQUANT,1,[define if you have imagequant installed.])
+     with_imagequant=yes
+     PACKAGES_USED="$PACKAGES_USED imagequant"
+    ],
+    [AC_MSG_WARN([libimagequant not found; disabling 8bpp PNG support])
+     with_imagequant=no
+    ]
+  )
+else
+  with_imagequant=no
+fi
+
 # look for libjpeg with pkg-config ... fall back to our tester
 AC_ARG_WITH([jpeg], 
   AS_HELP_STRING([--without-jpeg], [build without libjpeg (default: test)]))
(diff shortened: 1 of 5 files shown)


felixbuenemann:
@jcupitt Please review, especially the code around memory allocations and `VIPS_UNREF()` etc.

I'm also not sure if `vips_colourspace()` is the best way to convert to sRGB or if a combination of extract_band/bandjoin and cast with shift would be better.

It would also be great if we could add the alpha channel in one step, but I couldn't find a way to do that at the same time as the conversion.

To test grab a tarball of libimagequant from https://github.com/ImageOptim/libimagequant/releases and do the usual `configure` / `make install` steps.

felixbuenemann:
Btw. libimagequant is a very high quality quantizer, it is much slower and has higher memory usage than simpler algorithms like Neuquant (see [pngng implementation](https://github.com/stuart/pngnq/blob/master/src/neuquant32.c)).

libvips/foreign/pngsave.c
```diff
@@ -133,13 +136,37 @@ vips_foreign_save_png_class_init( VipsForeignSavePngClass *class )
...
+		VIPS_ARGUMENT_OPTIONAL_INPUT,
+		G_STRUCT_OFFSET( VipsForeignSavePng, colors ),
+		2, 256, 256 );
+
+	VIPS_ARG_INT( class, "Q", 14,
+		_( "Q" ),
+		_( "Q factor" ),
+		VIPS_ARGUMENT_OPTIONAL_INPUT,
+		G_STRUCT_OFFSET( VipsForeignSavePng, Q ),
+		0, 100, 75 );
```
lovell: The libimagequant docs suggest it will fail if the desired "quality" can't be met. Its default for `Q` is `0`, which tells it to proceed regardless of quality. Should we match that default here to reduce the possibility of unexpected errors?

libvips/foreign/pngsave.c
```diff
@@ -133,13 +136,37 @@ vips_foreign_save_png_class_init( VipsForeignSavePngClass *class )
 		VIPS_TYPE_FOREIGN_PNG_FILTER,
 		VIPS_FOREIGN_PNG_FILTER_ALL );
 
+	VIPS_ARG_INT( class, "colors", 13,
+		_( "Colors" ),
+		_( "Number of palette entries" ),
+		VIPS_ARGUMENT_OPTIONAL_INPUT,
+		G_STRUCT_OFFSET( VipsForeignSavePng, colors ),
+		2, 256, 256 );
```
lovell: Should a value of `0`, meaning don't quantise, be permitted here?

libvips/foreign/pngsave.c
felixbuenemann: In libimagequant there is a minimum and maximum quality. We always set the minimum quality to `0` so it never fails because quality can't be met. The `Q` parameter maps to maximum quality only.

It could be argued that the default quality should be `100` instead of `75`, but I chose that value to match other lossy compressors with the same option, like @jpegsave@.

However having a default of `75` also means that a call with @colors=256@ will likely produce less colors, unless `Q=100` is specified, so for predicting behavior `100` would be a better default.

libvips/foreign/pngsave.c
felixbuenemann: The range for `max_colors` in libimagequant is 2-256 so we can't reasonably permit `0` here, because it would also allow `1` (at least if we want to use the default option range checks).

Internally though `colors` is initialized to `0` which means don't quantize.

My thought was that people should simply omit the `colors` option if quantization is unwanted.

libvips/foreign/pngsave.c
felixbuenemann: @lovell I'm happy to hear your thoughts on what#s the better default for `Q`here, I'm slightly leaning towards changing it to `100`, but I'm still undecided.

libvips/foreign/pngsave.c
felixbuenemann: Oh well, we could get away with redefining the range to 0-255 and simply doing a `colors + 1` when the value is passed to libimagequant. That would be a bit weird for users though, cause they shouldn't be counting from zero.

lovell: The 2-256 range is fine as-is. As you say, anything else would be too confusing.

configure.ac
```diff
@@ -1142,6 +1142,24 @@ if test x"$with_png" != "xno"; then
   )
 fi
 
+# look for libimagequant with pkg-config (only if libpng is enabled)
+AC_ARG_WITH([imagequant],
+  AS_HELP_STRING([--without-imagequant], [build without imagequant (default: test)]))
+
+if test x"$with_imagequant" != "xno" && test x"$with_png" != "xno"; then
```
jcupitt: You conventionally write `x"no"` to mirror the left-hand side of the `!=`. Though of course it makes no practical difference.

libvips/foreign/pngsave.c
lovell: Ah, I didn't spot that it was a range, in which case yes, I agree with a default of `100` here.

configure.ac
felixbuenemann: Yeah, it’s copied over from the existing png test a few lines above.

libvips/foreign/pngsave.c
felixbuenemann: I've changed the default for `Q` to `100` and also made the parameter description more useful.

libvips/foreign/vipspng.c
```diff
@@ -877,12 +881,110 @@ write_png_block( VipsRegion *region, VipsRect *area, void *a )
...
 
+#ifdef HAVE_IMAGEQUANT
+static int
+quantize_image( VipsImage *in, VipsImage *out, VipsImage *palette_out,
+	int colors, int Q, double dither )
+{
+	/* Ensure input is sRGB. */
+	if( in->Type != VIPS_INTERPRETATION_sRGB) {
+		VipsImage *srgb;
+		if( vips_colourspace( in, &srgb, VIPS_INTERPRETATION_sRGB,
```
jcupitt: How about adding a test for CMYK and throwing an error? Otherwise colourspace will fail with a mysterious message.

jcupitt: ... if pngsave is asked to quanize an CMYK image, I mean.

felixbuenemann: What I really want to do here is go from ushort to uchar and grayscale (1/2) bands to uchar with 3/4 bands in the most efficient way. If there is a better way to do that I am happy to chage it.

CMYK should already be blocked by an assertion further up the call stack in the pngsave case.

libvips/foreign/pngsave.c
felixbuenemann: Not too fond of the name `quant`, how about `palette`?

libvips/foreign/vipspng.c
jcupitt: Duh of course you're right, I was thinking of jpg. Carry on!

libvips/foreign/pngsave.c
jcupitt: Yes, that sounds better.

lovell: Quick thought: should the `colors` property be made consistent with the British English "colour" spelling used everywhere else in the libvips API e.g. `colourspace`? (The approach taken by sharp is to provide aliases so both work.)

jcupitt: Good point, libvips is en_GB for API names. I noticed this too, then forgot to comment :( More coffee.

felixbuenemann:
I think I've addressed all of the feedback!

jcupitt:
Hey Felix, you probably saw, we've set up a libvips organisation:

https://github.com/libvips

The idea is to open up development and for me to step back a little (I have a new full-time job argh). Would you be interested in becoming a member, plus commit rights? I realize you are very busy too. Let me know if you're interested.

felixbuenemann:
Hey John, I hope you enjoy your new Job!

As my lack of GitHub activity shows things are pretty crazy for me at the moment.

I'm not sure how useful direct commit access would be, since I write way too little C code to be confident that it is correct, so I'll always go through a PR review anyways.

felixbuenemann:
Oh well, I might not be a C wizard, but I'm sure I can help with looking through issues or fix the CI ;-)

Go ahead and add me to the libvips org!

jcupitt:
OK, invite sent! Thank you!

I'm going through the PR process too, except for very small things.

PR #1061, opened 2018-08-14
 Fix missing area unref in getpoint()

dbouron:
vips_getpoint() doesn't unref VipsArea.


jcupitt:
Argh! Good catch @dbouron !

PR #1066, opened 2018-08-17
Close input early

jcupitt:
The current behaviour (close input handles on unref) works for languages like C / C++ / Python / Rust / etc. where things get unreffed automatically when they go out of scope.

On languages like Ruby / C# / node / etc. where things are unreffed on GC, files can stay open for a long time after you've finished with them. This interacts in an unfortunate way with the Windows default of refusing to remove open files.

This change closes file handles as soon as the sequential scan of the input file finishes, and therefore produces something closer to expected behaviour for GCd languages on Windows.

See https://github.com/kleisauke/net-vips/issues/12


libvips/foreign/tiff2vips.c
```diff
@@ -1880,6 +1940,11 @@ rtiff_stripwise_generate( VipsRegion *or,
 		y += hit.height;
 	}
 
+	/* Shut down the input file as soon as we can. 
+	 */
+	if( y >= or->im->Ysize )
```
kleisauke: I think that `r-> top` should be subtracted from `y` here, `if( y - r->top >= or->im->Ysize )` (?)

jcupitt:
fitsload is a random-access reader, so I don't think we can make it close early, unfortunately. Or not without adding an entire new codepath.

I added a thing to make openexrload shut down early for non-tiled images. Again, it'd be a big change to make it close early for tiled images :(

jcupitt:
Also, good catch on the tiffload logic, I'd got confused about y being in absolute or relative coordinates :(

kleisauke:
`tiffload` seems to work now, thanks! `openexrload` still seems broken (tested with [these images](https://github.com/openexr/openexr-images/tree/master/ScanLines)). 

Also,
Is there something wrong with `openexrload` and `fitsload`? Because I always get a transparent image (with different dimensions) when I tried to process [.exr images](https://github.com/openexr/openexr-images) or [.fits images](https://fits.gsfc.nasa.gov/fits_samples.html). This behavior is reproducible on Windows and Linux. I'm not sure whether this is only with libvips 8.7 or that older versions had this too. I'll test more.

jcupitt:
I tried adding `printf( "early exr shutdown\n" );` just before the `read_close()` at the end of `openexr2vips.c` and I see:

```
john@kiwi:~/pics/openexr/ScanLines$ vips copy Blobbies.exr x.jpg
early exr shutdown
```

So I think it is closing early, at least in this case.

You're right, openexr was not scaling up alpha to the 0 - 255 convention that libvips uses, I've fixed it.

For fits, try loading into nip2 and playing with the scale. It's a scientific format, so it won't make nice jpg images without some extra conversion work.

jcupitt:
For example, the first FITS image on that page has four colour bands, none of which correspond to R, G or B, and which are in a strange unit used in astronomy and not really related to anything visual :(

kleisauke:
Thanks for fixing OpenEXR! I no longer see transparent images. :)

I always expected something visually to be carried out, which is not always correct with some formats. I should check the FITS Wikipedia next time..

Regarding OpenEXR, with this patch:
```diff
diff --git a/libvips/foreign/openexr2vips.c b/libvips/foreign/openexr2vips.c
index 1111111..2222222 100644
--- a/libvips/foreign/openexr2vips.c
+++ b/libvips/foreign/openexr2vips.c
@@ -440,6 +440,7 @@ vips__openexr_read( const char *filename, VipsImage *out )
 				return( -1 );
 		}
 
+		printf( "early exr shutdown\n" );
 		read_close( read ); 
 	}
 
```

and this Python script:
```python
#!/usr/bin/python

import time
import os

import pyvips

pyvips.cache_set_max(0)

images_dir = "C:\\test-suite\\images"

file = os.path.join(images_dir, "Blobbies.exr")
image = pyvips.Image.new_from_file(file, access="sequential")
image.write_to_file("x.jpg")

print("File should close within 10 sec")
time.sleep(10)

print("10 sec passed, lets delete the file")

try:
    os.remove(file)
    print("Successfully deleted file from openexrload")
except IOError:
    print("Could not delete file from openexrload")

```

I see this on Windows:
```
File should close within 10 sec
10 sec passed, lets delete the file
Could not delete file from openexrload
early exr shutdown
```

So it looks like early close doesn't work for OpenEXR.

jcupitt:
Ah think I have it -- there was a close missing on header read.

kleisauke:
Seems to work on Windows!

kleisauke:
I think you can merge this. I've tested all loaders on Windows and apart from `fitsload` they all work (tested with removing the test image immediately after opening it).

Not relevant for this PR,
You might be interested in [the remaining patch](https://github.com/kleisauke/build-win64-mxe/blob/1655f29bd112ac518b05848830e754344807b6b6/8.7/vips-8-fixes.patch) (to successfully build libvips on Windows with support for NIfTI). Perhaps this will help Linux too (if znzlib is built as a shared library).

jcupitt:
Oh nice, sure, I'll add your patch.

Merged!

PR #1081, opened 2018-09-04
Silence 'missing sentinel' warnings in GCC v6+

lovell:
Another small C++ compiler warning a la https://github.com/jcupitt/libvips/pull/864
```
composite.cpp: In function 'int vips_composite_base_build(VipsObject*)':
composite.cpp:1196:25: warning: missing sentinel in function call [-Wformat=]
     width, height, NULL ) )
                         ^
```


jcupitt:
Of course! 

I did a quick grep and that seems to be the only one.

Issue #879, opened 2018-02-11
Smartcrop: cropbox is misplaced


kleisauke:
When using smartcrop, it looks like the cropbox is misplaced when changing the dimensions from 128x128 to 256x256. This is since libvips 8.5.7 (or higher). I couldn't reproduce it with 8.5.6.

Test image: https://t0.nl/skutsje.jpg
<sup>(See [wikipedia](https://en.wikipedia.org/wiki/Sk%C3%BBtsje) for an explanation of 'Skûtsje')</sup>

**libvips 8.5.7+**
```
vipsthumbnail -s 128 --smartcrop attention skutsje.jpg
```
![tn_skutsje](https://user-images.githubusercontent.com/12746591/36073849-530691c6-0f37-11e8-9422-47ea8b999994.jpg)

```
vipsthumbnail -s 256 --smartcrop attention skutsje.jpg
```
![tn_skutsje](https://user-images.githubusercontent.com/12746591/36073854-6255444c-0f37-11e8-86ca-5774c708a514.jpg)

**libvips 8.5.6**
```
vipsthumbnail -s 128 --smartcrop attention skutsje.jpg
```
![tn_skutsje](https://user-images.githubusercontent.com/12746591/36073887-fb1c20f6-0f37-11e8-88a9-1cdc524ce559.jpg)

```
vipsthumbnail -s 256 --smartcrop attention skutsje.jpg
```
![tn_skutsje](https://user-images.githubusercontent.com/12746591/36073891-0c37cc8c-0f38-11e8-9b83-d3443f240bf1.jpg)

This can be reproduced in the nip2 workspace, see:
[128x128](https://user-images.githubusercontent.com/12746591/36073902-3b16d098-0f38-11e8-9962-916db2671977.png) vs [256x256](https://user-images.githubusercontent.com/12746591/36073903-42562098-0f38-11e8-8f1b-1a4176db0b57.png)
<sup>from: https://github.com/jcupitt/libvips/issues/672#issuecomment-307174985</sup>

With the old nip2 workspace I got the expected result:
[128x128](https://user-images.githubusercontent.com/12746591/36073953-08708dea-0f39-11e8-8f43-eddda8ce9359.png) vs [256x256](https://user-images.githubusercontent.com/12746591/36073971-64533f90-0f39-11e8-8801-630583297cc6.png)
<sup>from: https://github.com/jcupitt/libvips/issues/619#issuecomment-287065036</sup>

lovell:
[...]

PR #809, opened 2017-11-19
Smartcrop: ~15% speedup via a linecache

lovell:
Hi John, I've been doing a little profiling of the "attention" smartcrop, in this case attempting to extract a 100x10000 region from a 10000x10000 input image.

Current behaviour (gcc 5.4 `-O3`):
```
$ time vips smartcrop 10000x10000.v out.v 100 10000
real    0m18.825s
user    0m18.572s
sys     0m0.284s
```

```
36,788,404,121  ???:vips_extract_band_buffer [/usr/local/lib/libvips.so.42.8.0]
15,708,124,763  /build/glibc-bfm8X4/glibc-2.23/math/../sysdeps/ieee754/dbl-64/e_atan2.c:__ieee754_atan2_avx [/lib/x86_64-linux-gnu/libm-2.23.so]
14,953,036,753  ???:vips_col_sRGB2scRGB_8 [/usr/local/lib/libvips.so.42.8.0]
12,562,353,888  ???:vips_ifthenelse_gen [/usr/local/lib/libvips.so.42.8.0]
12,262,089,216  ???:vips_bandjoin_buffer [/usr/local/lib/libvips.so.42.8.0]
10,367,136,011  ???:vips_XYZ2Lab_line [/usr/local/lib/libvips.so.42.8.0]
 6,880,982,698  ???:vips_sRGB2scRGB_gen [/usr/local/lib/libvips.so.42.8.0]
 6,878,396,905  ???:vips_col_scRGB2XYZ [/usr/local/lib/libvips.so.42.8.0]
 5,883,590,511  ???:vips_convi_gen'2 [/usr/local/lib/libvips.so.42.8.0]
```

I noticed there's a 3-way split in the middle of the `vips_smartcrop_attention` pipeline. Adding a linecache operation just before, as this PR demonstrates, yields a ~15% performance improvement:

```
$ time vips smartcrop 10000x10000.v out.v 100 10000
real    0m15.592s
user    0m15.544s
sys     0m0.084s
```

```
28,810,327,781  ???:vips_extract_band_buffer [/usr/local/lib/libvips.so.42.8.0]
15,708,124,763  /build/glibc-bfm8X4/glibc-2.23/math/../sysdeps/ieee754/dbl-64/e_atan2.c:__ieee754_atan2_avx [/lib/x86_64-linux-gnu/libm-2.23.so]
12,562,353,888  ???:vips_ifthenelse_gen [/usr/local/lib/libvips.so.42.8.0]
[...]

PR #1083, opened 2018-09-05
Smartcrop (attention): resize at start of pipeline to improve performance/accuracy

lovell:
Based on a proposal from #809, this PR vastly improves the performance of attention-based smartcrop and, based on my testing, slightly improves its accuracy. It's probably worth getting those interested in #879 to do some preliminary testing too (👋 @kleisauke).

Before:
```
$ time vips smartcrop 10000x10000.v out.v 100 10000
real    0m12.773s
user    0m12.528s
sys     0m0.292s
```
After:
```
$ time vips smartcrop 10000x10000.v out.v 100 10000
real    0m0.503s
user    0m0.388s
sys     0m0.156s
```

To ensure this change hasn't decreased accuracy, I've tested on the 5000 images of the MSRA salient object database - see https://github.com/lovell/sharp/tree/master/test/saliency for the approach.

Here's a graph of the counts (y) for each % accuracy (x), blue (1) is the current behaviour, green (2) is the behaviour with this change. I suspect/hope the slight overall improvement is due to a reduction in edge-detection noise.

In terms of the code change itself, the default Lanczos3 kernel performs better than the current use of a linear kernel. Also, the ordering of the `vips_object_local_array` is a bit messed up now but this change tries to keep the diff clean.

### DIFF
--- a/libvips/conversion/smartcrop.c
+++ b/libvips/conversion/smartcrop.c
@@ -204,6 +204,18 @@ vips_smartcrop_attention( VipsSmartcrop *smartcrop,
 	int x_pos;
 	int y_pos;
 
+	/* The size we shrink to gives the precision with which we can place
+	 * the crop
+	 */
+	hscale = 32.0 / in->Xsize;
+	vscale = 32.0 / in->Ysize;
+	sigma = VIPS_MAX( sqrt( pow( smartcrop->width * hscale, 2 ) +
+		pow( smartcrop->height * vscale, 2 ) ) / 10, 1.0 );
+	if ( vips_resize( in, &t[17], hscale,
+		"vscale", vscale,
+		NULL ) )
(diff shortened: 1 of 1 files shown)


jcupitt:
Oh very nice. I tinkered a bit with this too, but not to the extent you have here. Great!

Let's get 8.7 out of the door and merge this at the start of 8.8.

proyb6:
@lovell I wonder if you could shown CPU usage and memory consumption using gtime -v vips

jcupitt:
OK, merged! 

@proyb6 the speedup will be simply huge for large images. smartcrop should become almost free.

andrieslouw:
@kleisauke : Let's test this in production, as current load of 8.6's smartcrop is huge (~60% of our total I guess, but we could first check that assumption).

Issue #1089, opened 2018-09-17
vips_thumbnail crop seems incomplete


triptec:
I'm missing gravity in VipsInteresting. It has 'centre' but the rest (north-west, north, north-east etc..)would be good to have as well i think or have I misunderstood something?

triptec:
ah, ok but left and right would then also make sense?

jcupitt:
It's only ever on one axis, you can't position on both, so low/centre/high would mean both left-right and up-down, depending on the aspect ratios.

triptec:
You’re right, seems I didn’t think about this too much. Thanks

jcupitt:
OK, there's a PR to add low/high options. I'll close this issue.

PR #1103, opened 2018-09-22
add low/high to smartcrop

jcupitt:
smartcrop attention had a centre setting, but not low or high

see https://github.com/libvips/libvips/issues/1089


lovell:
Does `VIPS_INTERESTING_LOW` have the same effect as the existing `VIPS_INTERESTING_NONE`?

jcupitt:
Oh, good point, I'd not noticed `NONE`. Low and high seem clearer, I suppose, by analogy with `VipsAlign`.

jcupitt:
OK, thanks for looking it over!

Issue #1073, opened 2018-08-23
Can not read XMP data from PNG


iovdin:
https://storage.googleapis.com/frameright-crops/single_crop.png

``` 
$ vipsheader -a single_crop.png

single_crop.png: 512x512 uchar, 3 bands, srgb, pngload
width: 512
height: 512
bands: 3
format: uchar
coding: none
interpretation: srgb
xoffset: 0
yoffset: 0
xres: 2.834
yres: 2.834
filename: test/single_crop.png
vips-loader: pngload
```

``` 
$ exiftool single_crop.png

ExifTool Version Number         : 11.09
File Name                       : single_crop.png
Directory                       : test
File Size                       : 464 kB
File Modification Date/Time     : 2018:08:22 16:55:29+03:00
File Access Date/Time           : 2018:08:23 10:58:09+03:00
File Inode Change Date/Time     : 2018:08:22 17:37:31+03:00
File Permissions                : rw-r--r--
File Type                       : PNG
File Type Extension             : png
MIME Type                       : image/png
Image Width                     : 512
Image Height                    : 512
Bit Depth                       : 8
Color Type                      : RGB
Compression                     : Deflate/Inflate
Filter                          : Adaptive
Interlace                       : Noninterlaced
XMP Toolkit                     : Adobe XMP Core 5.6-c138 79.159824, 2016/09/14-01:09:01
Modify Date                     : 2018:08:22 13:55:29.219Z
Create Date                     : 2018:08:22 13:55:29.219Z
Metadata Date                   : 2018:05:25 15:37:10+02:00
Data                            : eyJvcmlnaW5hbFNpemUiOnsid2lkdGgiOjUxMiwiaGVpZ2h0Ijo1MTJ9LCJjcm9wcyI6W3sieCI6NjQsInkiOjc0LCJ3aWR0aCI6MzY
3LCJoZWlnaHQiOjM1MiwibGlicmFyeUlkIjpudWxsLCJsaWJyYXJ5SXRlbUlkIjpudWxsLCJyYXRpbyI6eyJ4IjozNjcsInkiOjM1Mn0sImZpeGVkUmF0aW8iOnRydWUsIm1pblNp
emUiOnsid2lkdGgiOjM2NywiaGVpZ2h0IjozNTJ9LCJyZXFzIjpudWxsfV0sIm1ldGFEYXRhRm9ybWF0IjoxfQ==
SRGB Rendering                  : Perceptual
Image Size                      : 512x512
[...]

PR #1104, opened 2018-09-22
Add xmp png support

jcupitt:
Add text read/write for png. PNG XMP is a special text segment. PNG EXIF should be another issue. Not sure about IPTC.

see https://github.com/libvips/libvips/issues/1073


jcupitt:
There's a better one now, see https://github.com/libvips/libvips/pull/1124

PR #1120, opened 2018-10-02
update gh-pages index links

xlotlu:
I didn't check for everything that might need updating. I only changed the first paragraph because it's high-visibility.


jcupitt:
Oooof, thank you! There always seem to be more links to find and fix.

xlotlu:
:)

jcupitt:
I changed a couple more in `index.md`.

PR #1124, opened 2018-10-06
add xmp load/save for png

jcupitt:
plus a test

see https://github.com/libvips/libvips/issues/1073


jcupitt:
OK, let's merge and test in master.

Issue #1132, opened 2018-10-09
Check if VImage's VipsImage* is null


Fingolfin1196:
As far as I know, the `VImage` (or, more appropriately, its `VObject` base class) don't offer a possibility to check if the contained `vobject` pointer is null or not, as far as I can tell. This is a problem, since all functions that use `vobject` check if it is null and throw errors if it is, which I can't prevent.
For some time now, I've added a small `bool is_null const` function that just returns `vobject == nullptr`, which serves its purpose, but it's the only reason I need to recompile the library every time and can't just use the repository version.

jcupitt:
Oh, interesting, I'm not sure I've ever needed to test for null. Could you give an example where this is useful?

I'm very happy to accept a PR to add an `is_null` test.

Fingolfin1196:
In the project that I use `libvips` in, getting an empty `VImage` from a function call represents that there is no result. When the `VImage` is null, I (obviously) can't query it's dimensions, bands etc., each of which would lead to segmentation faults. To handle this possibility, a null check would be welcome.

jcupitt:
Merged to git master, thanks!

PR #1133, opened 2018-10-10
Add is_null check to VImage

Fingolfin1196:
This pull request adds an is_null check, which is then used in VImage.
The usefulness is described in #1132.


jcupitt:
OK, looks nice, thank you!

PR #1155, opened 2018-11-04
Fallback to standard temp file if O_TMPFILE fails

### LINKED ISSUE #1151: O_TMPFILE doens't work on each FS
The changes introduced in https://github.com/jcupitt/libvips/commit/06802e41107d7125f12168b32fcbb045782d4c13 assumes that the filesystem mounted on `/tmp` has the capability to handle the O_TMPFILE flag. This is not always the case, making vips fail when trying to create these temporary files. 

Our use case was running vips in a docker container (docker for mac, nfs + overlayfs) but there might be others. I "fixed" this by removing the content from this commit from the code in my build (& applying a bit of C knowledge).

There should be a way to disable this behaviour, either at compile time or by adding a flag to the binary when running it. The later might be the more flexible option, the former the easier (but a bit brutal).

lovell:
This should fix #1151, makes the Windows `_O_TEMPORARY` bit work again and simplifies the logic so the default and fallback behaviour are both the same statement.

This doesn't deal with updating `image->filename` to use the `O_TMPFILE` filename. This means that further errors involving the file, if any, will still include `/tmp/vips-xxx.v` rather than the actual  `/tmp/random-os-assigned-filename`.

### DIFF
--- a/libvips/iofuncs/vips.c
+++ b/libvips/iofuncs/vips.c
@@ -188,6 +188,7 @@ vips__open_image_write( const char *filename, gboolean temp )
 	int fd;
 
 	flags = MODE_WRITE;
+	fd = -1;
 
 #ifdef O_TMPFILE
 	/* Linux-only extension creates an unlinked file. CREAT and TRUNC must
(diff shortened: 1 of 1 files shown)


jcupitt:
Nice job!

Oh dear, you're right, it wasn't being set on Windows, copy-paste error I guess.

jcupitt:
This feels like a fix that should go into 8.7.1, what do you think?

jcupitt:
OK, 8.7 HEAD patched too, I'll make an 8.7.1.

jcupitt:
8.7.1 is up!

https://github.com/libvips/libvips/releases/tag/v8.7.1

PR #1183, opened 2018-12-17
Use latest Xcode for macOS builds

aried3r:
https://docs.travis-ci.com/user/reference/osx/#os-x-version

Let's see if this works :)


jcupitt:
Living dangerously!

Issue #1195, opened 2018-12-31
Add a fallback path to CMYK2XYZ in case lcms2 is missing


jcupitt:
If there's no lcms2 available, we could have a simple fallback cmyk <-> xyz converter. 

@dbouron has a ncie patch here that would only need minor changes:

https://github.com/dbouron/libvips/commit/7231ffea468e191e899c3508370ca34daab0bf3d

See https://github.com/libvips/libvips/issues/1186

jcupitt:
See https://github.com/libvips/libvips/pull/1196

dbouron:
We can close this issue as #1196 has been merged

PR #1196, opened 2018-12-31
Add cmyk <-> srgb colourspace conversion

dbouron:
If no lcms2 library has been found, cmyk conversions are handled with default route to srgb.

#1195


jcupitt:
Hello, sorry for the delay in getting to this. 

How about doing CMYK -> XYZ rather than sRGB? You can treat XYZ and RGB as equivalent (not correct, but this is only an approximation), so it should be a simple change. You just need to get the range correct -- XYZ is 0 - 100, with D65 weighting, so:

https://github.com/jcupitt/libvips/blob/master/libvips/include/vips/colour.h#L58

We could then use the lcms2 CMYK2XYZ if lcms2 is detected, and this approximation otherwise.

I'm happy to tweak this, or would you rather do it?

dbouron:
Hi,
No problem, I will make the change tomorrow.
I agree, it will remove lots of #ifdef HAVE_LCMS2

dbouron:
According to HAVE_LCMS2 is defined or not, vips_CMKY2XYZ and vips_XYZ2CMYK type will be either an operation or a colour code

jcupitt:
It seems to build OK for me with the `--without-lcms` flag. Could you explain the problem a little more?

dbouron:
> It seems to build OK for me with the `--without-lcms` flag. Could you explain the problem a little more?

I have the following compilation:

`
XYZ2CMYK.c:150:16: error: unknown type name ‘VipsImage’
 vips_XYZ2CMYK( VipsImage *in, VipsImage **out, ... )
`

Did you rebase with master? Because, you've merged the hotfix few days ago

jcupitt:
Ah I understand, you mean the superclass of CMYK2XYZ will be either `VipsOperationClass` or `VipsColourCode`, is that right?

I agree, that's a bit strange, but I think it's harmless. The libvips API deliberately hides the GObject class hierarchy for operations, so it's just an implementation detail.

libvips/colour/CMYK2XYZ.c
```diff
@@ -150,20 +150,57 @@ vips_CMYK2XYZ_build( VipsObject *object )
...
+    int i;
+
+    for (i = 0; i < width; i++) {
+        float r = 255.0 - ((1.0 / 255.0) * p[0] * (255.0 - p[3]) + p[3]);
+        float g = 255.0 - ((1.0 / 255.0) * p[1] * (255.0 - p[3]) + p[3]);
+        float b = 255.0 - ((1.0 / 255.0) * p[2] * (255.0 - p[3]) + p[3]);
+        
+        float x, y, z;
+        
+        vips_col_scRGB2XYZ(r / 255.0, g / 255.0, b / 255.0, &x, &y, &z);
```
jcupitt: This will be rather slow, and I think it's not necessary. You can just set:

```C
x = VIPS_D65_X0 * r / 255.0;
```

etc.

libvips/colour/CMYK2XYZ.c
```diff
@@ -150,20 +150,57 @@ vips_CMYK2XYZ_build( VipsObject *object )
...
+        q[0] = x;
+        q[1] = y;
+        q[2] = z;
+
+        p += 4;
+        q += 3;
+    }
+}
+
+#endif /*HAVE_LCMS2*/
```
jcupitt: I don't like mixing up the two `CMYK2XYZ` operations with `ifdefs`. It would be very easy to make a mistake in the future.

I think it would be better to have two completely separate implementations with a single ifdef to flip between them. Something like:

```C
// prelude

#ifdef HAVE_LCMS2

// implementation based on icc_import

#else /*!HAVE_LCMS2*/

// implementation based on ColourCode with your own maths

#endif /*HAVE_LCMS2*/

/**
 * doc comment
 */
int
vips_CMYK2XYZ( VipsImage *in, VipsImage **out, ... )
{
// etc.
}
```

libvips/colour/CMYK2XYZ.c
jcupitt: I guess you could do all the scaling in a separate step:

```C
float c = p[0] / 255.0;
...
float r = 1.0 - (c * (1.0 - k) + k);
...
q[0] = VIPS_D65_X0 * r;
```

Perhaps that's a bit clearer? Maybe not worth it.

libvips/colour/XYZ2CMYK.c
```diff
@@ -2,7 +2,7 @@
  * vips_icc_export.
  *
  * 21/12/18
- *      - from CMYK2XYZ.c
+ *      - from CMYK2CMYK.c
```
jcupitt: Search-and-replace typo I guess.

libvips/colour/XYZ2CMYK.c
```diff
@@ -94,20 +94,64 @@ vips_XYZ2CMYK_build( VipsObject *object )
...
+
+    int i;
+
+    for (i = 0; i < width; i++, p += 3, q += 4) {
+        float R, G, B;
+        int r, g, b, og;
+        int c, m, y, k;
+        
+        vips_col_XYZ2scRGB(p[0], p[1], p[2], &R, &G, &B);
+        vips_col_scRGB2sRGB_8(R, G, B, &r, &g, &b, &og);
```
jcupitt: Again, I think you can just scale by `D65_X0` etc.

```C
float r = p[0] / VIPS_D65_X0;
...
c = 1.0 - r;
...
k = VIPS_MIN( c, VIPS_MIN( m, y ) );
...
c = (c - k) / (1.0 - k);
...
q[0] = VIPS_CLIP( 0.0, 255.0 * c, 255.0 );
```

libvips/colour/XYZ2CMYK.c
```diff
@@ -94,20 +94,64 @@ vips_XYZ2CMYK_build( VipsObject *object )
...
+
+        if (r == 0 && g == 0 & b == 0) {
```
jcupitt: Should be `&&` second time.

libvips/colour/XYZ2CMYK.c
```diff
@@ -94,20 +94,64 @@ vips_XYZ2CMYK_build( VipsObject *object )
...
+            q[3] = 255;
+            continue;
```
jcupitt: You'll need to set `q[0] / 1 / 2` as well. 

I don't like continue in loops except for errors, I think this would be clearer as an if / else, if it's still necessary.

libvips/colour/XYZ2CMYK.c
```diff
@@ -94,20 +94,64 @@ vips_XYZ2CMYK_build( VipsObject *object )
...
+        }
+        c = 255 - r;
+        m = 255 - g;
+        y = 255 - b;
+        k = c;
+        k = VIPS_MIN(k, VIPS_MIN(m, y));
+
+        q[0] = VIPS_CLIP(0, 255, (c - k) / (255.0 - k) * 255.0);
```
jcupitt: It's `VIPS_CLIP( minimum, value, maximum )`, so swap the last two args.

libvips/colour/colourspace.c
```diff
@@ -255,8 +257,8 @@ static VipsColourRoute vips_colour_routes[] = {
 	{ LABQ, LCH, { vips_LabQ2Lab, vips_Lab2LCh, NULL } },
 	{ LABQ, CMC, { vips_LabQ2Lab, vips_Lab2LCh, vips_LCh2CMC, NULL } },
 	{ LABQ, LABS, { vips_LabQ2LabS, NULL } },
-	{ LABQ, CMYK, { vips_LabQ2Lab, vips_Lab2XYZ, vips_XYZ2CMYK } },
-	{ LABQ, scRGB, { vips_LabQ2Lab, vips_Lab2XYZ, vips_XYZ2scRGB } },
+	{ LABQ, CMYK, { vips_LabQ2Lab, vips_Lab2XYZ, vips_XYZ2CMYK, NULL } },
```
jcupitt: Oh, great catch! Duh.

libvips/colour/XYZ2CMYK.c
dbouron: Good catch, I will fix it in next patch. Thx :)

libvips/colour/XYZ2CMYK.c
dbouron: Oups sorry

libvips/colour/XYZ2CMYK.c
dbouron: Yes I see, I did it because q[1..2] will have no impact on color rendering if q[3] == 255.
I will change that

libvips/colour/CMYK2XYZ.c
dbouron: Yes, it more readable and better for futur dev

dbouron:
Yes, fixed in next commit

Issue #1197, opened 2018-12-31
Compilation fails if no lcms2 has been found


dbouron:
Because of recent merged commits, compilation failed if HAVE_LCMS2 is not defined.
VipsImage struct is not declared because vips/vips.h is never include in both CMYK2XYZ.c and XYZ2XYZ.c files.

jcupitt:
OK, merged. Thanks!

PR #1198, opened 2018-12-31
Fix compilation error when lcms has not been found

dbouron:
If HAVE_LCMS2 is not defined, VipsImage is not declared in both
CMYK2XYZ and XYZ2CMYK source files.

#1197


jcupitt:
Duh, my fault, sorry.

PR #1207, opened 2019-01-09
PNG writer: ensure XMP metadata is ignored when strip=TRUE

lovell:
Using the latest code on the `master` branch I was seeing XMP metadata attached to PNG output even when `[strip]` was set.

This PR fixes the problem by moving all of the metadata-writing logic into a single `if( !strip ) { ... }` block. (Thanks to indentation the diff is a bit wonky so this change looks greater than it really is.)


jcupitt:
Oh, much nicer. Sill to have forgotten the `strip` flag before.

PR #1209, opened 2019-01-10
Add missing C++ functions

kleisauke:
The functions `has_alpha()` and `copy_memory()` were missing from the C++ binding.

### DIFF
--- a/cplusplus/VError.cpp
+++ b/cplusplus/VError.cpp
@@ -32,8 +32,6 @@
 #endif /*HAVE_CONFIG_H*/
 #include <vips/intl.h>
 
-#include <iostream>
-
 #include <vips/vips8>
 
 VIPS_NAMESPACE_START
--- a/cplusplus/VImage.cpp
+++ b/cplusplus/VImage.cpp
@@ -563,7 +563,7 @@ VImage::new_from_file( const char *name, VOption *options )
 }
 
 VImage 
-VImage::new_from_buffer( void *buf, size_t len, const char *option_string, 
+VImage::new_from_buffer( const void *buf, size_t len, const char *option_string, 
 	VOption *options )
 {
 	const char *operation_name;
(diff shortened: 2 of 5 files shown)


kleisauke:
Just noticed that a wrapper function for `vips_image_remove` is also missing. 

Shall I add this too? This allows users to use `image.remove(VIPS_META_ORIENTATION)` instead than `vips_image_remove(image.get_image(), VIPS_META_ORIENTATION)`.

kleisauke:
Not related to this PR:
A `std::string` overload function for `new_from_buffer` would also be useful, see this example:

```cpp
std::stringstream svg;
svg << R"(<svg xmlns="http://www.w3.org/2000/svg" version="1.1" width="100" height="100">)" << '\n'
    << R"(<circle cx="50" cy="50" r="50" />)" << '\n'
    << "</svg>";

std::string svg_mask = svg.str();

/*
This would be useful:
VImage mask = VImage::new_from_buffer(svg_mask, "", VImage::option()->set("access", VIPS_ACCESS_SEQUENTIAL));
*/
VImage mask = VImage::new_from_buffer(reinterpret_cast<void *>(&svg_mask[0u]), svg_mask.size(),
                                      "", VImage::option()->set("access", VIPS_ACCESS_SEQUENTIAL));
```

jcupitt:
Sure, a `std::string` for `new_from_buffer` sounds useful!

kleisauke:
Done with https://github.com/kleisauke/libvips/commit/c397543c8e1710d57d61e788a72dfd84e8aeabf0. Shall I open a separate pull request for that or include it in this PR?

jcupitt:
Sure, it's only a small change, let's do it all here.

kleisauke:
OK, all done.

jcupitt:
OK, all looks good! Nice work.

PR #1211, opened 2019-01-14
vips_image_temp_name: Fixed race condition / heap-buffer-overflow

lhecker:
`vips_image_temp_name` can be called concurrently and thus the usage of the static `name` array is not thread safe.
`-fsanitize=address` spuriously complains about `heap-buffer-overflow` errors due to this.

---

**A different issue follows, but I thought I'd might write it up already.**

I recently upgraded all dependencies of a project I'm working on from libvips 8.6.4 to 8.7.2, as well as Alpine from 3.7 to 3.8 (i.e. an upgrade of liborc from 0.4.27 to 0.4.28).
Ever since then I'm seeing spurious, seemingly racy, segmentation faults.

This PR is the first thing that came out of the bug hunt. :smile: 

I personally suspect that this is due to liborc, as I can see the following if I compile with `-fsanitize=address`:
```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==16786==ERROR: AddressSanitizer: SEGV on unknown address 0x60400f00000c (pc 0x7f8be84be206 bp 0x62a00103d8a0 sp 0x7f8bdc0d9548 T42)
==16786==The signal is caused by a READ memory access.
AddressSanitizer:DEADLYSIGNAL
    #0 0x7f8be84be205 in orc_code_chunk_merge (/usr/lib/x86_64-linux-gnu/liborc-0.4.so.0+0xd205)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV (/usr/lib/x86_64-linux-gnu/liborc-0.4.so.0+0xd205) in orc_code_chunk_merge
Thread T42 created by T41 here:
...
```

Interestingly on the production servers SIGSEGV always happens in a deep chain of recursive `vips_object_close` calls instead (600+ frames deep!).
The crash happens on `push` or `call` instructions which lead me to believe that this is a stackoverflow, but the difference between the stack pointers in the highest and lowest frames is only around 84kB.
Due to that I believe that the current libvips 8.7.2 & liborc 0.4.28 combination might silently currupt heaps/stacks and thus lead to segmentation faults for users, if libvips is used concurrently.

If you happen to be interested in that I'd be able to _privately_ provide you with 6 different backtraces. While I sadly won't be able to provide you with a proper binary and it's core dumps, I could write a small test case though, given enough time.

### DIFF
--- a/libvips/iofuncs/image.c
+++ b/libvips/iofuncs/image.c
@@ -1666,7 +1666,13 @@ vips_image_temp_name( char *name, int size )
 {
 	static int global_serial = 0;
 
-	int serial = g_atomic_int_add( &global_serial, 1);
+	int serial =
+#if GLIB_CHECK_VERSION( 2, 30, 0 )
+		g_atomic_int_add( &global_serial, 1);
+#else
+		g_atomic_exchange_and_add( &global_serial, 1);
+#endif
+
 	vips_snprintf( name, size, "temp-%d", serial );
 }
 
--- a/libvips/iofuncs/image.c
+++ b/libvips/iofuncs/image.c
@@ -1658,18 +1658,16 @@ vips_image_set_kill( VipsImage *image, gboolean kill )
 	image->kill = kill;
 }
 
-/* Make a name for a filename-less image. Use immediately, don't free the
- * result.
+/* Fills the given buffer with a temporary filename.
+ * Assuming that "int" might be 64 Bit wide a buffer size of 26 suffices.
  */
-static const char *
-vips_image_temp_name( void )
+void
+vips_image_temp_name( char *name, int size )
 {
-	static int serial = 0;
-	static char name[256];
+	static int global_serial = 0;
 
-	vips_snprintf( name, 256, "temp-%d", serial++ );
-
-	return( name );
+	int serial = g_atomic_int_add( &global_serial, 1);
+	vips_snprintf( name, size, "temp-%d", serial );
 }
 
 /**
@@ -1689,12 +1687,14 @@ VipsImage *
 vips_image_new( void )
 {
 	VipsImage *image;
+	char filename[26];
 
 	vips_check_init();
+	vips_image_temp_name( filename, sizeof(filename) );
 
 	image = VIPS_IMAGE( g_object_new( VIPS_TYPE_IMAGE, NULL ) );
 	g_object_set( image,
-		"filename", vips_image_temp_name(),
+		"filename", filename,
 		"mode", "p",
 		NULL );
 	if( vips_object_build( VIPS_OBJECT( image ) ) ) {
@@ -1741,7 +1741,9 @@ vips_image_new_mode( const char *filename, const char *mode )
 VipsImage *
 vips_image_new_memory( void )
 {
-	return( vips_image_new_mode( vips_image_temp_name(), "t" ) );
+	char filename[26];
+	vips_image_temp_name( filename, sizeof(filename) );
+	return( vips_image_new_mode( filename, "t" ) );
 }
 
 /**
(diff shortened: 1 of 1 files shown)


libvips/iofuncs/image.c
```diff
@@ -1658,18 +1658,16 @@ vips_image_set_kill( VipsImage *image, gboolean kill )
...
+vips_image_temp_name( char *name, int size )
 {
-	static int serial = 0;
-	static char name[256];
-
-	vips_snprintf( name, 256, "temp-%d", serial++ );
+	static int global_serial = 0;
 
-	return( name );
+	int serial = g_atomic_int_add( &global_serial, 1);
```
lovell: It looks like [g_atomic_int_add](https://developer.gnome.org/glib/stable/glib-Atomic-Operations.html#g-atomic-int-add) only started returning a value from glib v2.30+ so we might need to check this is available at install time. I think RHEL 6 (EOL 2020) ships with an older glib version, perhaps v2.28.

libvips/iofuncs/image.c
```diff
@@ -1689,12 +1687,14 @@ VipsImage *
 vips_image_new( void )
 {
 	VipsImage *image;
+	char filename[26];
```
lovell: 26 chars is probably enough but I wonder if there was a reason for the original 256 chars? Non-standard tmp directories?

libvips/iofuncs/image.c
jcupitt: Oops, sorry, I misread your patch, you've moved the static buffer to the caller, sorry.

OK, this sounds right. As Lovell says, we'll probably need to add to configure check and switch to `g_atomic_int_exchange_and_add()` before glib 2.30.

libvips/iofuncs/image.c
lhecker: It might just be a leftover imperfection from 8 years ago I guess?
The old code had the same hardcoded string which can't be longer than 26 characters.

jcupitt: Just habit, I think. 256 is a generic "small string" size.

The only problem perhaps is that 26 is an oddly specific value. People coming after us will wonder how it was chosen.

lhecker: I chose 26 because that's
```
strlen("temp-") + ceil(log10(1 << 63)) + /* minus sign */ 1 + /* null byte */ 1
```

Do you have a recommendation how to improve that?
Maybe using a constant in form of a macro definition?

libvips/iofuncs/image.c
lhecker: @jcupitt @lovell I pushed ab6126d. Please let me know if that suffices. 🙂

libvips/iofuncs/image.c
jcupitt: Maybe:

```C
#define TEMP_NAME_PREFIX "temp-"
/* Add 30 for an int in ascii, that should be enough for anyone (jk).
 */
#define TEMP_NAME_BUFFER_SIZE (sizeof (TEMP_NAME_PREFIX) + 30)

/* Make a name for a filename-less image. The buffer should be at least 
 * TEMP_NAME_BUFFER_SIZE characters. 
 */
static const char *
vips_image_temp_name( char *name, int size )
{
        static int global_serial = 0;

        int serial = g_atomic_int_add( &global_serial, 1);

        vips_snprintf( name, size, TEMP_NAME_PREFIX "-%d", serial++ );

        return( name );
}
```

I must stop, I'm not thinking clearly. I'm celebrating getting a grant of £150 (via Kirk) to organise a libvips London meeting, so I've had ahem a couple of glasses of wine. 

Anyway, this all looks great and I don't have much to add, so ... cheers!

jcupitt:
OK, merged. Thank you @lhecker !

libvips/iofuncs/image.c
```diff
@@ -1658,18 +1658,22 @@ vips_image_set_kill( VipsImage *image, gboolean kill )
...
-	static int serial = 0;
-	static char name[256];
+	static int global_serial = 0;
 
-	vips_snprintf( name, 256, "temp-%d", serial++ );
+	int serial =
+#if GLIB_CHECK_VERSION( 2, 30, 0 )
+		g_atomic_int_add( &global_serial, 1);
+#else
+		g_atomic_exchange_and_add( &global_serial, 1);
```
kleisauke: `g_atomic_exchange_and_add` -> [`g_atomic_int_exchange_and_add`](https://developer.gnome.org/glib/stable/glib-Atomic-Operations.html#g-atomic-int-exchange-and-add)

jcupitt: Oops! I'll fix it. Nice spot Kleis.

lhecker: Oh dear... I'm sorry for this mistake. :disappointed:

PR #1212, opened 2019-01-14
Expose libwebp reduction_effort parameter

lovell:
Skal from the team that maintains libwebp at Google opened https://github.com/lovell/sharp/issues/1545 to note that its `method` parameter is not currently made available.

His recommendation is to expose this as the more appropriately-named `reduction_effort` (I agree with this as the use of `method` is far too overloaded).

Hence this PR, which adds and documents a new `reduction_effort` parameter to the WebP savers.

Running the following commands:
```sh
$ vips copy input.jpg reduction_effort.default.webp
$ vips copy input.jpg reduction_effort.0.webp[reduction_effort=0]
$ vips copy input.jpg reduction_effort.4.webp[reduction_effort=4]
$ vips copy input.jpg reduction_effort.6.webp[reduction_effort=6]
```
results in:
```
365958 reduction_effort.0.webp
253340 reduction_effort.4.webp
253340 reduction_effort.default.webp
244142 reduction_effort.6.webp
```


jcupitt:
Looks good! Very nice.

PR #1214, opened 2019-01-16
Ensure ICC profile data is freed after use (master)

lovell:
The recent change to the `master` branch introduced by #1194 leaks the ICC profile data:
```
==28548== 2,429 bytes in 1 blocks are definitely lost in loss record 6,423 of 6,523
==28548==    at 0x483774F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==28548==    by 0x9424650: g_malloc (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.5800.1)
==28548==    by 0x8FCD261: vips_malloc (memory.c:176)
==28548==    by 0x8FD7722: vips__file_read (util.c:754)
==28548==    by 0x8FD77D6: vips__file_read_name (util.c:787)
==28548==    by 0x8EFBA6D: vips_profile_load_build (profile_load.c:137)
==28548==    by 0x8FBEF18: vips_object_build (object.c:367)
==28548==    by 0x8FCAA57: vips_cache_operation_buildp (cache.c:866)
==28548==    by 0x8FD074C: vips_call_required_optional (operation.c:880)
==28548==    by 0x8FD0E55: vips_call_by_name (operation.c:920)
==28548==    by 0x8FD12A8: vips_call_split (operation.c:1024)
==28548==    by 0x8EFBC95: vips_profile_load (profile_load.c:221)
==28548==    by 0x8F00EED: vips_icc_transform_build (icc_transform.c:1089)
```
The proposed change in this PR seems to fix things.

PR #1218, opened 2019-01-24
profile_load: ensure once-ness, prevent profile use-after-free (master)

lovell:
Hello, to add to #1214, this prevents the new static profiles leaking and a use-after-free.

### DIFF
--- a/libvips/colour/profile_load.c
+++ b/libvips/colour/profile_load.c
@@ -95,7 +95,7 @@ vips_fallback_profile_get_init( void )
 static void *
 vips_fallback_profile_get( const char *name, size_t *length )
 {
-	GOnce once = G_ONCE_INIT;
+	static GOnce once = G_ONCE_INIT;
 
 	GSList *p;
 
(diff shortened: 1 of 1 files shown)


libvips/colour/profile_load.c
```diff
@@ -95,7 +95,7 @@ vips_fallback_profile_get_init( void )
 static void *
 vips_fallback_profile_get( const char *name, size_t *length )
 {
-	GOnce once = G_ONCE_INIT;
+	static GOnce once = G_ONCE_INIT;
```
jcupitt: Ahem ooops!

libvips/colour/profile_load.c
```diff
@@ -240,8 +240,8 @@ vips__profile_set( VipsImage *image, const char *name )
 	if( profile ) { 
 		data = ((VipsArea *) profile)->data;
 		length = ((VipsArea *) profile)->length;
-		vips_image_set_blob( image, VIPS_META_ICC_NAME, 
-			(VipsCallbackFn) NULL, data, length );
+		vips_image_set_blob_copy( image, VIPS_META_ICC_NAME,
+			data, length );
```
jcupitt: Argh you're right, this is wrong too, but it will cause a copy. Really we want to just up the refcount on the blob.

How about:

```C
        GValue value = { 0 };

        g_value_init( &value, VIPS_TYPE_BLOB );
        g_value_set_boxed( value, profile );
        vips_image_set( image, VIPS_META_ICC_NAME, &value );
        g_value_unset( &value );
```

lovell: Thanks John, I've updated the PR with this approach.

jcupitt:
Great work Lovell.

PR #1220, opened 2019-02-01
Use pyvips to generate the C++ binding

kleisauke:
Since the vips8 Python interface that came with libvips is disabled by default. 

For easier reviewing, I've sorted all methods in the same way as the old ones (manually :sweat_smile:). Once the PR is approved, you can re-run `gen-operators.py` to sort it alphabetically. I couldn't find a way to keep the old order.


cplusplus/gen-operators.py
```diff
@@ -1,212 +1,257 @@
...
+                       if ((flags & _OUTPUT) != 0 and
+                           (flags & _REQUIRED) != 0) or
+                       ((flags & _INPUT) != 0 and
+                        (flags & _REQUIRED) != 0 and
+                        (flags & _MODIFY) != 0) and
+                       name != member_x]
+
+    has_output = len(required_output) >= 1
+
+    # TODO: Should we output the operation description as comment?
```
kleisauke: Would this be useful?

cplusplus/vips-operators.cpp
```diff
@@ -1,156 +1,156 @@
...
 
-void VImage::system( char * cmd_format , VOption *options )
+void VImage::system( char *cmd_format, VOption *options )
 {
-    call( "system" ,
-        (options ? options : VImage::option()) ->
-            set( "cmd-format", cmd_format ) );
+    call( "system",
+        (options ? options : VImage::option())->
+            set( "cmd_format", cmd_format ) );
```
kleisauke: The old one (`cmd-format`) seems to be incorrect. See: https://github.com/libvips/libvips/blob/10c4831a709063189d67b7d6076b1b26bf62b2b2/libvips/iofuncs/system.c#L289

cplusplus/gen-operators.py
```diff
@@ -1,212 +1,257 @@
...
-            arg = cppize(prop.name)
-            if flags & Vips.ArgumentFlags.OUTPUT and prop == result:
-                arg = '&' + arg
+        return ffi.NULL
 
-            print 'set( "%s", %s )' % (prop.name, arg),
+    type_map(type_from_name('VipsOperation'), add_nickname)
 
-    print ');'
+    # TODO: Should we add the 'missing' extract_area synonym by hand?
```
kleisauke: Other bindings have this too, but I'm not sure if this is useful for C++.

cplusplus/gen-operators.py
jcupitt: Sure, why not. I'm sure some C++ users would like it.

cplusplus/gen-operators.py
```diff
@@ -1,212 +1,257 @@
...
-    "VipsArrayDouble" : "std::vector<double>",
-    "VipsArrayImage" : "std::vector<VImage>",
-    "VipsBlob" : "VipsBlob *"
+    GValue.gbool_type: 'bool',
+    GValue.gint_type: 'int',
+    GValue.gdouble_type: 'double',
+    GValue.gstr_type: 'char *',
+    GValue.refstr_type: 'char *',
+    GValue.gflags_type: 'int',
+    GValue.gobject_type: 'char *',
```
jcupitt: This is curious. Do we need one for gobject? It seems to work OK without this line.

cplusplus/gen-operators.py
```diff
@@ -1,212 +1,257 @@
...
-    if result != None:
-        print ''
-        print '    return( %s );' % cppize(result.name)
+    # make list unique and sort
+    all_nicknames = list(set(all_nicknames))
+    all_nicknames.sort()
 
-    print '}'
-    print ''
+    print('Generating {0} ...'.format(filename))
```
jcupitt: This script is usually run by `Makefile.am` if `vips-operators.cpp` is missing. It's probably simpler to do the date stuff there. Or remove the stuff from the Makefile if we're going to get maintainers to run the script by hand.

cplusplus/gen-operators.py
```diff
@@ -1,212 +1,257 @@
...
 
-# we have a few synonyms ... don't generate twice
-generated = {}
+    with open(filename, 'w') as f:
+        f.write(preamble.format('headers' if declarations_only else "bodies", now))
```
jcupitt: If we're going to run from the Makefile, it's probably best to have a `--gen-headers` switch to this script, and then run the script from `include/vips/Makefile.am` with that flag set. Don't write to files in this script, just print, and get our caller to send the result to the right place.

cplusplus/gen-operators.py
```diff
@@ -1,212 +1,257 @@
...
+    return result
+
+
+preamble = """// {0} for vips operations
+// {1}
+// this file is generated automatically, do not edit!
+"""
 
-    print 'VImage::%s(' % nickname,
+now = datetime.datetime.now().strftime('%a %d %b %H:%M:%S %Y')
```
jcupitt: Again, it's probably simpler to create this in the Makefile.

cplusplus/vips-operators.cpp
jcupitt: I think it's harmless, see eg. https://github.com/libvips/libvips/blob/master/cplusplus/vips-operators.cpp#L1244 but I guess needs testing.

cplusplus/gen-operators.py
jcupitt: Yes, let's have `crop` too, if we can.

cplusplus/gen-operators.py
kleisauke: Removed (leftover from: https://github.com/libvips/php-vips/blob/master/examples/generate_phpdoc.py#L29).

cplusplus/gen-operators.py
kleisauke: Fixed with: https://github.com/libvips/libvips/pull/1220/commits/241f52fb54fa32ecc1f36238badcd255d69d5283.

The parameters are also described with that commit.

cplusplus/vips-operators.cpp
kleisauke: Hopefully it will be automatically converted from `-` to `_`. Your example also has an underscore:
https://github.com/libvips/libvips/blob/10c4831a709063189d67b7d6076b1b26bf62b2b2/libvips/create/mask_butterworth.c#L94

I'm going to test this.

kleisauke: Just tested this, and it seems to makes no difference (so both `_` or `-` will work). Tested with:
```cpp
// g++ -g -Wall example.cc `pkg-config vips-cpp --cflags --libs`

#include <vips/vips8>
#include <cassert>

auto main(int argc, const char *argv[]) -> int {
    if (vips_init(argv[0]) != 0) {
        vips_error_exit(nullptr);
    }

    auto im = vips::VImage::gaussmat(1, 0.1);
    assert(im.width() == 5);
    assert(im.height() == 5);
    assert(im.bands() == 1);
    assert(im.format() == VIPS_FORMAT_DOUBLE);
    assert(im.max() == 20.0);

    auto total = im.avg() * im.width() * im.height();
    auto scale = im.get_double("scale");
    assert(total == scale);

    auto p = im.getpoint(im.width() / 2, im.height() / 2);
    assert(p[0] == 20.0);

    vips_shutdown();
}
```

jcupitt:
I tested it too, and it all works fine. Nice work Kleis!

PR #1221, opened 2019-02-02
vips_thumbnail_buffer: add optional option_string argument (#1147)

### LINKED ISSUE #1147: vips_thumbnail_buffer: pass options to underlying image loader
When using [`vips_thumbnail_buffer`](https://jcupitt.github.io/libvips/API/current/libvips-resample.html#vips-thumbnail-buffer), there's no possibility to pass options to the underlying image loader (see [here](https://gist.github.com/kleisauke/0109790229d0b05e0ccc621a957aa5d8) for a C++ example).

Hints for implementation:
A simple way to solve this issue is to add a new `option_string` parameter (`"[name=value,...]"`) to `vips_thumbnail_buffer` and pass this string to the [`vips_image_new_from_buffer`](https://jcupitt.github.io/libvips/API/current/VipsImage.html#vips-image-new-from-buffer) calls:

https://github.com/libvips/libvips/blob/ecdbf66f4dde3865f02b4974c798848839fc414c/libvips/resample/thumbnail.c#L841-L845

(Not sure if this will break API compatibility, perhaps a compatibility stub needs to be added.)

kleisauke:
For example, this allows us to thumbnail any page from a PDF-formatted memory buffer:
```cpp
auto buffer_thumb = vips::VImage::thumbnail_buffer(blob, 200,
                                                   vips::VImage::option()->set("option_string", "[page=1]"));
```

Resolves #1147.

### DIFF
--- a/libvips/resample/thumbnail.c
+++ b/libvips/resample/thumbnail.c
@@ -983,7 +983,7 @@ typedef struct _VipsThumbnailBuffer {
 	VipsThumbnail parent_object;
 
 	VipsArea *buf;
-	char *open_args;
+	char *option_string;
 } VipsThumbnailBuffer;
 
 typedef VipsThumbnailClass VipsThumbnailBufferClass;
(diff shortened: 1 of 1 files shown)


jcupitt:
Nice! Yes, this is needed :( it was a mistake to drop that param.

PR #1224, opened 2019-02-05
vips_magicksave for GraphicsMagick

DarthSim:
Hi,

Looks like I made `vips_magicksave` working with GM. There is what I've done:

1. GraphicsMagick has [its own implementation](http://hg.code.sf.net/p/graphicsmagick/code/file/c38fc0e3e465/wand/magick_compat.c#l315) of `ImportImagePixels` in the wand part. Since it's pointlessly to link the wand part, I just borrowed the code. Using `SetImagePixels` may be a more effective solution, but the current one is better than nothing.

2. GM doesn't have `UndefinedPixel` value. Since we anyway will throw an error in `vips_foreign_save_magick_build ` if the band format is unsupported, I've decided that we can safely assign `CharPixel` to `magick->storage_type` at the start. Maybe, we even can remove this assignation, but I'm not sure about this.

3. Also, GM doesn't have `ImagesToBlob` function. From what I researched, it successfully saves multi-frame images with `ImageToBlob`, so I use one when `ImagesToBlob` is not found. Tested with GIF and TIFF, works as expected.

I didn't use C since university time, so I'm sorry if there are problems with code style and stuff :)

### DIFF
--- a/libvips/foreign/magick.c
+++ b/libvips/foreign/magick.c
@@ -78,6 +78,13 @@ magick_import_pixels( Image *image, const ssize_t x, const ssize_t y,
 		type, pixels, exception ) );
 }
 
+void *
+magick_images_to_blob( const ImageInfo *image_info, Image *images, size_t *length,
+	ExceptionInfo *exception )
+{
+	return( ImagesToBlob( image_info, images, length, exception ) );
+}
+
 void
 magick_set_property( Image *image, const char *property, const char *value,
 	ExceptionInfo *exception )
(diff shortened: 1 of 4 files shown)


libvips/foreign/magick.c
```diff
@@ -44,6 +44,8 @@
 
 #include <vips/vips.h>
 
+#include <assert.h>
```
jcupitt: Don't need this, use `g_assert()` instead.

libvips/foreign/magick.c
```diff
@@ -193,6 +195,16 @@ magick_import_pixels( Image *image, const ssize_t x, const ssize_t y,
 	return( ImportImagePixels( image, x, y, width, height, map,
 		type, pixels ) );
 #else /*!HAVE_IMPORTIMAGEPIXELS*/
+	assert(image != (Image *) NULL);
+	assert(image->signature == MagickSignature);
```
jcupitt: use g_assert() here.

libvips/foreign/magick.c
```diff
@@ -193,6 +195,16 @@ magick_import_pixels( Image *image, const ssize_t x, const ssize_t y,
...
+
+	Image *constitute_image=
+		ConstituteImage(width,height,map,type,pixels,&image->exception);
+	if (constitute_image) {
+		(void) CompositeImage(image,CopyCompositeOp,constitute_image,x,y);
+		DestroyImage(constitute_image);
+		return (image->exception.severity == UndefinedException);
+	}
```
jcupitt: Oh, fantastic. I didn't find the implementation in wand. This is much better.!

libvips/foreign/magicksave.c
```diff
@@ -465,8 +465,15 @@ vips_foreign_save_magick_buffer_build( VipsObject *object )
...
 
-	if( !(obuf = ImagesToBlob( magick->image_info, magick->images, 
-		&olen, magick->exception )) ) { 
+#ifdef HAVE_IMAGESTOBLOB
+	obuf = ImagesToBlob( magick->image_info, magick->images,
+		&olen, magick->exception );
+#else
+	obuf = ImageToBlob( magick->image_info, magick->images,
+		&olen, magick->exception );
+#endif /*HAVE_IMAGESTOBLOB*/
```
jcupitt: I think it'd be best to make a little wrapper in magick.c called `magick_images_to_blob()` and call that here. 

The aim of magick.c is to get all these ugly ifdefs out of the main body of the code.

jcupitt:
This is fantastic! Thank you @DarthSim, very nice!

I added some small comments. I can address this stuff if you're busy.

DarthSim:
Done

jcupitt:
Looks great! Thank you very much for doing this work @DarthSim.

DarthSim:
Glad to help :)
Libvips is awesome, thank you for your work!

PR #1235, opened 2019-02-20
Add support for HEIF files on Travis

jcupitt:


jcupitt:
I fixed up heifload/save with 1.3.2, so hopefully this goes on cleanly now. I revised the heif tests too.

PR #1241, opened 2019-02-27
change init message

jcupitt:
silly test commit

### DIFF
--- a/libvips/iofuncs/init.c
+++ b/libvips/iofuncs/init.c
@@ -709,7 +709,7 @@ static GOptionEntry option_entries[] = {
 		N_( "show progress feedback" ), NULL },
 	{ "vips-leak", 0, 0, 
 		G_OPTION_ARG_NONE, &vips__leak, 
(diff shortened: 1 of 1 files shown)


libvips/iofuncs/init.c
```diff
@@ -709,7 +709,7 @@ static GOptionEntry option_entries[] = {
 		N_( "show progress feedback" ), NULL },
 	{ "vips-leak", 0, 0, 
 		G_OPTION_ARG_NONE, &vips__leak, 
-		N_( "leak-check on exit" ), NULL },
+		N_( "leak-check on brexit" ), NULL },
```
jcupitt: what??

lovell: "A Seaborne Freight spokesperson claimed there couldn't possibly be any leaks as there are no ships."

jcupitt: I was having to do a git tutorial and needed a guinea pig :(

Issue #803, opened 2017-11-15
Add save to descriptor


UnitedMarsupials:
It makes sense that the target filename is examined to determine the desired file type.

But it should not be the only way to specify the type. Say, for example, my PHP-application needs to output the image to stdout or a network socket... Currently, the only way to do that is by writing the entire image into a buffer, and then deal with the buffer.

The `vips_image_write_to_*` ought to examine the the options list _first_ and only fall back to checking the filename extension, if `format` is not specified in the options...

jcupitt:
Hello, the `write_to_buffer` function is very simple: it looks at the suffix and finds the highest-priority saver that supports that extension.

You can call the savers directly, for example in Python:

```python
data = image.jpegsave_buffer(Q=85)
```

Unfortunately llibvips does not support writing to a socket -- it's just file or memory buffer. If your platform is linux, you can write to a named pipe, or to something like `/dev/stdout`, if that's any help.

UnitedMarsupials:
> You can call the savers directly

Heh, I can call into the relevant format-specific libraries directly too. But, if I'm using VIPS API, I'd like to stay within it...

> Unfortunately llibvips does not support writing to a socket

Why does it matter to the library, where the file-descriptor came from? Maybe, some savers require a _seekable_ object, but I doubt many of them do...

jcupitt:
It's still in the libvips API, just a different function. In PHP:

```php
$image->jpegsave("/dev/stdout");
```

To write a JPG to stdout. 

There's a branch that adds true streaming to and from sockets (if the format supports it), but it's never been merged, for various reasons. 

There was a long discussion here:

https://github.com/lovell/sharp/issues/30

jcupitt:
... but I agree that write to a file descriptor would be a nice, and fairly simple addition. Would that meet your needs?

UnitedMarsupials:
> Would that meet your needs?

How would I specify the filetype? If the same way it is done for when writing to a buffer, then yes :)

jcupitt:
Yes, something like `$image->writeToDescriptor(".jpg", stdout);` I suppose.

UnitedMarsupials:
> Yes, something like $image->writeToDescriptor(".jpg", stdout); I suppose.

Or `vips_image_write_to_filedescriptor()`, right? Yes, that works, thank you!

jcupitt:
Nope :( not yet.

ioquatix:
[...]

PR #1256, opened 2019-03-12
early feedback: loading and storing via FILE

srijs:
For my use-case, I'd like to be able to load and save images from/to stdin/stdout, similar to what's been described in #803.

I understand that eventually you'd like to see something based on `GFile` or `GInputStream`/`GOutputStream`, however I'm just starting out on this codebase and didn't want to pick such a large task just yet.

As a compromise, I've started to implement loading and storing functions that can deal with standard `FILE` objects.

What I'm mainly looking for here is to get early feedback on the approach, before I start to change more code.

Let me know what you think!

### DIFF
--- a/doc/function-list.xml
+++ b/doc/function-list.xml
@@ -822,6 +822,11 @@
   <entry>vips_pngsave()</entry>
 </row>
 <row>
+  <entry>pngsave_stream</entry>
+  <entry>save image to png stream</entry>
+  <entry>vips_pngsave_stream()</entry>
+</row>
+<row>
   <entry>pngsave_buffer</entry>
   <entry>save image to png buffer</entry>
   <entry>vips_pngsave_buffer()</entry>
--- a/libvips/foreign/foreign.c
+++ b/libvips/foreign/foreign.c
@@ -1856,6 +1856,7 @@ vips_foreign_operation_init( void )
 	extern GType vips_foreign_load_png_get_type( void ); 
 	extern GType vips_foreign_load_png_buffer_get_type( void ); 
 	extern GType vips_foreign_save_png_file_get_type( void ); 
+	extern GType vips_foreign_save_png_stream_get_type( void );
 	extern GType vips_foreign_save_png_buffer_get_type( void ); 
 	extern GType vips_foreign_load_csv_get_type( void ); 
 	extern GType vips_foreign_save_csv_get_type( void ); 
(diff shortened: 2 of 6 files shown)


jcupitt:
Hello @srijs, this is very interesting, sorry I've not responded. I'm horribly busy for the next 10 days or so, but I'll looks at this after that's done.

jcupitt:
I'd really like to do something like this for 8.8, but I think this is the wrong approach, sorry :( For example, you won't be able to write to an http stream with this, which has to be one of the main advantages.

I think we should add simple input and output stream classes. https://github.com/lovell/sharp/issues/30 outlines `VipsStream`, `VipsStreamInput` and `VipsStreamOutput`. They are only a few 100 lines of code, it's nothing like the complexity of `GFile` etc. (phew!). Here's the implementation from 2014:

https://github.com/libvips/libvips/blob/load-from-stream/libvips/iofuncs/stream.c

The feature we need that was missing last time we tried this is rewind: the input class needs to keep the first part of the image data it reads (the header) in a cache, and resupply it if necessary. This would let us open the same stream twice, once to get the header, and a second time to get the pixels. Load operations will need some way to signal to the input object when they are parsing the header. 

We then need to rework PNG load/save (to pick an example) to run on top of `VipsStream`, and reimplement the current file and memory load/save as tiny wrappers.

Here's how that was done in 2014: 

https://github.com/libvips/libvips/blob/load-from-stream/libvips/foreign/vipspng.c#L768

apuckey:
Hi,

I have been playing around with the php extensions and libvips. very impressed with the ram use however not having streams almost defeats the purpose when you have to load a 2gb file in ram to process :D 

Could you not just use a library like libcurl to handle the streaming to/from http streams ? added bonus is it works on windows as well using the same code i believe.

Ignore if you have already thought of it and discounted it. I'm not a c guru so could be totally off base here :)

cheers

jcupitt:
Hiya,

It's not quite as bad as it sounds -- for example, with a JPG , you only need to load the compressed image into memory. libvips will process it as:

```
read -> jpeg in memory -> 128 scanline buffer -> process -> 
    128 scanline buffer -> jpg in memory -> write
```

So you only need the memory for a few lines of uncompressed pixels. PNG will see a saving too, since images are typically 30% smaller than uncompressed pixels. It won't help much for TIFF I guess.

Of course you can do:

```
read to file in /tmp -> 128 scanline buffer -> process -> 
    128 scanline buffer -> file in /tmp -> write
```

Assuming you have enough `tmp`. It's actually no slower, on linux anyway.

True streaming, where you never need to have the entire compressed image buffered in memory, is on the wishlist but has not happened yet, and won't in the short-term.

jcupitt:
Here's a PR that adds true streaming support:

https://github.com/libvips/libvips/pull/1443

Though only for jpg at the moment. Any comments very welcome.

PR #1262, opened 2019-03-20
Convert tools/vipsprofile to Python 3+2

hroncok:
I don't know how to test this (I don't know or use libvips), so please give it a try.
Unless there is a typo, I'm pretty confident this cannot break anything, however I don't know yet if this the full change needed.

### DIFF
--- a/tools/vipsprofile
+++ b/tools/vipsprofile
@@ -1,15 +1,18 @@
-#!/usr/bin/python2
+#!/usr/bin/python
+from __future__ import division
+from __future__ import print_function
 
 import re
-import math
 import cairo
+from io import open
+
 
 class ReadFile:
     def __init__(self, filename):
         self.filename = filename
 
     def __enter__(self):
-        self.f = open(self.filename, 'r') 
+        self.f = open(self.filename, 'r', encoding='utf-8')
         self.lineno = 0
         self.getnext();
         return self
@@ -17,8 +20,10 @@ class ReadFile:
     def __exit__(self, type, value, traceback):
         self.f.close()
 
-    def __nonzero__(self):
+    def __bool__(self):
         return self.line != ""
+
+    __nonzero__ = __bool__
 
     def getnext(self):
         self.lineno += 1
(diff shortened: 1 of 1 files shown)


jcupitt:
Hello, thanks for this!

I tried it out and there are two small problems still.

1. The ReadFile class uses the py2 `__nonzero__` overload to detect end of file. It needs to add `__bool__ = __nonzero__` (or the reverse, I guess?) for py3.
2. There are the usual Python int/float division mixups. Probably `from __future__ import division` and move everything to py3 style is best.

Would you like to make these changes? I can do it if you're busy.

You can test the script like this:

```
$ vips gaussblur wtc.v x.tif 5 --vips-profile
recording profile in vips-profile.txt
$ vipsprofile 
reading from vips-profile.txt
loaded 28806 events
total time = 1.575308
name		alive	wait%	work%	unkn%	memory	peakm
    worker 20	   1.3	 0.0477	89.2	10.8	 0.47	 0.47	
    worker 21	   1.3	 0.0328	89.5	10.4	 0.47	 0.47	
    worker 22	   1.3	 0.062	87.9	12.1	 0.47	 0.47	
    worker 23	   1.3	 0.993	88.8	10.2	 0.47	 0.47	
    worker 24	   1.3	 0.261	89	10.8	 0.47	 0.47	
    worker 25	   1.3	 0.733	88	11.3	 0.47	 0.47	
    worker 26	   1.3	 0.267	88.1	11.6	 7.33	 7.33	
    worker 27	   1.3	 0.0407	89.1	10.9	 0.47	 0.47	
   wbuffer 28	   1.3	 92.1	7.85	0.00633	 0	 0	
   wbuffer 29	   1.3	 92.6	7.38	0.00687	 0	 0	
      main 30	   1.6	 81.9	0	18.1	 -8.14	 10.3	
peak memory = 20.9 MB
leak! final memory = 2.49 MB
writing to vips-profile.svg
```

I'm not sure why it's claiming a leak, I should check.

hroncok:
I'll get back to that later today. Thanks for fast feedback.

hroncok:
Do you have some `wtc.v` to test with?

Anyway, I tried to guess hard what should use floor division, hopefully I got it right.

jcupitt:
You can use any image. That was a 10k x 10k RGB (you need to give it something big to get a nice graph), but anything would work.

hroncok:
I've used benchmark/sample2.v and this is what I got:


vips-profile svg

jcupitt:
That looks great!

jcupitt:
I read your patch and I agree with everything. Nice job, and thank you very much for doing this work!

hroncok:
Thanks.

PR #1265, opened 2019-03-22
Fixed vips_rect function

crashoz:
Hello,
I was struggling to convert complex images and found out there is a copy-paste mistake in the _vips_rect_ function. It's a very small PR to fix it 😄 

Great lib!


jcupitt:
Wow, that's awful, thank you very much for this! 

The test suite uses the Python interface, not the C API, so it missed this problem.

PR #1293, opened 2019-04-22
fixes set_stacksize which broke during merge

abdollar:
- This fixes set_stacksize which broke during the merge.
- I changed the condition on checking the current stack size with the minimum to make it easier to read as it was reversed.
- I added logging using g_info so we can see the actual size the OS has set.

```
G_MESSAGES_DEBUG=VIPS VIPS_MIN_STACK_SIZE=10m vips copy stackheavy.pdf x.png
VIPS-INFO: 16:13:41.937: current stack size is set to 10485760
```


jcupitt:
Ooop, you're right, I meant to have a separate if() there, that was the whole point of breaking it up.

Thanks!

PR #1294, opened 2019-04-24
Change 'char *' to 'const char *' in the C++ binding

kleisauke:
This PR changes the `GValue` of type `G_TYPE_STRING` to `const char *` instead of `char *`. Because ISO C++11 does not allow conversion from string literal to `char *`. For example, this example code isn't valid:
```cpp
/**
 * g++ -g -Wall example.cpp `pkg-config vips-cpp --cflags --libs`
 */

#include <iostream>
#include <vips/vips8>

using vips::VImage;

auto main(int argc, const char *argv[]) -> int {
    if (vips_init(argv[0]) != 0) {
        vips_error_exit(nullptr);
    }

    VImage image = VImage::jpegload(
        "example.jpg", VImage::option()->set("access", VIPS_ACCESS_SEQUENTIAL));

    std::cout << "width: " << image.width() << std::endl;
    std::cout << "height: " << image.height() << std::endl;

    vips_shutdown();
}

```
```
example.cpp: In function ‘int main(int, const char**)’:
example.cpp:16:79: warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
         "example.jpg", VImage::option()->set("access", VIPS_ACCESS_SEQUENTIAL));
                                                                               ^
```
This warning could be hidden with `const_cast<char *>("example.jpg")` but I think this PR solves it in a neater way.


jcupitt:
Oh nice, and you fixed the docs too!

PR #1296, opened 2019-04-26
master: fix interlaced PNG input handling with newer libpng versions

lovell:
It looks like commit 57ce5a3 prevents interlaced PNG handling working with newer libpng versions.

```sh
$ vipsthumbnail interlaced.png

(vipsthumbnail:31018): VIPS-WARNING **: 22:37:46.939: bad adaptive filter value
vipsthumbnail: unable to thumbnail interlaced.png
```

This PR ensures the new call to `png_set_interlace_handling` occurs before the first call to `png_read_update_info`.

(Calling `png_set_interlace_handling` for non-interlaced PNG images appears to be OK, or at least is sensibly ignored internally by libpng.)

### DIFF
--- a/libvips/foreign/vipspng.c
+++ b/libvips/foreign/vipspng.c
@@ -471,6 +471,15 @@ png2vips_header( Read *read, VipsImage *out )
 			VIPS_META_ICC_NAME, profile, proflen );
 	}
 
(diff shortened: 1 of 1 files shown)


jcupitt:
Ooop. Nice!

I'll add a test for this.

PR #1331, opened 2019-06-05
Ensure WebP quality option is set for lossless

lovell:
This was reported in https://github.com/lovell/sharp/issues/1740


jcupitt:
Huh silly oversight, nice!

PR #1345, opened 2019-06-20
improve labpack processing

angelmixu:
At work we are unifying our loaders with libvips (or to say, we are deprecating our loaders for using libvips ones), and I found that labpack was not being processed correctly, so I found an already existing function that properly converted the color.

### DIFF
--- a/libvips/foreign/tiff2vips.c
+++ b/libvips/foreign/tiff2vips.c
@@ -741,7 +741,6 @@ rtiff_labpack_line( Rtiff *rtiff, VipsPel *q, VipsPel *p, int n, void *dummy )
 
 	float pf[3];
 	int x;
-	int i;
 
 	for( x = 0; x < n; x++ ) {
 		pf[0] = p[0] * 100.0f / 255.0f;
@@ -750,10 +749,7 @@ rtiff_labpack_line( Rtiff *rtiff, VipsPel *q, VipsPel *p, int n, void *dummy )
 
 		vips__Lab2LabQ_vec(q, pf, 1);
 
-		for (i = 3; i < samples_per_pixel; i++)
-			q[i+1] = p[i];
-
-		q += samples_per_pixel+1;
+		q += 4;
 		p += samples_per_pixel;
 	}
 }
@@ -768,7 +764,7 @@ rtiff_parse_labpack( Rtiff *rtiff, VipsImage *out )
 		rtiff_check_interpretation( rtiff, PHOTOMETRIC_CIELAB ) )
 		return( -1 );
 
-	out->Bands = rtiff->header.samples_per_pixel + 1;
+	out->Bands = 4; 
 	out->BandFmt = VIPS_FORMAT_UCHAR; 
 	out->Coding = VIPS_CODING_LABQ; 
 	out->Type = VIPS_INTERPRETATION_LAB; 
--- a/libvips/foreign/tiff2vips.c
+++ b/libvips/foreign/tiff2vips.c
@@ -739,13 +739,15 @@ rtiff_labpack_line( Rtiff *rtiff, VipsPel *q, VipsPel *p, int n, void *dummy )
 {
 	int samples_per_pixel = rtiff->header.samples_per_pixel;
 
+	float pf[3];
 	int x;
 
 	for( x = 0; x < n; x++ ) {
-		q[0] = p[0];
-		q[1] = p[1];
-		q[2] = p[2];
-		q[3] = 0;
+		pf[0] = p[0] * 100.0f / 255.0f;
+		pf[1] = (signed char)p[1];
+		pf[2] = (signed char)p[2];
+
+		vips__Lab2LabQ_vec(q, pf, 1);
 
 		q += 4;
 		p += samples_per_pixel;
--- a/libvips/foreign/tiff2vips.c
+++ b/libvips/foreign/tiff2vips.c
@@ -772,6 +772,56 @@ rtiff_parse_labpack( Rtiff *rtiff, VipsImage *out )
 	return( 0 );
 }
 
+
+/* Per-scanline process function for 8-bit VIPS_CODING_LAB to 16-bit LabS with alpha channels.
+ */
+static void
+rtiff_lab_whit_alpha_line(Rtiff* rtiff, VipsPel* q, VipsPel* p, int n, void* dummy)
+{
+	int samples_per_pixel = rtiff->header.samples_per_pixel;
+
+	unsigned char* p1;
+	short* q1;
+	int x;
+	int i;
+
+	p1 = (unsigned char*)p;
+	q1 = (short*)q;
+	int L;
+	for (x = 0; x < n; x++) {
+
+		q1[0] = ((unsigned int)(p1[0]) * 32767) / 255;
+		q1[1] = (short)(p1[1]) << 8;
+		q1[2] = (short)(p1[2]) << 8;
+
+		for (i = 3; i < samples_per_pixel; i++)
+			q1[i] = (p1[i] << 8) + p1[i];
+
+		q1 += samples_per_pixel;
+		p1 += samples_per_pixel;
+	}
+}
+
+/* Read an 8-bit LAB image with alpha bands into 16-bit LabS.
+ */
+static int
+rtiff_parse_lab_with_alpha(Rtiff* rtiff, VipsImage* out)
+{
+	if (rtiff_check_min_samples(rtiff, 4) ||
+		rtiff_check_bits(rtiff, 8) ||
+		rtiff_check_interpretation(rtiff, PHOTOMETRIC_CIELAB))
+		return(-1);
+
+	out->Bands = rtiff->header.samples_per_pixel;
+	out->BandFmt = VIPS_FORMAT_SHORT;
+	out->Coding = VIPS_CODING_NONE;
+	out->Type = VIPS_INTERPRETATION_LABS;
+
+	rtiff->sfn = rtiff_lab_whit_alpha_line;
+
+	return(0);
+}
+
 /* Per-scanline process function for LABS.
  */
 static void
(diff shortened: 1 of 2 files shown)


jcupitt:
Hi @angelmixu,

It's hard to find sample CIELAB TIFF images, but there is this:

http://www.brucelindbloom.com/downloads/ColorCheckerCalculator.tif.zip

It seems to load and save OK with the current libvips code. Do you have a sample image where it fails?

jcupitt:
The libvips LABQ format is strictly 4 bytes per pixel, so I don't think we can just copy the extra bands over.

Perhaps we should get rid of TIFF CIELAB -> LABQ and just load all CIELAB TIFFs as float? We could have the extra channels then.

angelmixu:
Hi! @jcupitt !

Sorry for the confusion, It did not fail per se, but we had an improvement for color precission when using LabQ coding.
In libVips the colors where parsed to 3 bytes (leaving the last byte unused), and with our tiff loader LabQ uses 4 bytes, a little more precission, as our color management code is improved with this.

What we had on our system was quite similar to the vips__Lab2LabQ_vec function, so I reused the function for this case and works fine.

I just tested on an image with extra channels and pushed the change for correctly parsing it.

[PreviewData.zip](https://github.com/libvips/libvips/files/3314417/PreviewData.zip)

Here I have attached a zip with two files:
- TesterLab.tif is a normal LabQ file
- LabQPreview is a LabQ with two extra channels

I have been testing with the normal LabQ file, and with my changes it copies the file correctly.
Althought, with the labQ with extra channels, it gets corrupted.

I'll work on the saver next week :)

angelmixu:
> The libvips LABQ format is strictly 4 bytes per pixel, so I don't think we can just copy the extra bands over.
> 
> Perhaps we should get rid of TIFF CIELAB -> LABQ and just load all CIELAB TIFFs as float? We could have the extra channels then.

If I apply our saver code to libVips LabQ2LabC functions, it would be something like this:

```
static void
LabQ2LabC( VipsPel *q, VipsPel *p, int n, int pBands )
{
	int x;
	int a;
	
	for( x = 0; x < n; x++ ) {
		/* Get most significant 8 bits of lab.
		 */
		*(q++) = *(p++);
		*(q++) = *(p++);
		*(q++) = *(p++);
				
		p += 1;

		for (a = 4; a < pBands; ++a )
			*(q++) = *(p++);
	}
}
```

It has been working correctly for the time being, but I should take a look when I start modifying our tiff saver.
But only changing the loader for the modified libvips one, for us works fine (as far as my tests go).

angelmixu:
Hi @jcupitt I've pushed the change for writing correctly the labQ.
For our tests look like it is working correctly :)

What do you think?

jcupitt:
Hi again @angelmixu,

> In libVips the colors where parsed to 3 bytes (leaving the last byte unused), 
> and with our tiff loader LabQ uses 4 bytes, a little more precission, as our 
> color management code is improved with this.

But there are only three bytes for LAB in the TIFF file, so the extra bits you have generated must just be noise, musn't they?

I agree that CIELAB with extra alpha channels is worth supporting. But VIPS_CODING_LABPACK is strictly four bytes per pixel, you can't add extra ones without breaking code elsewhere.

How about adding some new logic: if the TIFF loader sees CIELAB with more than three bytes per pixel, it loads as VIPS_CODING_NONE, but sets the interpretation to VIPS_INTERPRETATION_LabS, ie. 16-bit short? VIPS_INTERPRETATION_LAB (float pixels) would be another possibility.

kmartinez:
LabQ is an old but useful compression we made up years ago - to save storing CIELab as three floats. 4 bytes per pixel saved us a lot of disk load time compared to 12 bytes/pixel.
By the way I just re-enabled a small archive of images of various formats which may be useful to people:
http://degas.ecs.soton.ac.uk/vips/

angelmixu:
Hi @jcupitt ! Sorry for the delay!
> But there are only three bytes for LAB in the TIFF file, so the extra bits you have generated must just be noise, mustn't they?

We have been testing and doing some research and my coworkers have reached to the conclusion that you are right! lol

> I agree that CIELAB with extra alpha channels is worth supporting. But VIPS_CODING_LABPACK is strictly four bytes per pixel, you can't add extra ones without breaking code elsewhere.
> 
> How about adding some new logic: if the TIFF loader sees CIELAB with more than three bytes per pixel, it loads as VIPS_CODING_NONE, but sets the interpretation to VIPS_INTERPRETATION_LabS, ie. 16-bit short? VIPS_INTERPRETATION_LAB (float pixels) would be another possibility.

My coworkers say that this is a good solution :)
We usually don't work with floats in our engine and use bytes and shorts for having a better performance, so, using LABS would be best for us.

@kmartinez  thanks for the explanation!

Just wondering... if we get a normal CIELAB file, with one byte per band, so, three bytes, it could generate a VIPS_CODING_NONE, VIPS_INTERPRETATION_LAB, and VIPS_FORMAT_UCHAR, isn't it?
But as I'm guessing, VIPS_INTERPRETATION_LAB is associated with a float format and VIPS_INTERPRETATION_LABS with an unsigned short. 

At first, we don't need to use an extra byte, if the file format uses 3 bytes.
I can see the benefit of using LabQ when converting from float to bytes, but not in this case :/

(When I first approached libvips I was confused with this some times, because I thought _intrepretation_ would be the color space, and the _format_ would be the type, or size, of a pixel's band).

jcupitt:
LABS uses 0 - 32767 for L* (ie. L* == 100 is scaled to 32767), and a*b* as -32768 - 32767 (ie. -128 is scaled to -32768). Is that OK? You'll probably need to add some divides and multiplies when you interface to your own code.

You can't use `uchar` for LAB because you need the negative values or arithmetic goes wonky. You could use `char` and scale like `short` does in LABS, but you'd only get 0 - 127 for L and that's not enough precision.

angelmixu:
@jcupitt Sure! That's ok!
Lots of thanks, I'll close this issue then :)

jcupitt:
Would you like to try something to load 8-bit CIELAB + alpha as LABS? 

If you won't have time, it'd be worth making an enhancement issue instead so this idea isn't lost.

angelmixu:
Ah yes! I can try to!
In some days I'll upload something :)

angelmixu:
@jcupitt Hi! I just uploaded the loading conversion from 8bit LAB with alpha to LabS.
I have used Photoshop to generate a 16bit image from the 8bit original one, then, use vips command line tool to copy the original, so it saves a 16bit image, and compare back to the Photoshop 16bit generated image.

This looks like works fine, but there's a problem with the alpha channels.
I had to set the channels as EXTRASAMPLE_UNSPECIFIED (vips2tiff.c line 680) for being like the original file.
I haven't uploaded the code for the alpha channels as there should be another way to copy the values from the original tiff file.
Maybe a tag in the VipsImage?

What do you think?

jcupitt:
I think it's OK to leave them as UNASSALPHA.

I made a couple of small changes and rebased on top of master:

https://github.com/libvips/libvips/tree/angelmixu-8.7

Do you have some test images? We need 8 and 16-bit LAB TIFF with alpha.

angelmixu:
Hi @jcupitt !
I have created 2 test images with 2 extra alpha channels, one in 8bits and the other in 16bit:
[lab_alpha.zip](https://github.com/libvips/libvips/files/3501096/lab_alpha.zip)
If you need more images I can generate them.

Trying the branch you mentioned, using a vips copy, it generates the image with the first alpha channel as transparency, and the second the same as original.
The image in Photoshop is lighter than the original, as Photoshop has applied the transparency channel to the LAB channels.

I think we should maintain the type of the alpha channels as they are in the original image.

jcupitt:
OK, it seems to work with your test images.

Could we do the alpha change in another PR? At the moment, all of libvips is unpremultiplied, unassoc alpha. If we start adding other alpha types, it's a big change.

jcupitt:
Oh, could you send an sRGB version of your test image? Just to confirm that I am seeing the correct colours.

angelmixu:
> OK, it seems to work with your test images.
> 
> Could we do the alpha change in another PR? At the moment, all of libvips is unpremultiplied, unassoc alpha. If we start adding other alpha types, it's a big change.

Sure! I'll try to do something.
What would be the best approach? I was thinking on a new tag in VipsImage describing each channel type, this wouldn't mess with old code or images, and newer versions could benefit from it. We are already using Photoshop tags for saving some information like that, maybe we could reuse that information too?
Or do you think there's another better solution?




> Oh, could you send an sRGB version of your test image? Just to confirm that I am seeing the correct colours.

Sure! Do you mean, what Photoshop is showing on screen? Or converting the LAB image to RGB using Photoshop?
I'm out of work right now, and tomorrow is holidays here, so I'll upload them on Friday :)

jcupitt:
Re. alpha: I'd like to understand what kinds of extra information are required and why, and how that might interact with things like the thumbnailer, flatten, premultiply etc. etc.

jcupitt:
Yes, convert the LAB to sRGB in photoshop.

angelmixu:
Hi @jcupitt !
Here are both test files converted to sRGB, and the output also converted to sRGB:
[lab_alpha.zip](https://github.com/libvips/libvips/files/3508670/lab_alpha.zip)
I can notice a difference of color when converting the output file (with the applied transparency ) to RGB.

Regarding alpha channels, we usually store additional information in channels, or for example masks that must be applied in a process.
Our clients also send us standard files with extra channels, I'm not aware of their use, but I could try to ask.

As to what would happen to the thumbnailer, we usually resize all channels at the same time, and imaging applications only read color channels, so they don't change their behaviour.
For flatten, we still don't have any own process for this purpose, we have a PSD file loader that has some of the functionality, but in this case flatten is not available. Although, in the future we want to support Photoshop operations between channels, and have layers support.
For premultiply, and others... as far as I know, we only use color bands for processing, and alpha channels are just copied. But may be some special cases that I'm not aware of, I could check with my coworkers con Monday probably.

angelmixu:
Hi @jcupitt !
I just noticed this PR wasn't merged into master.

Would you need any further changes?

jcupitt:
Sorry @angelmixu I'll try to resolve this today.

jcupitt:
OK, merged to master! I had to rebase and edit it a little.

Thank you for working on this, @angelmixu.

angelmixu:
Thanks to you @jcupitt ! :)
I'll be working on improving the file formats (mainly tiff), I'll probably need some help/answers, so I'll post some issue in the next days/weeks.

Issue #1336, opened 2019-06-12
Enhancement: expose control over PNG IDAT chunk size


abarre:
Hi,

I try to consider the replacement of pngcrush by sharp. Sharp doesn't seem to offer a way to remove chunk for the PNG format, I used to use the following command for pngcrush :  

```pngcrush -rem alla -remove input.png output.png```

There is an extract of the [doc](http://pwet.fr/man/linux/commandes/pngcrush/) : 
```
-rem chunkname
Name of an ancillary chunk or optional PLTE to be removed. Be careful with this. Please don't use this feature to remove transparency, gamma, copyright, or other valuable information. To remove several different chunks, repeat: -rem tEXt -rem pHYs. Known chunks (those in the PNG spec or extensions document) can be named with all lower-case letters, so -rem bkgd is equivalent to -rem bKGD. But note: -rem text removes all forms of text chunks; Exact case is required to remove unknown chunks. To do surgery with a chain-saw, -rem alla removes all known ancillary chunks except for tRNS, and -rem allb removes all but tRNS and gAMA.
```

Since it's not provided by sharp, can you, please, clarify if there is a way to remove png chunk in libvips like it's done in pngcrush ?

**Comment by jcupitt** [comment]
Not directly, but you can remove metadata items by name. For example using `pyvips`:

```python
im = pyvips.Image.new_from_file('x.png')
im.remove('xmp-data')
im.write_to_file('y.png')
```

Will strip the XMP metadata (which can be many MB) from the image. Use `vipsheader -a x.png` to see all metadata on an image.

abarre:
Thank you @jcupitt, I will suggest this feature on sharp.

abarre:
Could you clarify if your reply concerns the following chunks :

- Background Color (bKGD)
- Timestamp (tIME)
- Latin-1 Text Annotations (tEXt, zTXt)
- International Text Annotations (iTXt)
- Histogram (hIST)
- Suggested Palette (sPLT)
- Significant Bits (sBIT)
- Physical Pixel Dimensions (pHYs)
- Physical Scale (sCAL)
- Image Offset (oFFs)
- Pixel Calibration (pCAL)
- Fractal Parameters (fRAc)
- GIF Conversion Info (gIFg, gIFx)
- GIF Plain Text (gIFt)
...

Ref : http://www.libpng.org/pub/png/book/chapter11.html#png.ch11.div.5
[...]

PR #1357, opened 2019-07-02
pngsave: expose control over output IDAT chunk size

lovell:
Using `png_set_compression_buffer_size` and the sample image discussed in #1336
 
```sh
$ vips copy fish.png fish-a.png[compression=9,filter=all,strip]
$ vips copy fish.png fish-b.png[compression=9,filter=all,strip,buffer_size=10000000]
```
This results in only one IDAT chunk:
```sh
$ pngchunks fish-a.png | grep "IDAT contains" | wc -l
61
$ pngchunks fish-b.png | grep "IDAT contains" | wc -l
1
```
...and therefore a slightly smaller file size:
```
495910 fish-a.png
495190 fish-b.png
```

### DIFF
--- a/libvips/foreign/pforeign.h
+++ b/libvips/foreign/pforeign.h
@@ -204,11 +204,13 @@ int vips__png_header_buffer( const void *buffer, size_t length, VipsImage *out )
 int vips__png_write( VipsImage *in, const char *filename, 
 	int compress, int interlace, const char *profile,
 	VipsForeignPngFilter filter, gboolean strip,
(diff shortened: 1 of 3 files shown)


jcupitt:
This is a tiny saving :( Do you think this is worthwhile? It seems a bit marginal to me.

lovell:
Yeah, the maintenance cost is likely to be higher than the benefit. These changes fell out of my investigation into #1336 so creating a PR was easy after that. I don't mind either way.

jcupitt:
OK, let's leave this for reference, but snoozed.

PR #1360, opened 2019-07-03
heifsave: expose compression option

lovell:
This implements an existing TODO in the heifsave code. I tested against the latest libheif master and avif branches and it appears to either work or return (expected) "unsupported" messages with the various values so this change will future-proof libvips against upstream libheif changes.

It also improves libheif error messaging and adds a couple of further suffixes that libheif can load.


jcupitt:
This looks great!

PR #1362, opened 2019-07-08
fix: make gif-loop consistent between GIF and WEBP

### LINKED ISSUE #1502: Make gif-loop consistent between GIF and WEBP
Just to bring attention to https://github.com/libvips/libvips/pull/1362 as it is a recently updated old PR. 🙂

deftomat:
We discovered that converting GIF into WEBP can unexpectedly change number of iterations.

The problem is that `gif-loop=2` in GIF means 3 iterations and just 2 iterations in WEBP.

So, this PR is trying to unify gif-loop for both formats.

Now, the default GIF value is 1 iteration and when `gifload` found *netscape extension block*, then it will change it to 0 (infinite loop) or to the real number of iterations. So, for example, when we found 2, we store it as 3 as viewer should iterate 3 times until it stops.

Now, it is possible to load GIF and immediately save it into WEBP without worrying that the number of iterations will reduce by 1.

I think that this change can be a breaking change for GIF but unfortunately, it is also a bug for WEBP. 

Maybe, we should use different name, for example: `loop`

closes #1502


jcupitt:
Hello @deftomat,

I hadn't realized that GIF loop == 1 actually meant play twice (ie. rewind once).

I think the `gif-loop` metadata item should keep the GIF meaning (otherwise the name is confusing and it could break old code), but I agree that we need a new `loop` tag with a single defined meaning that the imagemagick, libwebp and giflib loaders and savers all respect.

So how about:

* `gif-loop` is simply the value in a GIF file. We tag this as deprecated, but still set it in the loaders for compatibility.
* `loop` has webp meaning, ie. 0 == forever, 1 == play once, 2 == play twice, etc.

For reference, giflib: 0 == forever, 1 == play twice, 2 == play 3 times, etc. IM uses the same convention as webp.

What do you think?

jcupitt:
Handy for testing:

```python
#!/usr/bin/python3

import sys
import pyvips

im = pyvips.Image.new_from_file(sys.argv[1], n=-1, access="sequential")
im = im.copy()
im.set("gif-loop", int(sys.argv[3]))
im.write_to_file(sys.argv[2])
```

deftomat:
Hi. Yes, it sound good. I will update it.

deftomat:
@jcupitt I'm not exactly sure how we want to deprecate `gif-loop` for GIF. I understand that `gifload.c` will fill `loop` and `gif-loop` properties but what about `magicksave.c`? Which option should have the priority? User can forgot to update both options. Unfortunately, there is no `gif-loop` setter so I cannot update both `gif-loop` and `loop` when one of them change.

deftomat:
ping @jcupitt 🙂

jcupitt:
Hiya, sorry for not responding. I think the savers should all do this: 

1. If "loop" is set, use that, and don't look for "gif-loop". It has the webp meaning, so 0 == forever, 1 == play once, 2 == play twice etc.
2. Only if "loop" is not set, check "gif-loop". This has the gif meaning, so 0 == forever, 1 == play twice, 2 == play three times etc.

libvips has a deprecation system for header fields, so we can add "gif-loop" to that and it'll vanish from all interfaces, though it'll still work if you use it explicitly.

deftomat:
@jcupitt 
Sorry for keeping this PR open for so long.

I add `loop` field with the correct (webp) meaning to both webp and gif.

Also, I add `.editorconfig` as it could help contributors. Config will force supported editors to use tabs instead of spaces and to stops them from removing the trailing spaces. I was fighting with my editor as I had a massive amount of changes after each save 😞

jcupitt:
OK, let's merge. I can see a couple of tiny things, but I'll zap them in another patch. 

Thank you very much for working on this, @deftomat !

jcupitt:
I added some other file types to .editorconfig, adjusted some formatting, and added "gif-loop" to the list of deprecated (and hidden) metadata fields.

deftomat:
AWESOME, thanks 👍

Is there an automated way how to format files in this project? Because I always forgot to put spaces into IFs 😞

jcupitt:
I keep meaning to write an indent file, but I'm too lazy :(

PR #1366, opened 2019-07-15
Support std::vector within the image_get/set_array_int functions

kleisauke:
Can be useful, for example:
```c++
std::vector<int> delays = image.get_array_int("delay");
for (int i : delays) {
    std::cout << i << std::endl;
}

delays = {100, 200, 300, 400, 500};
image.set("delay", delays);
```

### DIFF
--- a/cplusplus/include/vips/VImage8.h
+++ b/cplusplus/include/vips/VImage8.h
@@ -358,6 +358,13 @@ public:
 	}
 
 	void 
+	set( const char *field, std::vector<int> value )
+	{
+		vips_image_set_array_int( this->get_image(), field, &value[0],
+			static_cast<int>( value.size() ) );
+	}
+
+	void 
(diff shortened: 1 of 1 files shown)


cplusplus/include/vips/VImage8.h
```diff
@@ -352,9 +352,10 @@ class VImage : VObject
 	}
 
 	void 
-	set( const char *field, int *value, int n )
```
jcupitt: How about supporting both? We don't need to remove the old setters -- overloading will pick the right one at compile-time.

kleisauke: Good idea! I'll change my PR.

kleisauke: OK, done!

jcupitt:
\o/

PR #1367, opened 2019-07-18
Add note about libheif

mcfedr:


jcupitt:
Thanks!

Now I look I think `install.md` in `gh-pages` needs quite a few other updates from `README.md` -- I'll have a go.

PR #1379, opened 2019-08-02
Add undefined sanitizer

omira-sch:
This adds the undefined sanitizer [UBSAN](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html) to the fuzzing build in travis. Currently, [ClusterFuzz](https://github.com/google/oss-fuzz/blob/master/docs/clusterfuzz.md) is fuzzing libvips with both ASAN and UBSAN sanitizers enabled by default. But in travis only ASAN was specified.

This PR also fix the declaration of FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION (see https://llvm.org/docs/LibFuzzer.html#fuzzer-friendly-build-mode). This must be a C macro instead of a env variable. This works in the oss-fuzz environemnt because they set CFLAGS correctly, but in travis we were missing this macro.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,11 +41,10 @@ matrix:
         - PYVIPS_VERSION=master
         - JPEG=/usr
         - JOBS=`nproc`
-        - CFLAGS="-fsanitize=address -fno-omit-frame-pointer -fopenmp"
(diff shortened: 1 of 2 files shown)


jcupitt:
Oh, sorry, I broke your PR by applying the other first. Shall I apply the UBSAN fix by hand?

omira-sch:
No, it's ok. In fact it's the oppostive. Merging #1381 first fixes this PR. I've rebased it and it's should pass now all the tests.

PR #1381, opened 2019-08-02
skip large images in the fuzzers not in the lib

omira-sch:
Hi. I have been tracking the project since the fuzzers are running in oss-fuzz.com to check if they are going well and I see you have already fixed several issues. Congrats.

You pushed a commit to skip large images using the macro FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION. This macro should be used to make code "friendly" to fuzzers, but only in very specific cases. For instance:
- to ignore failed checksums
- to make code deterministic (e.g. feed a random generator with the same seed value)

Any other issue with the fuzzers should be handled in the fuzzing functions themselves, rather than in the library. Otherwise, you will end up testing different execution paths when fuzzing than when using the library regularly, which is like testing two diverging versions of the same library.

This PR reverts your commit https://github.com/libvips/libvips/commit/0accdf858bbe32f07aeb6d79de860ad759b91f65 and move the checks to the fuzzers. It will also fix the build of https://github.com/libvips/libvips/pull/1379, which is failing now in travis: the test suite has some big images, and the library returns an error for big images because of the above commit.

I set the threshold for all the fuzzers to the same arbitrary value (1048576). It could make sense to review these values in the future analyzing the performance metrics of each fuzzers individually.

### DIFF
--- a/fuzz/jpegsave_buffer_fuzzer.cc
+++ b/fuzz/jpegsave_buffer_fuzzer.cc
@@ -11,10 +11,21 @@ extern "C" int
 LLVMFuzzerTestOneInput( const guint8 *data, size_t size )
 {
 	VipsImage *image;
-	size_t len;
 	void *buf;
+	size_t len, width, height, bands;
 
 	if( !(image = vips_image_new_from_buffer( data, size, "", NULL )) ) {
+		return( 0 );
+	}
+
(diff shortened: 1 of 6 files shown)


jcupitt:
Ah OK, yes that looks neater. Nice!

PR #1385, opened 2019-08-06
Add GIF optimization options to …

DarthSim:
This PR adds `optimize_gif_frames` and `optimize_gif_transparency` options to `vips_magicksave` and `vips_magicksave_buffer`.

If `optimize_gif_frames` is set, `OptimizePlusImageLayers` is applied to the ImageMagick image after rendering. `OptimizePlusImageLayers` crops each frame to the smallest size preserving the results of the GIF animation. It also adds or removes frames if it reduces the total number of pixels.

if `optimize_gif_transparency` is set, `OptimizeImageTransparency` is applied to the ImageMagick image. `OptimizeImageTransparency` takes a frame optimized GIF animation, and compares the overlayed pixels against the disposal image resulting from all the previous frames in the animation. Any pixel that does not change the disposal image (and thus does not affect the outcome of an overlay) is made transparent.

Example:

No optimizations allpied [124K]

animated_gif1_orig

`optimize_gif_frames` [102K]

animated_gif1_plus

`optimize_gif_frames` + `optimize_gif_transparency` [91K]

animated_gif1_trans

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -692,7 +692,7 @@ if test x"$magick6" = x"yes"; then
   # IM
   save_LIBS="$LIBS"
   LIBS="$LIBS $MAGICK_LIBS"
-  AC_CHECK_FUNCS([InheritException AcquireExceptionInfo SetImageProperty SetImageExtent AcquireImage GetVirtualPixels ResetImageProfileIterator ResetImageAttributeIterator ResetImagePropertyIterator MagickCoreGenesis SetImageOption BlobToStringInfo])
+  AC_CHECK_FUNCS([InheritException AcquireExceptionInfo SetImageProperty SetImageExtent AcquireImage GetVirtualPixels ResetImageProfileIterator ResetImageAttributeIterator ResetImagePropertyIterator MagickCoreGenesis SetImageOption BlobToStringInfo OptimizePlusImageLayers OptimizeImageTransparency])
   LIBS="$save_LIBS"
 fi
 
--- a/libvips/foreign/magick.c
+++ b/libvips/foreign/magick.c
@@ -205,6 +205,31 @@ magick_set_number_scenes( ImageInfo *image_info, int scene, int number_scenes )
 	image_info->scenes = strdup( page );
 }
 
+int
+magick_optimize_image_layers( Image **images, ExceptionInfo *exception )
+{
+	Image *tmp;
+
+	tmp = OptimizePlusImageLayers(*images, exception );
+
+	if ( exception->severity != UndefinedException )
+		return MagickFalse;
+
+	VIPS_FREEF( DestroyImageList, *images );
+
+	*images = tmp;
+
+	return MagickTrue;
+}
+
+int
+magick_optimize_image_transparency( const Image *images,
+	ExceptionInfo *exception )
+{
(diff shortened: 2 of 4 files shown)


jcupitt:
Hey @DarthSim this looks really nice!

I suppose my only comment would be: is it worthwhile to have two flags? How about just having one called `optimise_gif`? Would you ever want to do one of these and not the other?

DarthSim:
`optimize_gif_transparency` not always does a good job. Actually, it highly depends on whether an image has large still areas. If the image is fully animated or if it's badly dithered `optimize_gif_frames` + `optimize_gif_transparency` gives a worse result than `optimize_gif_frames` only. Here is an example:

No optimizations applied [104K]

animated_gif5_orig

`optimize_gif_frames` [100K]

animated_gif5

`optimize_gif_frames` + `optimize_gif_transparency` [105K]

animated_gif5_trans

There are resulting frames after all optimizations:

animated_gif5_trans_frames

As you see, only the last frame has small transparent parts that make compression worse.

If you don't like the idea of having two flags, I'd rather keep only frames optimization instead of combining the flags.

kleisauke:
The `OptimizePlusImageLayers` and `OptimizeImageTransparency` functions are not available within GraphicsMagick, so I think these functions needs to be detected at `configure` time and the options conditionally ignored if they are not present.

jcupitt:
Ah OK, thanks for the example @DarthSim. Sure, let's have two flags.

Good point about GraphicsMagick. I just tried building against GM and seems it's been broken for a while, annoyingly. Let's merge this PR, then fix GM build in another commit.

jcupitt:
I changed my mind -- I've fixed up GM build, could you rebase against master please @DarthSim?

I've cleaned up the configure tests for IM, so it should be simple to add tests for `OptimizePlusImageLayers` and `OptimizeImageTransparency`.

https://github.com/libvips/libvips/blob/master/configure.ac#L695

DarthSim:
@jcupitt Do you think ignoring those flags with GM would be enough, or we should throw a warning?

It looks like IM build is broken now. `HAVE_GETIMAGEMAGICK3` is undefined in my `config.h`, and `magick_ismagick` uses implementation with two args.

jcupitt:
I think silent ignore would be OK, but issue a warning if you like.

That's annoying about your build breaking -- git master builds cleanly for me with IM 6.9.10 and GM 1.4. What IM is it failing for?

DarthSim:
It fails with IM 7.0.8. `GetImageMagick` is wrapped with `if test x"$magick6" = x"yes"; then`, so it always false for IM7. The same problem with HAVE_CMYCOLORSPACE.

DarthSim:
I've hidden all dirty stuff inside `magick.c` and fixed GM compatibility.

jcupitt:
I've adjusted it again, and it builds cleanly and passes the test suite with im6, im7 and gm. Phew!

Sorry, I've broken your PR again :(

DarthSim:
No problem, fixed it :)

jcupitt:
That looks great!  Nice work.

DarthSim:
Awesome! Thanks!

PR #1387, opened 2019-08-06
Add oss-fuzz fixed testcases for regression testing

omira-sch:
I have downloaded oss-fuzz's testcases that are already fixed and copied them into the matching corpus directory. Only the interesting ones, ignoring OOM and timeouts. 

If ever happens we got a regression in the code, the build with the sanitizers will fail and Travis will alert us, because the corpus is tested as part of the "make check".

This PR also deletes the build artifact "libstandaloneengine.a" that I guess was pushed by mistake days ago, and removes fuzz directory from .gitignore. Any stale object file in fuzz is already ignored by the patterns *.o and *.a


jcupitt:
Nice, and I'll try to remember to add testcases in future.

PR #1397, opened 2019-08-19
add fuzzers for vips_smartcrop and vip_mosaic

omira-sch:
Two new fuzzers for high-level functions: vips_smartcrop and vip_mosaic. Smartcrop is interesting because is an optional operation for thumbnailing. Mosaic tests vips_merge and several subfunctions.

To fuzz vip_mosaic, it joins the original image with itself rotated 180 degrees and choose the join-coordinates and orientation randomly.

### DIFF
--- a/fuzz/Makefile.am
+++ b/fuzz/Makefile.am
@@ -6,7 +6,9 @@ FUZZPROGS = \
 	pngsave_buffer_fuzzer \
 	webpsave_buffer_fuzzer \
 	sharpen_fuzzer \
-	thumbnail_fuzzer
+	thumbnail_fuzzer \
+	smartcrop_fuzzer \
+	mosaic_fuzzer
 
 AM_DEFAULT_SOURCE_EXT = .cc
 
--- a/fuzz/mosaic_fuzzer.cc
+++ b/fuzz/mosaic_fuzzer.cc
@@ -1,0 +1,63 @@
+#include <vips/vips.h>
+
+struct mosaic_opt {
+	guint8 dir : 1;
+	guint16 xref;
+	guint16 yref;
+	guint16 xsec;
+	guint16 ysec;
+};
+
+extern "C" int
+LLVMFuzzerInitialize( int *argc, char ***argv )
+{
+	vips_concurrency_set( 1 );
+	return( 0 );
+}
+
+extern "C" int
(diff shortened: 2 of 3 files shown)


jcupitt:
LGTM!

I doubt if the mosaic funcs will ever see untrusted input, but why not. Smartcrop certainly will.

I mentioned you in a blog post, I hope that's OK. Do let me know if you'd like it changed, or a link added.

https://libvips.github.io/libvips/2019/08/18/libvips-in-oss-fuzz.html

omira-sch:
Nice blog post! Very clear and well explained. Just one minor suggestion, if you can refer me with my full name, like "Oscar Mira (@omira-sch)", that would be great.

I had in mind to open a feature-request to add more decoders to the fuzzing, especially those scientific formats that don't depend on ImageMagick. Just as a reminder that is something we can do in the mid-term.

jcupitt:
OK, I added your real name.

It would be good to fuzz things like PPM and CSV as well -- but they need a filesystem source.

valldrac:
Thanks. What do you mean by filesystem source?

jcupitt:
Only a few readers can work from a memory source -- many need a file.

There's been talk of adding our own simple IO layer to abstract this out, but no one's seen it though yet.

valldrac:
OK. The fuzzer can write the buffer to a temporary file and read back from there. Not the most efficient way of fuzzing, but it will do the job. 

I can write a PoC if you like... CSV or PPM?

jcupitt:
It's OK, I'll have a go, it's very simple. I thought perhaps oss-fuzz would disapprove of FS access.

valldrac:
Some fuzzers need to touch the FS and, because they run in a container environment, they don't interfere with each other.

To test readers, write the fuzzed buffer to a fixed path like /tmp/fuzzlibvips.ppm and read from there. Keep in mind that people building libvips locally will also execute the fuzzers during make test.

To test writers, better use /dev/null as output file (if the image is written sequentially).

jcupitt:
I had a stab:

https://github.com/libvips/libvips/commit/1de458556a27af735644623e0ee0dad6ea532e0b

and some corpus files to start it off:

https://github.com/libvips/libvips/commit/54ce3513fd7f3fabf6513739a6d11da9fe817b3f

omira-sch:
Looks good.

PR #1398, opened 2019-08-19
move fuzzing corpus into a single dir

omira-sch:
It moves the regression testcases (aka corpus) into a single directory named "common_fuzzer_corpus". I think we really don't need a directory for each fuzzer. It's a bit much. The filenames of the testcases already contains the name of the crashed fuzzer, so we don't lose that info. This way is more easy to manage and neat.

Because we are looking for crashes rather than correctness, there is nothing wrong on running all the fuzzers over all the testcases during "make check". It will take just a few seconds more in Travis.


jcupitt:
Sure, sounds reasonable!

PR #1417, opened 2019-09-01
configure.ac: drop --enable-cpp7 and --enable-pyvips8 options

bgilbert:
They're no longer connected to anything.


jcupitt:
Ooop, yes, left-over stuff. Thanks Benjamin!

PR #1430, opened 2019-09-17
Fix calculation of scaled SVG dimensions

DarthSim:
Currently, libvips incorrectly scales the width/height of SVG when `rsvg_handle_get_dimensions` doesn't return scaled ones. It floors scaled values while it should ceil them. As a result, the edge pixels can be lost.

Example:

We have the following SVG: https://jsfiddle.net/5crw3h70/1/

When I do `vipsthumbnail utkonos.svg --size 234x90`, it produces 234x59 image, while it should be 234x60. As you can see, the bottom pixels are cut:

img

Here is the result after my patch:

img

### DIFF
--- a/libvips/foreign/svgload.c
+++ b/libvips/foreign/svgload.c
@@ -273,8 +273,8 @@ vips_foreign_load_svg_parse( VipsForeignLoadSvg *svg, VipsImage *out )
 			 * cairo instead.
 			 */
 			svg->cairo_scale = scale;
-			width = width * scale;
-			height = height * scale;
+			width = VIPS_CEIL(width * scale);
+			height = VIPS_CEIL(height * scale);
 		} else {
 			/* SVG with width and height reports correctly scaled 
 			 * dimensions.


libvips/foreign/svgload.c
```diff
@@ -273,8 +273,8 @@ vips_foreign_load_svg_parse( VipsForeignLoadSvg *svg, VipsImage *out )
 			 * cairo instead.
 			 */
 			svg->cairo_scale = scale;
-			width = width * scale;
-			height = height * scale;
+			width = VIPS_CEIL(dimensions.em * scale);
+			height = VIPS_CEIL(dimensions.ex * scale);
```
jcupitt: I don't like using `dimensions.em` / `.ex` here -- it should be the same on both branches of the `if()`.

I this should be `width = VIPS_CEIL( dimensions.width * scale );`

jcupitt:
That's a great test image -- I've been looking for an SVG with fractional dimensions.

Nice catch, it's a shame this just missed 8.8.3 :(

libvips/foreign/svgload.c
DarthSim: Ohh, I should remember all the C stuff if I'm going to contribute libvips. I was pretty sure that `int * double == int` :)

Fixed

lovell:
If you didn't see it, there's some discussion about a slightly alternative solution at #1459 (warning: involves rounding).

lovell:
#1459 has now been merged into the master branch so this problem should be fixed in the future v8.9.0 release.

lovell:
Thank you @DarthSim for taking the time to submit a PR to address this!

PR #1433, opened 2019-09-20
Silence a couple of GCC compiler warnings

lovell:
I stumbled across these whilst investigating a unrelated compiler bug/feature on an ARM64 machine (I had enabled the "-Werror" switch).


jcupitt:
Nice!

PR #1443, opened 2019-10-12
ready to merge: Add stream class

jcupitt:
# Work in progress: experiment with true streaming for image load and save

Streams are a new IO class that image loaders can use instead of having separate paths for file and buffer IO. Streams can be attached to file descriptors (pipes, files, FIFOs, sockets etc.), files on disc, or areas of memory. They can be subclassed to add other IO types, such as (perhaps) node streams, https requests, etc.

Proposed API here:

https://github.com/libvips/libvips/blob/add-stream-object/libvips/include/vips/stream.h

## How it works

Streams give bytes to loaders, but cache bytes from the image header. This means you can connect a stream to a pipe (pipes do not allow seeking), then pass the stream to several loaders until one is able to parse it. That loader can then read out the image header and return any metadata.

Applications can examine metadata, decide how to open the image (perhaps setting shrink-on-load, or selecting a specific page), and then open the stream again. Because the header bytes have been cached, the loader will simply rewind the stream and reparse the cached bytes.

When the loader finally starts decoding pixels, it signals to the stream that it has finished with the header. At this point the stream will discard the header cache and simply pass buffers from `read()` directly to the decode library as they arrive. 

Streams offer four basic styles of IO to loaders:

- Unbuffered `read()`. Loaders like jpeg and png use this -- the file bytes are returned via a series of calls using the same semantics as the standard *nix `read` syscall.
- Buffered `read()`. Loaders like Radiance and csv use this. The stream has a set of functions to return the next line of text from the file, or to read the next N bytes. Behind the scenes, the stream keeps a largish (4096 at the moment) buffer of bytes to prevent many small `read()` calls.
- `read()` and `seek()`. Loaders like tiff need this. Streams emulate seek on pipes by reading bytes into a cache on demand.
- `mmap()`. Loaders like webp need this. All the data in the stream is read into memory somehow and presented to the loader as a single large blob.

Streams have one other interesting feature: they support minimisation. At the end of a section of computation (perhaps parsing a header, or extracting a chunk of pixels), they can be minimised. Any associated file resources are closed, and the current read point is recorded. If the stream is used again, it will be automatically reopened and the read point restored.

### Supported loaders

jpg, png, webp, hdr (Radiance) and tif (though only load, not save) all now work with streams. More of the loaders and savers will be converted.

tif save is tricky since pyramid write needs to make a set of temporary streams for the layers. This will need some thought.

### Process from `stdin`

```
cat k2.jpg | vips invert stdin x.jpg
```

The magic filename `"stdin"` opens a stream attached to file descriptor 0 (`stdin`), does `vips_image_new_from_stream()`, and passes that image into the operation. You can add load options, eg.:

```
cat k2.jpg | vips invert stdin[shrink=2] x.jpg
```

Will do a x2 shrink-on-load.

### Process to `stdout`

Give a filename with no name before the suffix to open a stream to `stdout` and write in that format. For example:

``` 
vips invert k2.jpg .jpg > x.jpg
```

You can chain operations with pipes:

```
vips invert k2.jpg .jpg | vips invert stdin x.jpg
```

### Streams as operation arguments

Operations which take streams directly as arguments use this syntax:

```
cat k2.jpg | vips thumbnail_stream [descriptor=0] x.jpg 200
```

## C API

```C
VipsStreami *streami = vips_streami_new_from_descriptor( 0 );
VipsStreamo *streamo = vips_streamo_new_to_descriptor( 1 );
VipsImage *a = vips_image_new_from_stream( streami, "shrink", 2, NULL );
VipsImage *b;
vips_invert( a, &b, NULL );
vips_image_write_to_stream( b, streamo );
g_object_unref( a );
g_object_unref( b );
g_object_unref( streami );
g_object_unref( streamo );
```

You can hook streams up to memory areas and files as well. You can subclass `VipsStreami`, override the `read()` (and optionally `seek()`) methods, and add your own IO processing.

## Python API

The `add-stream` branch of pyvips allows this:

```python
x = pyvips.Streami.new_from_descriptor(4132)
y = pyvips.Image.new_from_stream(x, "")
if y.width > 1000:
    # big image .. shrink on load
    # opening x again will automatically rewind it
    y = pyvips.Image.new_from_stream(x, "", shrink=2)
y = y.invert()
z = pyvips.Streamo.new_to_descriptor(2487)
y.write_to_stream(z)
```

Plus methods for buffer and file streams.

## To do before we merge to master

- [x] add support for the `::minimise` signal, and re-enable jpg `test_descriptors`
- [x] think again about the ugly `stdin` magic filename
- [x] try some other loaders, esp. TIFF (needs `seek`) and GIF (needs `mmap`).
- [x] add support to pyvips
- [x] add tests to test suite
- [x] fix filename encoding (we assume utf-8 at the moment)
- [x] add `mmap`able descriptors
- [x] add `seek`able descriptors
- [x] can we test for `mmap`able and `seek`able?
- [x] do we need eof?
- [x] need a `close()` vfunc? output is a bit ugly, and node streams might need it
- [x] make a streami subclass with signals for `read()` / `seek()` / `open()` / `close()` etc. to make it easy to extend behaviour in Python 
- [x] implement load / save to URI in Python to test extensibility
- [x] `vips_image_new_from_file()` / `_buffer()` / etc. should try with a stream first before falling back to the old behaviour
- [x] rename `VipsStreamiw` and `VipsStreamib`


lovell:
This looks very nice, thanks John.

I'll need to spend some time to see what sharp could do to take advantage of this. From what I can tell, sharp could inspect the input and/or output Node.js `Stream` instances for a possible `fd` property and use the new/proposed `vips_stream_input_new_from_descriptor` and/or `vips_stream_output_new_from_descriptor` when present.

jcupitt:
Hiya, yes, you can also subclass `VipsStreamInput`, override the `read()` method and add any other processing. 

I guess node would probably need a `close()` method too? There's `finalize()`, but that can happen much later.

libvips/foreign/foreign.c
```diff
@@ -1791,6 +1865,58 @@ vips_foreign_save( VipsImage *in, const char *name, ... )
...
+/* Can thsi class write this filetype to a stream?
+ */
+static void *
+vips_foreign_find_save_stream_sub( VipsForeignSaveClass *save_class, 
+	const char *suffix )
+{
+	VipsObjectClass *object_class = VIPS_OBJECT_CLASS( save_class );
+	VipsForeignClass *class = VIPS_FOREIGN_CLASS( save_class );
+
+	printf( "testing %s for %s\n", object_class->nickname, suffix );
```
kleisauke: Redundant `printf` statement.

libvips/foreign/foreign.c
```diff
@@ -1791,6 +1865,58 @@ vips_foreign_save( VipsImage *in, const char *name, ... )
 	return( result );
 }
 
+/* Can thsi class write this filetype to a stream?
```
kleisauke: Typo `thsi` -> `this`.

libvips/foreign/jpegload.c
```diff
@@ -34,9 +34,9 @@
  */
 
 /*
+ */
```
kleisauke: `DEBUG_VERBOSE` and `DEBUG` are now defined.

libvips/iofuncs/image.c
```diff
@@ -2167,6 +2167,48 @@ vips_image_new_from_buffer( const void *buf, size_t len,
...
+ * @...: %NULL-terminated list of optional named arguments
+ *
+ * Loads an image from the formatted stream @input, 
+ * loader recommended by vips_foreign_find_load_stream(). 
+ *
+ * Load options may be given in @option_string as "[name=value,...]" or given as
+ * a NULL-terminated list of name-value pairs at the end of the arguments.
+ * Options given in the function call override options given in the string. 
+ *
+ * See also: vips_image_write_to_buffer().
```
kleisauke: `See also: vips_image_write_to_buffer()` -> `See also: vips_image_write_to_stream()`.

libvips/iofuncs/stream.c
```diff
@@ -0,0 +1,948 @@
...
+
+		input->decode = TRUE;
+		VIPS_FREEF( g_byte_array_unref, input->header_bytes ); 
+		VIPS_FREEF( g_byte_array_unref, input->sniff ); 
+	}
+}
+
+/**
+ * vips_stream_input_sniff: 
+ * @bytes: number of bytes to sniff
```
kleisauke: `@bytes` -> `@length`.

libvips/iofuncs/stream.c
```diff
@@ -0,0 +1,948 @@
...
+ *
+ * Return a pointer to the first few bytes of the file.
+ */
+unsigned char *
+vips_stream_input_sniff( VipsStreamInput *input, size_t length )
```
kleisauke: Missing doc comment for `input`.

libvips/iofuncs/stream.c
```diff
@@ -0,0 +1,948 @@
...
+
+	return( stream ); 
+}
+
+/**
+ * vips_stream_output_new_memory:
+ *
+ * Optional args:
+ *
+ * @blob: #VipsBlob, memory area containing output
```
kleisauke: There are no optional args for this function.

kleisauke:
This is awesome, thanks John. I started to implement this in NetVips:
https://github.com/kleisauke/net-vips/tree/true-streaming

I'm able to [load and save from a C# stream](https://github.com/kleisauke/net-vips/blob/true-streaming/samples/NetVips.Samples/Samples/Stream.cs#L15-L24) and [thumbnailing a stream](https://github.com/kleisauke/net-vips/blob/true-streaming/samples/NetVips.Samples/Samples/ThumbnailStream.cs#L15-L24) seems to work too! 🎉 

Regarding subclassing `VipsStreamInput`  and `VipsStreamOutput` wouldn't it not be easier to add these functions?:
```
vips_stream_input_set_custom_reader
vips_stream_input_set_custom_rewinder
vips_stream_output_set_custom_writer
```

See the relevant patch:
https://github.com/libvips/build-win64-mxe/blob/true-streaming/8.9/patches/vips-8-fixes.patch

It is then much easier for language bindings to set up their custom read/write/rewind implementations.

jcupitt:
Hi Kleis, yes, it should make loaders and savers much simpler, hopefully. 

Plus if we can work S3 bucket directly to S3 bucket, we should see a useful drop in latency. It would be very interesting if we could measure that improvement somehow.

Yes, you could set custom functions for read/write/rewind, but with subclassing you can also hook into `_build` and `_finalize`, and perhaps add more parameters. I'm planning to add vfuncs for seek and mmap as well.

I suppose this is something we should experiment with once this is all fleshed out.

jcupitt:
I just read your NetVips changes. Very nice!

libvips/iofuncs/stream.c
```diff
@@ -382,13 +385,6 @@ vips_stream_input_class_init( VipsStreamInputClass *class )
 		G_STRUCT_OFFSET( VipsStreamInput, blob ),
 		VIPS_TYPE_BLOB );
 
-	VIPS_ARG_BOOL( class, "seekable", 4,
```
kleisauke: Was this intentionally removed? It will break:
```c
VipsStreamInput *input = vips_stream_input_new_from_options( "seekable=1" );
```
```
unable to make VipsStreamInput
stream_input: no property named `seekable'
```

jcupitt: Hiya, yes, it tests descriptors for seekable automatically now, so you don't need to tell it.

kleisauke: I see. I'll try the subclass approach in C#.

Do you know how I can do that easily? For e.g. I don't have access to macros, so `G_DEFINE_TYPE` will not work. I guess pyvips has the same problem, a `vips_stream_build` function would be useful to have for language bindings.

jcupitt: Oh huh you're right, you really need C/C++ to subclass. You can sort-of do it in Python, but it's a pretty awful hack.

Let's add a set of callbacks as per your patch ^^ up there.

kleisauke: Of course we have to focus on the other things first, this is a problem for later.

jcupitt: OK, added another point to the checklist!

libvips/include/vips/stream.h
```diff
@@ -0,0 +1,247 @@
...
+	/* TRUE is this descriptor supports lseek(). If not, then we save data
+	 * read during header phase in a buffer.
+	 */
+	gboolean seekable;
+
+	/*< private >*/
+
+	/* The current read point.
+	 */
+	off_t read_position;
```
kleisauke: Perhaps this should be `gint64`? This ensures that we can handle large streams on 32-bit machines.

jcupitt: Good idea!

jcupitt: ... do you know if lseek() returns a 32-bit value on 32-bit Windows? We might need to make an lseek() wrapper that calls whatever their 64-bit API is, just as we had to for vips_file_length().

libvips/iofuncs/stream.c
```diff
@@ -0,0 +1,1067 @@
...
+		return( input->baseaddr );
+	}
+
+	/* Have to read() the whole thing. header_bytes will keep a copy of
+	 * the file.
+	 */
+	VIPS_DEBUG_MSG( "    read() of entire source\n" );
+	if( vips_stream_input_rewind( input ) )
+		return( NULL );
+	while( vips_stream_input_read( input, buffer, 4096 ) > 0 )
```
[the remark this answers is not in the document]
jcupitt: True, I suppose I was thinking that seekable fd's must all be mappable. That won't always be true for subclasses which might implement the seek() method but not bother with map().

libvips/include/vips/stream.h
kleisauke: I think `lseeki64` is only able to do 64-bit seeks. I noticed there is a `vips__seek` function:
https://github.com/libvips/libvips/blob/be7c1404c321917c602c1fb3f54da009a92bef0d/libvips/iofuncs/util.c#L1031

Perhaps we could use that?

jcupitt: Oh duh, I totally forgot about that. Yes, I'll fix it up and use it. Thanks!

libvips/iofuncs/stream.c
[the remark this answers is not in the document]
kleisauke: Perhaps we could just move it to a memory stream within `vips_stream_input_map`? See for example:
https://github.com/libvips/build-win64-mxe/blob/7e3cdd37ba6b36895e0bb8d47ac32893142425af/8.9/patches/vips-8-fixes.patch#L214-L272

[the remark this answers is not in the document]
jcupitt: It's getting too complex. I'm going to refactor it a bit I think, argh.

kleisauke:
I had some spare time left to implement this in the C++ binding as well. See: 
https://github.com/libvips/libvips/compare/add-stream-object...kleisauke:stream-cplusplus

jcupitt:
Sweet! I'm just reworking it a bit, then I'll pull that branch in.

kleisauke:
I just rebased and squashed the commits within the [stream-cpluscplus](https://github.com/kleisauke/libvips/commits/stream-cplusplus) branch.

You might also be interested in this commit: https://github.com/kleisauke/libvips/commit/4bf2d4bef9ca1fcea679fe45c5fad2cda8328ff1.

jcupitt:
Sorry, I keep changing stuff and breaking everything. It should stabilize soon.

kleisauke:
Let me know if it's stabilized! Perhaps you are also interested in these commits:
https://github.com/kleisauke/libvips/commit/c6ef5a469f543bd44016d43f3977ca5a864b4157
https://github.com/kleisauke/libvips/commit/048939b5e342c836dc48a296c991b07a82e90beb
https://github.com/kleisauke/libvips/commit/289d1721a5ea1dd14d9b562d0c0062665d9ee5a7
and optionally:
https://github.com/kleisauke/libvips/commit/370784e8367a9cb738d1ab4e014090e0966d401d

jcupitt:
Thanks for those patches Kleis! I've pasted them in.

kleisauke:
No problem, thanks for including it! I only have a few small patches left to make the subclass / custom stream approach work properly:
https://github.com/kleisauke/libvips/commit/09fc0d63ffed95ed164f7ca92c16c4838f1f802b
https://github.com/kleisauke/libvips/commit/767a2202a0a2763f84095d5ef0c4acde4921e53a
https://github.com/kleisauke/libvips/commit/93a9c5f82611aa2cfd96f16a811c744f7c4647c4

jcupitt:
OK, I included your patches Kleis, nice!

I think it's all working now and I promise not to touch it again for a while :)

kleisauke:
Thanks! I noticed that `vips_radload_stream` and `vips_radload_buffer` is missing, was this intentional? I had a quick go: https://github.com/kleisauke/libvips/commit/bc99b26b4e0326267ffea3b34d5e4a67c7866c1b but it doesn't seem to work for non-seekable streams.

By the way, 
I discovered (a bit late) that there is [`GIOStream`]( https://developer.gnome.org/gio/stable/GIOStream.html) base class available which includes the [`GInputStream`](https://developer.gnome.org/gio/stable/GInputStream.html) and [`GOutputStream`](https://developer.gnome.org/gio/stable/GOutputStream.html) subclasses. Perhaps we could subclass `VipsStreami` and `VipsStreamo` to it? Support for streams in the SVG loader is then easier to implement (see for e.g. [`rsvg_handle_new_from_stream_sync`](https://developer.gnome.org/rsvg/stable/rsvg-Using-RSVG-with-GIO.html#rsvg-handle-new-from-stream-sync)).

jcupitt:
I think I just hadn't got around to the other radload operations. It was very tied to stdio, so the old radload just did to and from files and that's all I replaced. You're right, we should expose at least `radload_stream`.

I'll merge your very nice patch! You're right about non-seekable streams and streamib, I'll have a look.

I looked at `GIOStream`, but it's SO complicated. I was hoping to make something smaller and simpler for libvips, since our needs are pretty modest, except for the header cache logic.

Maybe we can wrap a `GInputStream` around a `VipsStreami` somehow? I've not looked.

kleisauke:
> You're right about non-seekable streams and streamib, I'll have a look.

Commit https://github.com/libvips/libvips/pull/1443/commits/ce4729c92b169ce48ddab4c670e98381cb2c05bd fixes this, thanks!

> I looked at `GIOStream`, but it's SO complicated.

Indeed, `GIOStream` is a complicated class. The asynchronous methods doesn't make it any easier either.

> Maybe we can wrap a `GInputStream` around a `VipsStreami` somehow?

That's an interesting approach. It seems that this is possible: https://github.com/kleisauke/libvips/commit/609fdb4fb6f22346c2ca8dc246d8e87dfe315b07. I'm able to convert a pipeable SVG stream to PNG with that commit (should work with other streams as well, but I haven't tested this yet). I also need to check the minimise logic within that commit.

jcupitt:
That is very cool Kleis! I like the way they've set up the seekable interface, it's actually quite convenient, at least from C.

I must rename `VipsStreamib`. It's not a `VipsStreami` subclass any more so the name is wrong.

kleisauke:
I just tested `svgload_stream` with other streams as well, and it seems to work. The minimise logic also seems to works properly, so I added this loader to the descriptors test with https://github.com/kleisauke/libvips/commit/41c08b0ffb183ee7b031c34f5eebc6e2adb6d974.

> I must rename `VipsStreamib`. It's not a `VipsStreami` subclass any more so the name is wrong.

I also don't like my naming of the `VipsStreamiw` (the w means wrapper) class, do you have suggestions for a better name? I guess that the whole class can be conditionally excluded, since it's only needed for librsvg (although it may be useful for future loaders that uses the `GInputStream` class).

I'll have a look at the `libnsgif` loader next, perhaps it can also support streams. By the way, are there any loaders (or savers) that we already know can't be converted?

jcupitt:
Hey that's great, I'll merge it. Is your C++ work ready to merge as well?

Does glib have a convention for naming `GInputSteam` subclasses? Perhaps `VipsGInputStream`?

libnsgif needs the whole source file available at once, it can't stream images (as far as I know), so it's not a great fit. It's rather like webp.

giflib isn't much better: you can only find out the number of frames by scanning the entire image, and you have to discover the number of frames as part of reading the header.

csv is similar: you must scan the whole thing to get the vertical height.

tiffsave won't work well: it needs to make temp files (or temp memory objects) for pyramid layers.

dzsave uses libgsf for output, so I supposed we'd need another adapter to link libgsf to VipsStreamo.

I can't remember what libheif does, that might be an interesting one to look at.

jcupitt:
My feeling is that we should probably just rework one example of each general type of load/save operation to check that we have enough flexibility in VipsStream, then merge to master. We can convert the rest later.

kleisauke:
> Hey that's great, I'll merge it. Is your C++ work ready to merge as well?

Thanks! The [stream-cpluscplus](https://github.com/kleisauke/libvips/tree/stream-cplusplus) is ready to merge as well, I think. I just rebased it.

> Does glib have a convention for naming `GInputSteam` subclasses? Perhaps `VipsGInputStream`?

`VipsGInputStream` sounds OK. `VipsInputStream` should be fine too, but maybe that'll lead to confusion with `VipsStreami`.  According to [this](https://developer.gnome.org/SubclassGObject/#Naming_and_conventions) it should be `NamespaceClass`.

> libnsgif needs the whole source file available at once, it can't stream images (as far as I know), so it's not a great fit. It's rather like webp.

I see, it's a pity that some loaders needs a single large blob as input.

> I can't remember what libheif does, that might be an interesting one to look at.

I had a quick look. Looks like they're supporting streaming:
https://github.com/strukturag/libheif/blob/f3e58e21a87be3704c5c4837984be7eb235717d5/libheif/heif.h#L395-L399
https://github.com/strukturag/libheif/blob/f3e58e21a87be3704c5c4837984be7eb235717d5/libheif/heif.h#L347-L371

> My feeling is that we should probably just rework one example of each general type of load/save operation to check that we have enough flexibility in VipsStream, then merge to master. We can convert the rest later.

Indeed, the rest can be converted later. The VipsStream(i/o) classes seems pretty stable already. An advantage of merging into master is that we can see if OSS-Fuzz discovers anything.

libvips/foreign/svgload.c
```diff
@@ -473,22 +473,12 @@ G_DEFINE_TYPE( VipsForeignLoadSvgStream, vips_foreign_load_svg_stream,
 gboolean
 vips_foreign_load_svg_stream_is_a( VipsStreami *input )
 {
-	ssize_t n;
+	unsigned char *data;
 
-	if( vips_streami_rewind( input ) )
+	if( !(data = vips_streami_sniff( input, SVG_HEADER_SIZE )) )
```
kleisauke: This fails if the SVG image is smaller than `SVG_HEADER_SIZE` (1000). Since `vips_streami_sniff` will sniff exactly the amount of needed bytes. See for example:
```bash
$ wget https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/intertwingly.svg
$ stat --printf="%s" intertwingly.svg
962
$ cat intertwingly.svg | vips invert stdin x.jpg
VipsForeignLoad: stream is not in a known format
$ printf '<?xml version="1.0" encoding="UTF-8" standalone="no"?>\n' > x.svg; cat intertwingly.svg >> x.svg;
$ stat --printf="%s" x.svg
1017
$ cat x.svg | vips invert stdin x.jpg
$ vipsheader x.jpg
x.jpg: 100x100 uchar, 3 bands, srgb, jpegload
```

jcupitt: Ah I wondered what was going on there. Let's change sniff then to return the number of bytes read.

cplusplus/include/vips/VInterpolate8.h
```diff
@@ -49,7 +43,7 @@ class VInterpolate : VObject
 	}
 
 	static 
-	VInterpolate new_from_name( const char *name, VOption *options = 0 );
+	VInterpolate new_from_name( const char *name );
```
kleisauke: Should I add a compat stub for `VInterpolate::new_from_name` preserving the `VOption *` parameter? I think this change isn't ABI compatible:
https://kleisauke.nl/compat_reports/libvips-cpp/8.8.3_to_8.9.0/compat_report.html

The `libvips.so` seems to be fine:
https://kleisauke.nl/compat_reports/libvips/8.8.3_to_8.9.0/compat_report.html

(only internal API symbols were removed)

jcupitt: Yes, I was wondering about that.

Although the options aren't used right now, `VipsInterpolate` was originally designed to support parameterised interpolators and they might come back. If they do, we'd need to re-add the options param and it would break ABI again.

I'd be inclined to revert this change, unless there is a good reason.

libvips/foreign/svgload.c
kleisauke: Fixed with https://github.com/libvips/libvips/pull/1443/commits/998b5eea4593d92adaf2141854d6a7a55e453f43, thanks!

cplusplus/include/vips/VInterpolate8.h
kleisauke: Ah, I didn't know that `VipsInterpolate` was originally designed to support parameterised interpolators. Let's revert this change!

cplusplus/VImage.cpp
```diff
@@ -169,7 +169,7 @@ VOption::set( const char *name, const char *value )
 
 // input image
 VOption *
-VOption::set( const char *name, VImage value )
+VOption::set( const char *name, const VImage &value )
```
jcupitt: Sorry to return to this old change.

I guess this is for efficiency, is that right? By design, `VInterpolate` is just a pointer (`sizeof(VImage) == 8` etc.) so I don't think there's any saving for switching to call-by-reference, and it adds complexity.
 
I'm inclined to revert this, unless there is a good reason for call by reference.

cplusplus/include/vips/VInterpolate8.h
jcupitt: OK, reverted with https://github.com/libvips/libvips/commit/92521c0dc6ac5d4078107e35f5ea16c96ac2cbf8

cplusplus/VImage.cpp
kleisauke: It's intended for efficiency. Indeed it makes no sense here and I'm also not sure whether it's ABI-compatible.

Feel free to revert this. Perhaps we should also pass-by-value within the `new_from_stream` and `write_to_stream` functions?:
https://github.com/libvips/libvips/pull/1443/files#diff-812eb578264cc528d97b8955370e75aaR517
https://github.com/libvips/libvips/pull/1443/files#diff-812eb578264cc528d97b8955370e75aaR573

jcupitt: Hiya, yes, I flipped them all over to call-by-value in https://github.com/libvips/libvips/pull/1443/commits/123c52be7041f8d1ea9e937643b3a0df317b02ea

jcupitt: Yes, it would break ABI: `type&` parameters have extra code at the caller.

Though C++ ABI breaks are very hard to avoid, so it  probably doesn't matter.

cplusplus/include/vips/VImage8.h
```diff
@@ -222,10 +222,10 @@ class VOption
 	VOption *set( const char *name, int value );
 	VOption *set( const char *name, double value );
 	VOption *set( const char *name, const char *value );
-	VOption *set( const char *name, const VImage &value );
-	VOption *set( const char *name, const VInterpolate &value ); 
-	VOption *set( const char *name, const VStreamI &value );
-	VOption *set( const char *name, const VStreamO &value );
+	VOption *set( const char *name, const VImage value );
```
kleisauke: `const` could be removed, since it doesn't affect the signature of a function. See:
https://clang.llvm.org/extra/clang-tidy/checks/readability-avoid-const-params-in-decls.html

(if you want, I can make a PR to fix all `clang-tidy` linter errors)

cplusplus/VImage.cpp
kleisauke: Thanks! fwiw, the reason I switched them to a const reference was due to this check by `clang-tidy`:
https://clang.llvm.org/extra/clang-tidy/checks/performance-unnecessary-value-param.html

Somehow it detects that `VImage` is expensive to copy.

jcupitt: Huh strange. There is a copy constructor that is triggered to adjust the ref counts, perhaps that's enough to make the warning happen?

cplusplus/include/vips/VImage8.h
jcupitt: VImage etc. are really smart pointers, so the const is suppsed to suggest to users that the implicit reference is const. Perhaps it's not helpful :(

kleisauke: Top-level `const` is very useful within the implementation (i.e., code in the body of the function) and the compiler will enforce it, however it is ignored from the interface viewpoint. For example:
```cpp
VOption *set( const char *name, const VImage value );
```
and
```cpp
VOption *set( const char *name, VImage value );
```
declares the same function.

(https://stackoverflow.com/q/17208570/10952119 might be relevant here)

cplusplus/VImage.cpp
kleisauke: It seems that when I remove the VObject destructor (`~VObject()`) and copy constructor
(`VObject( const VObject &a ) : vobject( a.vobject )`) the warnings disappear, so I think you're right.

cplusplus/include/vips/VImage8.h
jcupitt: Ye exactly, call-by-value is always const. My idea was to add the const decorators for human readers to try to make it plain that the underlying image was not being modified.

kleisauke: Ah I see, that makes sense. As long as there are no compiler warnings, I'm fine with it.

fwiw, I've looked at the rest of the warnings and think they're safe to ignore. I only noticed that we could replace a `memset` and initialize a GValue directly within the member initialization list:
```diff
--- cplusplus/include/vips/VImage8.h
+++ cplusplus/include/vips/VImage8.h
@@ -197,10 +197,9 @@
 		}; 
 
 		Pair( const char *name ) : 
-			name( name ), input( false ), vimage( 0 )
+			name( name ), input( false ), vimage( 0 ), 
+			value( G_VALUE_INIT )
 		{
-			// argh = {0} won't work wil vanilla C++
-			memset( &value, 0, sizeof( GValue ) ); 
 		}
 
 		~Pair()
```

Unfortunately the `G_VALUE_INIT` macro is only available since GLib 2.30 :(

jcupitt:
@kleisauke I moved VipsStreamiw inside svgload, since that's the only place it's used, for now anyway. We can move it out and add to the public API if we need it elsewhere.

I changed the name to `VipsGInputStream`, as we discussed.

I was puzzled by this section:

https://github.com/libvips/libvips/blob/add-stream-object/libvips/foreign/svgload.c#L281

Why not just call `vips_streami_seek(streami, count, SEEK_CUR);` to skip forward? Is it to avoid errors for seeking past the end of the file?

We should find a more efficient way to do this: calling `vips_streami_seek( streami, 0, SEEK_END );` will force the whole of a pipe into memory, and that'll break streaming.

jcupitt:
I swapped it for `vips_streami_seek(streami, count, SEEK_CUR);` and it seems to work. 

https://github.com/libvips/libvips/commit/c7f622d6466dcd3f9fc01150b035196da6f83943

kleisauke:
Ah, that section is a bit weird. It's inspired from this:
https://github.com/GNOME/glib/blob/d3768c89ae4d86e6113104eb9da8e89b07de3c30/gio/glocalfileinputstream.c#L192-L231

It's indeed to avoid out-of-bounds seeks. We could also guard against that with (if it's memory stream):
```c
count = VIPS_MIN( count, streami->length - streami->read_position );
```

Although it looks like `vips_g_input_stream_skip` is not called by librsvg, so maybe it doesn't matter.

I guess we could also move the gio header to svgload, now that `VipsGInputStream` is there?:
https://github.com/libvips/libvips/blob/73985463fdee3ac9b14b3dc857b5240e42480048/libvips/include/vips/vips.h#L91

We may still need to check whether our minimum version for glib needs to be bumped.

jcupitt:
I added a test for gio and moved the include.

jcupitt:
I added `VipsStreamiu` (u for user I guess) -- a `VipsStreami` subclass with action signals you can implement in order to attach new input sources.

Here's a demo program:

```C
/* compile with
 *
 * gcc -g -Wall stream.c `pkg-config vips --cflags --libs`
 */

#include <vips/vips.h>

typedef struct _MyFile {
	const char *filename; 
	const unsigned char *contents; 
	size_t length;
	size_t read_position;
} MyFile;

static gint64
read_cb( VipsStreamiu *streamiu, void *buffer, gint64 length, MyFile *file )
{
	gint64 bytes_read = 
		VIPS_MIN( length, file->length - file->read_position );

	printf( "read_cb: buffer = 0x%p, length = %zd\n", buffer, length );

	memcpy( buffer, file->contents + file->read_position, bytes_read );
	file->read_position += bytes_read;

	return( bytes_read );
}

static gint64
seek_cb( VipsStreamiu *streamiu, gint64 offset, int whence, MyFile *file )
{
	gint64 new_pos;

	printf( "seek_cb: offset = %zd, whence = %d\n", offset, whence );

	switch( whence ) {
	case SEEK_SET:
		new_pos = offset;
		break;

	case SEEK_CUR:
		new_pos = file->read_position + offset;
		break;

	case SEEK_END:
		new_pos = file->length + offset;
		break;

	default:
		vips_error( "demo", "%s", "bad 'whence'" );
		return( -1 );
	}

	file->read_position = VIPS_CLIP( 0, new_pos, file->length );

	return( file->read_position );
}

int
main(int argc, char **argv)
{
	MyFile file;
	VipsStreamiu *streamiu;
	VipsImage *image;

	if( VIPS_INIT( NULL ) )
		return( -1 );

	file.filename = argv[1];
	file.contents = NULL;
	file.length = 0;
	file.read_position = 0;

	if( !g_file_get_contents( file.filename, 
		(char **) &file.contents, &file.length, NULL ) ) 
		vips_error_exit( "unable to load %s", file.filename );

	streamiu = vips_streamiu_new();
	g_signal_connect( streamiu, "seek", G_CALLBACK( seek_cb ), &file );
	g_signal_connect( streamiu, "read", G_CALLBACK( read_cb ), &file );

	if( !(image = vips_image_new_from_stream( VIPS_STREAMI( streamiu ), "",
		"access", VIPS_ACCESS_SEQUENTIAL,
		NULL )) ) 
		vips_error_exit( NULL );

	if( vips_image_write_to_file( image, argv[2], NULL ) )
		vips_error_exit( NULL );

	VIPS_UNREF( streamiu );
	VIPS_UNREF( image );

	return( 0 );
}
```

So it should be easy to hook up to Python, since we already have `signal_connect`.

I do one for output as well.

jcupitt:
And an output class you can override. 

This sample program uses both and seems to work!

```C
/* compile with
 *
 * gcc -g -Wall stream.c `pkg-config vips --cflags --libs`
 */

#include <sys/types.h>
#include <sys/stat.h>
#include <fcntl.h>
#include <unistd.h>
#include <vips/vips.h>

typedef struct _MyInput {
	const char *filename; 
	const unsigned char *contents; 
	size_t length;
	size_t read_position;
} MyInput;

typedef struct _MyOutput {
	const char *filename; 
	int fd;
} MyOutput;

static gint64
read_cb( VipsStreamiu *streamiu, 
	void *buffer, gint64 length, MyInput *my_input )
{
	gint64 bytes_read = VIPS_MIN( length, 
		my_input->length - my_input->read_position );

	printf( "read_cb: buffer = 0x%p, length = %zd\n", buffer, length );

	memcpy( buffer, 
		my_input->contents + my_input->read_position, bytes_read );
	my_input->read_position += bytes_read;

	return( bytes_read );
}

static gint64
seek_cb( VipsStreamiu *streamiu, 
	gint64 offset, int whence, MyInput *my_input )
{
	gint64 new_pos;

	printf( "seek_cb: offset = %zd, whence = %d\n", offset, whence );

	switch( whence ) {
	case SEEK_SET:
		new_pos = offset;
		break;

	case SEEK_CUR:
		new_pos = my_input->read_position + offset;
		break;

	case SEEK_END:
		new_pos = my_input->length + offset;
		break;

	default:
		vips_error( "demo", "%s", "bad 'whence'" );
		return( -1 );
	}

	my_input->read_position = VIPS_CLIP( 0, new_pos, my_input->length );

	return( my_input->read_position );
}

static gint64
write_cb( VipsStreamou *streamou, 
	const void *data, gint64 length, MyOutput *my_output )
{
	gint64 bytes_written;

	printf( "write_cb: data = 0x%p, length = %zd\n", data, length );

	bytes_written = write( my_output->fd, data, length );

	return( bytes_written );
}

static void
finish_cb( VipsStreamou *streamou, MyOutput *my_output ) 
{
	printf( "finish_cb:\n" );

	close( my_output->fd );
	my_output->fd = -1;
}

int
main(int argc, char **argv)
{
	MyInput my_input;
	MyOutput my_output;
	VipsStreamiu *streamiu;
	VipsStreamou *streamou;
	VipsImage *image;

	if( VIPS_INIT( NULL ) )
		return( -1 );

	my_input.filename = argv[1];
	my_input.contents = NULL;
	my_input.length = 0;
	my_input.read_position = 0;

	if( !g_file_get_contents( my_input.filename, 
		(char **) &my_input.contents, &my_input.length, NULL ) ) 
		vips_error_exit( "unable to load from %s", my_input.filename );

	streamiu = vips_streamiu_new();
	g_signal_connect( streamiu, "seek", G_CALLBACK( seek_cb ), &my_input );
	g_signal_connect( streamiu, "read", G_CALLBACK( read_cb ), &my_input );

	if( !(image = vips_image_new_from_stream( VIPS_STREAMI( streamiu ), "",
		"access", VIPS_ACCESS_SEQUENTIAL,
		NULL )) ) 
		vips_error_exit( NULL );

	my_output.filename = argv[2];
	my_output.fd = -1;

	if( (my_output.fd = vips__open( my_output.filename, 
		O_WRONLY | O_CREAT | O_TRUNC, 0644 )) == -1 )
		vips_error_exit( "unable to save to %s", my_output.filename );

	streamou = vips_streamou_new();
	g_signal_connect( streamou, "write",
		G_CALLBACK( write_cb ), &my_output );
	g_signal_connect( streamou, "finish", 
		G_CALLBACK( finish_cb ), &my_output );

	if( vips_image_write_to_stream( image, ".png", 
		VIPS_STREAMO( streamou ), NULL ) ) 
		vips_error_exit( NULL );

	VIPS_UNREF( image );
	VIPS_UNREF( streamiu );
	VIPS_UNREF( streamou );

	return( 0 );
}
```

jcupitt:
I'll try hooking them up to Python, then I think we're all done and can merge to master. Phew.

jcupitt:
You can now subclass streams in Python to implement your own sources and sinks.

Example code:

<details>
```python
#!/usr/bin/env python3

import sys
import pyvips

class Mystreami(pyvips.Streamiu):
    def __init__(self, pointer):
        super(Mystreami, self).__init__(pointer)

        # these must be attached before we build the streamiu, see `new`
        self.signal_connect('read', self.read_cb)
        self.signal_connect('seek', self.seek_cb)

    @staticmethod
    def new(name):
        """Make a new input stream from a filename. 

        """

        gtype = pyvips.type_from_name('VipsStreamiu')
        pointer = pyvips.GObject.new_pointer_from_gtype(gtype)
        self = Mystreami(pointer)

        self.name = name
        # use memoryview so we can efficiently copy slices out
        self.loaded_bytes = memoryview(open(name, 'rb').read())
        self.length = len(self.loaded_bytes)
        self.read_point = 0

        return self.build()

    def read_cb(self, buf):
        print(f'read: {len(buf)} bytes ...')

        p = self.read_point
        bytes_available = self.length - p
        bytes_to_copy = min(bytes_available, len(buf))
        buf[:bytes_to_copy] = self.loaded_bytes[p:p + bytes_to_copy]
        self.read_point += bytes_to_copy

        return bytes_to_copy

    def seek_cb(self, offset, whence):
        print(f'seek: offset = {offset}, whence = {whence} ...')
        if whence == 0:
            # SEEK_SET
            new_read_point = offset
        elif whence == 1:
            # SEEK_CUR
            new_read_point = self.read_point + offset
        elif whence == 2:
            # SEEK_END
            new_read_point = self.length + offset
        else:
            raise Exception(f'bad whence {whence}')

        self.read_point = max(0, min(self.length, new_read_point))
        print(f'   new read_point = {self.read_point}')

        # uncomment this to simulate a pipe input instead
        # return -1

        return self.read_point

class Mystreamo(pyvips.Streamou):
    def __init__(self, pointer):
        super(Mystreamo, self).__init__(pointer)

        # these must be attached before we build the streamou, see `new`
        self.signal_connect('write', self.write_cb)
        self.signal_connect('finish', self.finish_cb)

    @staticmethod
    def new(name):
        """Make a new output stream from a filename. 

        """

        gtype = pyvips.type_from_name('VipsStreamou')
        pointer = pyvips.GObject.new_pointer_from_gtype(gtype)
        self = Mystreamo(pointer)

        self.name = name
        self.f = open(name, 'wb')

        return self.build()

    def write_cb(self, buf):
        print(f'write: {len(buf)} bytes ...')

        return self.f.write(buf)

    def finish_cb(self):
        print(f'finish: ...')
        self.f.close()

streamiu = Mystreami.new(sys.argv[1])
image = pyvips.Image.new_from_stream(streamiu, '', access='sequential')

image = image.invert()

streamou = Mystreamo.new(sys.argv[2])
image.write_to_stream(streamou, '.jpg')
```
</details>

So I think that's everything done. Let's leave it a few days for any final polishing and testing, then merge to master.

jcupitt:
Matching PR for the pyvips changes: https://github.com/libvips/pyvips/pull/139

jcupitt:
All done! Thank you everyone! This will be in 8.9. The pyvips changes are merged too.

PR #1452, opened 2019-10-22
#1376 - Adding base docker image for popular web formats

### LINKED ISSUE #1376: Is there interest in automated releases to docker-hub?
I was wondering if there is interest in publishing an official build to docker-hub, and if so would a pull request to create it be accepted?

augustocdias:
Fixes #1376 
I don't have much knowledge about Travis, but I think this should work.
You'd need to create the env vars for `DOCKER_PASSWORD`, `DOCKER_USERNAME` and `VIPS_VERSION` (I don't know if it follows the same as pyvips and it will be used to download the release tar.gz from github)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,14 +20,14 @@ script:
     DYLD_LIBRARY_PATH=$PWD/libvips/.libs
     LD_PRELOAD=$ASAN_DSO
     $PYTHON -m pytest -v test/test-suite
-  - docker build --squash --build-arg VIPS_VERSION -t vips/vips-web .
-  - docker tag vips/vips-web:latest vips/vips-web:$VIPS_VERSION
+  - docker build --squash --build-arg VIPS_VERSION=$TRAVIS_TAG -t vips/vips-web .
+  - docker tag vips/vips-web:latest vips/vips-web:$TRAVIS_TAG
 
 deploy:
   provider: script
   script: 
     - docker push vips/vips-web:latest
-    - docker push vips/vips-web:$VIPS_VERSION
+    - docker push vips/vips-web:$TRAVIS_TAG
   on:
     branch: master
     tags: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: cpp
 
+services:
+  - docker
+
 before_script:
   - $PYTHON -m pip download --no-deps https://github.com/libvips/pyvips/archive/$PYVIPS_VERSION.tar.gz
   - tar xf $PYVIPS_VERSION.tar.gz
@@ -10,12 +13,23 @@ before_script:
     --with-jpeg-libraries=$JPEG/lib
     --with-magick=$WITH_MAGICK
   - make -j$JOBS -s 
+  - echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
 script:
   - make -j$JOBS -s -k V=0 VERBOSE=1 check
   - LD_LIBRARY_PATH=$PWD/libvips/.libs
     DYLD_LIBRARY_PATH=$PWD/libvips/.libs
     LD_PRELOAD=$ASAN_DSO
     $PYTHON -m pytest -v test/test-suite
+  - docker build --build-arg VIPS_VERSION -t vips/vips-web .
+  - docker tag vips/vips-web:latest vips/vips-web:$VIPS_VERSION
+
+deploy:
+  provider: script
+  script: 
+    - docker push vips/vips-web:latest
+    - docker push vips/vips-web:$VIPS_VERSION
+  on:
+    branch: master
 
 matrix:
   allow_failures:
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ script:
     DYLD_LIBRARY_PATH=$PWD/libvips/.libs
     LD_PRELOAD=$ASAN_DSO
     $PYTHON -m pytest -v test/test-suite
-  - docker build --build-arg VIPS_VERSION -t vips/vips-web .
+  - docker build --squash --build-arg VIPS_VERSION -t vips/vips-web .
   - docker tag vips/vips-web:latest vips/vips-web:$VIPS_VERSION
 
 deploy:
@@ -30,6 +30,7 @@ deploy:
     - docker push vips/vips-web:$VIPS_VERSION
   on:
     branch: master
+    tags: true
 
 matrix:
   allow_failures:
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,8 +15,7 @@ RUN apk add --update --no-cache --virtual .build-deps \
             lcms2-dev \
             libpng-dev \
             orc-dev \
-            libwebp-dev \
-            libheif-dev 
+            libwebp-dev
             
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM alpine
+
+ARG VIPS_VERSION
+
+RUN apk update
+
+RUN apk add build-base pkgconfig glib-dev gobject-introspection-dev expat-dev tiff-dev libjpeg-turbo-dev libexif-dev giflib-dev librsvg-dev lcms2-dev libpng-dev orc-dev libwebp-dev libheif-dev
```
lovell: libheif-dev brings in x265-dev which contains patent-encumbered software.

Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM alpine
```
lovell: We should use a tagged, stable version of alpine such as `3.10`.

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+ARG VIPS_VERSION
+
+RUN apk update
+
+RUN apk add build-base pkgconfig glib-dev gobject-introspection-dev expat-dev tiff-dev libjpeg-turbo-dev libexif-dev giflib-dev librsvg-dev lcms2-dev libpng-dev orc-dev libwebp-dev libheif-dev
+
+RUN apk add libimagequant-dev --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing
```
lovell: libimagequant-dev is GPL - are we OK with this?

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+RUN tar xvzf vips-${VIPS_VERSION}.tar.gz -C vips --strip-components 1
+
+WORKDIR /vips
+
+RUN ./configure --enable-debug=no --without-python
+
+RUN make
+
+RUN make install
+
```
lovell: Could remove the `/vips` directory after this point.

Dockerfile
lovell: Some of these dependencies are build-time only so can be removed from the final image. The `apk` utility allows the creation of virtual groups of dependencies that can be used for this purpose. This would allow the separation of e.g. `tiff` (required at build and runtime) from `tiff-dev` (required only at build time).

.travis.yml
```diff
@@ -10,12 +13,23 @@ before_script:
...
     --with-magick=$WITH_MAGICK
   - make -j$JOBS -s 
+  - echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
 script:
   - make -j$JOBS -s -k V=0 VERBOSE=1 check
   - LD_LIBRARY_PATH=$PWD/libvips/.libs
     DYLD_LIBRARY_PATH=$PWD/libvips/.libs
     LD_PRELOAD=$ASAN_DSO
     $PYTHON -m pytest -v test/test-suite
+  - docker build --build-arg VIPS_VERSION -t vips/vips-web .
```
lovell: Could add the `--squash` flag to remove intermediate layers introduced by the use of multiple RUN directives.

.travis.yml
```diff
@@ -10,12 +13,23 @@ before_script:
...
+  - docker tag vips/vips-web:latest vips/vips-web:$VIPS_VERSION
+
+deploy:
+  provider: script
+  script: 
+    - docker push vips/vips-web:latest
+    - docker push vips/vips-web:$VIPS_VERSION
+  on:
+    branch: master
```
lovell: Adding `tags: true` here will ensure we only deploy tagged commits and therefore stable software.

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+RUN ldconfig
+
+WORKDIR /
+
+RUN rm -rf vips
```
lovell: Should we add `CMD /usr/local/bin/vips` as the default command?

Dockerfile
augustocdias: I don't know about this... if you think it should not be there, I can remove

Dockerfile
augustocdias: Done

Dockerfile
augustocdias: Done

.travis.yml
augustocdias: Done

Dockerfile
augustocdias: Done

.travis.yml
augustocdias: Done

Dockerfile
jcupitt: We'd be distributing (by making a docker image available) a binary in which GPL software was linked to non-GPL software, so it would be a GPL licence violation.

Users who want to include GPL software in their stack need to build the stack themselves. I think (??) distributing a docker image which builds a stack which violates the GPL is not a GPL violation, so we could do that.

jcupitt: Oh hmm that's not quite correct. Including GPL packages just makes the whole thing GPL, so any downstream users of the docker image would be GPL too. Many commercial users would find that unacceptable.

augustocdias: I'll remove then

Dockerfile
augustocdias: Did the changes

Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+
+RUN ldconfig /etc/ld.so.conf.d
+
+WORKDIR /
+
+RUN rm -rf vips
+
+RUN apk del .build-deps
+
+RUN apk add --update --no-cache libgsf glib gobject-introspection expat tiff libjpeg-turbo libexif giflib librsvg lcms2 libpng orc libwebp libheif
```
lovell: The `libheif` package depends on `x265`, which contains patent-encumbered software. Are we OK with this?

toadkicker:
I added @jcupitt and @lovell to the dockerhub; feel free to boot me off there. Just been following this issue and this looks like a great solution.

Dockerfile
augustocdias: I can remove as well... I don't know much about libraries's licenses... I just put what made sense to me.

jcupitt: Yes, we probably have to remove libheif as well, unfortunately.

.travis.yml
```diff
@@ -10,12 +13,24 @@ before_script:
...
     LD_PRELOAD=$ASAN_DSO
     $PYTHON -m pytest -v test/test-suite
+  - docker build --squash --build-arg VIPS_VERSION -t vips/vips-web .
+  - docker tag vips/vips-web:latest vips/vips-web:$VIPS_VERSION
+
+deploy:
+  provider: script
+  script: 
+    - docker push vips/vips-web:latest
+    - docker push vips/vips-web:$VIPS_VERSION
```
lovell: Given this will be for tagged commits only, we should be able to use [TRAVIS_TAG](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables) instead of having to manually maintain a separate `VIPS_VERSION` environment variable.

Dockerfile
augustocdias: removed

.travis.yml
augustocdias: I changed. Thanks for the tip

jcupitt:
We've now moved to github actions, so I'll close this to keep things tidy.

PR #1455, opened 2019-10-27
Fix dependency to introspect program on Windows MINGW

larskanis:
I added libvips to the official MINGW-packages here https://github.com/msys2/MINGW-packages/pull/5885 and added this patch to libvips in https://github.com/msys2/MINGW-packages/pull/5907 . So this is the upstream PR. Hope it's correct.


jcupitt:
Hello @larskanis, looks fine, thank you very much!

PR #1459, opened 2019-11-01
Ensure SVG dimensions are rounded correctly when scaling or setting density via cairo

lovell:
See the example in https://github.com/lovell/sharp/issues/1947, which at a density of 220 (scale = 3.055556) is currently incorrectly rendered at 64x42. With the change in this PR it correctly renders at 64x43.

### DIFF
--- a/libvips/foreign/svgload.c
+++ b/libvips/foreign/svgload.c
@@ -280,8 +280,8 @@ vips_foreign_load_svg_parse( VipsForeignLoadSvg *svg, VipsImage *out )
 			 * cairo instead.
 			 */
 			svg->cairo_scale = scale;
-			width = width * scale;
-			height = height * scale;
+			width = VIPS_ROUND_UINT( width * scale );
+			height = VIPS_ROUND_UINT( height * scale );
 		} else {
 			/* SVG with width and height reports correctly scaled 
 			 * dimensions.


kleisauke:
Could this be a duplicate of https://github.com/libvips/libvips/pull/1430?

lovell:
Ah, yes, this is totally a duplicate - sorry, I didn't spot that.

I'm now unsure if we should be using VIPS_CEIL or VIPS_ROUND_UINT - I guess we need to match cairo's internal behaviour.

kleisauke:
I had a quick look at the source code of the `rsvg-convert` tool to see how they handle this. Seems like they're rounding to the nearest integer, see:
https://gitlab.gnome.org/GNOME/librsvg/blob/8712cc65f9b32de91e5ac76cf8671bf1c6838680/rsvg-convert.c#L110-111

So I think `VIPS_ROUND_UINT` is correct here.

m1h4:
Doesn't this seem counter-intuitive? If, for example, you have a shape that spans to 3.25, a canvas that can accommodate the rasterized version of that shape (especially with antialiasing in effect) should span to 4, no? The final pixel (to which .25 of the shape intrudes into) will contain some shading information, no?

dsdsd

Unless I'm misunderstanding the purpose of the affected calculations…

lovell:
I think I agree with VIPS_ROUND_UINT as this relates to the scaling of the viewport rather than the scaling of shapes on the canvas.

The use of VIPS_CEIL could introduce the same problems that the current (implied) use of floor is causing, e.g. an image with shapes relying on bleed/overscan to be cut off might end up with an extra strip of near-background pixels at the edges.

m1h4:
But, unless those are completely blank, you would want to preserve that information, right? Otherwise, the output again might have noticeable clipping artifacts (just less noticeable than the current version).

Basically saying that flooring is obviously not the correct approach here, but the subtle differences between rounding and ceiling could make a difference. Maybe just following librsvg's example is the right thing to do here, I'm just trying to be cautious.

m1h4:
After some hands-on with both solutions and also researching other libs/apps it seems that rounding gives the most consistent results. Since there are often situations where the subtle antaliasing info gets cut at the left and top edges, having it maximally preserved at the bottom and right (via ceiling) is then unnecessary and gives the effect that the image was padded.

jcupitt:
OK, let's merge then. Thank you everyone.

PR #1483, opened 2019-11-28
block metadata changes on shared images

jcupitt:
If images are shared (ref count > 1), block changes to the set of metadata items on the image. These can cause crashes in highly threaded programs.

See https://github.com/lovell/sharp/issues/1986

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ script:
   - LD_LIBRARY_PATH=$PWD/libvips/.libs
     DYLD_LIBRARY_PATH=$PWD/libvips/.libs
     LD_PRELOAD=$ASAN_DSO
-    $PYTHON -m pytest -v test/test-suite
+    $PYTHON -m pytest -sv --log-cli-level=WARNING test/test-suite
 
 matrix:
   allow_failures:
--- a/ChangeLog
+++ b/ChangeLog
@@ -25,6 +25,7 @@
 - fix use of resolution-unit metadata on tiff save [kayarre]
 - support TIFF CIELAB images with alpha [angelmixu]
 - support TIFF with premultiplied alpha in any band 
+- block metadata changes on shared images [pvdz]
 
 17/9/19 started 8.8.4
 - improve compatibility with older imagemagick versions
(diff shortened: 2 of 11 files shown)


jcupitt:
To do:

- [ ] fix up test suite (current fails in a couple of places with this change)
- [ ] check pyvips and ruby-vips tests too
- [ ] some internal libvips metadata manipulation currently fails
- [ ] check things like nip2 as well
- [ ] we could make the locks on metadata changes even tighter: eg. only allow changes during the CONSTRUCT phase of image object build

kleisauke:
I noticed that it doesn't print the GLib warning messages on Travis, commit https://github.com/kleisauke/libvips/commit/fe77eae343dcd5cb05a48826ec17aeb47f2f5c70 should resolve this. Here's the output:
https://travis-ci.org/kleisauke/libvips/jobs/618704046#L2584-L2881
https://gist.github.com/kleisauke/326368a898d6b8b5afd056fc168ec9a4

It looks like Travis currently segfaults due to this line:
https://github.com/libvips/libvips/blob/49a8f0779dc3fc974bee003f14c8332fdbe986bb/libvips/foreign/dzsave.c#L596-L599

jcupitt:
A ruby-vips test is failing. It looks like it might need an API change, unfortunately.

kleisauke:
NetVips also needs to be updated. This example program:
```csharp
var image = Image.NewFromFile("equus_quagga.jpg", access: "sequential");
Console.WriteLine(image.RefCount);

image = image.Copy();
Console.WriteLine(image.RefCount);
```

Prints a reference count of `2` before commit https://github.com/kleisauke/net-vips/commit/3cd1bdf55119c9107c2de38d726cbcd5d3a6e436, so modifying metadata in combination with this PR didn't work.

I've fixed this by early release the `GValue` (`g_value_unset`) within [`VipsObject.Get`](https://github.com/kleisauke/net-vips/blob/3cd1bdf55119c9107c2de38d726cbcd5d3a6e436/src/NetVips/VipsObject.cs#L102-L111) and [`VipsObject.Set`](https://github.com/kleisauke/net-vips/blob/3cd1bdf55119c9107c2de38d726cbcd5d3a6e436/src/NetVips/VipsObject.cs#L123-L129). Perhaps ruby-vips needs a similar approach?
https://github.com/libvips/ruby-vips/blob/master/lib/vips/object.rb#L148-L168

jcupitt:
You're right, I added `GValue#unset` to ruby-vips and called it for GValues I knew were going out of scope. This reduced unnecessary refs enough to make copy / set work. 

It feels messy :( but at least there's no API break.

jcupitt:
nip2 also passes. Let's commit!

PR #1486, opened 2019-11-29
Update C++ binding and function list

kleisauke:
Depends on https://github.com/libvips/pyvips/pull/141.


doc/gen-function-list.py
```diff
@@ -53,15 +55,50 @@ def add_nickname(gtype, a, b):
...
 
-    # add 'missing' synonyms by hand
-    all_nicknames.append('crop')
-
     # make list unique and sort
     all_nicknames = list(set(all_nicknames))
     all_nicknames.sort()
 
+    # make dict with overloads
+    overloads = {
```
kleisauke: I don't think there's a way to automatically generate these overloads?

doc/gen-function-list.py
```diff
@@ -53,15 +55,50 @@ def add_nickname(gtype, a, b):
...
+
+    overloads['relational_const'] = [o + '_const' for o in overloads['relational']] + ['relational_const1'] + \
+                                    [o + '_const1' for o in overloads['relational']]
+
     for nickname in all_nicknames:
-        gen_function(nickname)
+        result = gen_function(nickname)
+        if nickname in overloads:
+            result = result.replace('()', '(), ' + (', '.join('vips_{}()'.format(n) for n in overloads[nickname])))
+        print(result)
```
jcupitt: Very nice!

doc/gen-function-list.py
kleisauke: (I copied and pasted them from the old function list, so maybe I forgot a few)

jcupitt: Ooop, didn't post my comment. Sorry, no, I don't think so, unfortunately.

Issue #1473, opened 2019-11-19
Processing an 8bit LAB+alpha TIFF image, outputs a 16bit image


angelmixu:
Hi!

I'm improving the TIFF file format for the needs of our company in this branch: https://github.com/angelmixu/libvips/tree/improve-file-formats
(BTW, there seems to be a libtiff bug where a CMYK MultiInk file with more than 4 color bands generates a warning)

And after this PR https://github.com/libvips/libvips/pull/1345 when copying an 8bit TIFF LAB+alpha to another file, it generates a 16bit file.
What the PR is doing is that if it find an 8bit LAB with alpha, it generates a LABS with alphas, but then... if you want to save that process in an 8bit LAB image with alphas... there's no way to do it because the image is already a LabS file.

What would be the best way to process this 8bit file?

I was thinking on maybe tag the file as 8bit in some way, or maybe specify the output to be 8 bit in an option of the format.
What do you think?

jcupitt:
I guess the `squash` option could squash lab16 down to lab8.

You could also load lab8 + alpha as float lab. That might be simpler.

PR #1489, opened 2019-12-04
Improve TIFF file format

angelmixu:
Hi! I fixed a compilation issue with Visual Studio 2019, in libvips/arithmetic/unaryconst.c for using memset the string.h header is needed. (I have activated in my project to make an error when an unknowns function would be interpreted as an int).

I've taken into account the inkset in TIFF files, thus differenciating between CMYK and MULTIBAND images.
Also, in the TIFF writer I've used the Type of the VipsImage to choose the output, instead of the number of bands.

When writing a multiband TIFF file, I've set the colour_bands to 4, as libtiff generates a warning when loading the files (as I have seen in the code in libtiff, I think it's a bug in the library), but conceptually, I'd say colour_bands should be the Bands number of the image.

What do you think? :)

Apart from this issue: https://github.com/libvips/libvips/issues/1473
I would need a way to specify some of the extrabands to be EXTRASAMPLE_UNSPECIFIED , and/or maintain the extra samples tag per band when generating the output image.
In the documentation it says "If an extra sample is used to encode information that has little or nothing to do with alpha, ExtraSample=0 (EXTRASAMPLE_UNSPECIFIED) is recommended."
Anyone can point me into the best direction to implement this?

References:
https://www.awaresystems.be/imaging/tiff/tifftags/extrasamples.html
https://www.awaresystems.be/imaging/tiff/tifftags/photometricinterpretation.html
https://www.awaresystems.be/imaging/tiff/tifftags/inkset.html


libvips/foreign/vips2tiff.c
```diff
@@ -654,50 +654,52 @@ wtiff_write_header( Wtiff *wtiff, Layer *layer )
...
+		}
+		else if( wtiff->im->Type == VIPS_INTERPRETATION_CMYK &&
+			wtiff->im->Bands >= 4 ) {
+			photometric = PHOTOMETRIC_SEPARATED;
+			TIFFSetField( tif, 
+				TIFFTAG_INKSET, INKSET_CMYK );
+			colour_bands = 4;
+		}
+		else if (wtiff->im->Type == VIPS_INTERPRETATION_MULTIBAND ) {
+			photometric = PHOTOMETRIC_SEPARATED;
```
jcupitt: I don't think you can assume MULTIBAND images are the same as TIFF MULTIINK. I often use 7 or 8 band TIFFs and I SEPARATED is not correct for them. I think we have to delete this case.

jcupitt:
OK, merged minus the multiband -> multiink code. Thank you @angelmixu !

angelmixu:
Thanks to you @jcupitt  :)

How would you distinguish between a CMYK + alpha channels, and a MULTIBAND image?
The only tag that would say if it's a CMYK or a MULTIBAND (when loading the image) is the INKSET, as it's specified here: https://www.awaresystems.be/imaging/tiff/tifftags/inkset.html
Also, if you have a TIFF file with two channels, it can be a B&W with an alpha, or a MULTICHANNEL.
Maybe I'm missing something :/

jcupitt:
Do you have a sample MULTIINK tiff I could test with? I think you have sent me one in the past, but I've lost it :( 

Yes, `PHOTOMETRIC_MINISWHITE` and `PHOTOMETRIC_MINISBLACK` become libvips greyscale images. `MULTIBAND` is generic "not something I have a special mode for" tag.

angelmixu:
Hi @jcupitt ! Here is a test file.
It's mimiking some of our clients workflow.
First I made a PSD file with Photoshop, and then with our engine, converted to a TIFF multiband file.

[test_multiband.zip](https://github.com/libvips/libvips/files/3944298/test_multiband.zip)

jcupitt:
libtiff seems to dislike multi-ink images too -- I see the warning:

```
TIFFReadDirectory: Warning, Sum of Photometric type-related color channels and ExtraSamples doesn't match SamplesPerPixel. Defining non-color channels as ExtraSamples..
```

As a vips image, MULTIBAND is OK, I think. You see all your data anyway.

What's your goal? Do you want to be able to load one of these muiti-ink images into libvips, save again, and have NumberOfInks and InkSet set correctly in the output? That would require some extra metadata on the image, I think.

angelmixu:
Yes, it seems to be a bug in libtiff, it's in this line:
https://gitlab.com/libtiff/libtiff/blob/master/libtiff/tif_dirread.c#L4108

The function _TIFFGetMaxColorChannels returns 4 channels as it's a 
`PHOTOMETRIC_SEPARATED`. In this case, if it's not a CMYK, I think the number of color channels should be the number of Bands when inkset is set to INKSET_MULTIINK.

In the case of the image I attached, the samplesperpixel variable is 5, and doesn't have extrachannels (as all channels are colors), so it goes inside and shows the warning.

Our goal is to be able to load TIFF images and know if they are a CMYK or a group of Spot Colors per channel. Our clients use multi channel images for generating some kind of different colorations to images, and use extra channels as masks, or additional information like areas of use, and more.
At this moment, we use the Phototop Tag to store channel information, its spot color, its name, if it's a mask, an alpha, etc. It would be awesome if the file formats of libvips would generate a tag for storing channel information of the image, it would help us a lot to know what channel has which information.

For example, if loading a 4 band RGB file, it would help to know if the last channel is a transparency/alpha or an additional mask channel.

I was thinking on generating some kind of tag for VipsImage called maybe VIPS_CHANNEL_INFO, and store all the data there. Now we have our own structure for this information, but maybe it's not the best for libvips. I could try to develop this if you think it's fine :)

PS: there's also a bug with the RichTIFFIPTC tag, it shows a warning too because in libtiff the default data type is wrong. I should probably send a fix to libtiff or ask them about it.

jcupitt:
You can add your own metadata to vips images, for example:

```C
VipsImage *image = vips_image_new_from_file( "x.tif", NULL );
vips_image_set_int( image, "my-int-tag", 12 );
```

Now that image will have that tag -- it'll be copied about as the image is processed, and saved automatically to files in vips format.

You can test for it in a saver, eg.:

```C
if( vips_image_get_typeof( image, "my-int-tag" ) ) {
   ... some action
}
```

There are a set of obvious types you can use, or you can define new ones:

https://libvips.github.io/libvips/API/current/libvips-header.html#vips-image-set

angelmixu:
Great! :)
I was thinking on setting a blob.
Maybe something like this could be used when generating a VipsImage from the image loaders, and then converting to the image format on the savers.

```
struct ChannelInfo {
    const char* channel_name;
    const char* channel_type; //an enum saying if it's a base color, spot color, alpha, mask, etc.
    const unsigned char color [4]; // the spot color, and maybe another variable for if it's in rgb, hsb, lab, etc.
}

struct ChannelDescriptor {
    ChannelInfo channelInformation[];
    int channel_count;
}

ChannelDescriptor* channel_descriptor = malloc(...);
// fill the data

vips_image_set_blob( image, "vips-channel-info", &channel_descriptor);
```

(I'm doing this without having the actual code in front of me)

jcupitt:
You can't put strings into blobs like that, unless you have some packing system like EXIF.

I'd make a set of tags, perhaps:

```
vips_image_set_string( image, "ink-0-name", "some name" );
vips_image_set_int( image, "ink-0-type", 3 );
```

etc. The type could be a string too I suppose.

jcupitt:
If you do it as a set of tags, you can then manipulate them in something like Python, which is very handy.

angelmixu:
Sounds good! I'll give it a try when I have some time.
Thanks! :)

PR #1504, opened 2019-12-18
fix: do not try to use "undefined" delay during WEBP animation save

deftomat:
Expression could be truthy even when no `delay` was provided by user.

Now, the `delay_length` is 0 by default to avoid this situation.


jcupitt:
Ooop, good catch, thank you!

PR #1510, opened 2019-12-30
Fix crash on backward memcpy()

ledeuns:
In some environment, building fails because of backward memcpy().


jcupitt:
Ooop, that's a silly error. 

Thank you very much @ledeuns !

ledeuns:
Giving credits when due. Fix courtesy of tb@openbsd :)

jcupitt:
Oh heh well thank you Theo.

PR #1514, opened 2020-01-06
Reset orientation if exif data is incorrect

Antti:
If data in the orientation field `exif-ifd0-Orientation` is incorrect, ignore the orientation value.
Other software seems to be ignoring orientation value if it's not within an expected range, libvips was falling back to orientation 8.

### DIFF
--- a/libvips/foreign/exif.c
+++ b/libvips/foreign/exif.c
@@ -525,9 +525,8 @@ vips__exif_parse( VipsImage *image )
 		int orientation;
 
 		orientation = atoi( str );
-		if ( orientation < 0 || orientation > 8 ) {
+		if( orientation < 1 || orientation > 8 )
 			orientation = 1;
-		}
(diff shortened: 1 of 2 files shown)


libvips/foreign/exif.c
```diff
@@ -525,7 +525,9 @@ vips__exif_parse( VipsImage *image )
 		int orientation;
 
 		orientation = atoi( str );
-		orientation = VIPS_CLIP( 1, orientation, 8 );
+		if ( orientation < 0 || orientation > 8 ) {
```
jcupitt: Should this be < 1? I don't think 0 is valid.

https://sirv.com/help/resources/rotate-photos-to-be-upright/#exif-values

Style nitpick: libvips style is no space before the `(`, and no `{}` for single line `if()`s. A ternary might be neater.

jcupitt:
Good point, thanks for reporting this.

jcupitt:
Thanks @Antti !

PR #1551, opened 2020-02-12
Tile cache improvements (tiff load performance increase)

angelmixu:
Hi!

At my job we had an issue with slow tiff load times on certain files.
We took a look and after some testing we saw that increasing the tile cache improve the process time from 22 seconds to 2 seconds.

I made a parameter when loading tiff file for specifying the maximum memory when using the tile cache.
It then converts the maximum memory to the maximum number of tiles when using vips_tile_cache.

I don't know if this is a good way of doing it, or maybe there would be a more generic way (across other file formats) for using this variable.

What do you think?

BTW, as in master branch on Visual Studio 2019 is not building, I provided some fixes.

### DIFF
--- a/libvips/deprecated/im_tiff2vips.c
+++ b/libvips/deprecated/im_tiff2vips.c
@@ -69,14 +69,14 @@ im_istifftiled( const char *filename )
 
 static int
 im_tiff_read_header( const char *filename, VipsImage *out, 
-	int page, int n, gboolean autorotate )
+	int page, int n, gboolean autorotate, int max_tile_cache_memory )
 {
 	VipsSource *source;
 
 	if( !(source = vips_source_new_from_file( filename )) )
 		return( -1 );
 	if( vips__tiff_read_header_source( source, 
-		out, page, n, autorotate ) ) {
+		out, page, n, autorotate, max_tile_cache_memory) ) {
 		VIPS_UNREF( source );
 		return( -1 );
 	}
@@ -87,13 +87,13 @@ im_tiff_read_header( const char *filename, VipsImage *out,
 
 static int
 im_tiff_read( const char *filename, VipsImage *out, 
-	int page, int n, gboolean autorotate )
+	int page, int n, gboolean autorotate, int max_tile_cache_memory)
 {
 	VipsSource *source;
 
 	if( !(source = vips_source_new_from_file( filename )) )
 		return( -1 );
-	if( vips__tiff_read_source( source, out, page, n, autorotate ) ) {
+	if( vips__tiff_read_source( source, out, page, n, autorotate, max_tile_cache_memory) ) {
 		VIPS_UNREF( source );
 		return( -1 );
 	}
--- a/libvips/deprecated/im_tiff2vips.c
+++ b/libvips/deprecated/im_tiff2vips.c
@@ -112,11 +112,13 @@ tiff2vips( const char *name, IMAGE *out, gboolean header_only )
 	char *p, *q;
 	int page;
 	int seq;
+    int max_tile_cache_memory;
 
 	im_filename_split( name, filename, mode );
 
 	page = 0;
 	seq = 0;
+    max_tile_cache_memory = 0;
 	p = &mode[0];
 	if( (q = im_getnextoption( &p )) ) {
 		page = atoi( q );
@@ -125,6 +127,11 @@ tiff2vips( const char *name, IMAGE *out, gboolean header_only )
 		if( im_isprefix( "seq", q ) )
 			seq = 1;
 	}
+    if( (q = im_getnextoption( &p )) ) {
+    }
+    if( (q = im_getnextoption( &p )) ) {
+        max_tile_cache_memory = atoi( q );
+    }
 
 	/* We need to be compatible with the pre-sequential mode 
 	 * im_tiff2vips(). This returned a "t" if given a "p" image, since it
@@ -146,11 +153,11 @@ tiff2vips( const char *name, IMAGE *out, gboolean header_only )
 	}
 
 	if( header_only ) {
-		if( im_tiff_read_header( filename, out, page, 1, FALSE ) )
+		if( im_tiff_read_header( filename, out, page, 1, FALSE, max_tile_cache_memory ) )
 			return( -1 );
 	}
 	else {
-		if( im_tiff_read( filename, out, page, 1, FALSE ) )
+		if( im_tiff_read( filename, out, page, 1, FALSE, max_tile_cache_memory ) )
 			return( -1 );
 	}
 #else
--- a/libvips/foreign/pforeign.h
+++ b/libvips/foreign/pforeign.h
@@ -86,9 +86,9 @@ int vips__tiff_write_buf( VipsImage *in,
 gboolean vips__istiff_source( VipsSource *source );
 gboolean vips__istifftiled_source( VipsSource *source );
 int vips__tiff_read_header_source( VipsSource *source, VipsImage *out, 
-	int page, int n, gboolean autorotate );
+	int page, int n, gboolean autorotate, int max_tile_cache_memory );
 int vips__tiff_read_source( VipsSource *source, VipsImage *out,
-	int page, int n, gboolean autorotate );
+	int page, int n, gboolean autorotate, int max_tile_cache_memory);
 
 extern const char *vips__foreign_tiff_suffs[];
 
--- a/libvips/foreign/ppmload.c
+++ b/libvips/foreign/ppmload.c
@@ -403,7 +403,7 @@ vips_foreign_load_ppm_map( VipsForeignLoadPpm *ppm, VipsImage *image )
 
 	gint64 header_offset;
 	size_t length;
-	const void *data;
+	const VipsPel *data;
 
 	vips_sbuf_unbuffer( ppm->sbuf );
 	header_offset = vips_source_seek( ppm->source, 0, SEEK_CUR );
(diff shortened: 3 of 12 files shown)


jcupitt:
Hi @angelmixu, that's interesting. Could you share an image that shows this speed problem?

kleisauke:
Are you doing a lot of crops on the same image? If so, forcing the entire image to memory can speed up the processing time.

@jcupitt Do you know if increasing the tile cache for TIFF images also help for users who have implemented their own `vips_dzsave()` (with custom logic). A issue was recently reported by sharp (https://github.com/lovell/sharp/issues/2074) and NetVips (https://github.com/kleisauke/net-vips/issues/54) on this matter.

angelmixu:
@jcupitt I was trying to make a test for posting in this PR, but I don't have permission to post the image online, may I send it to you privately in any way?
Also, our software has custom transformations that I don't know how to replicate on standard libvips.

The process of the image is the following:
* loads a 10band multiband TIFF file (in this commit https://github.com/libvips/libvips/commit/40bffa8cc6c5982ae8ebc36f48a46bed668c99a1 the reverted code on vips2tiff enables the saving of this kind of image files, but the file we are using comes directly from a client)
* for each band of the image there's a color assigned and an operation between each channel, then generate an RGB image with the resulting composition
* reduce or increase resulting image size
* save to jpeg into a different resolution

@kleisauke yes, that's it! it's a tiled tiff image with lots of bands, and we have to access several times for different purposes. Thus, maintaining the image in memory speeds up a lot of the accesses.

jcupitt:
I tried making a test image:

```
$ vips xyz x.v 1000 1000
$ vips bandjoin "x.v x.v x.v x.v x.v" x.tif[tile]
$ vipsheader x.tif
x.tif: 1000x1000 uint, 10 bands, srgb, tiffload
```

Is that the kind of image? I guess it's tiled if you're using a tile cache. I made a test program:

```python
#!/usr/bin/python3

import sys
import pyvips

im = pyvips.Image.new_from_file(sys.argv[1])

# make a three-band image from the 10-band image with a linear recombination
im = im.recomb([[1] + [0] * 9,
                [0, 1] + [0] * 8,
                [0, 0, 1] + [0] * 7])

im = im.resize(0.7)

im.write_to_file(sys.argv[2])
```

And it runs like this:

```
$ time ./process.py x.tif y.jpg
real	0m0.152s
user	0m0.459s
sys	0m0.043s
```

Is that roughly correct?

jcupitt:
Sure, send me the image by email if you like -- I'm jcupitt at gmail.com.

Of perhaps run `tiffinfo` and post the output? We should be able to make an equivalent image from that.

jcupitt:
If you want to keep an image in memory for repeated use, I would just do:

```python
im = im.copy_memory()
```

If the image is huge and you can't keep the whole thing in memory, you could use a large sparse tile cache:

```python
im = im.tilecache(max_tiles=5000, threaded=True)
```

And it'll keep the most recent 5000 tiles. That might give better performance than `copy_memory()` (or might not heh).

I'd like to not add any more loader flags, if possible. It's much better (when we can) to solve problems by composing operators rather than by adding more features.

angelmixu:
The example looks correct (we use the C interface), the recomb process in this case is simple, ours is very complex (I'm not the owner of the code and can't explain correctly the process, but it has lots of stuff, looks at adjacent pixels, converts to some middle space colours, etc.).

In our case this would output a multiband, not an srgb file, the commit I mentioned before has the changes for generating a multiband image. We differenciate between multiband images, cmyk, and other formats as for us is very important (I build a branch with this modification for using on our software).
```
$ vipsheader x.tif
x.tif: 1000x1000 uint, 10 bands, srgb, tiffload
```
I have sent you an e-mail with the image, and here is the info you requested:

vipsheader returns this:
`12505.tif: 1814x1134 uchar, 10 bands, multiband, tiffload`

and tiffinfo:
```
% tiffinfo 12505.tif 
TIFFReadDirectory: Warning, Sum of Photometric type-related color channels and ExtraSamples doesn't match SamplesPerPixel. Defining non-color channels as ExtraSamples..
TIFF Directory at offset 0x2d9ea8 (2989736)
  Image Width: 1814 Image Length: 1134
  Tile Width: 256 Tile Length: 256
  Resolution: 28.3465, 28.3465 pixels/cm
  Bits/Sample: 8
  Compression Scheme: LZW
  Photometric Interpretation: separated
  Ink Names: Lucido, Lucido, Lucido, Lucido, Lucido, Lucido, Lucido, Lucido, Lucido, Lucido
  Orientation: row 0 top, col 0 lhs
  Samples/Pixel: 10
  Planar Configuration: single image plane
  InkSet: 2
  NumberOfInks: 10
  Photoshop Data: <present>, 770 bytes
```

> If the image is huge and you can't keep the whole thing in memory, you could use a large sparse tile cache:
> 
> ```python
> im = im.tilecache(max_tiles=5000, threaded=True)
> ```

Yes, we usually work with very big files, like one or two GBs in size!
The tilecache of the image is set per image?
Wouldn't the tile_cache function override this when called inside the tiff file format?
I'll take a look and try it :)

Lots of thanks!!!

jcupitt:
There's an flag on loaders to open via memory:

```C
if( !(image = vips_image_new_from_file("somthing.tif", "memory", TRUE, NULL)) )
  error ... ();
```

It'll do the equivalent of `copy_memory` during load.

https://libvips.github.io/libvips/API/current/VipsForeignSave.html#VipsForeignLoad--memory

angelmixu:
Oh, interesting, and in the case of a very big file, that can't be placed in memory, could I specify a maximum memory usage?
Or I better try to use the tile cache?

jcupitt:
In that case, yes, try adding a tile cache node.

angelmixu:
```
        int tile_width  = 0;
        int tile_height = 0;
        int n_lines     = 0;
        vips_get_tile_size( _ref->im,
                            &tile_width,
                            &tile_height,
                            &n_lines);
        
        int max_tiles = VIPS_MAX(1, (max_tile_cache_memory / ((tile_width * tile_height) * _ref->im->Bands)) );
        max_tiles = VIPS_MIN(max_tiles,1000000);
        
        VipsImage* tile_cache_out;
        if( vips_tilecache( _ref->im, &tile_cache_out,
                            "tile_width", tile_width,
                            "tile_height", tile_height,
                            "max_tiles", max_tiles,
                            "threaded", true,
                            NULL ) ) {
            verror(vips_error_buffer());
        }
        g_object_unref(_ref->im);
        _ref->im = tile_cache_out;
```

This code works perfectly, I can get the needed speed.
But there's a drawback, I don't get the correct tile size, it returns the Xsize as width, and 1 as height.

In this case, when going inside vips_get_tile_size function, the dhint is VIPS_DEMAND_STYLE_THINSTRIP, when this file is a tiled image, and when stepping into the rtiff_read_tilewise function, the tile width and height are 256.

Should the tiff2vips at some point set the dhint of the VipsImage to VIPS_DEMAND_STYLE_SMALLTILE?

Can I use this tilecache for every tile type, or should I only use it for TIFF files?

jcupitt:
Hello again,

I'd ignore `tile_width` and `tile_height`. They tell you about the demand style this pipeline will (probably) use, not what's right for your code.

`n_lines` is useful -- that tells you how libvips plans to do output buffering. If it's 256 (for example) that means libvips will keep 256 scanlines of output in memory for output buffering, and that you'll need at least 256 * width pixels in your cache to avoid thrashing.

I'd just use:

```C
        VipsImage* tile_cache_out;

        if( vips_tilecache( _ref->im, &tile_cache_out,
                            "tile_width", 256,
                            "tile_height", 256,
                            "max_tiles", 5000,
                            "threaded", true,
                            NULL ) ) {
            verror(vips_error_buffer());
        }
        g_object_unref(_ref->im);
        _ref->im = tile_cache_out;
```

Perhaps with some logic to make sure the 5000 is at least enough to keep `n_lines` worth of pixels.

angelmixu:
This looks like it's working fine, and gets the same performance as using 5000 max_tiles.
In this file it gets 44 max_tiles and processes it in 1 second :D

```  
        int tile_width  = 0;
        int tile_height = 0;
        int n_lines     = 0;
        vips_get_tile_size( _ref->im,
                            &tile_width,
                            &tile_height,
                            &n_lines);
        
        int memory_needed = VIPS_MIN(max_tile_cache_memory, _ref->im->Bands * _ref->im->Xsize * _ref->im->Ysize);
        int max_tiles = VIPS_MAX(1, (memory_needed / (n_lines * _ref->im->Xsize)) );
        max_tiles = VIPS_MIN(max_tiles,1000000);
     
        VipsImage* tile_cache_out;
        if( vips_tilecache( _ref->im, &tile_cache_out,
                            "tile_width", n_lines,
                            "tile_height", n_lines,
                            "max_tiles", max_tiles,
                            "threaded", true,
                            NULL ) ) {
            verror(vips_error_buffer());
        }
        g_object_unref(_ref->im);
        _ref->im = tile_cache_out;
```

jcupitt:
There's a danger of int overflow there with large images. I think I'd do something like:

```C
int tile_width = 256;
int tile_height = 256;
int tiles_across = 1 + image->Xsize / tile_width;
int tiles_down = 1 + n_lines / tile_height;
int max_tiles = VIPS_CLIP( 10, 2 * tiles_across * tiles_down, 10000 );
```

`VIPS_CLIP( LOW, VALUE, HIGH )` is just max(min) and constrains `VALUE` to lie between `LOW` and `HIGH`.

angelmixu:
Oh cool! Thanks!
I'll try it! :)

Should I close this PR then?

PR #1552, opened 2020-02-16
Add force subsample argument to JPEG save

Elad-Laufer:
If Q >= 90 then subsampling is disabled, this argument forces it


jcupitt:
Hello @Elad-Laufer,

We already have `no-subsample` and `Q`. Perhaps we should deprecate `no-subsample` and have a three way switch: force on, force off, automatic (set by Q), with auto as the default.

lovell:
If there's a plan to offer more fine-grained control over subsampling, perhaps we should consider allowing 4:2:2 as well as 4:4:4 ("no-subsample") and 4:2:0 ("force-subsample")?

jcupitt:
I looked into this when we added `jpeg-chroma-subsample` to jpegload: the only two supported settings for JPEG are 4:4:4 (no subsampling) and 4:2:0 (subsample both Cb and Cr by 2). It's very confusing :(

http://poynton.ca/PDFs/Chroma_subsampling_notation.pdf

jcupitt:
(or that's my understanding from that useful PDF)

lovell:
@jcupitt Thanks for the info, no worries, let's keep this as simple as possible.

libvips/foreign/jpegsave.c
```diff
@@ -194,6 +198,13 @@ vips_foreign_save_jpeg_class_init( VipsForeignSaveJpegClass *class )
 		G_STRUCT_OFFSET( VipsForeignSaveJpeg, quant_table ),
 		0, 8, 0 );
 
+    VIPS_ARG_ENUM( class, "subsample_mode", 19,
+            _( "Subsample mode" ),
+            _( "Select chroma subsample operation mode" ),
+            VIPS_ARGUMENT_OPTIONAL_INPUT,
+            G_STRUCT_OFFSET( VipsForeignSaveJpeg, subsample_mode ),
+            VIPS_TYPE_FOREIGN_JPEG_SUBSAMPLE,
+            VIPS_FOREIGN_JPEG_SUBSAMPLE_AUTO );
```
jcupitt: I think we should tag `no_subsample` as deprecated. That'll hide it from APIs and docs, but it'll still work from old code.

libvips/foreign/jpegsave.c
```diff
@@ -230,7 +241,7 @@ vips_foreign_save_jpeg_target_build( VipsObject *object )
 		jpeg->Q, jpeg->profile, jpeg->optimize_coding, 
 		jpeg->interlace, save->strip, jpeg->no_subsample,
```
jcupitt: Let's not pass no_subsample. Instead, a little higher up, have something like:

```C
if arg-set("no_subsample") && !arg_set("subsample_mode")
  subsample_mode = no_subsample ? OFF : ON
```

ie. old code using the `no_subsample` arg is able to set the default for the new `subsample_mode` value.

libvips/foreign/jpegsave.c
```diff
@@ -297,7 +307,7 @@ vips_foreign_save_jpeg_file_build( VipsObject *object )
...
```
jcupitt: Again, remove `no_subsample`.

libvips/foreign/jpegsave.c
```diff
@@ -367,7 +377,7 @@ vips_foreign_save_jpeg_buffer_build( VipsObject *object )
...
```
jcupitt: Remove 'no_subsample`.

libvips/foreign/jpegsave.c
```diff
@@ -440,7 +450,7 @@ vips_foreign_save_jpeg_mime_build( VipsObject *object )
...
```
jcupitt: Remove.

libvips/foreign/pforeign.h
```diff
@@ -166,7 +166,7 @@ int vips__jpeg_write_target( VipsImage *in, VipsTarget *target,
 	gboolean optimize_coding, gboolean progressive, gboolean strip,
 	gboolean no_subsample, gboolean trellis_quant,
```
jcupitt: Oh heh and again.

libvips/foreign/vips2jpeg.c
```diff
@@ -627,17 +628,24 @@ write_vips( Write *write, int qfac, const char *profile,
...
-	 */
-	if( no_subsample ||
-		qfac >= 90 ) { 
-		int i;
-
-		for( i = 0; i < in->Bands; i++ ) { 
-			write->cinfo.comp_info[i].h_samp_factor = 1;
-			write->cinfo.comp_info[i].v_samp_factor = 1;
-		}
+	switch (subsample_mode) {
```
jcupitt: libvips indentation style is K&R, so:

```C
switch( subsample_mode ) {
case VIPS_...ON:
        break;
```

ie. hard tabs set to 8, 80 char line length.

libvips/foreign/vips2jpeg.c
```diff
@@ -627,17 +628,24 @@ write_vips( Write *write, int qfac, const char *profile,
...
+        case VIPS_FOREIGN_JPEG_SUBSAMPLE_ON:
+            break;
+        case VIPS_FOREIGN_JPEG_SUBSAMPLE_AUTO:
+            /* Turn off chroma subsampling. Follow IM and do it automatically for
+            * high Q.
```
jcupitt: Asterisks should line up.

```C
/* Like 
 * this.
 */
```

libvips/foreign/vips2jpeg.c
```diff
@@ -627,17 +628,24 @@ write_vips( Write *write, int qfac, const char *profile,
...
+            */
+            if( qfac < 90 ) {
+                break;
+            }
+        case VIPS_FOREIGN_JPEG_SUBSAMPLE_OFF:
+        default: {
```
jcupitt: Newline before `{`, like a func definition, and `for` formatted as:

```C
for( i = 0; i < xxx; i++ ) { 
```

libvips/foreign/vips2jpeg.c
```diff
@@ -797,12 +806,18 @@ vips__jpeg_write_target( VipsImage *in, VipsTarget *target,
 	 */
         target_dest( &write->cinfo, target );
 
+    /* Retain old behavior for now
+     */
+    if( no_subsample ) {
+        subsample_mode = VIPS_FOREIGN_JPEG_SUBSAMPLE_OFF;
+    }
+
```
jcupitt: This `if()` can go if we have the extra logic in `_build()`.

test/test-suite/test_foreign.py
```diff
@@ -254,6 +254,27 @@ def jpeg_valid(im):
             # format area at the end
             assert y.startswith("hello world")
 
+    @skip_if_no("jpegload")
+    def test_jpegsave(self):
+        im = pyvips.Image.new_from_file(JPEG_FILE)
+
+        # higher Q should mean a bigger buffer
+        q10 = im.jpegsave_buffer(Q=10)
+        q10_subsample_off = im.jpegsave_buffer(Q=10, subsample_mode=2)
```
jcupitt: I would use the string value here, ie. `"off"`.

test/test-suite/test_foreign.py
```diff
@@ -254,6 +254,27 @@ def jpeg_valid(im):
...
+        q10_subsample_no = im.jpegsave_buffer(Q=10, no_subsample=1)
```
jcupitt: I don't think we should have deprecated API being tested in the test suite, let's remove this.

Same for the case below. Just test `on`, `off`, `auto`.

libvips/foreign/vips2jpeg.c
Elad-Laufer: done

libvips/foreign/vips2jpeg.c
Elad-Laufer: done

libvips/foreign/vips2jpeg.c
Elad-Laufer: done

test/test-suite/test_foreign.py
Elad-Laufer: done

test/test-suite/test_foreign.py
Elad-Laufer: done

libvips/foreign/jpegsave.c
Elad-Laufer: done

libvips/foreign/jpegsave.c
Elad-Laufer: Should I add the check to each `_build(...)` function? currently I only check once at `vips__jpeg_write_target(...)`.

Is there a way to perform the check in single point without code duplication?

jcupitt: Yes, add a build method to the base class for jpegsave and do the test after chaining up.

jcupitt: tiffsave has something a little like this:

https://github.com/libvips/libvips/blob/master/libvips/foreign/tiffsave.c#L153

A `_build` method on the base class and use of `isset` to set defaults.

libvips/foreign/vips2jpeg.c
Elad-Laufer: done

libvips/foreign/jpegsave.c
Elad-Laufer: done

libvips/foreign/jpegsave.c
Elad-Laufer: done

libvips/foreign/jpegsave.c
Elad-Laufer: done

libvips/foreign/jpegsave.c
Elad-Laufer: done

libvips/foreign/pforeign.h
Elad-Laufer: done

jcupitt:
I revised the formatting slightly and added some docs.

PR #1564, opened 2020-02-29
deprecate unused option --delete

UweOhse:
This patch assumes that the --delete option of vipsthumbnail has been ignored by intend. If that's the case it should be hidden, to not cost someone else an hour.


jcupitt:
Oh dear, I'm very sorry you wasted time on this. Looking back, it seems that the "delete" option was removed when thumbnail became a vips operation (rather than a separate program), and that was back in 2016.

I think the current recommended equivalent would be:

```
vipsthumbnail something.jpg -o x.jpg[profile=none]
```

I'll have a look at the related docs.

Thank you for fixing this!

PR #1566, opened 2020-03-04
autogen.sh: Honor NOCONFIGURE

jtojnar:
Linux distros will usually want to run configure separately so the script should respect their wishes. See also https://people.gnome.org/~walters/docs/build-api.txt


jcupitt:
Oh, interesting, I'd not come across that. Thanks!

PR #1590, opened 2020-03-24
Use g_get_num_processors() when GLib >= 2.48.1

kleisauke:
Also backport the patch from https://bugzilla.gnome.org/show_bug.cgi?id=748530

### DIFF
--- a/libvips/iofuncs/threadpool.c
+++ b/libvips/iofuncs/threadpool.c
@@ -382,7 +382,7 @@ get_num_processors( void )
 		/* Prefer affinity-based result, if available 
 		 */
 		if( af_count > 0 )
-			nproc = count;
+			nproc = af_count;
 	}
 }
 #endif /*OS_WIN32*/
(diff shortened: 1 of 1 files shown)


libvips/iofuncs/threadpool.c
```diff
@@ -355,24 +362,33 @@ get_num_processors( void )
...
+		for( af_count = 0; process_cpus != 0; process_cpus >>= 1 )
 			if( process_cpus & 1 )
-				count++;
+				af_count++;
 
-		if( count > 0 )
+		/* Prefer affinity-based result, if available 
+		 */
+		if( af_count > 0 )
 			nproc = count;
```
lovell: Does this line need to be changed too?
```diff
- nproc = count;
+ nproc = af_count;
```

kleisauke: Yes, it had to be changed too. Fixed with https://github.com/libvips/libvips/pull/1590/commits/ceacd0bab9296f02d79c3ec07f8e6fcb8ff383f9.

Thanks!

jcupitt:
Nice job!

Issue #1575, opened 2020-03-13
GIO dependency is through libRSVG


charmander:
When building with `--without-rsvg` (and without everything else), linking fails on GIO functions, e.g. https://github.com/libvips/libvips/blob/764b196d50ebc4b12d9b1a6b1fcb521d447eb2a7/libvips/iofuncs/ginputsource.c#L201

I’m not familiar enough with Autotools to know whether this is a correct fix, but I added it to configure.ac here:

```autoconf
PKG_CHECK_MODULES(REQUIRED, glib-2.0 >= 2.6 gmodule-2.0 gobject-2.0 gio-2.0)
PACKAGES_USED="$PACKAGES_USED glib-2.0 gmodule-2.0 gobject-2.0 gio-2.0"
```

Sorry for not making sure exactly which optional library exclusions are required or listing the exact functions – rebuilding takes a while on my device.

jcupitt:
That's odd, it's supposed to test for GIO and disable rsvg if it's not available. I'll have a look.

jcupitt:
OK, I think that fixes it. Thanks!

lovell:
libvips _is_ dependent on gio (e.g. `g_input_stream_get_type`) independently of librsvg so I think the original suggestion to modify `PKG_CHECK_MODULES` is correct. Happy to create a PR for this if you agree John.

jcupitt:
I think it's guarded by `HAVE_GIO`, but I've not actually tried building without gio :( It would actually be tricky to test.

gio is glib 2.15 and later, so perhaps we should bump the minimum version of glib to 2.15 as well. We could remove a few other ifdefs too, which would be a useful cleanup.

lovell:
Thanks John, there's an example build failure at https://travis-ci.org/github/google/oss-fuzz/jobs/665815922 that might help demonstrate.

jcupitt:
`vips.pc` has `gio-2.0` in `Requires.private`, so I think that ought to work. Is that link line not being generated by `pkg-config`?

I'm still completely confused about the differences between `Libs`, `Requires`, `Requires.private` and `Libs.private` :( I think this is the issue that led to the current state:

https://github.com/libvips/libvips/issues/196

lovell:
The `vips.pc` file appears to correctly include `gio-2.0` in `Requires.private`.

[...]

PR #1596, opened 2020-03-26
Increase minimum glib-2.0 dependency to v2.15 (released March 2012)

lovell:
As discussed in https://github.com/libvips/libvips/issues/1575

This change ensures that gio will always be available and can therefore be added to the list of dependencies.

It fixes a link-time error on the current master branch when building without any transitive dependency on gio e.g.:

```sh
./configure --without-rsvg --without-gsf --without-magick --without-heif --without-orc --without-OpenEXR --without-lcms --without-pangoft2 --without-imagequant --enable-introspection=no 
```


jcupitt:
LGTM! This is a nice simplification.

PR #1597, opened 2020-03-27
allow null image as source for assignment and copy constructor

### LINKED ISSUE #1595: Overzealous tests in VImage constructor and assignment operator ?
The following snippets will always fail:

```cpp
vips::VImage i1;
vips::VImage i2( i1 ); // fails here
```

```cpp
vips::VImage i1;
vips::VImage i2
i1 = i2; // fails here
```

from what I can gather, the culprits are, respectively, [this assert](https://github.com/libvips/libvips/blob/master/cplusplus/include/vips/VImage8.h#L101) and [this assert](https://github.com/libvips/libvips/blob/master/cplusplus/include/vips/VImage8.h#L125) if glib asserts are on and [this referencing](https://github.com/libvips/libvips/blob/master/cplusplus/include/vips/VImage8.h#L108) and [this referencing](https://github.com/libvips/libvips/blob/master/cplusplus/include/vips/VImage8.h#L132) if they are off.

Is there a reason not to accept NULL vobject in right-hand operands while allowing it at construction?

For the little story, I'm testing my implementation of vips under heavy load and came by these lines of code (I seem to have an unrelated issue but thought the behavior highlighted by the snippets was probably not intended).

jaubourg:
This fixes #1595.

I tested it didn't break anything in my own environment but I was unable to locate unit tests related to VImage in the libvips repo (maybe I just missed them) so I'm deferring to Travis here ;)

### DIFF
--- a/cplusplus/include/vips/VImage8.h
+++ b/cplusplus/include/vips/VImage8.h
@@ -81,7 +81,7 @@ public:
 		}
 #endif /*VIPS_DEBUG_VERBOSE*/
 
-		if( !steal ) {
+		if( !steal && vobject ) {
 #ifdef VIPS_DEBUG_VERBOSE
 			printf( "   reffing object\n" ); 
 #endif /*VIPS_DEBUG_VERBOSE*/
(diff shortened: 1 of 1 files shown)


jcupitt:
There are no unit tests for vimage run as standard, unfortunately. There are some tests in cplusplus/examples and they pass. 

Anyway, thanks for this!

PR #1627, opened 2020-04-27
Ensure vips__temp_name creates unique filenames

kleisauke:
There was a possible chance that `vips__temp_name` could generate the same filename twice if it is called concurrently.

This PR ensures that the serial increment is an atomic operation, similar to #1211.


jcupitt:
It could still collide across processes, I suppose, but sure, why not!

Issue #709, opened 2017-08-10
vips_thumbnail: Prefer width over height or vice versa


kleisauke:
I like to see a (optional) `prefer` option in `vips_thumbnail` which can be set to a new enum (let's call it `VIPS_PREFER_WIDTH` or `VIPS_PREFER_HEIGHT` for example).

Use case scenario:
To prevent reduction or enlargement in the horizontal axis, I'm setting the target resize width to `VIPS_MAX_COORD` (which is currently `10000000`), when a user only have specified a fixed height. I'm doing vice versa for the target resize height (so setting it to `VIPS_MAX_COORD` when a fixed width is specified). This feels somewhat hackish.

Example which currently fails due to reduction in the vertical axis (the dimensions of the input JPEG-file are 1600x622):
https://gist.github.com/kleisauke/8e090b4f83679cd536cd3bc7aebe60a2
```bash
thumbnail width = 499
expected width = 500
thumbnail height = 194
expected height = 194
```

This is fixable if we change the script to this:
https://gist.github.com/kleisauke/a7df1817fa7ed8f01a627a7eccd342c4
```bash
thumbnail width = 500
expected width = 500
thumbnail height = 194
expected height = 194
```

When a `prefer` option is in `vips_thumbnail` we can do:
https://gist.github.com/kleisauke/1dc424dfdf1b2b65192be4189a4c6364

janko-m:
It would be nice if omitting one dimension would automatically resize just by the other dimension, rather than the dimension default being a square. That would be ideal because it would be more useful, as it's easy to specify square dimensions explicitly, though I know it would bring backwards incompatibility.

In the `image_processing` gem I'm using this code: https://github.com/GustavoCaso/image_processing/blob/c27afdf10ec2a38889de2d4e0124d24a622ea018/lib/image_processing/vips/processor.rb#L65-L78

jcupitt:
The official way to handle this is to default the unspecified dimension to infinity, so I think I'd write it as something like:

```ruby
      def infer_dimensions((width, height), (current_width, current_height))
        raise Error, "either width or height must be specified" unless width || height

        # or what libvips uses as +inf anyway
        infinity = 10000000

        width = width.nil? ? infinity : width
        height = height.nil? ? infinity : height

        [width, height]
      end
```

Perhaps the `VIPS_MAX_COORD` constant should be defined by ruby-vips to stop this magic number spreading.

janko-m:
[...]

PR #1639, opened 2020-05-03
Default height to VIPS_MAX_COORD for vips_thumbnail

kleisauke:
This to prevent reduction in the vertical axis when the height is omitted. Marked this PR as draft since it changes the behavior of `vips_thumbnail` into a resize based on a specific axis (which is more common, I think) instead than a resize based on a square bounding box.

This PR resolves (partially) #709, but a better way to address that issue is to make `width` an optional parameter. This allows users to do this, for example:
```c
// Resize to a width of 200 pixels, no matter what the height of the input image is
if( vips_thumbnail( filename, &image,
	"width", 200,
	NULL ) )
	return( -1 );

// Resize to a height of 200 pixels, no matter what the width of the input image is
if( vips_thumbnail( filename, &image,
	"height", 200, 
	NULL ) )
	return( -1 );

// Identity transform
if( vips_thumbnail( filename, &image, NULL ) )
	return( -1 );
```

Instead of the current behavior:
```c
// Resize to a width of 200 pixels, no matter what the height of the input image is
if( vips_thumbnail( filename, &image, 200,
	"height", VIPS_MAX_COORD,
	NULL ) )
	return( -1 );

// Resize to a height of 200 pixels, no matter what the width of the input image is
if( vips_thumbnail( filename, &image, VIPS_MAX_COORD,
	"height", 200,
	NULL ) )
	return( -1 );

// Identity transform
if( vips_thumbnail( filename, &image, VIPS_MAX_COORD,
	"height", VIPS_MAX_COORD, 
	"size", VIPS_SIZE_DOWN,
	NULL ) )
	return( -1 );
```

This PR resolves kleisauke/net-vips#71 when merged.

### DIFF
--- a/libvips/resample/thumbnail.c
+++ b/libvips/resample/thumbnail.c
@@ -28,6 +28,8 @@
  * 	- smarter heif thumbnail selection
  * 12/10/19
  * 	- add thumbnail_source
+ * 3/5/20 kleisauke
+ *	- prevent reduction in the vertical axis when the height is omitted
  */
 
 /*
@@ -560,9 +562,6 @@ vips_thumbnail_build( VipsObject *object )
 	if( vips_object_argument_isset( object, "no_rotate" ) ) 
 		thumbnail->auto_rotate = !thumbnail->no_rotate;
 
-	if( !vips_object_argument_isset( object, "height" ) )
-		thumbnail->height = thumbnail->width;
-
 	/* Open and do any pre-shrinking.
 	 */
 	if( !(t[0] = vips_thumbnail_open( thumbnail )) )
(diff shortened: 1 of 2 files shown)


jcupitt:
I'm not sure, this seems like quite a large behaviour change :-( It's likely to break a lot of downstream code.

How about adding a new operation with the new behaviour? `thumbnail_axis` perhaps?

kleisauke:
I'm also a little worried about this behavior/compatibility change. I created this PR mainly for reference purposes. I'll add the `VIPS_MAX_COORD` constant to NetVips instead.

A new `thumbnail_axis` function might be a little odd if it only sets the `thumbnail->height` here
https://github.com/libvips/libvips/blob/54bfa23025397b10e7fede996518e9268521c41e/libvips/resample/thumbnail.c#L563-L564
to `VIPS_MAX_COORD` instead. A new declaration which makes the `width` and `height` optional, for e.g.:
```c 
int
vips_thumbnail_axis( const char *filename, VipsImage **out, ... )
```

Is better, but having 2 functions that can do exactly the same thing is a bit weird.

Anyways, feel free to label this PR as [`snoozing`](https://github.com/libvips/libvips/labels/snoozing). :smile:

jcupitt:
Yes, it would be odd :( Really we'd want to replace all the `vips_thumbnail*()` functions with your proposed new API, but we'd need to do it under a new name to avoid breakage.

`vips_thumbnail2()`? `vips_load_and_resize()`? They all seem ugly :(

jaceppo (approved):

PR #1642, opened 2020-05-05
Ensure magick7 buffer+file checks use consistent min length guard

lovell:
This change add the same `len > 10` check to the buffer-based logic to bring it in line with the existing file-checking logic:

https://github.com/libvips/libvips/blob/a592d99bb2e8a90be00121bf81cba338232183b5/libvips/foreign/magick7load.c#L752-L760

This prevents a call to `GetImageMagick` with a zero-length buffer, which crashes on FreeBSD.


jcupitt:
Nice!

PR #1647, opened 2020-05-10
revise kernel mask calculations

jcupitt:
We were seeing small displacements at some shrink factors because of
rounding in the mask size calcualation. This PR takes the rounding into
account when positioning the mask.

See https://github.com/libvips/libvips/pull/1592#issuecomment-626363031


jcupitt:
I'll merge this, it seems like an clear improvement.

PR #1657, opened 2020-05-18
Improve compile-time libheif and runtime AVIF detection

lovell:
libheif will soon support h265 and/or AVIF (using the same API) so this PR allows libvips to look for either at compile time (see https://github.com/strukturag/libheif/pull/222 for the upstream change).

It also adds the newer `ftypavif` magic bytes (older AVIF images recycled the use of `ftypmif1`, but that is no longer recommended).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -855,14 +855,16 @@ if test x"$with_heif" != x"no"; then
   PKG_CHECK_MODULES(HEIF, libheif,
     [with_heif=yes
      have_h265_decoder=`$PKG_CONFIG libheif --variable builtin_h265_decoder`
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -872,6 +874,8 @@ if test x"$with_heif" != x"no"; then
      with_heif=no
      have_h265_decoder=
      have_h265_encoder=
+     have_avif_decoder=
+     have_avif_encoder=
```
jcupitt, line 878: is this a double-paste?

jcupitt:
Ooop, sorry about the incorrect comment, deleted.

PR #1678, opened 2020-06-09
Revise spngloader

randy408:
Revised spngloader to be structurally similar to the libpng loader and the [test case](https://github.com/randy408/libspng/blob/master/tests/test_spng.h#L93) which guarantees the outputs will be identical for all possible color and bit depth combinations.

There might be some minor syntax errors, I haven't tried to build it.


jcupitt:
Thanks Randy, I'll use this as a base.

PR #1679, opened 2020-06-10
heifload: ignore nclx colour profiles, ensure assumed uchar format

lovell:
lcms can handle standard and reduced profiles but not "nclx" (there's some info about this at https://github.com/strukturag/libheif/issues/119#issuecomment-521958102 ). At some point we should try to handle these, but this change at least prevents failure.

The heif loader assumes uchar but libheif 1.7.0 now provides HDR (double float per channel). A future possible enhancement might be to try query the format up front, but again this change prevents failure for now.

Also fixes a small bug where length was sometimes being truncated when reading, which was causing larger inputs to reach EOF too soon.


jcupitt:
It all looks good!
 
I'm not sure why travis is failing though -- do we need to update the libheif there? Perhaps we need to set a new minimum version.

lovell:
The failure is:
```
==5276==ERROR: AddressSanitizer: allocator is out of memory trying to allocate 0x60 bytes
==5276==ERROR: AddressSanitizer failed to allocate 0x2000 (8192) bytes of InternalMmapVector (error code: 12)
```
...so I'm unsure about this change:
```diff
- heif->length = vips_source_seek( heif->source, 0L, SEEK_CUR );
+ result = vips_source_seek( heif->source, 0L, SEEK_CUR );
```
It's now unclear where `heif->length` is updated so perhaps something like `heif->length = heif->source->length` is required in there somewhere?

lovell:
I've updated this PR to remove the broken bit.

Issue #1574, opened 2020-03-12
heifload: bad image dimensions on decode


daverupp:
Good day
I have some weired behavior when trying to generate a thumbnail out of a heic image, maybe someone can help me out here.

when I try this command:
/vipsthumbnail --size=900x900 20200303_150422.heic --vips-info
everything works as expected

when I try this command:
/vipsthumbnail --size=324x324 20200303_150422.heic --vips-info

I got following error:
(vipsthumbnail:15141): VIPS-WARNING **: 15:58:00.237: error in tile 0 x 120./vipsthumbnail: unable to thumbnail 20200303_150422.heic
heifload: bad image dimensions on decode
heifload: bad image dimensions on decode
heifload: bad image dimensions on decode
heifload: bad image dimensions on decode

Do we need to check something before generating the thumbnail?
the original image exif is:
original image is 
Exif Image Width                : 4608
Exif Image Height               : 3456

jcupitt:
Hello @daverupp,

It's working for me with my test images, eg.:

```
$ vipsthumbnail Example1.heic -s 324x324
```

Could you share the image that's failing for you? Also, which libvips version, which platform, how did you make the binary etc.

daverupp:
Sorry for the poor information...
we build on ubuntu 18.04 libvips 8.9.0 built from source. 

Please see attached the HEIC image, that is producing errors... 
(actually it is taken by a Samsung Galaxy with the picture format HEIF/HEIC) This is a new setting you can choose for android 10 to save some space... so I think, this will be soon quite a standard and would be super cool, if you will find out something about that error... thank you so much for looking into it.
(Sorry, it has to be a zip...)

[20200313_115653.heic.zip](https://github.com/libvips/libvips/files/4329578/20200313_115653.heic.zip)

jcupitt:
Thanks! Yes, I see an error too. I'll have a look.

jcupitt:
It looks like a libheif bug, something to do with fetching thumbnails from rotated images, perhaps? I tried git master libheif and that fails too, so you may have found a current issue.

I'll try to prepare a bug report for the very helpful libheif maintainer.

jcupitt:
Submitted libheif issue for reference: https://github.com/strukturag/libheif/issues/211

daverupp:
Thank you very much for taking care.
So, I close here the thread and I hope, libheif maintainers will have a look into this.

all the best
dave

jcupitt:
[...]

PR #1680, opened 2020-06-10
Heifload fails to load an image with orientation

Elad-Laufer:
A test case which demonstrates a load failure of an heif image with orientation 6, this is a regression as in 8.9 this works as expected.


lovell:
I've been fixing up heifload for recent changes in libheif - does applying #1679 help?

jcupitt:
Hello again, this seems to be working with git master:

```
$ vipsheader heic-orientation-6.heic 
heic-orientation-6.heic: 4032x3024 uchar, 3 bands, srgb, heifload
$ vipsheader heic-orientation-6.heic[autorotate]
heic-orientation-6.heic: 3024x4032 uchar, 3 bands, srgb, heifload
```

The test suite is failing because `heif_valid` checks the value of the RGB pixel at (10, 10), and it's different between the two images.

Elad-Laufer:
rewrote the test to follow our failed use case, in which the the save fails and not the load itself.

Elad-Laufer:
The error I get is as follows:

E           Error: unable to call jpegsave_buffer
E             heifload_buffer: bad image dimensions on decode
E           heifload_buffer: bad image dimensions on decode

Elad-Laufer:
> I've been fixing up heifload for recent changes in libheif - does applying #1679 help?

Ran the (new) test with the these changes, does not change the result.

lovell:
Possibly related to #1574

Elad-Laufer:
It seems that enabling autorotate on load works allows the image to be saved without an error.

Elad-Laufer:
if( image_width != heif->page_width ||
			image_height != heif->page_height ) {
			vips_error( class->nickname, 
				"%s", _( "bad image dimensions on decode" ) );
			return( -1 );
		}

Seems that in some scenarios it does not account for the orientation tag and does not transpose width and height.

PR #1682, opened 2020-06-11
Load PNGs with libspng

jcupitt:
This swaps the PNG loader from libpng to libspng, if it is available.

https://libspng.org/

Benefits:

- about 25% faster
- much simpler and more modern API
- should be more robust
- users should see no difference, except the speed improvement

PNG save stays with libpng for now.


libvips/foreign/spngload.c
```diff
@@ -0,0 +1,800 @@
...
+			return( -1 );
+	}
+	else {
+		t[0] = vips_image_new();
+		vips_foreign_load_png_set_header( png, t[0] );
+
+		/* Initialize for progressive decoding.
+		 */
+		if( (error = spng_decode_image( png->ctx, NULL, 0, 
+			png->fmt, SPNG_DECODE_PROGRESSIVE )) ) {
```
randy408, line 519: Should be `SPNG_DECODE_PROGRESSIVE | SPNG_DECODE_TRNS`, the pixels specified to be transparent in the tRNS chunk will not be transparent without this flag, should also do this for the other `spng_decode_image()` call.

jcupitt: fixed with f96f2d30

jcupitt:
OK, let's merge and hammer out any issues in master.

PR #1686, opened 2020-06-12
Add BT.709 ICC profile and use for HEIF "nclx"

lovell:
HEIF containers allow the use of CICP (Coding-Independent Code-Points) instead of ICC profiles, but the default "nclx" behaviour appears to be the same as "Rec. 709" / "BT.709" as used by most TVs.

https://en.wikipedia.org/wiki/Rec._709

This PR adds the profile from http://www.color.org/chardata/rgb/BT709.xalter as a new "built-in" profile and attaches it where required.

Here's an example image before (without any profile):

before

...and after (with BT709):

after

https://github.com/AOMediaCodec/av1-avif/issues/84 goes into a _lot_ more detail.


jcupitt:
709 is such a boring profile :( it's just srgb with a slightly different transfer function. Why couldn't they just use srgb?

Still, we obviously must have it for avif. Let's do it!

jcupitt:
Hang on, having read the linked issue more carefully, they are just talking about 709 chromaticities, not the transfer function.  709 and srgb have the same chromaticities.

Look at point 6b here:

https://github.com/AOMediaCodec/av1-avif/issues/84#issuecomment-626480194

They are using the srgb `transferCharacteristics` as the default, so the effective default ICC profile is actually srgb, not bt709.

Should we revert and think about this again?

lovell:
Ah yes, sorry John, let's revert.

That comment led me to discover https://github.com/joedrago/colorist/blob/84ff75b82e6d707a282ddcb7bb32e6fad4ebae3c/lib/src/format_avif.c#L355 which provides an implementation of the nclx to ICC conversion.

jcupitt:
Good find on the nclx converter. I suppose we'll need something like that.

ValZapod:
> just srgb with a slightly different transfer function. Why couldn't they just use srgb?

No, it is not. sRGB is almost 2.2 gamma while bt.709 is 2.4 gamma on ideal display like OLED..

jcupitt:
Hi @ValZapod, that's true, but AVIF are not using the 709 transfer function, just the chromaticities, which are the same as sRGB. Or that's my understanding from the linked thread.

ValZapod:
> are not using the 709 transfer function

They are. They can that is. sRGB is 13 for transfer function in nclx while BT.709 is 1.

jcupitt:
I found this:

https://github.com/AOMediaCodec/libavif/wiki/CICP#unspecified

So for the unspecified case they are using 13, the sRGB transfer function. So (I think?) libvips is correct to tag AVIF images with no profile as being sRGB and not 709.

ValZapod:
Yes, it is correct.

PR #1695, opened 2020-06-28
Revert "Update include path for libspng"

randy408:
The include path will stay the same for v0.6.0, sorry about that.

Reverts libvips/libvips#1694


jcupitt:
No problem, thanks for keeping me updated.

We've merged the final big PR for 8.10, we'll let it fuzz for a couple of weeks and then release 8.10.

Issue #192, opened 2014-11-06
vips in the browser


andrzejbl:
Hi, 

I'm searching for a lib that will allow me to use minimum amount of memory while downsizing images.
I want to compile it to javascript so it will run in browser. It means that disk I/O also uses memory. 

Could you tell be what is the lowest memory usage when I'm resizing 36mpix image?

cancerberoSgx:
Did anybody has success building a webassembly version of the library ? I'm the author of a ImageMagick webasesembly port using emscripten and there I compiled almost all of its delegate libraries some of which are also dependencies of this library - before I give it a try I would like to know if you think there is viable or if anybody has already tried or any experience, thanks!

jcupitt:
Hi @cancerberoSgx, 

It has been talked about a few times, but I don't know of anyone who has actually done it. Please post an update if you get it working, I'd be interested.

kleisauke:
I started working on this as part of a university research project. 

After 3 days of trial and error, I managed to create a working proof of concept:
https://kleisauke.nl/wasm-vips/

It's able to thumbnail a [4120x2747 image](https://kleisauke.nl/wasm-vips/zebra.jpg) to 500x333 within ~200 ms, which is quite nice. I compiled libvips using Emscripten with a minimal set of dependencies (gettext, zlib, ffi, glib, expat and webp). 

Note that this example uses the SharedArrayBuffer capability in browsers. That API allows sharing memory between the main thread and web workers as well as atomic operations for synchronization, which enables Emscripten to implement support for the Pthreads (POSIX threads) API.
It may not work in all browsers, and due to the recent Spectre and Meltdown vulnerabilities, browsers other than Chrome [have disabled SharedArrayBuffer by default](https://caniuse.com/#feat=sharedarraybuffer).

Most of the work consisted of patching the dependencies. I was able to compile libffi with [this patch](https://github.com/atrosinenko/libffi/compare/master...atrosinenko:emscripten), glib requires some minor patches in the build system. libvips itself requires a major patch (> 1100 lines) to [fix all function pointer issues](https://emscripten.org/docs/porting/guidelines/function_pointer_issues.html), for example:

<details><summary>Details</summary>

````patch
diff --git a/libvips/iofuncs/gate.c b/libvips/iofuncs/gate.c
index 1111111..2222222 100644
--- a/libvips/iofuncs/gate.c
+++ b/libvips/iofuncs/gate.c
[...]

PR #1697, opened 2020-06-30
Support for use in WASM environments (#192)

kleisauke:
This PR enables the use of libvips in WebAssembly environments (see #192). It consist of fixing function pointer cast issues, 
which can only be detected at runtime (as far as I know). This PR has been tested on a WASM-ported version of the libvips test suite that includes 192 different tests.

It is possible that not all function pointer cast issues are fixed by this PR, as the WASM build does not contain all dependencies supported by libvips. I've managed to support JPEG, PNG, TIFF and WebP images by using libjpeg-turbo, libpng, libtiff and libwebp. EXIF metadata in JPEG files and ICC profile support is also available by using libexif and lcms2. For more details, see https://github.com/libvips/libvips/issues/192#issuecomment-590093454.

Here is a brief overview of the related changes within this PR:
- Add `vips_area_free_cb` (that matches the `int (*VipsCallbackFn) (void *a, void *b)` signature), in favour of the `vips_free` callback.
- Deprecate `vips_free`, just use `g_free` for non-callback functions and `vips_area_free_cb` for callback functions (e.g. `vips_blob_new` or `vips_image_set_blob`).
- `vips_crop` needs an adapter function to work properly. This is basically the same as [this patch for glib](https://gist.github.com/kleisauke/acfa1c09522705efa5eb0541d2d00887#file-glib-emscripten-patch-L495-L553). I've `#ifdef`'d this change with `__EMSCRIPTEN__` since this is a quite a hackish workaround.
- All `GThreadFunc` callbacks now have the correct signature (i.e. `void *(*GThreadFunc) (void *data)`).
- Sometimes the last `void *` parameter was missing from the `VipsSListMap2Fn` and `VipsRegionFillFn` callbacks.
- Add a `void *user_data` parameter to the `GCompareFunc` and `GFunc` callback functions. Needed by `g_slist_sort` and `g_slist_foreach`. [This SO answer](https://stackoverflow.com/a/14044244/10952119) might be relevant here.
- Signals using the `vips_INT__VOID`, `g_cclosure_marshal_VOID__POINTER` and `g_cclosure_marshal_VOID__VOID` closures are called with an additional `void *data` argument.

`abi-compliance-checker` result:
https://kleisauke.nl/compat_reports/vips/master_to_wasm-function-pointers/compat_report.html

### DIFF
--- a/libvips/colour/profile_load.c
+++ b/libvips/colour/profile_load.c
@@ -96,7 +96,7 @@ vips_profile_load_build( VipsObject *object )
 	else if( (data = vips__file_read_name( load->name, 
 		vips__icc_dir(), &length )) ) 
 		profile = vips_blob_new( 
-			(VipsCallbackFn) g_free, data, length );
+			(VipsCallbackFn) vips_area_free_cb, data, length );
 	else {
 		vips_error( class->nickname, 
 			_( "unable to load profile \"%s\"" ), load->name );
--- a/libvips/conversion/extract.c
+++ b/libvips/conversion/extract.c
@@ -172,6 +172,14 @@ vips_extract_area_build( VipsObject *object )
 	return( 0 );
 }
 
+#ifdef __EMSCRIPTEN__
+static void
+vips_crop_class_init_adapter( VipsExtractAreaClass *class, void *dummy )
+{
+	vips_extract_area_class_init( class );
+}
+#endif
+
 static void
 vips_extract_area_class_init( VipsExtractAreaClass *class )
 {
@@ -226,6 +234,14 @@ vips_extract_area_class_init( VipsExtractAreaClass *class )
 
 }
 
+#ifdef __EMSCRIPTEN__
+static void
+vips_crop_init_adapter( VipsExtractArea *extract, void *dummy )
+{
+	vips_extract_area_init( extract );
+}
+#endif
+
 static void
 vips_extract_area_init( VipsExtractArea *extract )
 {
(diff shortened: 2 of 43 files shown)


jcupitt:
This patch is much smaller than I was fearing, phew! It seems straightforward.

I was wondering about the best way to prevent this kind of issue creeping back into libvips in the future. Do you have any thoughts on this? I suppose your set of WASM tests would be a good start.

kleisauke:
I was counting the amount of lines of the [`.patch` file](https://patch-diff.githubusercontent.com/raw/libvips/libvips/pull/1697.patch), and not the amount of changed lines, so the PR arrived much smaller. :smile:

Indeed, the WASM tests should be a good start. I just found out that GCC supports a compiler flag called  `-Wcast-function-type` that warns when a function pointer is cast to an incompatible function, which might help to prevent these issues as well. I had an attempt to add this to Travis, see:
https://github.com/libvips/libvips/compare/master...kleisauke:cast-function-type
https://travis-ci.org/github/kleisauke/libvips/jobs/707609405

It reports some false-postives with the `GCompareFunc` callback as that [is called with an extra `user_data` argument](https://stackoverflow.com/a/14044244/10952119). But it also found 2 function pointer cast issues that the WASM test suite didn't catch, see:
https://github.com/libvips/libvips/blob/84d0d5a7a32416806fee463155361a347b06e6cb/tools/vips.c#L111-L112
```c
static void *
list_class( GType type, void *data )
```

https://github.com/libvips/libvips/blob/e728e5638d41210582ee1389294f4dc1b229c52e/libvips/foreign/dzsave.c#L223-L224
```c
static void *
vips_gsf_tree_close( VipsGsfDirectory *tree, void *a, void *b )
```

jcupitt:
I think I fixed those function casts, thanks!

Do you plan a PR with those travis improvements?

kleisauke:
Great! PR https://github.com/libvips/libvips/pull/1716 improves the CI by enabling the `-Wcast-function-type` warning and updating Ubuntu/GCC to the latest version.

PR #1699, opened 2020-06-30
Further reduce animated WebP image size by ~5% when min_size is set

lovell:
When `min_size` has been explicitly set to minimise the size of an animated WebP image, also setting `allow_mixed` allows the encoder to evaluate and mix both lossy and lossless frames.

See https://github.com/webmproject/libwebp/commit/f29bf582df2349b29264e188670750b54850651e

The original claim is a 5% reduction, but I've seen anywhere from a ~2%-15% reduction depending on the input. There doesn't appear to be any increase in compression time and I've not seen any increase in file size so this seems like a safe setting.


jcupitt:
Nice!

Issue #1084, opened 2018-09-11
look at libnsgif


jcupitt:
It seems libnsgif is quite a bit better than giflib:

https://www.netsurf-browser.org/projects/libnsgif/

We should look into swapping our gif load library.

jcupitt:
libnsgif uses the NetSurf build system and is awkward to make in isolation. Here's a fork of libnsgif with an autotools build system ready for adding to the libvips stack.

https://github.com/jcupitt/libnsgif

jcupitt:
libvips branch for working on libnsgif support

https://github.com/jcupitt/libvips/tree/add-libnsgif-support

kleisauke:
It seems that giflib recently removed the Autotools build system in favour of a hand-written Makefile. This will cause problems for cross-compilation.

GEGL (used by GIMP) seems to have moved to libnsgif (by [copying it to the libs folder](https://github.com/GNOME/gegl/tree/a07cbdf3c9fb5e78ce709d8950ab086e54ca2997/libs/libnsgif)). GDK-PixBuf is also considering switching to libnsgif: https://gitlab.gnome.org/GNOME/gdk-pixbuf/issues/46.

jcupitt:
Argh! How annoying.

I posted a begging note:

https://sourceforge.net/p/giflib/feature-requests/6/

Though I doubt if it'll help. I'll fix up the half-done libnsgif loader.

jcupitt:
Branch with a half-done gif loader using libnsgif

https://github.com/libvips/libvips/tree/add-libnsgif

lovell:
I'd be very happy to help test a libnsgif-based loader when it's ready to do so.

Recently I've experienced not only build system changes to giflib patch releases but also API/ABI breakages. Homebrew is pinned to giflib v5.1.4 due to these and the binaries provided by sharp are locked likewise.

https://sourceforge.net/p/giflib/bugs/132/
https://sourceforge.net/p/giflib/bugs/133/

jcupitt:
And at least that second problem is because esr abandoned autotools, sigh.

I'll add fixing up that libnsgif loader to the to-do list. Unless anyone else fancies a go?

jcupitt:
This branch now has a working GIF loader based on libnsgif:

https://github.com/libvips/libvips/tree/add-libnsgif

[...]

PR #1709, opened 2020-07-07
Add libnsgif support

jcupitt:
See this issue for notes on this PR: https://github.com/libvips/libvips/issues/1084

This is probably too late for 8.10. Let's look at this again during the 8.11 cycle.


jcupitt:
I moved the libnsgif source into the tree and removed the old giflib code, so we now have no external dependencies for GIF load.

I added a `--without-nsgif` flag to configure to disable GIF load if you wish.

jcupitt:
Unfortunately it seems libnsgif has non-functioning `DISPOSE_PREVIOUS` handling again, I'll investigate.

tlsa:
@jcupitt Thanks for the patch, I have appied it upstream.

http://git.netsurf-browser.org/libnsgif.git/commit/?id=3665353c4da1dfb7b4a1d9fa412f0dec2cb673d2

jcupitt:
Hi @tlsa, thanks for applying the patch! Hopefully I didn't break some other case :( GIF has to be the worst image format ever :(

lovell:
Apologies for not being able to test this earlier, before merging.

Using [rotating-squares.gif](https://github.com/lovell/sharp/blob/master/test/fixtures/rotating-squares.gif) as sample input shows a difference in the reported metadata.

Before:
```
$ vipsheader -a rotating-squares.gif 
...
n-pages: 30
loop: 0
delay: 30 30 30 30 30 30 30 30 30 30 30 30 30 30 30 30 30 30 30 30 30 30 30 30 30 30 30 30 30 30 
```
After:
```
$ vipsheader -a rotating-squares.gif 
...
n-pages: 30
delay: 30
```

lovell:
The missing `loop` property can be fixed with the following diff, but I'm less sure about the right approach for `delay`:
```diff
--- a/libvips/foreign/nsgifload.c
+++ b/libvips/foreign/nsgifload.c
@@ -285,6 +285,7 @@ vips_foreign_load_nsgif_set_header( VipsForeignLoadNsgif *gif,
        vips_image_set_int( image, VIPS_META_N_PAGES, 
                gif->frame_count_displayable );
        vips_image_set_int( image, "gif-loop", gif->anim->loop_count );
+       vips_image_set_int( image, "loop", gif->anim->loop_count );
 
        vips_image_set_array_int( image, "delay", gif->delay, gif->n );
```

jcupitt:
Drat, sorry Lovell, I should have tested more carefully.

I think it's all fixed with https://github.com/libvips/libvips/commit/89c12b0e62ff9ce019a388f2fe53ac83a498366e if you have time to test.

lovell:
@jcupitt Your fix looks good, thank you.

lovell:
@jcupitt Is it possible to recover from `GIF_INSUFFICIENT_FRAME_DATA`? There are ~25 places in `libnsgif.c` that can return this status, which makes it feel more like something that should always fail fast.

jcupitt:
Oh interesting, I'm not sure. Do you have an example gif which behaves badly?

lovell:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=32607 is a 23 byte GIF file that claims to be 16448x73536.

jcupitt:
Oh wow OK that's bad. 

I'm planning to work on jp2k compression in tiff today, so if anyone else has time to fix this, please go ahead!

PR #1716, opened 2020-07-14
Travis improvements

kleisauke:
This PR improves the CI by upgrading Ubuntu and GCC to the latest version available. The `-Wcast-function-type` warning is now enabled as discussed in https://github.com/libvips/libvips/pull/1697.

I had to add bash to the LSan suppression file, otherwise the build will somehow fail during `make check`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,13 +9,10 @@ addons:
   apt:
     update: true
     sources: &common_sources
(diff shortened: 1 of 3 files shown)


jcupitt:
Ah, much nicer! Bionic is a good step forward.

Sorry about the stupid tree decl.

PR #1718, opened 2020-07-15
Update link for vips-gmic

Zeranoe:
vips-gmic doesn't appear to be a libvips repo anymore.


jcupitt:
Ooop, you're right, this link was changed in error. Thanks!

PR #1727, opened 2020-07-19
Fix undefined-shift in scanline_read_old

kleisauke:
See for details: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24230.

As an aside,
I somehow couldn't reproduce this within the Travis build, see:
https://travis-ci.org/github/kleisauke/libvips/jobs/709682348

Perhaps there's a `-fsanitize=` check missing within our flags?
https://github.com/libvips/libvips/blob/72ca38cab0616795dc7b67a688c9a0c2027a9ecc/.travis.yml#L113

For reference, here are the undefined sanitizer flags set by OSS-Fuzz:
https://github.com/google/oss-fuzz/blob/b4bf7839051cc7bcc9f4d1989eaab82865bae367/infra/base-images/base-builder/Dockerfile#L93


libvips/foreign/radiance.c
```diff
@@ -546,7 +546,7 @@ scanline_read_old( VipsSbuf *sbuf, COLR *scanline, int width )
 			scanline[0][BLU] == 1 ) {
 			int i;
```
jcupitt, line 547: Should i be a guint? Otherwise 255 << 24 will make a huge -ve number and break the loop.

I don't think it's dangerous, but it would be a decode error.

kleisauke: Ah, I overlooked that one. Commit https://github.com/libvips/libvips/commit/aac17486152b90de3f5d22cca1b30de1be6333fb fixes this. It is now able to decode `thumbnail_fuzzer-5111890150424576` without any problems.

kleisauke:
It seems that UBSan didn't exits with a non-zero code on error. Commit https://github.com/kleisauke/libvips/commit/5f0abaceb4d5177c56b1fecb4f4b91535cb569fc on the [`oss-fuzz-24230-travis`](https://github.com/kleisauke/libvips/commits/oss-fuzz-24230-travis) branch resolves this. However it revealed new UBSan errors on the existing corpus, see:
https://travis-ci.org/github/kleisauke/libvips/jobs/709752013
```
flatten.c:190:4: runtime error: 1.64485e+07 is outside the range of representable values of type 'unsigned short'
scRGB2sRGB.c:114:11: runtime error: 8224 is outside the range of representable values of type 'unsigned char'
mosaic_fuzzer.cc:47:56: runtime error: member access within misaligned address 0x60b0000004c3 for type 'struct mosaic_opt', which requires 2 byte alignment
```

I could resolve the last error with commit https://github.com/kleisauke/libvips/commit/161e66746eadc571a5e1fa46a94c5d1e02236959, see:
https://travis-ci.org/github/kleisauke/libvips/jobs/709931677

I'm not sure what's causing these other errors, maybe a `VIPS_CLIP` is missing?

jcupitt:
No, I'm not sure either. Let's merge this nice improvement for now and look at the other error in a new issue.

kleisauke:
Here's a work-in-progress branch to fix the remaining UBSan errors:
https://github.com/libvips/libvips/compare/master...kleisauke:fix-ubsan-errors

Relevant Travis build:
https://travis-ci.org/github/kleisauke/libvips/jobs/710731533

A `VIPS_CLIP` was missing in `vips_scRGB2sRGB_line_8`. I also added this in `vips_scRGB2BW_line_8`, just to be on the safe side. I fixed the `vips_flatten` UBSan error with some saturated casts for the int types (copied from `vips_cast`).

It still needs some commit polishing and I'm not sure if there's a better way to fix that UBSan error in `vips_flatten`. What do you think?

PR #1735, opened 2020-07-21
Performance: improve scRGB to sRGB conversion by ~8%

lovell:
Hello, I think we can safely remove the `isinf` checks as the values are clipped before use (this represents about half of the savings).

The `CLIP` macro was missing an `else`, which I think was preventing the compiler from detecting this as a "clip a value" operation.

Out-of-gamut tracking is optional and at a slight cost so we can remove it where the return value is ignored.

Before:
```
225,846,621  ???:vips_col_scRGB2sRGB_8 [/usr/local/lib/libvips.so.42.12.3]
 31,367,336  ???:vips_scRGB2sRGB_gen [/usr/local/lib/libvips.so.42.12.3]
```
After:
```
198,858,660  ???:vips_col_scRGB2sRGB_8 [/usr/local/lib/libvips.so.42.12.3]
 31,365,334  ???:vips_scRGB2sRGB_gen [/usr/local/lib/libvips.so.42.12.3]
```

(All these timings are using gcc v9.2.1 with the `-O3` flag.)

### DIFF
--- a/libvips/colour/LabQ2sRGB.c
+++ b/libvips/colour/LabQ2sRGB.c
@@ -238,7 +238,7 @@ vips_col_XYZ2scRGB( float X, float Y, float Z, float *R, float *G, float *B )
 /* Turn scRGB into sRGB. Return og=1 for out of gamut - rgb will contain an 
  * approximation of the right colour.
  *
- * Return -1 for NaN, Inf etc. 
+ * Return -1 for NaN.
  */
 static int
 vips_col_scRGB2sRGB( int range, int *lut, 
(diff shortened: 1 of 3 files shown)


jcupitt:
Nice!

kleisauke:
Great! I wonder if we can safely change these `VIPS_FCLIP`  to `VIPS_CLIP` as well?

https://github.com/libvips/libvips/blob/7c413f88725d8055660e0bd551bedbb68e8501a1/libvips/colour/scRGB2sRGB.c#L143
https://github.com/libvips/libvips/blob/10c4831a709063189d67b7d6076b1b26bf62b2b2/libvips/colour/scRGB2BW.c#L116

jcupitt:
Yes, I think so. 

```C
q[j] = VIPS_CLIP( 0, (int) (p[j] * 256.0), USHRT_MAX ); 
```

PR #1745, opened 2020-07-26
WebP save orientation tag test

Elad-Laufer:
This failing test demonstrates what seems to an issue in which the orientation tag is not written to WebP.

We encountered this issue when testing the latest libvips master, built on Debian. Also reproducible on Mac with latest brew dependencies.

### DIFF
--- a/test/test-suite/test_foreign.py
+++ b/test/test-suite/test_foreign.py
@@ -17,7 +17,7 @@ from helpers import \
     GIF_ANIM_DISPOSE_PREVIOUS_FILE, \
     GIF_ANIM_DISPOSE_PREVIOUS_EXPECTED_PNG_FILE, \
     temp_filename, assert_almost_equal_objects, have, skip_if_no, \
-    TIF1_FILE, TIF2_FILE, TIF4_FILE
+    TIF1_FILE, TIF2_FILE, TIF4_FILE, IMAGES
 
 
 class TestForeign:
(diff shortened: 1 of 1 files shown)


jcupitt:
Yes, libvips loads ICC, XMP and EXIF from webp, but doesn't save it. Let's tag this as an enhancement.

jcupitt:
Actually, this should work. I see:

```
$ vipsheader -f orientation Portrait_6.jpg 
6
$ vips copy Portrait_6.jpg x.webp
$ vipsheader -f orientation x.webp
6
```

Perhaps your libvips has been built without libwebpmux support?

jcupitt:
... I'm reluctant to add a test for this, since libwebp metadata support is optional.

The webp devs split the library into two parts: libwebp, which just handles image encode and decode, and libwebpmux, which handles image metadata. libvips can be built with libwebp and without libwebpmux. 

I see:

```
$ ldd $(which vips) | grep webp
	libwebpmux.so.3 => /lib/x86_64-linux-gnu/libwebpmux.so.3 (0x00007f0966f32000)
	libwebpdemux.so.2 => /lib/x86_64-linux-gnu/libwebpdemux.so.2 (0x00007f0966d2d000)
	libwebp.so.6 => /lib/x86_64-linux-gnu/libwebp.so.6 (0x00007f0966ac4000)
```

jcupitt:
I'll have a look at the brew formula, perhaps they need to add a dep on libwebpmux and demux.

PR #1758, opened 2020-08-02
Add C++ bindings for new_from_memory_steal()

Zeranoe:
new_from_memory_steal() will create a new image with the input
buffer and will "move" the data into the image. The buffer is then
managed by the image, and will be freed when it goes out of scope.


libvips/iofuncs/image.c
```diff
@@ -3872,6 +3872,43 @@ vips_band_format_iscomplex( VipsBandFormat format )
...
+ * vips_image_free_buffer:
+ * @image: the image that contains the buffer
+ * @buffer: the orignal buffer that was stolen
+ *
+ * Free the externally allocated buffer found in the input image. This function
+ * is intened to be used with g_signal_connect, so it can free an externally
+ * allocated buffer.
+ */
+void
+vips_image_free_buffer( VipsImage* image, void* buffer )
```
jcupitt, line 3885: You don't need the switch, you can just write eg.:

```
static void
vips_free_cb( VipsObject *object, char *buf )
{
        free( buf );
}
```

cplusplus/include/vips/VImage8.h
```diff
@@ -550,6 +550,23 @@ class VImage : VObject
...
+	new_from_memory_steal( void *data, size_t size,
+		int width, int height, int bands, VipsBandFormat format )
+	{
+		VipsImage *image;
+
+		if( !(image = vips_image_new_from_memory( data, size, 
+			width, height, bands, format )) )
+			throw( VError() );
+
+		if( g_signal_connect( image, "postclose",
```
jcupitt, line 563: I don't think this can fail -- I wouldn't check the result.

jcupitt:
This looks great! We're in feature freeze now for 8.10, so let's delay this a week for the 8.11 cycle.

jcupitt:
Looks nice, merged! Thank you very much for doing this work.

jcupitt:
I noted this in the ChangeLog and crfedited you, and added a doxy comment.

PR #1765, opened 2020-08-05
Add C++ bindings for VRegion

Zeranoe:
VRegion wraps a regular region, allowing for C++ to quickly access the underlying data.


jcupitt:
Hi @Zeranoe,

Thanks for this PR, it looks very interesting. We're in feature-freeze right now, ready for the upcoming 8.10. Let's look at this again in a week or so.

My instinct would be to wrap `VipsRegion` in the same way as `VipsImage` (ie. as a sub class of `VObject`) -- it feels a bit weird to have two different ways of doing it. Let's discuss this after 8.10.

jcupitt:
Hi again, would you be willing to rework this as a VObject subclass, rather like VImage?

Zeranoe:
> Hi again, would you be willing to rework this as a VObject subclass, rather like VImage?

I can do that, but I'm curious what the advantage is of having it as a VObject subclass?

jcupitt:
Of course it's debatable, but VipsRegion and VipsImage are both subclasses of VipsObject in the C API, so it would feel strange not to have VRegion as a subclass of VObject in the C++ binding. Plus it would give you automatic unreffing of the VipsRegion object, which (I think?) your template does not have.

You've used a template because you want automatic casting of the data pointer. How about making a very small template over a VRegion object for this?

Zeranoe:
> Of course it's debatable, but VipsRegion and VipsImage are both subclasses of VipsObject in the C API, so it would feel strange not to have VRegion as a subclass of VObject in the C++ binding. Plus it would give you automatic unreffing of the VipsRegion object, which (I think?) your template does not have.

I've addressed this with b3f2b59f18f20ec2b59f426bc89ba5351e5bec65.

> How about making a very small template over a VRegion object for this?

Could you elaborate on this point? I pictured the `VRegion` to be instantiated in the client's code, not in libvips.

jcupitt:
I would copy-paste the code for VImage to make VRegion, then put a simple template over the top of that to make your typed pointer. Your implementation is missing things like unref (for example). Subclassing VObject would get you that for free.

VObject only exists at compile-time, there's no run-time cost, so it would all be in the client code and not in libvips.

jcupitt:
Actually, checking again, VConnection might be an simpler model. Just copy-paste that and link it to VipsRegion instead.

Zeranoe:
Are you suggesting something like:
```
VRegion( VipsRegion *region, VSteal steal = STEAL ) : VObject( (VipsObject *) region, steal )
```

While doable, I think this reduces the usability of `VRegion` because of the requirement to call the C APIs `vips_region_new()`.

It would make sense to me to add a `new_region()` member function to `VImage`, which calls the underlying C API and the `VRegion` constructor.

jcupitt:
Yes, exactly. You could put the `vips_region_new()` in the constructor as well, so there would be no need to call it explicitly. Yes, you could put a little stub in `VImage` as well to make one.

Zeranoe:
I made some changes, let me know what you think.

jcupitt:
This feature is now in git master libvips, I'll close.

PR #1785, opened 2020-08-17
Ensure magick buffer+file checks use consistent min length guard

lovell:
The change in #1642 for magick v7 is also required in the magick v6 loader.

The sharp unit tests picked this up - I think the assertion in `GetImageMagick` has recently been backported.


jcupitt:
Great! I've added this to 8.10 as well, so it should be in 8.10.1.

Issue #1792, opened 2020-08-26
SEGV during vips_shutdown


kleisauke:
I'm currently investigating a SEGV during `vips_shutdown` after the libvips 8.10 update. I had no luck reproducing this on a test server, but managed to get a GDB backtrace from a production server.

```
(weserv:2337727): GLib-ERROR **: 00:01:04.340: file gthread-posix.c: line 1219 (g_system_thread_wait): error 'No such process' during 'pthread_join (pt->system_thread, NULL)'
2020/08/25 00:01:12 [alert] 2773464#0: worker process 2337749 exited on signal 11 (core dumped)
2020/08/25 00:01:12 [alert] 2773464#0: worker process 2337736 exited on signal 11 (core dumped)
2020/08/25 00:01:12 [alert] 2773464#0: worker process 2337710 exited on signal 11 (core dumped)
2020/08/25 00:01:12 [alert] 2773464#0: worker process 2337730 exited on signal 11 (core dumped)
2020/08/25 00:01:12 [alert] 2773464#0: worker process 2337750 exited on signal 11 (core dumped)
2020/08/25 00:01:12 [alert] 2773464#0: worker process 2337748 exited on signal 11 (core dumped)
2020/08/25 00:01:12 [alert] 2773464#0: worker process 2337727 exited on signal 5 (core dumped)
2020/08/25 00:01:12 [alert] 2773464#0: worker process 2337724 exited on signal 11 (core dumped)
2020/08/25 00:01:20 [alert] 2773464#0: worker process 2337725 exited on signal 11 (core dumped)
2020/08/25 00:01:20 [alert] 2773464#0: worker process 2337723 exited on signal 11 (core dumped)
2020/08/25 00:01:20 [alert] 2773464#0: worker process 2337726 exited on signal 11 (core dumped)
2020/08/25 00:01:20 [alert] 2773464#0: worker process 2337728 exited on signal 11 (core dumped)
```

```
Core was generated by `nginx: worker pr'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fe97d1615a2 in __deallocate_stack () at /lib64/libpthread.so.0
#1  0x00007fe97d1636e3 in __pthread_timedjoin_ex () at /lib64/libpthread.so.0
#2  0x00007fe97a5778eb in g_system_thread_wait () at /lib64/libglib-2.0.so.0
#3  0x00007fe97a5591de in g_thread_join () at /lib64/libglib-2.0.so.0
[...]

PR #1795, opened 2020-08-28
Fix a small memory leak in sinkscreen

kleisauke:
Noticed this while debugging https://github.com/libvips/libvips/issues/1792.

### DIFF
--- a/libvips/iofuncs/sinkscreen.c
+++ b/libvips/iofuncs/sinkscreen.c
@@ -451,6 +451,9 @@ vips__render_shutdown( void )
 		else
 			g_mutex_unlock( render_dirty_lock );
 	}
+
+	VIPS_FREEF( vips_g_mutex_free, render_dirty_lock );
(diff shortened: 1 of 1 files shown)


libvips/iofuncs/sinkscreen.c
```diff
@@ -451,6 +451,9 @@ vips__render_shutdown( void )
 		else
 			g_mutex_unlock( render_dirty_lock );
 	}
+
+	VIPS_FREEF( vips_g_mutex_free, render_dirty_lock );
+	vips_semaphore_destroy( &n_render_dirty_sem );
```
kleisauke, line 456: Perhaps I should move this within the above `if( render_dirty_lock )` statement?

jcupitt:
Probably -- I'll move it as part of moving the init back to first use.

jcupitt:
I moved this, and applied this patch to 8.10 as well.

Issue #1788, opened 2020-08-23
Add more parameters which are supported by libheif


adityapatadia:
This library supports AVIF encoding via libheif. That library has many useful parameters like `speed`, `threads` which are not supported by libvips. They should be added since AVIF support is going live in Chrome on 25h August and usage of that is expected to increase: https://github.com/strukturag/libheif/blob/2dff6b4282d5e03feb5bd8dea5b09710e635cf68/libheif/heif_encoder_aom.cc

jcupitt:
Hello @adityapatadia,

This sounds sensible. I'm reluctant to add /every/ parameter, since we will have to support and maintain them years into the future, and many are likely to be removed / added to / changed as AVIF evolves.

Is there a list of core, or essential parameters? We should certainly make sure we support all of them.

adityapatadia:
There is also a library `libavif` which has official support so may be that can be used to support encoding of AVIF images and libheif can be used to encode HEVC only. https://github.com/AOMediaCodec/libavif

adityapatadia:
> Is there a list of core, or essential parameters? We should certainly make sure we support all of them

From the link `heif_encoder_aom.cc` file, it seems essential parameters are `speed`, `threads`, `chroma_subsampling`, `quality` and `realtime`. They can greatly change the output type, quality, speed and are likely going to be supported forever.

adityapatadia:
Currently 3 parameters are supported: https://libvips.github.io/libvips/API/current/VipsForeignSave.html#vips-heifsave

lovell:
libheif uses a plugin-based approach for its encoders/decoders so there are already a number of optional string/integer/boolean parameters (using libheif's own type system) for x265, de265, aom and rav1e, plus soon will be further optional parameters for dav1d and svt-av1.

The current (exhaustive) list of encoder parameters in svt-av1, which I suspect may be become the defacto AVIF encoder on x64, can be seen at https://github.com/OpenVisualCloud/SVT-AV1/blob/master/Docs/svt-av1_encoder_user_guide.md#list-of-all-configuration-parameters

I would agree that "speed" and "chroma" are good candidates to add as they have equivalents in the WebP and JPEG savers respectively. The "realtime" setting appears to produce file sizes much larger than the JPEG equivalent, so would question its value.

One thing to note however is that "speed" is not entirely straightforward as each libheif encoder plugin accepts different values; aom is 0-8 whereas rav1e is 0-10 whereas x265 is "superfast" to "veryslow".

[...]

PR #1819, opened 2020-09-16
libheif: expose speed parameter

lovell:
This parameter is currently only used for AV1 compression.

By limiting to a 0-8 range we can provide a consistent API for both the aom and rav1e encoders. (The rav1e encoder accepts speed values of 9 and 10 but these use 64x64 blocks and fast key frame detection, which are concepts more relevant to low quality video than still images.)

As discussed in #1788

(The SVT-AV1 encoder, not yet available via libheif, also uses the same 0 slowest to 8 fastest range for its `EncoderMode` speed setting so this approach is hopefully future proof for that too.)

### DIFF
--- a/libvips/foreign/heifsave.c
+++ b/libvips/foreign/heifsave.c
@@ -79,6 +79,10 @@ typedef struct _VipsForeignSaveHeif {
 	 */
 	VipsForeignHeifCompression compression;
 
+	/* CPU effort (0-8).
+	 */
+	int speed;
+
 	/* The image we save. This is a copy of save->ready since we need to
 	 * be able to update the metadata.
 	 */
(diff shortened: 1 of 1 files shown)


jcupitt:
Looks good!

lovell:
@jcupitt Would you consider backporting this to the 8.10(.2) branch? I'm planning to add libheif+aom to the binaries provided by sharp and exposing control over the speed and therefore CPU time might help reduce the planet's energy consumption by a noticeable amount as everyone rushes to re-encode all their images as AVIF.

jcupitt:
It's usually bad to add API to stable, but I agree this seems like a special case. Let's do it!

kleisauke:
For reference, libavif takes a different approach and remaps the speed values 0-10 to CPU used (0-8) and aom usage (good quality / real-time mode), see:
https://github.com/AOMediaCodec/libavif/blob/3f2bf9512057e33e0746619b58f8cefb3ff5d1a4/src/codec_aom.c#L355-L377

PR #1828, opened 2020-09-22
heifload: verify ISO BMFF/3GPP2 signature in is_a check

lovell:
Ensures the first 4 bytes contain one of these two standard signatures, which (hopefully) all valid HEIF/HEIC/AVIF files will start with.


jcupitt:
Nice! I cherry-picked to 8.10 as well.

PR #1831, opened 2020-09-23
libheif: simplify is_a check of first 4 bytes

lovell:
I found a valid example image with 0,0,0,28 as the first 4 bytes that was failing the new check :(

This PR is a clean-up of #1828 to convert the first 4 bytes to a 32-bit unsigned integer and then verify it is a multiple of 4, up to 32. Apologies for not doing it this way the first time around.

### DIFF
--- a/libvips/foreign/heifload.c
+++ b/libvips/foreign/heifload.c
@@ -260,14 +260,11 @@ static const char *heif_magic[] = {
 static int
 vips_foreign_load_heif_is_a( const char *buf, int len )
 {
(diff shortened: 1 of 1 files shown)


jcupitt:
Is there a potential byte order issue here? You might need something like:

https://developer.gnome.org/glib/stable/glib-Byte-Order-Macros.html#GUINT-FROM-BE:CAPS

lovell:
Yes, thank you, updated accordingly.

PR #1832, opened 2020-09-25
Fix test failure on ARM-based Windows

kleisauke:
The optional parameters of `vips_gaussnoise` were incorrectly
passed within `vips_fractsurf`. This was discovered when running
the libvips testsuite on Windows 10 IoT (ARM32).


jcupitt:
Oh, funny! An ancient typo. I'll patch 8.10 as well.

PR #1833, opened 2020-09-25
heifload: prevent read beyond end of source buffer

lovell:


lovell:
Thanks, should we patch this on the 8.10 branch?

jcupitt:
I've revised the patch a bit -- thinking about it, the underlying problem is a mismatch between the way libheif handles EOF and the way that VipsSource does it.

VIpsSource uses the unix model where read() signals EOF. libheif has a separate wait_for_file_size() call (used before it triggers the read()) which is supposed to check that the read will be OK (ie. there will be enough bytes available). We were testing for EOF in the libheif read handler and then signalling in wait_for_file_size(), but that's the wrong way around -- wait_for_file_size() is called *before* read, so the EOF we spot in read() can't be fed back in time.

So a better fix is to test for EOF directly in wait_for_file_size() and remove the EOF stuff from the read handler. This also removes the `->source->length` reference, which looked a bit unsafe. On the downside, it's likely to cause more buffering, sadly.

(the unix model is much better since read and EOF happen in the same place and you can't get races, but that's another argument, of course)

jcupitt:
I patched 8.10 and master.

lovell:
Your approach is much neater, thank you.

jcupitt:
Oh no problem, thanks for fixing this!

There seems to be a lot of interest in AVIF. That $2,000 is very generous.

I tried libvips smartcrop on that mcconnell-obama test image, and it picks mcconnell every time :( It actually find the skin correctly, but mcconnell's face is larger (has more pixels of skin), so it prefers him. The red tie is very saturated, and that helps too.

I'll try to tune it a little. It might make an interesting blog post.

PR #1845, opened 2020-10-04
Adapt test suite for AVIF support

kleisauke:
This PR adapts the test suite for AVIF images allowing it to be run with the prebuilt binaries provided by sharp and NetVips in the near future. See https://github.com/lovell/sharp/issues/2289 for more context.

Here is a brief overview of the related changes within this PR:
- `heic-orientation-6.heic` was converted to AVIF using the following command:
  ```bash
  $ vips copy heic-orientation-6.heic avif-orientation-6.avif[speed=0,compression=av1]
  ````
- One TODO item has been added that is related to the above command. We currently need to hint the compression to `av1` when we save to `*.avif`, perhaps this could be done automatically?
- I disabled the lossless mode test in `test_heifsave` because it takes about ~45 seconds on my PC.
- The `README.md` file has been updated to indicate that we support AVIF images.
- The EXIF test in `test_heifsave` has been uncommented.
- Some typo fixes and other minor improvements.

Marked this PR as draft, as it needs libheif >= v1.8.0 built against aom (or rav1e/dav1d) which is not generally available in the major distros (e.g. Homebrew currently lacks support for AVIF images in libheif, causing a failure in the macOS Travis build).


jcupitt:
This is great! Lovell and I have just back-ported the "speed" param to 8.10, fwiw.

kleisauke:
Great! I plan to make a PR at Homebrew to build libheif against aom, but I'm not sure if Homebrew prefers the reference implementation over the speedy ones (i.e. rav1e and dav1d).

@lovell As Firefox has recently switched from aom to dav1d as default AV1 decoder, do you think it might be worth to experiment with dav1d within the pre-built binaries?

lovell:
@kleisauke We should probably go with either the all-in-one aom or a dav1d+rav1e combo. IIRC the latter doesn't yet support images smaller than 16x16.

kleisauke:
The only thing that could possibly prevent us from switching to the dav1d+rav1e combo is that only dav1d is continuously fuzzed by OSS-Fuzz.

kleisauke:
macOS CI failures should be gone after PR https://github.com/Homebrew/homebrew-core/pull/68275 has landed.

hjmallon:
It has landed. 🤞

kleisauke:
Great! It passes CI with commit https://github.com/libvips/libvips/pull/1845/commits/fdf9d0c6b0da33bf77ab74d7e7c2493bdc84c205 and https://github.com/libvips/libvips/pull/1845/commits/15e05675fbcc6449c3bcb7869666e8913127e77c, so this is ready for review now.

(unfortunately doing `brew update` within Travis increases the build time, hopefully this will be resolved when we move the macOS test to GitHub Actions)

jcupitt:
I can't see any problems -- let's do it!

PR #1850, opened 2020-10-09
heifload: expose heif-compression metadata

lovell:
Requires libheif 1.4.0 for `heif_main_brand` but versions prior to that didn't support AV1. The approach in this PR should hopefully provide useful information for newer versions whilst degrading gracefully for older versions.

```sh
$ vipsheader -a test.heic | grep heif
vips-loader: heifload
heif-primary: 0
heif-compression: hevc
```
```sh
$ vipsheader -a test.avif | grep heif
vips-loader: heifload
heif-primary: 0
heif-compression: av1
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -888,6 +888,18 @@ if test x"$with_heif" = x"yes"; then
   LIBS="$save_LIBS"
 fi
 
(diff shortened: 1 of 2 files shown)


jcupitt:
Useful!

PR #1852, opened 2020-10-10
Speed up VIPS_ARGUMENT_COLLECT_SET

kleisauke:
By using [`G_VALUE_COLLECT_INIT`](https://developer.gnome.org/gobject/stable/gobject-Varargs-Value-Collection.html#G-VALUE-COLLECT-INIT:CAPS), according to [this upstream bug report](https://bugzilla.gnome.org/show_bug.cgi?id=603590) this could give a speed-up varying from 15% to 30% (measured with Callgrind).


jcupitt:
Oh, nice. And what a simple change.

PR #1857, opened 2020-10-13
Separate lock for PDFium

DarthSim:
libvips uses the global lock for PDFium operations. This causes many other operations to freeze until PDFium stuff is done. For example, imgproxy Pro gets completely stuck when tries to load some heavy PDF (https://github.com/imgproxy/imgproxy/issues/484). Looks like it makes sense to use a separate mutex for PDFium since its main purpose is to not allow parallel usage of PDFium.

I'm not sure if I chose the right place to init the mutex.

### DIFF
--- a/libvips/foreign/pdfiumload.c
+++ b/libvips/foreign/pdfiumload.c
@@ -128,6 +128,8 @@ static char *vips_pdfium_errors[] = {
 	"page not found or content error"
 };
 
+static GMutex *vips_pdfium_mutex = NULL;
+
 static void
 vips_pdfium_error( void )
 {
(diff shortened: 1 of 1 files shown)


jcupitt:
Looks great! You're right, using the global lock there was not a good idea.

jcupitt:
I moved the mutex init inside the ONCE used for pdfium init and credited you on the ChangeLog.

kleisauke:
Do we also need to add `VIPS_FREEF( vips_g_mutex_free, vips_pdfium_mutex );` in `vips_foreign_load_pdf_dispose`?

jcupitt:
PDFium is very, very non-thread safe -- you can only call into the library once and only from one thread, so it must have a global lock. This patch just swaps the vips global lock for a global lock for pdfium. PDFium is never un-inited, so the lock never needs freeing.

It's because PDFium is derived from the FoxIt PDF reader, a Windows desktop program, and was never intended to be used as a library. That you can use it as a library at all is pretty amazing.

DarthSim:
> I moved the mutex init inside the ONCE used for pdfium init

This is definitely a better place. Thank you!

PR #1872, opened 2020-11-02
Resolve rounding issues in reduce

kleisauke:
This PR is a split-off from https://github.com/libvips/libvips/pull/1769, targeting the [`8.10`](https://github.com/libvips/libvips/tree/8.10) branch.

It fixes two small bugs that have been found with [jsummers/resamplescope](https://github.com/jsummers/resamplescope). The first commit (https://github.com/kleisauke/libvips/commit/49d8051e22d1b06f64a7f85b2c70db2497cece54) restores the round-to-nearest behaviour in `reduce{h,v}`, that was removed with PR https://github.com/libvips/libvips/pull/1592. This change can be observed in the following graphs:

| Before | After |
| :---: |  :---: |
|  |  |
| Rounding error | |

The second commit (https://github.com/kleisauke/libvips/commit/fdc140e8e90ec42e1655135fadaf1479bd2c852a) rounds the sum values to the nearest integer in `*_notab` version. I'm not entirely sure why this is necessary and whether this is the right place to do it, but it seems to fix the double-precision graph:
| Before | After |
| :---: |  :---: |
|  |  |
| Rounding / offset error | |

The last commit (https://github.com/kleisauke/libvips/commit/4533375f63a94f265feef3bc1fc4365b1b7d66f3) rounds the `x` and `y` values of the `vips_embed` call up and take this offset into account (i.e. by subtracting it from `hoffset` / `voffset`). Somehow I cannot illustrate this change within the graphs, but can confirm that this fixes the border-issues reported on sharp (https://github.com/lovell/sharp/issues/2408 and https://github.com/lovell/sharp/issues/2411). It also does not cause regressions on the [pyvips-issue-148](https://github.com/kleisauke/pyvips-issue-148) and [vips-issue-703](https://github.com/kleisauke/vips-issue-703) test repositories.

### DIFF
--- a/libvips/resample/bicubic.cpp
+++ b/libvips/resample/bicubic.cpp
@@ -79,15 +79,16 @@ typedef VipsInterpolate VipsInterpolateBicubic;
 typedef VipsInterpolateClass VipsInterpolateBicubicClass;
 
 /* Precalculated interpolation matrices. int (used for pel
- * sizes up to short), and double (for all others).
+ * sizes up to short), and double (for all others). We go to
+ * scale + 1 so we can round-to-nearest safely.
  */
 
 /* We could keep a large set of 2d 4x4 matricies, but this actually
  * works out slower since for many resizes the thing will no longer
  * fit in L1.
  */
-static int vips_bicubic_matrixi[VIPS_TRANSFORM_SCALE][4];
-static double vips_bicubic_matrixf[VIPS_TRANSFORM_SCALE][4];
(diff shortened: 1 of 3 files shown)


lovell:
Thanks Kleis, this change looks good to me.

jcupitt:
Looks good here too, nice work Kleis.

Sorry for being so slow on this. I finished my huge three-year project last night! \o/ so I should finally have some free time again.

PR #1880, opened 2020-11-09
Update Examples.md

amalbala:
Just some issues I found while testing the examples:

- Reference on header for the file try255.py
- Typo on parameter from bigtif to bigtiff
- Use explicit call to python interpreter from command line


lovell (approved):
Thank you!

jcupitt:
Nice! I'll change the `/usr/bin/env` too.

PR #1892, opened 2020-11-20
gifload: ensure total height of all pages is sanitised

lovell:
Helps prevent an integer overflow, plus adds a >60k page GIF sample to aid future fuzz testing.

We should probably backport this to 8.10 also.


libvips/foreign/gifload.c
```diff
@@ -700,8 +700,12 @@ vips_foreign_load_gif_scan_extension( VipsForeignLoadGif *gif )
 static int
 vips_foreign_load_gif_set_header( VipsForeignLoadGif *gif, VipsImage *image )
 {
+	const int total_height = gif->file->SHeight * gif->n;
+	if ( total_height <= 0 || total_height > VIPS_MAX_COORD ) {
```
jcupitt, line 704: You could probably still get int overflow here. How about doing the multiplication as gint64? That would be safe.

jcupitt: You need to cast one of the args to the * as well, or the mult will be done as int :(

lovell: Thanks, I've updated this, plus added an appropriate message to the error log.

jcupitt:
Great!

jcupitt:
I backported to 8.10 too.

Issue #1847, opened 2020-10-07
Big endian test failure (ruby-vips)


LocutusOfBorg:
Hello, this is not a big issue, because the test always failed there, so the Debian/Ubuntu autopkgtest infrastructure is not showing it as "regressed"

However I would like to point this possible issue in the code (or in the test code)
```
https://travis-ci.org/github/kleisauke/ruby-vips/jobs/733637121

  1) Vips::Image can load an image from a file

     Failure/Error: expect(x.avg).to eq(128)

       expected: 128

            got: 9.388699710976274e-44

       (compared using ==)

     # ./spec/image_spec.rb:27:in `block (2 levels) in <top (required)>'
```

kleisauke:
fwiw, the pyvips test suite runs succefully on IBM Z (s390x), see:
https://github.com/kleisauke/libvips/commit/673ebdd374e5ebb1ce7ce02198340427da51b906
https://travis-ci.org/github/kleisauke/libvips/jobs/734248676

Although, it fails on this test during `make check`:
https://github.com/libvips/libvips/blob/29b6ba12947a179c650f318354ffd1f57903097d/test/test_formats.sh#L149
```
testing sample.jpg tif[compression=jpeg] ... save / load difference is 263.000000
```

I'm going to try to port that Ruby test case to Python, to see if I can reproduce it on pyvips as well.

kleisauke:
I could also reproduce this on Python with libvips master, see commit https://github.com/kleisauke/libvips/commit/9c2eb79b4e62699b88e89cbb14c0d00e2576c8b3 and [Travis build](https://travis-ci.org/github/kleisauke/libvips/jobs/734262083). Here's a reduced test case: https://github.com/kleisauke/libvips/commit/66b6b2cd8f35c6b6f7dc9cb3b02c43cbf428e494.

kleisauke:
Here's a (possible) patch that seems to work for me: https://github.com/kleisauke/libvips/commit/0690bfea69bd8127cdb527dd324369a5d6e45ce3. The successful Travis build on s390x can be viewed here: https://travis-ci.org/github/kleisauke/libvips/jobs/734595838.

kleisauke:
@jcupitt Do you know if we could safely move the endianness check to compilation time? See for example: https://github.com/kleisauke/libvips/commit/52e77fbe9ad3b4e1c8acb72303a7055b3c2e50e5.

kleisauke:
[...]

PR #1899, opened 2020-11-23
Ensure vipsload only byte swaps if necessary

kleisauke:
MSB-ordered vips images were always byte swapped on both little- and big endian systems. And LSB-ordered vips images were loaded without a byte swap. This works correctly on little endian systems, but will not work on big endian systems where the byte swap must be done vice versa.

This PR ensures that the byte swap only takes place when needed. I also simplified the MSB-ordered image check (by using a helper) and moved the endianness check to compile time.

Resolves https://github.com/libvips/libvips/issues/1847.

(This PR targets the 8.10 branch, but I can change it to the master branch if necessary)

### DIFF
--- a/libvips/iofuncs/image.c
+++ b/libvips/iofuncs/image.c
@@ -859,10 +859,7 @@ vips_image_build( VipsObject *object )
 		if( (magic = vips__file_magic( filename )) ) {
 			/* We may need to byteswap.
 			 */
-			guint32 us = vips_amiMSBfirst() ? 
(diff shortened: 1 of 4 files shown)


jcupitt:
Sure, let's do this to 8.10, it's a bug-fix. Thanks for fixing up this mess!

(this is a very embarassing bug, but in my defence the old byte order code predated libvips using glib, which is why it's terrible,  and hadn't been tested in more than 20 years)

PR #1918, opened 2020-12-04
webpload: ensure first frame is not blended

lovell:
A possible fix for https://github.com/lovell/sharp/issues/2341#issuecomment-738726237


jcupitt:
Looks good! Is a first frame with mode BLEND out of spec? I agree it makes sense to treat it as COPY though.

jcupitt:
Thinking again, perhaps the problem is that `blend_pixel` in webp2vips is doing round down, not round to nearest? If I add rounding, the difference also vanishes.

Let's keep both changes: rounding in blend, and your don't-blend-on-first-frame, since it now becomes an optimization.

PR #1921, opened 2020-12-07
fix heif load fails with 0 length metadata

### LINKED ISSUE #1901: Libvips fails to load the file attached
As reported in #1888.

I'm getting a: `source: bad seek to 1272289` when loading the following heic file:

[IMG_2631.zip](https://github.com/libvips/libvips/files/5593259/IMG_2631.zip)

It's from "the wild" so I have no clue how it was created/saved.

I'm using a custom build with:

libde265 1.0.8
libheif 1.9.1
libvips 8.10.2
x265 3.4

ddennedy-gpsw:
fixes #1901

### DIFF
--- a/libvips/foreign/heifload.c
+++ b/libvips/foreign/heifload.c
@@ -468,6 +468,8 @@ vips_foreign_load_heif_set_header( VipsForeignLoadHeif *heif, VipsImage *out )
 		printf( "metadata type = %s, length = %zu\n", type, length ); 
 #endif /*DEBUG*/
 
+		if( !length )
+			continue;
 		if( !(data = VIPS_ARRAY( out, length, unsigned char )) )
 			return( -1 );
 		error = heif_image_handle_get_metadata( 


libvips/foreign/heifload.c
```diff
@@ -468,6 +468,8 @@ vips_foreign_load_heif_set_header( VipsForeignLoadHeif *heif, VipsImage *out )
 		printf( "metadata type = %s, length = %zu\n", type, length ); 
 #endif /*DEBUG*/
 
+		if( !length )
+			continue;
```
lovell, line 472: Is it worth adding a warning here?
```c
g_warning( "heifload: ignoring empty %s metadata", type );
```

lovell (commented):
Excellent work tracking this down, thank you. I've left one question inline.

It's probably also worth implementing @jcupitt's suggestion to silently ignore bad seeks to avoid writing an error to the log that we are otherwise ignoring, but that can be a separate commit/PR.

ddennedy-gpsw:
> It's probably also worth implementing @jcupitt's suggestion to silently ignore bad seeks to avoid writing an error to the log

I do not read that as a suggested change but as simply something he tried to make it succeed. This error is logged in `vips_source_seek()` which is low level and common. Bad seeks should be logged here. It could be argued that it should be logged at warning level, but it is returning -1; so it is not clear.

> Is it worth adding a warning here?

`g_info()` seems more appropriate here; please let me know which is preferred.

jcupitt:
This is great @ddennedy-gpsw, thank you for nailing this thing!

I think `g_warning()` would be better than `g_info()`, but I wonder if a warning is necessary? They are usually for serious and probably unexpected behaviour, and I doubt if anyone would expect empty metadata to load.

jcupitt:
Therefore (I meant to say) I think it looks good to commit.

ddennedy-gpsw:
Please merge this.

jcupitt:
I added a note to the changelog too.

kleisauke:
This PR should probably be cherry-picked in the 8.10 branch.

jcupitt:
Ooop! Good point.

Issue #1808, opened 2020-09-06
heifload ignores clean aperture information


kleisauke:
This issue is similar to #1574, but found a way to generate a HEIC image with a non-faulty `clap` box. This image works fine with `heif-convert` but is not processable with libvips.

Example image: https://t0.nl/x.heic

Generated with (with libheif v1.8.0 and libvips v8.10)
```bash
$ wget https://images.weserv.nl/zebra.jpg
$ vips copy zebra.jpg x.heic
```

Reproduction
```bash
$ wget https://t0.nl/x.heic
$ vips copy x.heic x.jpg

(vips:38053): VIPS-WARNING **: 10:49:00.847: heifload: ignoring nclx profile
...
(vips:38053): VIPS-WARNING **: 10:49:01.383: error in tile 0 x 0
...
heifload: bad image dimensions on decode

$ heif-convert x.heic x.jpg
File contains 1 images
Written to x.jpg
```

Image info
```bash
$ vipsheader x.heic

(vipsheader:36955): VIPS-WARNING **: 10:45:23.958: heifload: ignoring nclx profile
x.heic: 4120x2746 uchar, 3 bands, srgb, heifload

$ heif-info x.heic -d
...
Box: ispe -----
size: 20   (header size: 12)
version: 0
flags: 0
image width: 4120
image height: 2747

Box: clap -----
size: 40   (header size: 8)
clean_aperture: 4120/1 x 2747/1
offset: 0/2 ; -1/2
```

Note that the above `ispe` box contains an image height of 2747 pixels which is cropped to 2746 pixels with the `clap` box info. This is therefore a valid HEIC image.

jcupitt:
This seems to be working for me on ubuntu 20.04 with libvips 8.10 and the system-installed libheif, x265 etc.:

```
$ wget https://images.weserv.nl/zebra.jpg
--2020-09-06 13:25:04--  https://images.weserv.nl/zebra.jpg
Resolving images.weserv.nl (images.weserv.nl)... 104.27.158.82, 172.67.143.81, 104.27.159.82, ...
Connecting to images.weserv.nl (images.weserv.nl)|104.27.158.82|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 1854461 (1.8M) [image/jpeg]
Saving to: ‘zebra.jpg’

zebra.jpg           100%[===================>]   1.77M  2.00MB/s    in 0.9s    

[...]

PR #1935, opened 2020-12-17
webpload: prevent divide-by-zero when blending pixels

lovell:
Adds a test case to prevent future regression.

As discussed in commit 6eaf1ed


jcupitt:
Yes, good idea to have a test. This is my fault, sorry :( I'll start preparing 8.10.5. 

@kleisauke, any thoughts for other urgent stuff we should include?

kleisauke:
Not that I know, and sorry for not having caught this earlier.

jcupitt:
OK, we have https://github.com/libvips/libvips/releases/tag/v8.10.5

I'll do some win builds and then remove the pre-release tag.

kleisauke:
(Less urgent, but nice to have, is testing https://github.com/libvips/libvips/issues/1808#issuecomment-745392356. However, such AVIF images [cannot be loaded in Chrome](https://github.com/chromium/chromium/blob/c0fe1d638d94c08b80e7033ede733e63a30bbf0d/third_party/blink/renderer/platform/image-decoders/avif/avif_image_decoder.cc#L571-L577) either, so it does not block AVIF support in the prebuilt binaries, I think)

jcupitt:
Having to decode the image just to get the header fields feels very wrong :( Perhaps there's some other way we could fix the clap box issue?

kleisauke:
NetVips unit tests pass on Windows with the [MXE-based v8.10.5 builds](https://github.com/libvips/build-win64-mxe/releases/tag/v8.10.5). I still need to check Linux and macOS, but I don't think this will bring any regression.

@lovell The [`-build2` suffix](https://github.com/lovell/sharp-libvips/blob/1fdeeb752ab37bbf1bbc3d49e38689d6ed7c476f/build/win.sh#L26) needs to be removed from sharp-libvips Windows build script in order to pass CI.

PR #1938, opened 2020-12-22
Avoid seeking on bad file descriptors

kleisauke:
See https://github.com/lovell/sharp/issues/2494 for context.

### DIFF
--- a/libvips/iofuncs/source.c
+++ b/libvips/iofuncs/source.c
@@ -318,7 +318,9 @@ vips_source_build( VipsObject *object )
 	if( vips_object_argument_isset( object, "blob" ) ) {
 		size_t length;
 
-		source->data = vips_blob_get( source->blob, &length );
+		if( !(source->data = vips_blob_get( source->blob, &length )) )
+			return( -1 );
+
 		source->length = VIPS_MIN( length, G_MAXSSIZE );
 	}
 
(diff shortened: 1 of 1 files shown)


libvips/iofuncs/source.c
```diff
@@ -153,7 +153,8 @@ vips_source_test_features( VipsSource *source )
 	 * vips_source_seek() etc. or we might trigger seek emulation.
 	 */
 	if( source->data ||
-		class->seek( source, 0, SEEK_CUR ) != -1 ) { 
+		(VIPS_CONNECTION( source )->descriptor != -1 &&
```
jcupitt, line 156: I'm not sure here -- the subclass might not be using `descriptor`.

How about adding this test to `vips_source_seek_real()`, the default implementation? It could be eg.:

```C
    if( connection->descriptor != -1 )
        return( vips__seek_no_error( connection->descriptor, offset, whence ) );

    return( -1 );
```

kleisauke: You're right, I forgot that for a moment. Resolved with https://github.com/kleisauke/libvips/commit/e437805fc19cdf9af89685576a08b3a502a2954d.

Do you think if `vips_source_read_real` needs something similar? I had a look, but I think there is always a file descriptor present when that function is called.

jcupitt:
Great! Funny how easy it is to miss stuff like this.

kleisauke:
Indeed, I wonder if this would ever be discovered if it didn't cause a runtime error.

For further readers; note that the severity of this issue is low, as the invalid parameter error only seems to happen on Windows when linking against UCRT, which our Windows 32- and 64-bit binaries does not link against.

libvips/iofuncs/source.c
jcupitt: Ooooof, I'm not sure. I *think* it's OK.

PR #1947, opened 2020-12-29
Switch VIPS_FCLIP to VIPS_CLIP in vips_scRGB2{BW,sRGB}

kleisauke:
As discussed in https://github.com/libvips/libvips/pull/1735#issuecomment-662439737. I didn't actually benchmark this, but it seems like a straightforward change (hence targeting the [`8.10`](https://github.com/libvips/libvips/tree/8.10) branch).


jcupitt:
Sure, why not. Though of course this is only the alpha, so I don't suppose it'll make much difference.

PR #1948, opened 2020-12-29
Fix UBSan errors

kleisauke:
As discussed in https://github.com/libvips/libvips/pull/1727#issuecomment-660932493. Commit 81132311674ea54061bb5c2e36db9a2fe8fee19d fixes this error<sup>1</sup>:
```
Running: common_fuzzer_corpus/smartcrop_fuzzer-5687924892368896
mosaic_fuzzer.cc:47:56: runtime error: member access within misaligned address 0x60b0000004c3 for type 'struct mosaic_opt', which requires 2 byte alignment
```
https://travis-ci.org/github/kleisauke/libvips/jobs/709752013#L2599

Commit 1dd6698464584893e0da5e83378799a8b3f58caf fixes this error:
```
Running: common_fuzzer_corpus/sharpen_fuzzer-5678720198639616
scRGB2sRGB.c:114:11: runtime error: 8224 is outside the range of representable values of type 'unsigned char'
```
https://travis-ci.org/github/kleisauke/libvips/jobs/709752013#L2315

Commit 1a6854a6d3678b033c37bfb2b60db876e7022d4b fixes this error<sup>2</sup>: 
```
Running: common_fuzzer_corpus/jpegsave_buffer_fuzzer-5658586599915520
flatten.c:190:4: runtime error: 1.64485e+07 is outside the range of representable values of type 'unsigned short'
```
https://travis-ci.org/github/kleisauke/libvips/jobs/709752013#L2270

And the last commit (b8f18e119cc4b85c2b7749ae9aaa8a36cc457f44) fixes this error: 
```
Running: common_fuzzer_corpus/jpegsave_buffer_fuzzer-5658586599915520
heifload.c:282:27: runtime error: load of misaligned address 0x60f000000229 for type 'guint32' (aka 'unsigned int'), which requires 4 byte alignment
```
https://github.com/libvips/libvips/runs/3554490177#step:13:366

Marked this PR as draft due to these endnotes:
- [x] <sup>1</sup>: but also changes the way how `vips_mosaic` is fuzzed (i.e. by separating the fuzzer parameters from the image), so perhaps we need to regenerate the existing corpus?
- [x] <sup>2</sup>: this was copied-pasted from `vips_cast`, I'm not sure if there's a better way to fix this?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,7 +110,7 @@ jobs:
           echo "ASAN_OPTIONS=suppressions=${{ github.workspace }}/suppressions/asan.supp" >> $GITHUB_ENV
           echo "LSAN_OPTIONS=suppressions=${{ github.workspace }}/suppressions/lsan.supp" >> $GITHUB_ENV
           echo "TSAN_OPTIONS=suppressions=${{ github.workspace }}/suppressions/tsan.supp" >> $GITHUB_ENV
-          echo "UBSAN_OPTIONS=suppressions=${{ github.workspace }}/suppressions/ubsan.supp:print_stacktrace=1" >> $GITHUB_ENV
+          echo "UBSAN_OPTIONS=suppressions=${{ github.workspace }}/suppressions/ubsan.supp:halt_on_error=1:abort_on_error=1" >> $GITHUB_ENV
           echo "LD_LIBRARY_PATH=$LLVM_PREFIX/lib:`dirname $ASAN_DSO`" >> $GITHUB_ENV
           echo "$LLVM_PREFIX/bin" >> $GITHUB_PATH
           # workaround for https://github.com/google/sanitizers/issues/89
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,7 +128,7 @@ jobs:
         run: make V=0 -j$JOBS
 
       - name: Check libvips
-        run: make V=0 check || (cat test/test-suite.log && exit 1)
+        run: make V=0 VERBOSE=1 check
 
       - name: Install libvips
         run: sudo make V=0 install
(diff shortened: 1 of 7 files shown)


kleisauke:
I think the above endnotes have been resolved.
- The existing corpus (in the [`fuzz/common_fuzzer_corpus`](https://github.com/libvips/libvips/tree/master/fuzz/common_fuzzer_corpus) directory) doesn't needs to be regenerated, because it does not contain any corpora for the mosaic fuzzer.
- I don't think there's a better way to resolve this.

This is ready for review now.

jcupitt:
This looks great Kleis. 

The flatten change is a shame -- the casts will be quite a bit slower.

The overflows only happen if there is a mismatch between the expected range of the alpha (0-255, or 0-65535) and the available range of the numeric type, which will be a rare case, I think. For example, if the interpretation is srgb (0-255 alpha), but it's a ushort image (pixels can have much higher values).

How about detecting mismatch, and if that happens, flattening to float and then calling vips_clip() to go back to the original type? It should fix the error, and we won't lose speed, except in these rare cases.

kleisauke:
Oh huh, I could no longer reproduce the flatten change locally with clang 12. But CI doesn't seem to pass when reverting this.

jcupitt:
OK, shall we merge? I could make another PR with a possible flatten fix after this, if you like.

kleisauke:
Feel free to merge, another PR to fix that error in flatten sounds good to me. FWIW, I'm testing locally with clang 12.0.1. CI seems to be testing with version 12.0.0.

jcupitt:
OK, let's do it. Well done for crunching through the fuzzer results.

jcupitt:
I made an issue with a reproducer for the flatten error. I'll have a go at a PR.

PR #1961, opened 2021-01-12
heifsave: add option to control chroma subsampling

lovell:
As discussed in #1788, providing control over the use of chroma subsampling is a genuinely useful feature for still images created via libheif and is consistently supported by all of its encoder plugins.

This PR also adds a test for this now that the CI environment supports the AVIF format.

### DIFF
--- a/cplusplus/include/vips/VImage8.h
+++ b/cplusplus/include/vips/VImage8.h
@@ -2977,6 +2977,8 @@ static VImage heifload_source( VSource source, VOption *options = 0 );
  *   - **Q** -- Q factor, int.
  *   - **lossless** -- Enable lossless compression, bool.
  *   - **compression** -- Compression format, VipsForeignHeifCompression.
+ *   - **speed**: -- CPU effort, 0 slowest - 8 fastest, AV1 compression only, int.
+ *   - **subsample_mode** -- Select chroma subsample operation mode, VipsForeignSubsample.
  *   - **strip** -- Strip all metadata from image, bool.
  *   - **background** -- Background value, std::vector<double>.
  *   - **page_height** -- Set page height for multipage save, int.
@@ -2993,6 +2995,8 @@ void heifsave( const char *filename, VOption *options = 0 ) const;
  *   - **Q** -- Q factor, int.
  *   - **lossless** -- Enable lossless compression, bool.
  *   - **compression** -- Compression format, VipsForeignHeifCompression.
+ *   - **speed**: -- CPU effort, 0 slowest - 8 fastest, AV1 compression only, int.
+ *   - **subsample_mode** -- Select chroma subsample operation mode, VipsForeignSubsample.
  *   - **strip** -- Strip all metadata from image, bool.
  *   - **background** -- Background value, std::vector<double>.
  *   - **page_height** -- Set page height for multipage save, int.
@@ -3009,6 +3013,8 @@ VipsBlob *heifsave_buffer( VOption *options = 0 ) const;
  *   - **Q** -- Q factor, int.
  *   - **lossless** -- Enable lossless compression, bool.
  *   - **compression** -- Compression format, VipsForeignHeifCompression.
+ *   - **speed**: -- CPU effort, 0 slowest - 8 fastest, AV1 compression only, int.
+ *   - **subsample_mode** -- Select chroma subsample operation mode, VipsForeignSubsample.
  *   - **strip** -- Strip all metadata from image, bool.
  *   - **background** -- Background value, std::vector<double>.
  *   - **page_height** -- Set page height for multipage save, int.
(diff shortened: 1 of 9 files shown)


jcupitt:
I think the last component of enum values can't be a number or you get problems in some bindings, so it would need to be something like `_s420`. This is why angles like `VIPS_ANGLE_D90` have the `D` there.

Do we need such precise control? How about reusing `VipsForeignJpegSubsample` (on, off, auto) rather than making a new enum? We could call it `VipsForeignSubsample` and have a compat typedef for `VipsForeignJpegSubsample`.

lovell:
Yep, makes sense, I've no idea how much if any need there will be for 4:2:2, and can imagine most systems will be optimised for 4:2:0 or 4:4:4.

Do we want the same "quality" > 90 condition to apply for "auto"? In my not-very-scientific testing, HEIC and AVIF generally produce better "quality" than JPEG at a lower "quality" value, if that makes sense (they generally make better decisions about what to throw away and/or better decisions about what can be predicted).

cplusplus/include/vips/VImage8.h
```diff
@@ -2977,9 +2977,9 @@ static VImage heifload_source( VSource source, VOption *options = 0 );
  *   - **Q** -- Q factor, int.
  *   - **lossless** -- Enable lossless compression, bool.
  *   - **compression** -- Compression format, VipsForeignHeifCompression.
+ *   - **speed**: -- CPU effort, 0 slowest - 8 fastest, AV1 compression only, int.
+ *   - **subsample_mode** -- Select chroma subsample operation mode, VipsForeignSubsample.
  *   - **strip** -- Strip all metadata from image, bool.
- *   - **background** -- Background value, std::vector<double>.
```
lovell, line 2981: I think `background` and `page_height` might have been copypasta as they are not part of heifsave.

libvips/foreign/heifsave.c
```diff
@@ -363,6 +368,17 @@ vips_foreign_save_heif_build( VipsObject *object )
 		return( -1 );
 	}
 
+	chroma = heif->subsample_mode == VIPS_FOREIGN_SUBSAMPLE_OFF ||
+		( heif->subsample_mode == VIPS_FOREIGN_SUBSAMPLE_AUTO &&
+			heif->Q > 90 ) ? "444" : "420";
```
lovell, line 373: Have used same Q>90 condition as JPEG to prevent subsampling, easy to change this if required.

libvips/foreign/heifsave.c
```diff
@@ -496,6 +519,7 @@ vips_foreign_save_heif_init( VipsForeignSaveHeif *heif )
 	heif->Q = 50;
 	heif->compression = VIPS_FOREIGN_HEIF_COMPRESSION_HEVC;
 	heif->speed = 5;
+	heif->subsample_mode = VIPS_FOREIGN_SUBSAMPLE_AUTO;
```
lovell, line 522: This is a breaking change as the current default is to always use 4:2:0 subsampling.

libvips/include/vips/foreign.h
```diff
@@ -371,19 +371,21 @@ int vips_openslideload_source( VipsSource *source, VipsImage **out, ... )
...
-	VIPS_FOREIGN_JPEG_SUBSAMPLE_OFF,
-	VIPS_FOREIGN_JPEG_SUBSAMPLE_LAST
-} VipsForeignJpegSubsample;
+	VIPS_FOREIGN_SUBSAMPLE_AUTO,
+	VIPS_FOREIGN_SUBSAMPLE_ON,
+	VIPS_FOREIGN_SUBSAMPLE_OFF,
+	VIPS_FOREIGN_SUBSAMPLE_LAST
+} VipsForeignSubsample;
+
+typedef VipsForeignSubsample VipsForeignJpegSubsample;
```
lovell, line 388: Is this enough for backwards compatibility?

lovell (commented):
I've updated this PR to make `VipsForeignSubsample` more generic and added a few comments inline to help explain the approach.

cplusplus/include/vips/VImage8.h
jcupitt: They are defined (and implemented) in the base class for all savers. We should document them in all savers or none ... I guess none is simpler.

libvips/foreign/heifsave.c
jcupitt: I think this is a good choice!

libvips/include/vips/foreign.h
jcupitt: I was worrying about this in bed last night. Some bindings generate enums from introspection, and this change will prevent the old `VipsForeignJpegSubsample` enum being created. That in turn will cause an API break :+1: 

I think we'll need to keep both enums, but only use `VipsForeignSubsample` and add a note to the doc comment for `VipsForeignJpegSubsample` that it's deprecated and that people should use `VipsForeignSubsample` in new code.

libvips/foreign/heifsave.c
jcupitt: Ah true. Though I minor one I suppose.

cplusplus/include/vips/VImage8.h
lovell: I've kept them in for now, we can remove these in a later PR if required.

lovell (commented):
The `VipsForeignJpegSubsample` enum remains and is marked as deprecated in favour of the more generic `VipsForeignSubsample` enum.

jcupitt:
Read through it, looks great!

thomkrupa:
@jcupitt hey, do you have any ETA for releasing this PR? 

I don't want to rush you. I just want to manage the internal work that depends on this PR. We need this to fix: https://github.com/lovell/sharp/issues/2562

Thanks for your work!

jcupitt:
Hiya, it'll be in 8.11, which (guess!) will be April or May 2021.

PR #1986, opened 2021-02-08
Migrate from Travis CI to GitHub Actions

kleisauke:
This PR migrates the remaining bits (i.e. the sanitizers and macOS tests) from Travis CI to GitHub Actions. The old test workflow was renamed to a more accurate term, but I could revert this to facilitate reviewing.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,144 @@
...
+      - name: Rebuild the shared library cache
+        if: contains(matrix.os, 'ubuntu')
+        run: sudo ldconfig
+
+      - name: Install pyvips
+        run: pip3 install pyvips[test]
+
+      - name: Run test suite
+        env:
+          VIPS_LEAK: 1
```
jcupitt, line 142: Oh, interesting, does `VIPS_LEAK=1 python ...` not work? Or is this just neater?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,144 @@
...
+            libmatio-dev libnifti-dev liborc-0.4-dev \
+            liblcms2-dev libpoppler-glib-dev librsvg2-dev \
+            libgif-dev libopenexr-dev libpango1.0-dev \
+            libgsf-1-dev libopenslide-dev libffi-dev
+
+      - name: Install macOS dependencies
+        if: contains(matrix.os, 'macos')
+        run: |
+          brew install \
+            autoconf automake libtool \
```
jcupitt, line 64: I think you don't need the end of line backslash, though perhaps it's clearer to have it.

.github/workflows/ci.yml
kleisauke: That works too, but I thought I'd add it here, otherwise the length of the command gets quite long (if `LD_PRELOAD` is also included).

.github/workflows/ci.yml
kleisauke: Yes, I thought the `run: |` command needs escaping where `run:` (without `|`) doesn't, but I'm not sure.

lovell: Could use `>` to fold (replace newline with space):
```yaml
  run: >
    brew install
      autoconf automake libtool
```

kleisauke: I tried to use the fold sytnax, see: https://github.com/kleisauke/libvips/commit/1ca8e00165a7ad496c9f014b656bb4b27d5197e2. Unfortunally, [this fails](https://github.com/kleisauke/libvips/actions/runs/551502470) since the fold syntax requires the same indentation for all lines in order to turn newlines into spaces.

Somehow the normal syntax (`run:`, without `|` or `>`), does seem to work just fine, see: https://github.com/libvips/libvips/pull/1986/commits/b93db2f2bb7795f49f3421f5bbfa33fca273650c (not sure if this is a bug or a feature).

.github/workflows/ci.yml
jcupitt: Ah gotcha. Yes, it's neater, I agree.

jcupitt:
YAML is a very mysterious file format :(

jcupitt:
... and nice job, this is a great thing to have fixed!

PR #2116, opened 2021-02-24
Set minimum glib-2.0 dependency version of 2.40 (March 2014)

lovell:
This does mean libvips 8.11.0 will drop support for RHEL6 (glib 2.28.8), but that reached EOL in November 2020.


jcupitt:
Ahhhh the lovely feeling of removing ifdefs! Nice.

PR #2119, opened 2021-02-25
Compress fallback colour profiles with zlib (zopfli, via pigz)

lovell:
At runtime, this change decompresses colour profiles on the fly, as required. This reduces the shared library binary size (and therefore memory consumption) by 580KB, which at `-O2` represents a 17% saving.

Before:
```
5892519 profiles.c
3362808 libvips.so.42.13.0
```

After:
```
2360469 profiles.c
2782232 libvips.so.42.13.0
```

I'm unsure if the existing libvips test suite covers this, but the sharp tests do and they pass.

### DIFF
--- a/libvips/colour/profile_load.c
+++ b/libvips/colour/profile_load.c
@@ -66,10 +66,30 @@ vips_profile_fallback_get( const char *name, size_t *length )
 
 	for( i = 0; (fallback = vips__profile_fallback_table[i]); i++ ) 
 		if( g_ascii_strcasecmp( fallback->name, name ) == 0 ) {
-			if( length )
-				*length = fallback->length;
+			void *data;
+			GConverter *converter;
+			GConverterResult res;
+			gsize bytes_read;
+			gsize bytes_written;
 
-			return( fallback->data );
+			data = g_malloc0( fallback->length );
+			converter = G_CONVERTER( g_zlib_decompressor_new(
+				G_ZLIB_COMPRESSOR_FORMAT_ZLIB ) );
+
+			res = g_converter_convert( converter,
+				fallback->data, fallback->length,
+				data, fallback->length,
+				G_CONVERTER_INPUT_AT_END,
+				&bytes_read, &bytes_written, NULL );
+			g_object_unref( converter );
+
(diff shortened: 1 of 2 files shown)


libvips/colour/profile_load.c
```diff
@@ -66,10 +66,31 @@ vips_profile_fallback_get( const char *name, size_t *length )
 
 	for( i = 0; (fallback = vips__profile_fallback_table[i]); i++ ) 
 		if( g_ascii_strcasecmp( fallback->name, name ) == 0 ) {
-			if( length )
+			void *data;
+			GConverter *converter;
+			GConverterResult res;
+			gsize bytes_read;
+			gsize bytes_written;
+			GError *error;
```
jcupitt, line 74: You need to init this to NULL. Though I think I'd just pass NULL to g_converter_convert(), since you're not using the error message from GError.

libvips/colour/profile_load.c
```diff
@@ -66,10 +66,31 @@ vips_profile_fallback_get( const char *name, size_t *length )
...
+
+			data = g_malloc0( fallback->length );
+			converter = G_CONVERTER( g_zlib_decompressor_new(
+				G_ZLIB_COMPRESSOR_FORMAT_ZLIB ) );
+
+			res = g_converter_convert(converter,
+				fallback->data, fallback->length,
+				data, fallback->length,
+				G_CONVERTER_INPUT_AT_END,
+				&bytes_read, &bytes_written, &error );
```
jcupitt, line 84: ie. probably just pass NULL instead of &error.

libvips/colour/wrap-profiles.sh
```diff
@@ -15,7 +15,7 @@ echo "#include \"profiles.h\"" >> $out
 echo "" >> $out
 
 profile_names=
-for file in $in/*; do 
+for file in $in/*.icm; do
```
jcupitt, line 18: Ooop! Good idea.

jcupitt (changes requested):
I think there's a potential crash with error uninited, otherwise this looks great!

jcupitt:
LGTM! Nice!

Issue #1941, opened 2020-12-24
Was this linear light change from 8.9.2 to 8.10.0 intentional?


jjonesrs:
While testing the upgrade from 8.9.2 to 8.10.5, I noticed that the contents of the resized JPEGs changed. I narrowed down the version to the 8.10.0 release.

I reproduced the change outside of our pipeline using `vipsthumbnail.exe`.

Here's the command:
`vipsthumbnail.exe shoes.png --linear --size 80x -o "shoes80x80-8-9-2.jpg[background=255]"`

Version 8.9.2 produces:
![shoes80x80-8-9-2](https://user-images.githubusercontent.com/16615108/103048795-a566c300-4544-11eb-9d3f-b46dafa1cbe3.jpg)

Version 8.10.0 produces:
![shoes80x80-8-10-0](https://user-images.githubusercontent.com/16615108/103048833-c0d1ce00-4544-11eb-8900-c4d321376feb.jpg)

Version 8.10.0 without the `--linear` option produces:
![shoes80x80-8-10-0-no-linear](https://user-images.githubusercontent.com/16615108/103048900-f70f4d80-4544-11eb-97ee-e7d1365556e8.jpg)

If this change represents intended behavior, then I'll switch our pipeline away from using linear light.

Here's the original image for reference:
![shoes](https://user-images.githubusercontent.com/16615108/103048845-c9c29f80-4544-11eb-9d35-f8179385844e.png)

jcupitt:
Hi @jjonesrs, oh dear, that's bad. I'll have a look -- thanks for the report!

jcupitt:
Yes, as quick workaround, I'd move away from linear light for now.

jcupitt:
That was painful, but I think I've fixed this. Could I use a small detail from your `shoes.png` image for the libvips test suite? It would help prevent this regressing again.

jcupitt:
I added a test to that branch using the tip of one of the shoes. Please let me know if that's not OK.

https://github.com/libvips/libvips/blob/revise-premultiply/test/test-suite/images/rgba.png

jcupitt:
There's a PR for this now, I'll close. Thanks for the report! The fix should be in 8.10.6.

PR #2120, opened 2021-02-27
Revise premultiply

jcupitt:
This PR revises unpremultiply (again), and tries to simplify thumbnail a bit. It fixes resizing RGBA images in scRGB space, and hopefully doesn't regress an earlier RGBA resize issue. There's a test as well, to try to stop this happening again.

Background: https://github.com/libvips/libvips/issues/1941

Related issue: https://github.com/libvips/libvips/pull/1675


libvips/resample/thumbnail.c
```diff
@@ -760,16 +785,16 @@ vips_thumbnail_build( VipsObject *object )
...
-			return( -1 );
-		have_premultiplied = TRUE;
-
 		/* vips_premultiply() makes a float image. When we
 		 * vips_unpremultiply() below, we need to cast back to the
-		 * pre-premultiply format.
+		 * pre-premultiplied format.
 		 */
+		g_info( "premultiplying alpha" ); 
+		have_premultiplied = TRUE;
```
kleisauke, line 793: Not sure if this is neater, but instead of keeping a separate `have_premultiplied` flag, you could also set `unpremultiplied_format` to `VIPS_FORMAT_NOTSET` and check for `unpremultiplied_format != VIPS_FORMAT_NOTSET` instead.

libvips/resample/thumbnail.c
```diff
@@ -827,9 +851,10 @@ vips_thumbnail_build( VipsObject *object )
 			in = t[7];
 		}
 	}
-	else if( thumbnail->export_profile &&
-		(vips_image_get_typeof( in, VIPS_META_ICC_NAME ) || 
-		 thumbnail->import_profile) ) {
+	else if( thumbnail->export_profile ) {
```
kleisauke, line 854: Do you think this works when someone does:
```bash
$ vipsthumbnail --export-profile=srgb x.jpg
````
where `x.jpg` is an image _without_ embedded ICC profile? I think the previous check was there to handle this case (perhaps as a speed improvement).

File: libvips/resample/thumbnail.c [hunk 3]
```diff
@@ -844,6 +869,19 @@ vips_thumbnail_build( VipsObject *object )
 			return( -1 );
 		in = t[7];
 	}
+	else if( thumbnail->linear ) {
+		/* Linear mode, no colour management. We went to scRGB for
+		 * precessing, so we now revert to the input
```
kleisauke, line 874: nit: typo
```suggestion
		 * processing, so we now revert to the input
```

libvips/resample/thumbnail.c
jcupitt: Good idea!

libvips/resample/thumbnail.c
jcupitt: I think this is OK -- `have_imported` is now only set TRUE if `vips_image_get_typeof( in, VIPS_META_ICC_NAME ) || thumbnail->import_profile` earlier in the function, and we only hit this if if `!have_imported`.

jcupitt: But you're right, this won't work if `k2.jpg` has no profile and perhaps it should:

```
$ vipsthumbnail k2.jpg --export-profile cmyk
vipsthumbnail: unable to thumbnail k2.jpg
icc_transform: no input profile
```

It should assume that `k2.jpg` is in srgb.

jcupitt: I committed a thing for this case.

lovell:
Profiling suggests the unpremultiply operation might be ~17% slower with this change.

Using callgrind to unpremultiply a 6MP RGBA image with code compiled using gcc 9.3.0 and `-O3`.

Before:
```
  175,523,571  ???:vips_unpremultiply_gen [/usr/local/lib/libvips.so.42.12.7]
```
After:
```
  212,753,893  ???:vips_unpremultiply_gen [/usr/local/lib/libvips.so.42.12.7]
```
Would the following diff be functionally equivalent?
```diff
--- a/libvips/include/vips/util.h
+++ b/libvips/include/vips/util.h
@@ -91,11 +91,7 @@ vips_recip( double x )
 {
        static const double epsilon = 1e-10;
 
-       int sign = x < 0.0 ? -1.0 : 1.0;
-
-       return( sign * x >= epsilon ?
-               1.0 / x :
-               sign / epsilon );
+       return 1 / VIPS_CLIP( -epsilon, x, epsilon );
 }
```
If so, that's only 8% slower:
```
  190,352,189  ???:vips_unpremultiply_gen [/usr/local/lib/libvips.so.42.12.7]
```

jcupitt:
I removed the recip thing -- it wasn't really necessary, and it was slower than I expected. Thanks for the speed check!

lovell (approved):

jcupitt:
\o/ good stuff, thanks!

Issue #2128, opened 2021-03-03
vips tiffsave doesn't set predictor correctly


Adios:
## Bug report

**Describe the bug**

I try to re-compress an image in zip with `--predictor` set to `horizontal` as in
```
vips tiffsave GitHub-Mark.png vips_horizontal.tif --compression deflate --predictor horizontal
```
but it's seems to be ignored in the output image:
```
$ tiffinfo vips_horizontal.tif 
TIFF Directory at offset 0x3b3c (15164)
  Image Width: 560 Image Length: 560
  Resolution: 28.34, 28.34 pixels/cm
  Bits/Sample: 8
  Sample Format: unsigned integer
  Compression Scheme: AdobeDeflate
  Photometric Interpretation: RGB color
  Orientation: row 0 top, col 0 lhs
  Samples/Pixel: 3
  Rows/Strip: 128
  Planar Configuration: single image plane
```
And no matter `--predictor horizontal` or `--predictor none`, they output the same file size, _as if they were compressed without setting a predictor_.

**To Reproduce**

1. use image: https://github.githubassets.com/images/modules/logos_page/GitHub-Mark.png
2. `vips tiffsave GitHub-Mark.png vips_horizontal.tif --compression deflate --predictor horizontal`
3. `tiffinfo vips_horizontal.tif`
4. `vips tiffsave GitHub-Mark.png vips_none.tif --compression deflate --predictor none`

**Expected behavior**

1. Should see `Predictor: horizontal differencing 2 (0x2)` in the `tiffinfo` output
2. **vips_horizontal.tif** & **vips_none.tif** should have different sizes (in most cases)

**Actual behavior**

1. No `Predictor: horizontal differencing 2 (0x2)` in the `tiffinfo` output
2. **vips_horizontal.tif** & **vips_none.tif** have the same size

**Environment**

 - `Linux 5.11.2-arch1-1`

```
$ pacman -Q libtiff libvips imagemagick
libtiff 4.2.0-1
libvips 8.10.3-1
imagemagick 7.0.11.2-1
```

**Additional context**

As a comparison, use ImageMagick:
```
magick convert GitHub-Mark.png -compress zip -define tiff:predictor=2 gm_horizontal.tif
```
Could see the `Predictor` filed is set,
```
$ tiffinfo gm_horizontal.tif 
TIFF Directory at offset 0x2494 (9364)
  Image Width: 560 Image Length: 560
  Bits/Sample: 8
  Compression Scheme: AdobeDeflate
[...]

PR #2129, opened 2021-03-04
Add tiff zstd predictor support

Adios:
Hi, thank you for the quickly fix to #2128. And based on that I tried to make `--predictor` also available to ZSTD compression in tiff.

Before the commit, `vips` cannot set `predictor` on ZSTD compressed TIFF:
Screenshot from 2021-03-04 23-41-58

By this commit, `vips` now can set `predictor` on ZSTD TIFF:
Screenshot from 2021-03-04 23-37-30

### DIFF
--- a/libvips/foreign/tiffsave.c
+++ b/libvips/foreign/tiffsave.c
@@ -590,8 +590,8 @@ vips_foreign_save_tiff_buffer_init( VipsForeignSaveTiffBuffer *buffer )
  * User @level to set the ZSTD compression level. Use @lossless to
  * set WEBP lossless mode on. Use @Q to set the WEBP compression level.
  *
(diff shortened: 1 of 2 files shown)


jcupitt:
Oh, nice, I didn't know zstd used the predictors.

Though I think it would be better to have all the predictors together, I'll make a small change.

jcupitt:
I moved the predictors together with https://github.com/libvips/libvips/commit/495b54dfcfc2ce3aeead23603b691df841af0902

Rather than posting screenshots, it's best to copy-paste text and use triple backticks around it, eg.:

```
    ```
    some text
    ```
```

Renders as

```
some text
```

Adios:
No problem. I was also thinking if it would be better to have all the predictors together.

PR #2133, opened 2021-03-06
Improve detection of bad profiles

jcupitt:
Broken profiles can make colour transform difficult.

This patch adds a lot more sanity checking, so bad profiles are detected and ignored, allowing fallbacks to occur.

Still to do:

- [x] Having to supply a fallback profile is bad, we should fall back to srgb / cmyk / mono etc. automatically
- [x] The error messages could be better

See https://github.com/libvips/libvips/discussions/2084


jcupitt:
I tried to improve fallback handling. You now see:

```
john@kiwi:~/pics$ vipsthumbnail IEC61966-2.1.jpg -e srgb -o x.jpg -n perceptual

(vipsthumbnail:35262): VIPS-WARNING **: 17:22:27.470: input profile does not support perceptual intent
john@kiwi:~/pics$ 
```

I think this is now ready for review and merge!

PR #2134, opened 2021-03-06
Cleanup configure.ac and #ifdef's

kleisauke:
This PR cleanups `configure.ac` and a bunch of `#ifdef`'s. I used this renaming/remove strategy:
- `OS_WIN32` -> `G_OS_WIN32` (GLib definition when compiling for native Win32)
- `BINARY_OPEN` ->  `G_PLATFORM_WIN32` (GLib definition when compiling for native Win32 or Cygwin environments)
- `HAVE_G_WIN32_GET_COMMAND_LINE` -> `G_OS_WIN32`
- `VIPS_OS_DARWIN` -> <kbd>Delete</kbd> (appears to be unused)

This implies that the old definitions should no longer be used. Note that since the new definitions originates from GLib, it is necessary to include `vips/vips.h` prior to using them.

After this renaming, the following decoupled variables were removed from `configure.ac`:
- `vips_os_win32`
- `vips_binary_open`
- `have_g_win32_get_command_line`
- `vips_os_darwin`

Plus a couple of no longer defined variables:
- `GTHREAD_CFLAGS` / `GTHREAD_LIBS` - removed with https://github.com/libvips/libvips/commit/117dbd888a15f9be73661aeebd32ce762c82e5f6.
- `LIBWEBPMUX_CFLAGS`  / `LIBWEBPMUX_LIBS` - removed with https://github.com/libvips/libvips/commit/801a2a71443c33679ae9a18a1c9f8db8764152e4.

Finally, I removed some redundant `#if GLIB_CHECK_VERSION( x, y, z )` guards that are no longer necessary after commit https://github.com/libvips/libvips/commit/117dbd888a15f9be73661aeebd32ce762c82e5f6.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -141,69 +141,35 @@ AM_CONDITIONAL(ENABLE_DEPRECATED, [test x"$enable_deprecated" = x"yes"])
 # this gets pasted into version.h as a #define
 AC_SUBST(VIPS_ENABLE_DEPRECATED)
 
-AC_MSG_CHECKING([for native Win32])
-case "$host" in
-  *-*-mingw*)
-    vips_os_win32=yes
-    ;;
-  *)
-    vips_os_win32=no
-    ;;
-esac
-AC_MSG_RESULT([$vips_os_win32])
-
-if test x"$vips_os_win32" = x"yes"; then
-  AC_DEFINE(OS_WIN32,1,[native win32])
-
-  # makes gcc use win native alignment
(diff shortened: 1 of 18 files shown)


lovell (approved):

jcupitt:
Nice! Yes, it had become very untidy.

I'll fix up the summary text at the end of configure too, it's a bit unclear.

PR #2144, opened 2021-03-12
Ensure max_band vector is aligned on a 16-byte boundary

kleisauke:
Fixes https://github.com/mstorsjo/llvm-mingw/issues/190.

### DIFF
--- a/libvips/conversion/composite.cpp
+++ b/libvips/conversion/composite.cpp
@@ -163,10 +163,17 @@ vips_composite_base_dispose( GObject *gobject )
 	G_OBJECT_CLASS( vips_composite_base_parent_class )->dispose( gobject );
 }
 
-/* Our sequence value. This must be aligned on a 16-byte boundary when
- * HAVE_VECTOR_ARITH is defined.
+/* Our sequence value.
  */
 typedef struct {
+#ifdef HAVE_VECTOR_ARITH
+	/* max_band as a vector, for the RGBA case. This must be
+	 * defined first to ensure that the member is aligned
+	 * on a 16-byte boundary.
+	 */
+	v4f max_band_vec;
+#endif /*HAVE_VECTOR_ARITH*/
+
 	VipsCompositeBase *composite;
 
 	/* Full set of input regions, each made on the corresponding input
@@ -192,12 +199,6 @@ typedef struct {
 	/* For each enabled image, an input pointer.
 	 */
 	VipsPel **p;
-
-#ifdef HAVE_VECTOR_ARITH
-	/* max_band as a vector, for the RGBA case.
-	 */
-	v4f max_band_vec;
-#endif /*HAVE_VECTOR_ARITH*/
 
 } VipsCompositeSequence;
 
--- a/libvips/conversion/composite.cpp
+++ b/libvips/conversion/composite.cpp
@@ -55,12 +55,16 @@
 
 #include <stdio.h>
 #include <string.h>
-#if _MSC_VER
+#ifdef _MSC_VER
 #include <cstdlib>
 #else
 #include <stdlib.h>
 #endif
 #include <math.h>
+
+#if defined(HAVE__ALIGNED_MALLOC) || defined(HAVE_MEMALIGN)
+#include <malloc.h>
+#endif
 
 #include <vips/vips.h>
 #include <vips/internal.h>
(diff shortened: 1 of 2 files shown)


jcupitt:
I found this SO answer:

https://softwareengineering.stackexchange.com/questions/256179/the-advantage-of-using-attribute-aligned

Which seems to suggest that gcc and clang both allow:

```C
int a[16] __attribute__((aligned(16))) = {  1,  2,  3,  4,  5,  6,  7, 8,  9, 10, 11, 12, 13, 14, 15, 16 };
__vector signed int *va = (__vector signed int *) a;
```

It might be a little neater.

lovell:
LGTM, thanks Kleis. Can we assume `HAVE_VECTOR_ARITH` will be true for most modern compilers and remove the conditionality? (This might be something for 8.11.0.)

lovell (approved):

kleisauke:
OK, I added the aligned attribute to `v4f` with commit https://github.com/libvips/libvips/pull/2144/commits/e7faebf6af6b77c34d04a4e4efb7e47f124b8d39. I don't think we can remove those `HAVE_VECTOR_ARITH` checks altogether, because it might break MSVC (I'm also not sure if the aligned attribute is supported there).

jcupitt:
Looks like it's `__declspec(align(16))` in MSVC:

https://docs.microsoft.com/en-us/cpp/cpp/align-cpp?view=msvc-160

Though of course I don't know if the other vector stuff is supported.

libvips/conversion/composite.cpp
```diff
@@ -194,8 +193,47 @@ typedef struct {
 	 */
 	VipsPel **p;
 
+#ifdef HAVE_VECTOR_ARITH
+	/* max_band as a vector, for the RGBA case.
+	 */
+	v4f max_band_vec;
+#endif /*HAVE_VECTOR_ARITH*/
```
jcupitt, line 200: How about putting this struct member first?

At the moment, we could perhaps change the size of one the members before this and break alignment. I think it only works at the moment because of the exact number of pointers we have.

A comment about the need to have the vector as the first element would probably be useful.

kleisauke: Did you mean to put this member first within the `VipsCompositeSequence` or move it back to `VipsCompositeBase` and put it there as first? I tried the latter version (see commit https://github.com/kleisauke/libvips/commit/dcee3d11fe701c854b8e7ce1964f6ee0951f0b70, which is the smallest change), but that didn't seem to work.

Do you think this could work without those aligned malloc implementations?

jcupitt: I meant first in `VipsCompositeSequence`. 

My thinking was that the aligned malloc guarantees that the start of the struct will be on a 16-byte boundary, but `max_band_vec` will be aligned on something like (pointer-to-base + 3 * sizeof(pointer) + sizeof(int) + 2 * sizeof(pointer)). If we put the struct members with alignment requirements at the start of the struct, we can guarantee they will stay aligned even if the struct changes.

I'm not sure I understand how these structs are being laid out :( You'd think this would be:

```
base: composite
base + 4: input_regions
base + 8: composite_regions
base + 12: n
base + 16: enabled
base + 20: p
base + 24: max_band_vec
```

Which would obviously break alignment. I guess some padding somewhere must be making max_band_vec start at byte 32.

kleisauke: Ah yes, I'm also not entirely sure how these structs are laid out, so moving the `max_band_vec` member to the first position in the struct makes sense. Done with commit 305714c9 (tested on Windows 32-bit without any problems).

fwiw, I think the above mentioned commit also reproduced the crash on Linux x64, see:
https://travis-ci.org/github/kleisauke/libvips/jobs/764105312#L2420

jcupitt:
This is great work, thank you Kleis!

kleisauke:
No problem! Should I try to compile libvips with MSVC to check whether the vector support and/or the aligned attribute is supported there? fwiw, I saw that this is excluded in the repository provided by @angelmixu (sorry for the ping):
https://github.com/angelmixu/libvips-vs2017/blob/c2f993ae01261ee2b486c4e1823805c6c41e8e15/vs2017project/config.h#L350

jcupitt:
If you're interested, sure, have a look. I'd be pleased (and surprised, heh) if MSVC supported this kind of vector code.

PR #2159, opened 2021-03-24
Set correct description for heifsave argument

der-eismann:
Stumbled upon this while tinkering with AVIF


jcupitt:
Oh, nice, thank you. Sigh copy-paste errors :(

PR #2166, opened 2021-03-28
nsgifload: ensure GIF_INSUFFICIENT_FRAME_DATA is unrecoverable

lovell:
As discussed in https://github.com/libvips/libvips/pull/1709#issuecomment-808895573, this should fix all of the OSS Fuzz issues found so far today (currently 10).

### DIFF
--- a/libvips/foreign/nsgifload.c
+++ b/libvips/foreign/nsgifload.c
@@ -355,15 +355,8 @@ vips_foreign_load_nsgif_header( VipsForeignLoad *load )
 #ifdef VERBOSE
 	print_animation( gif->anim );
 #endif /*VERBOSE*/
-	if( result != GIF_OK && 
-		result != GIF_WORKING &&
-		result != GIF_INSUFFICIENT_FRAME_DATA ) {
+	if( result != GIF_OK && result != GIF_WORKING ) {
 		vips_foreign_load_nsgif_error( gif, result ); 
-		return( -1 );
-	}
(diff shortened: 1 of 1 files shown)


jcupitt:
This feels very severe to me -- it'll block all truncated GIFs, and that's a common case that used to be handled.

Is there some other way we can fix this? Perhaps some sanity checking on the dimensions?

kleisauke:
Looks like the parameters of the `calloc` call in `vips_foreign_load_nsgif_bitmap_create` is overflowing. Here's the salient part of that OSS-Fuzz issue:
```bash
ERROR: AddressSanitizer: calloc parameters overflow: count * size (-1643047120 * 4) cannot be represented in type size_t (thread T0)
<snip />
  | #1 0x787bab in vips_foreign_load_nsgif_bitmap_create libvips/libvips/foreign/nsgifload.c:510:16
<snip />
```
Perhaps a sanity check would help there?

jcupitt:
Oh, good idea, perhaps

```C
        if( width <= 0 ||
                width > VIPS_MAX_COORD ||
                height <= 0 ||
                height > VIPS_MAX_COORD ) {
                vips_error ...
                return( NULL );
        }

        return calloc( (size_t) width * height, 4 );
```

Though it would still be easy to bust through the 2gb fuzz limit, of course.

lovell:
Thanks for the feedback, take 2 of this PR at https://github.com/libvips/libvips/pull/2168

PR #2168, opened 2021-03-28
nsgifload: check dimensions before memory allocation

lovell:
GIF dimensions are 16-bit unsigned according to https://en.wikipedia.org/wiki/GIF

As images may be truncated, memory is zero initialised.


jcupitt:
Looks good!

PR #2176, opened 2021-03-31
nsgifload: ensure frame offset is within existing dimensions

lovell:
This fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=32746 although there may be a better solution.

We should submit a fix for this upstream too.

### DIFF
--- a/libvips/foreign/libnsgif/libnsgif.c
+++ b/libvips/foreign/libnsgif/libnsgif.c
@@ -357,6 +357,11 @@ static gif_result gif_initialise_frame(gif_animation *gif)
         width = gif_data[5] | (gif_data[6] << 8);
         height = gif_data[7] | (gif_data[8] << 8);
 
+        /* Ensure frame offset is within existing dimensions */
+        if (offset_x >= gif->width || offset_y >= gif->height) {
+                return GIF_FRAME_DATA_ERROR;
+        }
+
         /* Set up the redraw characteristics. We have to check for extending
          * the area due to multi-image frames.
          */


jcupitt:
Ouch. I tried stepping though the test case and I think it's actually a legal GIF (horrifyingly). 

It doesn't set the base size of the GIF, which I think means the animation should size to the largest frame, then frame 1 tries to draw a rect of 8224 x 8224 pixels at 31487 x 8224. This enlarges the frame to 39711 x 16448 (x_offset + width, y_offset + height), which is still within the allowed dimensions of a GIF.

libnsgif calls vips_foreign_load_nsgif_bitmap_create() which then int overflows when it computes width * height * 4.

How about instead changing vips_foreign_load_nsgif_bitmap_create() like this:

```
        return g_malloc0( (gsize) width * height * 4 );
```

Now there's no int overflow and valgrind runs cleanly. Load fails a couple of bytes later and exits with "no frames in GIF" when it discovers there's not actually 3gb of data there.

lovell:
Yes, I see what you mean, that is quite horrifying.

Given animated GIFs can rapidly expand the canvas in this manner, perhaps we should reduce the maximum permitted pixel count or dimensions below the 16 bit unsigned spec and instead select an arbitrarily smaller number? The current code can (attempt to) allocate over 17GB per frame.

jcupitt:
Yes, that sounds good to me. Do you have any feeling for what a sensible limit might be? 1000 x 1000 pixels? Or is that too small?

jcupitt:
I suppose the ideal would be if we could stop libnsgif allocating memory until gif_decode_frame(). Then `new_from_file` would return the size, and the caller would have a chance to bail out.

At the moment, the frame is allocated during gif_initialise() so there's no opportunity for the caller to detect a memory bomb GIF.

jcupitt:
How about this? https://github.com/libvips/libvips/commit/9bdf5e8cda3e0c63584984282b1e36d97c50bb1a

It moves frame memory allocate to gif_decode_frame(), and gif_initialise() just tracks the max size.

lovell:
Oh yes, that's much better. I guess we should limit the pixel count too - I'll do a separate PR for that.

jcupitt:
Or we could leave it to the application -- it now won't allocate any memory until you try to read pixels, so the caller will get a chance to check dimensions for decode bombs. Better than wiring it into the library, IMO.

lovell:
OK, there's a possible PR at #2179 but I don't mind either way and happy for that to be closed.

By default sharp already limits input pixel count based on dimensions in image header metadata, so your proposed fix is definitely the right way to do this.

PR #2179, opened 2021-04-01
nsgifload: enforce maximum GIF dimensions of 16383

lovell:
As discussed in https://github.com/libvips/libvips/pull/2176

The GIF spec allows dimensions up to 65535 (16-bit unsigned), but this equates to 17GB/frame.

A common task is to convert animated GIF to animated WebP, and the latter supports dimensions up to 16383 (14-bit unsigned), so that seems like a sensible limit and equates to 1GB/frame.

Also makes the error message consistent with other loaders.


jcupitt:
Sounds very sensible. We could make the libnsgif change (delaying memory allocation until frame load) too --- they seem like complementary ideas.

lovell:
Yes, I agree, we should apply both changes.

PR #2190, opened 2021-04-07
colour: add a fallback for the increasingly-common P3 profile

lovell:
I'm starting to see greater adoption for the P3 profile, which offers ~25% wider gamut compared with non-linear sRGB, plus CSS will soon support `p3` in its [color-gamut](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/color-gamut) directive.

This PR uses the 736 bytes (607 bytes when compressed) CC0-licensed "magic" compatibility ICC v2 profile from https://github.com/saucecontrol/Compact-ICC-Profiles#display-p3

(We could skip P3 and go straight to [BT.2020](https://en.wikipedia.org/wiki/Rec._2020), but I'm unsure which will become the de facto standard for the web.)


jcupitt:
Ah good idea. `profiles.c` is generated automatically from the set of profiles in `libvips/colour/profiles/`, so just copy `p3.icm` in there and rebuild.

lovell:
Great, thanks John. This PR contains the CC0 `libvips/colour/profiles/p3.icm` binary file and the automatically-rebuilt `profiles.c`. Is anything else required?

jcupitt:
Sorry Lovell, I didn't see the profile in the diff, just the patched `profiles.c`. I need stronger specs :(

PR #2195, opened 2021-04-12
Text colour rendering

jcupitt:
Add full-colour text rendering to `vips_text()` with an `rgba` flag.

Example:

```
$ vips text x.png "😀<span foreground='red'>red</span><span background='cyan'>blue</span>" --dpi 300 --rgba
```

To make:

x

Seems to work!

Non-rgba rendering works too, though there are some differences. For example, before emojii rendered in greyscale, but now you just get a mask shape.

```
$ vips text x2.png "😀<span foreground='red'>red</span>" --dpi 300
```

To make:

x2

Perhaps this is OK?

This is a change to libvips dependencies, so packagers will need to update.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,7 +62,6 @@ jobs:
         if: contains(matrix.os, 'macos')
         run: |
           brew update 
-          brew upgrade
           brew install
             autoconf automake libtool
             gtk-doc gobject-introspection
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,9 @@ jobs:
 
       - name: Install macOS dependencies
         if: contains(matrix.os, 'macos')
-        run:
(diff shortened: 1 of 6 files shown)


lovell (approved):
LGTM, thanks John. The Homebrew failures are due to today's Bintray "brownout" - see https://twitter.com/bintray/status/1381592942444388352 (can probably add a `brew update` to the CI script to fix this).

.github/workflows/ci.yml
```diff
@@ -60,7 +60,9 @@ jobs:
 
       - name: Install macOS dependencies
         if: contains(matrix.os, 'macos')
-        run:
+        run: |
+          brew update 
+          brew upgrade
```
lovell, line 65: I think only `brew update` is required here.

jcupitt: OK, I'll try that, thanks!

PR #2202, opened 2021-04-18
CI: ensure brew install commands are on a single line, split into base dependencies vs image formats

lovell:
Due to the (necessary) addition of `brew update`, this is now considered a multi-command script so every new line must be its own command.

(CI is currently failing on the `master` branch and in recent PRs due to this.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,17 +61,10 @@ jobs:
       - name: Install macOS dependencies
         if: contains(matrix.os, 'macos')
         run: |
-          brew update 
-          brew install
-            autoconf automake libtool
-            gtk-doc gobject-introspection
-            cfitsio fftw giflib
-            glib libexif libgsf
-            libheif libjpeg-turbo libmatio
-            librsvg libspng libtiff
-            little-cms2 openexr openslide
-            orc pango poppler webp
-            openjpeg
+          brew update
+          brew upgrade
+          brew install autoconf automake libtool fftw fontconfig gtk-doc gobject-introspection glib libexif libgsf little-cms2 orc pango
+          brew install cfitsio imagemagick libheif libjpeg-turbo libmatio librsvg libspng libtiff openexr openjpeg openslide poppler webp
 
       - name: Install Clang 10
         env:


lovell:
(Apologies for the commit noise, I'm not entirely sure why this is still failing, will investigate.)

kleisauke:
It seems that the CI failure is due to the `-fopenmp` being passed to a linker that doesn't support OpenMP. This option is defined in [ImageMagick's pkg-config files](https://github.com/Homebrew/homebrew-core/blob/9c321a9179ce0ee44f1fc2c72c6138f0fc461653/Formula/imagemagick.rb#L73-L74), which I tried to remove with commit https://github.com/kleisauke/libvips/commit/e62374b6864211fce16f55574959532dd77aaac1, but that didn't seem to work. Another option would be to use Homebrew's LLVM, or to avoid installing ImageMagick on the macOS CI runner (see for e.g. commit https://github.com/kleisauke/libvips/commit/3ff5b3bd498d3ec8a059725c0030056d8dc90054).

It's also necessary to explicitly link against fontconfig (see commit https://github.com/kleisauke/libvips/commit/fa5cc1f61b2b2b54d7c2c0300028a9bedf70b58f), to avoid this error on macOS: 
```
Undefined symbols for architecture x86_64:
  "_FcConfigAppFontAddFile", referenced from:
      _vips_text_build in libcreate.a(text.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
(it looks like pangocairo does not list fontconfig as a dependency).

Feel free to cherry-pick these commits (I can also open a separate PR if that's preferred).

kleisauke:
Oh, I just noticed that installing `imagemagick@6` also works (see for e.g. commit https://github.com/kleisauke/libvips/commit/81f759a4a35a218179be879f483af972a84cb060), since that [does not depend on OpenMP](https://github.com/Homebrew/homebrew-core/blob/27d1037ca6643d061f29cac97e791b6b3b3c2b3b/Formula/imagemagick%406.rb#L51).

lovell:
Good spot, thanks Kleis, I'll close this PR and let you open a new one from your `fix-ci-2` branch.

PR #2203, opened 2021-04-18
Update libnsgif

tlsa:
This series syncs with the [upstream libnsgif](http://git.netsurf-browser.org/libnsgif.git/)'s changes to optimize LZW decompression.

Patches created with the following (from `libnsgif/src` directory):

```
git format-patch --relative bba28df0cc0c75338a63645c6a55aebfebe91c74
```

And applied with:

```
git am --directory=libvips/foreign/libnsgif/ ../libnsgif/src/*.patch
```

### DIFF
--- a/libvips/foreign/libnsgif/libnsgif.c
+++ b/libvips/foreign/libnsgif/libnsgif.c
@@ -621,8 +621,8 @@ static gif_result gif__recover_previous_frame(const gif_animation *gif)
         return GIF_OK;
 }
 
-static inline gif_result
-gif__decode(gif_animation *gif,
+static gif_result
+gif__decode_complex(gif_animation *gif,
                 unsigned int frame,
                 unsigned int width,
                 unsigned int height,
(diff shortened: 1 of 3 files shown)


jcupitt:
Very cool, thank you @tlsa !

We have `update.sh` in `libvips/foreign/libnsgif` which attempts to update our copy of libnsgif from your git master. Would running that have the save effect as this PR?

tlsa:
> Very cool, thank you @tlsa !
> 
> We have `update.sh` in `libvips/foreign/libnsgif` which attempts to update our copy of libnsgif from your git master. Would running that have the save effect as this PR?

Oh, I did not see that script.  It should have the same ultimate effect, although it would be harder to inspect which commits you have from upstream from looking at the log.

jcupitt:
OK, let's merge this and then try `update.sh` as well to make sure it's all up to date. Thanks for the PR!

jcupitt:
Updated from your repo: https://github.com/libvips/libvips/commit/4cd5da230314f2009bb15d152845d573330c5e31

PR #2211, opened 2021-04-21
CI: fix failures in macOS runner

kleisauke:

(This PR is based upon #2202)


jcupitt:
Huh that's odd, you'd think pango would always require fontconfig.

Anyway, looks great, nice job!

PR #2213, opened 2021-04-21
C++: add set/get for an array of doubles

kleisauke:
Useful for retrieving and setting background colour metadata, for example:
```c++
std::vector<double> background = image.get_array_double("background");
for (double i : background) {
    std::cout << i << std::endl;
}

background = {255.0, 255.0, 255.0};
image.set("background", background);
```


lovell (approved):
LGTM

jcupitt:
\o/

PR #2223, opened 2021-04-26
Add jp2k compression to tiff load and save

jcupitt:
This patch adds jpeg2000 compression to tiff load and save using openjpeg. Images are compatible with QuPath and bioformats.


jcupitt:
Clsoed in favour of https://github.com/libvips/libvips/pull/2224

PR #2235, opened 2021-05-03
Reuse threads by using the thread pool of GLib (#2038)

kleisauke:
This PR revamps the threading system of libvips by using [GLib's threadpool](https://developer.gnome.org/glib/stable/glib-Thread-Pools.html) as discussed in #2038. This implementation allows reuse of already started threads.

Here is a brief overview of the related changes within this PR (see the individual commits for more details).
- Add `vips_threadpool_push` for inserting an new function into the list of tasks to be executed by libvips' thread pool.
- Add `vips__threadpool_shutdown` to shutdown the thread pool during [`vips_shutdown()`](https://libvips.github.io/libvips/API/current/libvips-vips.html#vips-shutdown).
- Add `vips__buffer_shutdown` to free the calculated pixel buffer cache early during [`vips_thread_shutdown()`](https://libvips.github.io/libvips/API/current/libvips-vips.html#vips-thread-shutdown).
- Remove `vips__n_active_threads`, as this is no longer necessary (it was used to check whether all threads are joined).
- Deprecate `vips_g_thread_new` and `vips_g_thread_join` which was a wrapper around the old GThread API.
- Remove mutex lock for `VipsThreadStartFn`.
- Move `sink_disc` and `sink_screen` threads to the thread pool.
- Add suppressions file for ThreadSanitizer.
- Swap `g_private_set` with `g_private_replace` where possible.
- Use `gatomicrefcount` in `sinkscreen.c`.

This was benchmarked on [Linux](https://github.com/kleisauke/vips-microbench/blob/master/results/reuse-threads.md) and [Windows](https://gist.github.com/kleisauke/cf6a115c65894d27eda4a8d9379776f7) with relative speed-ups varying between ~4% and ~15% depending on the benchmark and the OS it was run on. A soak-test based on https://github.com/libvips/libvips/pull/1240#issuecomment-467876334 can be found [here](https://gist.github.com/kleisauke/fbc5a8bb418da2f660c06bdd07115e10).

This PR was tested last week within production on [images.weserv.nl](https://images.weserv.nl/) without any problems. We noticed an immediate reduction in the number of interrupts and `fork()` / `clone()` system calls, see:
Interrupts and context switches | `fork()` / `clone()` system calls
:--: | :--:
Interrupts | fork() / clone() syscalls
_(Click to enlarge)_ | _(Click to enlarge)_

<details>
  <summary>API Changes</summary>

The following symbols are added to the public API:
* ```c
  int vips_threadpool_push( const char *name, GFunc func, gpointer data );
  ```
  _For inserting an new function into the list of tasks to be executed by libvips' thread pool._

The following symbols are added to the private API:
* ```c
  void vips__threadpool_shutdown( void );
  ```
  _To shutdown the thread pool during [`vips_shutdown()`](https://libvips.github.io/libvips/API/current/libvips-vips.html#vips-shutdown)._

* ```c
  void vips__buffer_shutdown( void );
  ```
  _To free the calculated pixel buffer cache early during [`vips_thread_shutdown()`](https://libvips.github.io/libvips/API/current/libvips-vips.html#vips-thread-shutdown)._

The following symbols are removed from the private API:
* ```c
  int vips__n_active_threads;
  ```
  _Used to check whether all threads are joined, no longer necessary._

The following symbols are deprecated and should no longer be used:

* ```c
  GThread *vips_g_thread_new( const char *domain, GThreadFunc func, gpointer data );
  ```
  _Wrapper around the [GThread API](https://developer.gnome.org/glib/stable/glib-Threads.html), use [`g_thread_try_new`](https://developer.gnome.org/glib/stable/glib-Threads.html#g-thread-try-new) instead._

* ```c
  void *vips_g_thread_join( GThread *thread );
  ```
  _Wrapper around the [GThread API](https://developer.gnome.org/glib/stable/glib-Threads.html), use [`g_thread_join`](https://developer.gnome.org/glib/stable/glib-Threads.html#g-thread-join) instead._

`abi-compliance-checker` result:
https://kleisauke.nl/compat_reports/vips/master_to_reuse-threads/compat_report.html
</details>

### DIFF
--- a/libvips/iofuncs/threadpool.c
+++ b/libvips/iofuncs/threadpool.c
@@ -201,7 +201,8 @@ vips_thread_main_loop( gpointer thread_data, gpointer pool_data )
 	VipsThreadExec *exec = (VipsThreadExec *) thread_data;
 
 	/* Set this to something (anything) to tag this thread as a vips 
-	 * worker.
+	 * worker. No need to call g_private_replace as there is no
+	 * GDestroyNotify handler associated with a worker.
 	 */
 	g_private_set( is_worker_key, thread_data );
 
--- a/libvips/iofuncs/threadpool.c
+++ b/libvips/iofuncs/threadpool.c
@@ -25,6 +25,7 @@
  * 	- free threadpool earlier 
  * 02/02/20 kleisauke
  *	- reuse threads by using GLib's threadpool
+ * 	- remove mutex lock for VipsThreadStartFn
  */
 
 /*
@@ -666,11 +667,8 @@ vips_task_run( gpointer data, gpointer user_data )
 
 	VIPS_GATE_START( "vips_task_run: thread" );
 
-	// TODO: Could we move this to vips_task_work_unit()?
-	g_mutex_lock( task->allocate_lock );
 	if( !(state = task->start( task->im, task->a )) )
 		task->error = TRUE;
-	g_mutex_unlock( task->allocate_lock );
 
 	/* Process work units! Always tick, even if we are stopping, so the
 	 * main thread will wake up for exit. 
@@ -686,9 +684,7 @@ vips_task_run( gpointer data, gpointer user_data )
 			break;
 	} 
 
-	g_mutex_lock( task->allocate_lock );
 	VIPS_FREEF( g_object_unref, state );
-	g_mutex_unlock( task->allocate_lock );
 
 	/* We are exiting: tell the main thread. 
 	 */
@@ -815,8 +811,8 @@ vips_threadpool_push( const char *name, GFunc func, gpointer data )
  * is allocated to it to build the per-thread state. Per-thread state is used
  * by #VipsThreadpoolAllocate and #VipsThreadpoolWork to communicate.
  *
- * #VipsThreadState is a subclass of #VipsObject. Start functions are called
- * from allocate, that is, they are single-threaded.
+ * #VipsThreadState is a subclass of #VipsObject. Start functions can be
+ * executed concurrently.
  *
  * See also: vips_threadpool_run().
  *
@@ -899,9 +895,9 @@ vips_threadpool_push( const char *name, GFunc func, gpointer data )
  * The object returned by @start must be an instance of a subclass of
  * #VipsThreadState. Use this to communicate between @allocate and @work. 
  *
- * @allocate and @start are always single-threaded (so they can write to the 
- * per-pool state), whereas @work can be executed concurrently. @progress is 
- * always called by 
+ * @allocate is always single-threaded (so it can write to the 
+ * per-pool state), whereas @start and @work can be executed concurrently.
+ * @progress is always called by 
  * the main thread (ie. the thread which called vips_threadpool_run()).
  *
  * See also: vips_concurrency_set().
--- a/libvips/iofuncs/threadpool.c
+++ b/libvips/iofuncs/threadpool.c
@@ -55,8 +55,7 @@
 
  */
 
-/* 
-#define DEBUG_OUT_OF_THREADS
+/*
 #define VIPS_DEBUG
 #define VIPS_DEBUG_RED
  */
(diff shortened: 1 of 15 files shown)


jcupitt:
Exciting! I'll play with it.

lovell:
This is great, thanks Kleis. I've done some perf testing with this on Linux and also see a small performance improvement but from your testing it looks like Windows is where this will really shine.

jcupitt:
I've read it (only skimmed really, argh, must find more time) and tested nip2 and vipsdisp and they all seem fine.

You see a linux speedup on `time pytest`, since that makes a lot of short-lived pipelines. 

Before: real 1m55s, user 2m15s, sys 31s
After: real 1m41s, user 2m5s, sys 26s

So about 10 - 15% improvement. But that's probably an extreme case.

Let's merge and do any further tuning or polishing in master. Great work Kleis!

libvips/iofuncs/threadpool.c
```diff
@@ -999,6 +870,12 @@ vips__threadpool_init( void )
 
 	if( g_getenv( "VIPS_STALL" ) )
 		vips__stall = TRUE;
+
+	if( vips__concurrency == 0 )
+		vips__concurrency = vips__concurrency_get_default();
+
+	vips__pool = g_thread_pool_new( vips_thread_main_loop, NULL,
+		-1, FALSE, NULL );
```
jcupitt, line 878: Should this `-1` be `vips__concurrency`?

jcupitt: Ooop, no -1 is correct.

kleisauke: AFAIK, such limiting is only necessary for exclusive thread pools, which is [required on WASM](https://github.com/kleisauke/wasm-vips/blob/fc889f48c911ef7e728f0ddfcebdce80be72e70e/build/patches/vips-1492-emscripten.patch), for example.

I also tried this in the past for non-exclusive thread pools (see commits https://github.com/libvips/libvips/commit/c61c59991d890cdeef18f302fee7cab82e3093bc, https://github.com/libvips/libvips/commit/2683368978e907feae675f5147cb80b9412cc1a1 and https://github.com/libvips/libvips/commit/cdbc891205c4e15dcfb04bda74cd19569bf65d22) but I eventually reverted that with commit https://github.com/libvips/libvips/commit/9a9e52cbccd283c59731bbbcefdb458ce2603a2a as it was difficult to do correctly and it didn't work with `dzsave`. The discussion has some more details: https://github.com/libvips/libvips/discussions/2038#discussioncomment-376998.

kleisauke: Indeed, the previous `vips_g_thread_join` calls were replaced with a semaphore. Given that we are using the threadpool semantics of GLib now, do you think we should introduce wrappers (or environment variables) for the [`g_thread_pool_set_max_unused_threads`](https://developer.gnome.org/glib/stable/glib-Thread-Pools.html#g-thread-pool-set-max-unused-threads) (default 2) and [`g_thread_pool_set_max_idle_time`](https://developer.gnome.org/glib/stable/glib-Thread-Pools.html#g-thread-pool-set-max-idle-time) (default 15s) functions? This allows users to set the number of unused threads kept alive and the maximum idle time of a thread.

The `vips_threadpool_push` function was named this way because it's just a small wrapper around [`g_thread_pool_push`](https://developer.gnome.org/glib/stable/glib-Thread-Pools.html#g-thread-pool-push). I'm not sure if `vips_thread_new` is better, since it's not directly bound to a thread (as threads can be reused). For the same reason, I renamed `VipsThread` to `VipsTask` within `threadpool.c`. How about renaming it to `vips_task_push` instead? Perhaps we should make this function internal, because I'm not sure if it's intended to be used outside of libvips.

jcupitt: OK, let's make it internal, it'll stop me worrying about the name too, heh.

jcupitt: ... I couldn't resist renaming it to `vips__threadpool_execute()`.

kleisauke: Great, thanks!

PR #2256, opened 2021-05-17
Only build libjxl as a loadable module when it can be found

kleisauke:
This was noticed within the CI, see: https://github.com/libvips/libvips/runs/2599127347#step:12:417.


jcupitt:
Oopo! Good.

PR #2293, opened 2021-06-09
Delete Cairo context on rsvg_handle_render_cairo error

DarthSim:
Even if `rsvg_handle_render_cairo` failed, we should delete the Cairo context. Otherwise, the context will leak.


jcupitt:
Nice! Thank you.

I'll merge to 8.11 too, you're just in time.

PR #2305, opened 2021-06-15
Allow openjp2 version 2.3

remicollet:
Because this is the version available in Fedora / RHEL / CentOS ...

https://rpms.remirepo.net/rpmphp/zoom.php?rpm=openjpeg2

Only very recent version (e.g. Fedora 34) have version 2.4

Build and test suite passes with 2.3

Notice: security PoV, version 2.3.1 includes fix for various CVEs (and security of libopenjpg is probably not your problem)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -848,9 +848,9 @@ AC_ARG_WITH([libopenjp2],
 
 # 2.4 is the first one to have working threading and tiling
 if test x"$with_libopenjp2" != x"no"; then
-  PKG_CHECK_MODULES(LIBOPENJP2, libopenjp2 >= 2.4,
+  PKG_CHECK_MODULES(LIBOPENJP2, libopenjp2 >= 2.3,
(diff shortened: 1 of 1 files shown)


remicollet:
Notice: this work at least in Fedora 33 and RHEL 8

RHEL 7 not working as still have an old issue with poppler pulling openjpeg version 1, so this cannot work. (but this is already a known issue)

jcupitt:
Hi Remi,

OpenJPEG 2.3 didn't work with chrominance subsampling, I think, so I set 2.4 to be the minimum version. I think 2.4 has working threading too.

remicollet:
@ok, so closing 

FYI I search for this when digging for https://github.com/remicollet/remirepo/issues/178

In fact the real issue is fixed by or #2308 

So, I Will only use libopenjpg2 on Fedora >= 34

PR #2310, opened 2021-06-18
Add atan2 to math2

indus:
Ref: https://github.com/libvips/libvips/discussions/2299

### DIFF
--- a/cplusplus/include/vips/VImage8.h
+++ b/cplusplus/include/vips/VImage8.h
@@ -1468,6 +1468,35 @@ public:
 		return( math2_const( VIPS_OPERATION_MATH2_WOP, 
 			other, options ) );
 	}
+	
+	/**
+	 * Calculate atan2 of each pixel.
+	 */
+	VImage 
+	atan2( VImage other, VOption *options = 0 ) const
+	{
+		return( math2( other, VIPS_OPERATION_MATH2_ATAN2, options ) );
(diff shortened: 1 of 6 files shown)


jcupitt:
That's great! Thank you very much for this work.

I suppose there could be `atan2(constant, image)` as well, though I wonder if anyone would use it. Let's leave that until someone complains.

A test would be good. The `pow` one would be easy to copy-paste:

https://github.com/libvips/libvips/blob/master/test/test-suite/test_arithmetic.py#L89

`ChangeLog` and `configure.ac` could be updated too, but I can do that if you like.

indus:
I can try to add the test.

But I just realized that [this check](https://github.com/libvips/libvips/pull/2310/files#diff-42fde0354143dd8111a38c69989cab973d7d6e85b8094e0886645abfd4ddcd6eR156-R157) doesn't make sense in my case (I just copied it over from the polar function.)
Is there a general rule what range the output should have.

jcupitt:
`polar` is aiming to create angles in [0, 360), of course.

I'm not sure about atan2 -- it should probably follow the usual convention, but maybe consistency with the rest of libvips is more important?

jcupitt:
I added a test and merged to master. This will be in 8.12, with you credited in the changelog.

Thank you very much for doing this work, and sorry I sat on it for so long.

PR #2318, opened 2021-06-25
svgload: skip images with invalid dimensions

lovell:
The following SVG file:
```svg
<svg viewBox="0 0 0 0"></svg>
```
is currently loaded by libvips as a 1x1 image with non-ignorable warnings generated from within cairo:
```sh
$ VIPS_WARNING=0 vipsheader zero.svg 

(vipsheader:11559): GLib-GObject-WARNING **: 08:43:30.968: value "0" of type 'gint' is invalid or out of range for property 'width' of type 'gint'

(vipsheader:11559): GLib-GObject-WARNING **: 08:43:30.968: value "0" of type 'gint' is invalid or out of range for property 'height' of type 'gint'
zero.svg: 1x1 uchar, 4 bands, srgb, svgload
```

The "correct" dimensions of this image is 0x0, which is invalid in libvips, so the proposed change in this PR returns an error.

```sh
$ vipsheader zero.svg 
vipsheader: svgload: bad dimensions
```

As a bonus it also adds support for the case where `vips_image_pipelinev` returns false, which was previously ignored, and there's a unit test too.

The target branch is 8.11 as this feels more like a bug fix than a breaking change.

Originally reported at https://github.com/lovell/sharp/issues/2766


lovell:
Oh, when VIPS_LEAK is set, the tests are failing. Perhaps we shouldn't be returning the result of `vips_image_pipelinev` after all? I'll investigate.

lovell:
I had another look but I'm still unsure where the possible leak is coming from. Am I right in expecting the existing loader mechanics to take care of minimise/dispose when `vips_foreign_load_*_header` functions return false?

jcupitt:
This test program seems to work OK:

```C
/* compile with
 *
 *      gcc -g -Wall svgloadbuffer.c `pkg-config vips --cflags --libs`
 */

#include <vips/vips.h>

int
main( int argc, char **argv )
{       
        VipsImage *im;
        const char *svg = "<svg viewBox=\"0 0 0 0\"></svg>";

        if( VIPS_INIT( argv[0] ) )
                vips_error_exit( NULL );

        if( vips_svgload_buffer( (void *) svg, strlen( svg ), &im, NULL ) )
                vips_error_exit( NULL );
        if( vips_image_write_to_file( im, argv[1], NULL ) ) {
                g_object_unref( im );
                vips_error_exit( NULL );
        }
        g_object_unref( im );

        return( 0 );
}       
```

I see:

```
$ VIPS_LEAK=1 ./a.out x.v
svgload_buffer: bad dimensions
error buffer: svgload_buffer: bad dimensions
$ 
```

So I suppose it must be a pyvips bug, how odd. I'll investigate.

libvips/foreign/svgload.c
```diff
@@ -300,16 +307,15 @@ vips_foreign_load_svg_parse( VipsForeignLoadSvg *svg, VipsImage *out )
 	 */
         vips_image_pipelinev( out, VIPS_DEMAND_STYLE_FATSTRIP, NULL );
```
jcupitt, line 308: We might as well check the return here.

test/test-suite/test_foreign.py
```diff
@@ -895,6 +895,12 @@ def svg_valid(im):
         assert abs(im.width * 2 - x.width) < 2
         assert abs(im.height * 2 - x.height) < 2
 
+        try:
+            im = pyvips.Image.new_from_buffer(b'<svg viewBox="0 0 0 0"></svg>', "")
+            assert False
+        except pyvips.error.Error as err:
+            assert str(err).rstrip().endswith("bad dimensions")
```
jcupitt, line 902: SO says this is best done as:

```python
        with pytest.raises(Exception):
            svg = b'<svg viewBox="0 0 0 0"></svg>'
            im = pyvips.Image.new_from_buffer(svg, "")
```

libvips/foreign/svgload.c
lovell: I initially added a `return`, but then removed it to stay consistent with the behaviour of all the other loaders. Perhaps I should do a follow-up PR that ensures all calls to `vips_image_pipelinev` pass the status back?

jcupitt:
Ah, it works OK with pyvips 2.1.15, it's a bug @kleisauke fixed a while ago. I'll push a new pyvips version and that should do it.

libvips/foreign/svgload.c
jcupitt: Ah I see, you're right, sorry.

Let's leave it without the test (I don't think it can ever fail meaningfully) and do that fixup in master.

jcupitt:
OK, there's a new pyvips with the fix, if you ping this branch it should now pass.

jcupitt:
\o/

Issue #2317, opened 2021-06-24
upgrade from version 8.10 to 8.11 causing libheif to break


daverupp:
hello @jcupitt. 

## Bug report

When we build the new libvips verison 8.11 we are not able to generate jpeg's out of HEIC images anymore.

**Actual behavior**
we see following error when we try to generate a thumbnail out of heic:

./vipsthumbnail: unable to thumbnail Example1.heic
VipsForeignLoad: "Example1.heic" is not a known file format

**Environment**
(please complete the following information)
 - OS: ubuntu 18.04
 - Vips: libvips 8.11.4

**Additional context**

with the new build and looking at the config afterwards we see this difference compared to old version 8.10
HEIC/AVIF load/save with libheif:       yes (dynamic module: yes)

please also see the attached log file with compiler information etc...
[libvips_issue.txt](https://github.com/libvips/libvips/files/6710253/libvips_issue.txt)

do we need to build libvips 8.11 differently with libheif?

jcupitt:
Hi @daverupp,

I think this has been fixed in 8.11 branch, would you be able to test it?

https://github.com/libvips/libvips/archive/refs/heads/8.11.zip

The problem is the that the new module loading system can't always find the run-time loaders.

There are a few workarounds for 8.11.0: you can either configure libvips with `--disable-modules` to build all loaders directly into the libvips library, or you can set the `VIPSHOME` environment variable to point to the install prefix.

We plan to make 8.11.1 in a day or two would (hopefully) should fix this.

daverupp:
hello @jcupitt,

thank you very much. the workaround worked :) 
we will wait for the new release and then try again and hopefully everything works.

perfect. wish you a nice day.
cheers dave

jcupitt:
I don't suppose you'd be able to test 8.11.1 and confirm that fix works? I'd like an independent confirmation before I push it out.

daverupp:
let me quickly check. we will quickly try :)

daverupp:
we tried the latest master, there the issue is still the same.
Do you have a 8.11.1 release version we could test?

jcupitt:
Here's the head of the 8.11 branch as a tarball:

http://www.rollthepotato.net/~john/vips-8.11.1.tar.gz

It has a new feature: set the `VIPS_INFO` environment variable and it'll print some startup info about how it's attempting to find the modules. For example:

```
john@banana ~ $ VIPS_INFO=1 irb
irb(main):001:0> require 'vips'
[...]

PR #2323, opened 2021-06-28
Disable modules by default when building static libraries

kleisauke:
Might help: https://github.com/libvips/libvips/issues/2317. Users may still opt-in with `--enable-modules` when building static libraries.


jcupitt:
Nice!

PR #2335, opened 2021-07-02
Use autoconf macros for pkgconf where possible

kleisauke:
In addition, this PR also adds support for the `--with-pkgconfigdir=` argument while configuring libvips. This can be useful for Windows where one wants the dynamic modules installed (and be found) in the `bin/` directory, while still preserving the location of the pkg-config files (see for e.g. commit https://github.com/libvips/build-win64-mxe/commit/365870359dde8c7b971617d348ae8de41f5f690f).

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -18,7 +18,6 @@ EXTRA_DIST = \
 	depcomp \
 	README.md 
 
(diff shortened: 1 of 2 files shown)


jcupitt:
Oh nice, that's much neater.

PR #2342, opened 2021-07-08
VImage: reserve vector before push_back loop

neheb:
Avoids unnecessary resizes.

Found with clang-tidy's performance-inefficient-vector-operation

Signed-off-by: Rosen Penev <rosenp@gmail.com>

### DIFF
--- a/cplusplus/VImage.cpp
+++ b/cplusplus/VImage.cpp
@@ -767,6 +767,7 @@ std::vector<VImage>
 VImage::bandsplit( VOption *options ) const
 {
 	std::vector<VImage> b;
+	b.reserve(bands());
 
 	for( int i = 0; i < bands(); i++ )
 		b.push_back( extract_band( i ) );


cplusplus/VImage.cpp
```diff
@@ -767,8 +767,9 @@ std::vector<VImage>
 VImage::bandsplit( VOption *options ) const
 {
 	std::vector<VImage> b;
+	b.reserve(bands());
 
-	for( int i = 0; i < bands(); i++ )
+	for( int i = 0; i < int(b.capacity()); i++ )
```
lovell, line 772: I think this line needs to stay as-is. The call to `reserve()` might over-allocate so the vector capacity will not necessarily equal the number of bands.

neheb: Changed.

lovell (approved):

jcupitt:
Great, thank you!

PR #2343, opened 2021-07-09
clang-tidy fixes

neheb:


libvips/colour/LCh2UCS.c
```diff
@@ -94,7 +94,7 @@ vips_col_L2Lcmc( float L )
 	if( L < 16.0 )
 		Lcmc = 1.744 * L;
 	else
-		Lcmc = 21.75 * log( L ) + 0.3838 * L - 38.54;
+		Lcmc = 21.75 * logf( L ) + 0.3838 * L - 38.54;
```
jcupitt, line 97: logf() and friends are a C99 feature and many older compilers don't implement C99 fully. I think we'd need some tests in configure for this commit. For example, we test for inline (another C99 feature).

cplusplus/VImage.cpp
```diff
@@ -427,73 +427,69 @@ set_property( VipsObject *object, const char *name, const GValue *value )
 void
 VOption::set_operation( VipsOperation *operation )
 {
-	std::list<Pair *>::iterator i;
-
-	for( i = options.begin(); i != options.end(); ++i )
-		if( (*i)->input ) {
+	for( auto&& option : options )
```
jcupitt, line 430: auto and this form of fopr loop are C++11 features, which many older compilers do not support. At some point we should switch to requiring a C++11 compiler, but I don't know if we're at that point yet. 

Perhaps we should have a discussion about transitioning to C++11?

cplusplus/VImage.cpp
```diff
@@ -189,7 +189,7 @@ VOption::set( const char *name, const char *value )
 
 // input vips object (image, source, target, etc. etc.)
 VOption *
-VOption::set( const char *name, const VObject value )
+VOption::set( const char *name, const VObject& value )
```
jcupitt, line 194: `VObject` is designed to be a single pointer, so there's no performance advantage to passing by reference. 

Plus this would change the ABI, though perhaps that's not so important for C++.

cplusplus/VImage.cpp
```diff
@@ -773,7 +773,7 @@ VImage::bandsplit( VOption *options ) const
 VImage
 VImage::bandjoin( VImage other, VOption *options ) const
 {
-	VImage v[2] = { *this, other };
+	VImage v[2] = { *this, std::move(other) };
```
jcupitt, line 774: `VObject` is just a pointer, so there's no need to use `std::move`.

cplusplus/VImage.cpp
```diff
@@ -831,37 +831,37 @@ VImage::operator()( int x, int y ) const
 }
 
 VImage
-operator+( const VImage a, const VImage b )
+operator+( const VImage& a, const VImage& b )
```
jcupitt, line 832: `VImage` is just a pointer so there's no performance advantage here, and it will change the ABI.

cplusplus/VImage.cpp
```diff
@@ -64,7 +64,7 @@ to_vectorv( int n, ... )
 	va_list ap;
 
 	va_start( ap, n );
-	for( int i = 0; i < n; i++ )
+	for( size_t i = 0; i < n; i++ )
```
jcupitt, line 68: `n` is an int, so I think I should be an int as well.

cplusplus/VImage.cpp
```diff
@@ -93,7 +93,7 @@ negate( std::vector<double> vector )
 {
 	std::vector<double> new_vector( vector.size() );
 
-	for( unsigned int i = 0; i < vector.size(); i++ )
+	for( size_t i = 0; i < vector.size(); i++ )
```
jcupitt, line 99: I suppose strictly this should be something like:

```suggestion
for( std::vector<double>::size_type i = 0; i < vector.size(); i++ )
```

What do you think?

cplusplus/VImage.cpp
```diff
@@ -767,8 +767,9 @@ std::vector<VImage>
 VImage::bandsplit( VOption *options ) const
 {
 	std::vector<VImage> b;
+	b.reserve(bands());
 
-	for( int i = 0; i < bands(); i++ )
+	for( int i = 0; i < int(b.capacity()); i++ )
```
jcupitt, line 765: As @lovell says, this should probably stay as `bands()` since vectors are allowed to over-allocate.

jcupitt (commented):
Thank you for doing this work @neheb.

Some of your changes assume a compiler standard more recent than the libvips minimum (we aim for C89 and C++03). We should maybe have a discussion about bumping minimum compiler versions.

All the `VObject` classes are designed to just be pointers (they are compile-time only), so the pass by reference changes won't affect performance, and indeed could make code slower, conceivably. There are ABI issues to think about there too.

I left a few other small comments.

cplusplus/VImage.cpp
kleisauke: I think that `clang-tidy`'s [`performance-unnecessary-value-param`](https://clang.llvm.org/extra/clang-tidy/checks/performance-unnecessary-value-param.html) check is indeed wrong here. Somehow it detects that `VObject` (and derivatives) is expensive to copy. If you remove the `VObject` destructor (`~VObject()`) and copy constructor
(`VObject( const VObject &a ) : vobject( a.vobject )`) these warnings will disappear.

jcupitt: Perhaps pass by reference would avoid the ref/unref pair? It still changes the ABI though.

cplusplus/VImage.cpp
neheb: I disagree. It's being used as an array index. The compiler may or may not promote it to a 64-bit integer on 64-bit systems.

cplusplus/VImage.cpp
neheb: Done.

cplusplus/VImage.cpp
```diff
@@ -60,11 +60,12 @@ VIPS_NAMESPACE_START
...
 {
-	std::vector<double> vector( n );
+	std::vector<double> vector;
+	vector.reserve( n );
 	va_list ap;
 
 	va_start( ap, n );
-	for( int i = 0; i < n; i++ )
+	for( std::vector<double>::size_type i = 0; i < size_t(n); i++ )
 		vector[i] = va_arg( ap, double );
```
kleisauke, line 69: This is undefined behavior because `vector` starts as an empty vector and `vector.reserve( n );` only changes its `capacity()`, see: https://stackoverflow.com/a/13029377.

If you want to add elements with `push_back()` or `insert()`, use `reserve`.  If you want to access elements using `[]` or `at()`, use `resize` (or initialize the vector with the `vector(size_type N)` constructor).

cplusplus/VImage.cpp
jcupitt: But `n` really is an int, if you look at the prototype. There's no point using `size_t` as the array index, the extra range can never be used.

C/C++'s rules around signed / unsigned comparisons are an almost endless source of bugs. In my opinion you should always use simply use int, unless there is a very good reason not to.

jcupitt:
I changed a couple of things and merged, thank you again, @neheb 

I also added a `make tidy` rule that runs `clang-tidy`. I enabled `clang-analyzer-*,portability-*` and it runs cleanly now. I tried enabling the performance checks, but I think there are too many false-positives for it to be useful.

cplusplus/VImage.cpp
jcupitt: Good point, I switched it back to reserving space with the constructor.

PR #2354, opened 2021-07-18
added the jxl save functions to foreign.h

adil-benameur:
Just added the save function for jxl in foreign.h :
- vips_jxlsave
- vips_jxlsave_buffer
- vips_jxlsave_target


jcupitt:
Ooop, that was a daft oversight. Thank you very much!

jcupitt:
I pushed this to 8.11 too, so this fix should be in 8.11.3.

Issue #2309, opened 2021-06-16
There is no simple way to prevent the truncated JPEG loading


bozaro:
# Original issue
I want to deny uploading truncated JPEG to microservice: this error that can silenly creep in on different layers on the way from the user to the microservice.

The original idea was to use the `fail` flag on JPEG loading. This flag treats all warnings as errors. As a result, I get an error while loading a truncated JPEG, as I am trying.

Unfortunately, some warnings appear in pictures that do not strictly comply with the JPEG standard.
For example, some Samsung Android devices like Samsung Galaxy J1 SM-J120F make JPEG images with `Invalid SOS parameters for sequential JPEG` warning. These pictures are processed by many libraries without any complaints.

These pictures also stop loading with `fail` flag.

The situation is especially complicated by the fact that when paraller image processing, there is no rellation between warning message and image.

# My workaround
I have not found an easy way to solve this problem and use this logic:
 - check image format by file signature;
 - for JPEG make double check with directly `libjpeg` library calls with ignoring all warnings except JWRN_JPEG_EOF;
 - load image for processing with `libvips` with `fail == FALSE` flag.

# Possible solutions
I don't see good simple solution.
I see the following options:
 - ignore some errors inside `libvips` - this might solve my problem, but I don't think `libvips` should be doing this: the warning severity is subjective;
 - add some warnings to error handler - move the decision about the problem severity to the developer side.

jcupitt:
Hi @bozaro,

Yes, it's a tricky issue. Some truncations will not cause errors or warnings, for example (try removing the final few bytes of a JPEG). Every load library takes a slightly different view of what counts as a warning and what is an error, so it's especially hard for libvips to present them in a uniform way.

You probably saw the bit of code in the libvips JPEG loader:

https://github.com/libvips/libvips/blob/master/libvips/foreign/jpeg2vips.c#L252

Perhaps we need a new loader flag (`fail-on-truncated`?) just for this case, since truncated files are so common. It could default to the value of `fail`, for compatibility. Does that sound reasonable?

bozaro:
It seems to me appropriate would be two mechanisms complementary to each other:
 - warning callback for move the decision about the problem severity to the developer side;
 - `fail-on-truncated` for simple prevent truncated file shortcut.

But I am not using `libvips` directly and I can't imagine how well it fits into his internal structure.

jcupitt:
[...]

PR #2360, opened 2021-07-24
add fail-on : better control over loader error handling

jcupitt:
done jpeg and it seems to work

see https://github.com/libvips/libvips/issues/2309


jcupitt:
The idea here is to give more control over the behaviour of loaders when they see an error, and specifically to make detecting truncated images (a very common case) simple.

Thinking about this in reverse, what kinds of load failure can libvips detect in all load types? 

- *warning* Most loaders have a warning mechanism for non-fatal errors.
- *error* All loaders have a way of signalling serious decode errors.
- *truncation* We handle the read call, so libvips can detect image truncation itself.

Loaders vary in what counts as an error and what a warning, so for example image truncation (very serious) is only a warning for some.

This PR adds a new load parameter called `fail-on` (and deprecates `fail`) which sets the level of read error which will cause the loader to stop and return an error code.

```
/**                     
 * VipsFailOn:
 * @VIPS_FAIL_ON_NONE: ignore all errors and warnings
 * @VIPS_FAIL_ON_WARNING: stop on load warning
 * @VIPS_FAIL_ON_ERROR: stop on load error 
 * @VIPS_FAIL_ON_TRUNCATED: stop on image truncated
 *      
 * What kind of load failure to stop on, from never stop to only stop on image
 * truncation. 
 */
typedef enum {
        VIPS_FAIL_ON_NONE,
        VIPS_FAIL_ON_WARNING,
        VIPS_FAIL_ON_ERROR,
        VIPS_FAIL_ON_TRUNCATED,
        VIPS_FAIL_ON_LAST,
} VipsFailOn;
```

The enums are in order, so loaders can do eg.:

```C
    if( load->fail_on >= VIPS_FAIL_ON_ERROR ) {
        ... error
    }
```

And they will fail if the fail level is set to error or truncated (a type of error).

I've done jpeg, but all loaders will need work.

Behaviour with this PR:

```
john@kiwi:~/pics$ vips avg truncated.jpg 
97.240126

(vips:45601): VIPS-WARNING **: 11:06:31.796: read gave 2 warnings

(vips:45601): VIPS-WARNING **: 11:06:31.796: VipsJpeg: Premature end of JPEG file

john@kiwi:~/pics$ vips avg truncated.jpg[fail-on=warning]

(vips:45623): VIPS-WARNING **: 11:07:06.256: error in tile 0 x 0

(vips:45623): VIPS-WARNING **: 11:07:06.256: error in tile 0 x 8
VipsJpeg: Premature end of JPEG file
VipsJpeg: out of order read at line 0
john@kiwi:~/pics$ vips avg truncated.jpg[fail-on=error]
VipsJpeg: Premature end of input file
john@kiwi:~/pics$ vips avg truncated.jpg[fail-on=truncated]
VipsJpeg: Premature end of input file
john@kiwi:~/pics$ vips avg truncated.jpg[fail-on=none]
97.240126

(vips:45656): VIPS-WARNING **: 11:07:38.120: read gave 2 warnings

(vips:45656): VIPS-WARNING **: 11:07:38.120: VipsJpeg: Premature end of JPEG file
```

jcupitt:
Forgot to say:

- The old `fail` flag is deprecated, but the member is still there internally and is inited to `fail_on >= WARNING`, so everything still works as before.
- The size of the `VipsForeignLoad` base class has changed, so we lose binary compat for loadable load modules. But I don't think we ever promised binary compat for subclasses of libvips objects, so it's probably OK.
-  We can't just put the fail-on-eof functionality into `VipsSource`, since some loaders (eg. CSV) need EOF to find the end of the data. It has to be done separately for each loader.
- I'll convert a few more and see if any obvious problems emerge.

libvips/foreign/foreign.c
```diff
@@ -1061,6 +1063,10 @@ vips_foreign_load_build( VipsObject *object )
 	if( sequential )
 		load->nocache = TRUE;
 
+	/* For compat, default the deprecated "fail" field to anything.
+	 */
+	load->fail = load->fail_on > VIPS_FAIL_ON_NONE;
```
kleisauke, line 1068: What if someone still sets the deprecated `fail` param to `true`, without setting the new `fail-on` param? Perhaps something like this would work?:
```suggestion
	if( !vips_object_argument_isset( object, "fail" ) )
		load->fail = load->fail_on > VIPS_FAIL_ON_NONE;
	else if( !vips_object_argument_isset( object, "fail-on" ) )
		load->fail_on = load->fail ? VIPS_FAIL_ON_WARNING : VIPS_FAIL_ON_NONE;
```
(untested)

kleisauke: Also for compat, perhaps the `load->fail_on > VIPS_FAIL_ON_NONE` line should be `load->fail_on >= VIPS_FAIL_ON_WARNING`?

kleisauke (approved):
LGTM! I added a comment about improving compatibility with the old `fail` parameter.

libvips/foreign/foreign.c
jcupitt: It's extremely confusing :( I kept getting these tests backwards.

`fail_on` controls the sensitivity of loaders to error conditions, from 0 meaning very insensitive, to 3 meaning extremely sensitive indeed. 

`fail_on > none` means any setting other than the default (ie. insensitive). `fail_on > warn` would only enable `fail` if `fail_on` was at the highest sensitivity. I think.

jcupitt: ... and good spot on `fail` compatibility!

kleisauke: No problem! With my last comment I meant that the now deprecated `fail` was at the highest sensitivity (I think). See for example:
https://github.com/libvips/libvips/blob/5cd676a94527c4c1aabdcce534b23d193675fcc5/libvips/foreign/jpeg2vips.c#L811
This was `jpeg->fail` before this PR, hence I thought perhaps we should initialize the compat here:
https://github.com/libvips/libvips/blob/5cd676a94527c4c1aabdcce534b23d193675fcc5/libvips/foreign/foreign.c#L1069
With `load->fail = load->fail_on >= VIPS_FAIL_ON_WARNING` instead. We could also _not_ initialize `load->fail` at all (if we are sure we are not using it anywhere).

jcupitt: Oh heh that's true, we can just get rid of `fail` now everything is converted. I'll do that.

jcupitt:
\o/

PR #2361, opened 2021-07-25
WIP -- experiment with ci install to user-writeable prefix

jcupitt:
Hopefully to make macos-m1 testing work.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,34 +14,44 @@ jobs:
         os: [ ubuntu-20.04 ]
         sanitize: [ false ]
         prefix: "/usr/local"
+        shell: bash
+        sudo: "sudo"
         build:
-          - { cc: gcc, cxx: g++, linker: ld, shell: bash }
+          - { cc: gcc, cxx: g++, linker: ld }
         include:
           - name: "Linux x64 (Ubuntu 20.04) - Clang 10 with ASan and UBSan"
             os: ubuntu-20.04
             sanitize: true
             prefix: "/usr/local"
-            build: { cc: clang-10, cxx: clang++-10, linker: ld.lld-10, shell: bash }
+            shell: bash
+            sudo: "sudo"
+            build: { cc: clang-10, cxx: clang++-10, linker: ld.lld-10 }
+
           - name: "macOS (10.15) - Xcode 12.3"
             os: macos-10.15
             sanitize: false
             prefix: "/usr/local"
-            build: { cc: clang, cxx: clang++, linker: ld.lld, shell: bash }
+            shell: bash
+            sudo: "sudo"
+            build: { cc: clang, cxx: clang++, linker: ld.lld }
+
           - name: "macOS (10.15) - Xcode 12.3 - M1"
             os: macos-m1
             sanitize: false
             prefix: "/Users/Administrator/vips"
-            build: { cc: clang, cxx: clang++, linker: ld.lld, shell: "/usr/bin/arch -arch arm64e /bin/bash -l {0}" }
+            shell: "/usr/bin/arch -arch arm64e /bin/bash -l {0}"
+            sudo: "sudo"
+            build: { cc: clang, cxx: clang++, linker: ld.lld }
 
     defaults:
       run:
-        shell: ${{ matrix.build.shell }} 
+        shell: ${{ matrix.shell }} 
+
+    env:
+      PREFIX: ${{ matrix.prefix }}
+      SUDO: ${{ matrix.sudo }}
 
     steps:
-      - name: Set prefix
-        env: 
-          PREFIX: ${{ maxtrix.prefix }}
-
       - name: Checkout code
         uses: actions/checkout@v2
 
@@ -131,6 +141,7 @@ jobs:
       - name: Configure libvips
         run:
           ./autogen.sh
+            --prefix=$PREFIX
             --disable-dependency-tracking
             --disable-deprecated || (cat config.log && exit 1)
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -13,27 +13,46 @@ jobs:
...
             sanitize: false
-            build: { cc: clang, cxx: clang++, linker: ld.lld, shell: "/usr/bin/arch -arch arm64e /bin/bash -l {0}" }
+            build: { cc: clang, cxx: clang++, linker: ld.lld }
+            shell: "/usr/bin/arch -arch arm64e /bin/bash -l {0}"
+            prefix: "/Users/Administrator/vips"
+            sudo: ""
 
     defaults:
       run:
         shell: ${{ matrix.build.shell }}
```
kleisauke, line 48:
```suggestion
        shell: ${{ matrix.shell }} 
```

.github/workflows/ci.yml
```diff
@@ -13,27 +13,46 @@ jobs:
...
     defaults:
       run:
         shell: ${{ matrix.build.shell }} 
 
     steps:
+      - name: Set prefix
+        env:
+          PREFIX: ${{ matrix.prefix }}
+          SUDO: ${{ matrix.sudo }}
+
```
kleisauke, line 55:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -123,6 +142,7 @@ jobs:
       - name: Configure libvips
         run:
           ./autogen.sh
+            --prefix=$PREFIX
```
kleisauke, line 145:
```suggestion
            --prefix=${{ matrix.prefix }}
```

kleisauke (commented):
I think these suggestions should fix the CI failures.

.github/workflows/ci.yml
kleisauke: (it was failing with "every step must define a `uses` or `run` key")

jcupitt: Oh of course. I saw the error, but there was no line number. I've moved the env to the job rather than the step.

.github/workflows/ci.yml
jcupitt: You're right, I forgot to move that out of build.

.github/workflows/ci.yml
kleisauke: Ah yes, that's much neater. Perhaps we can move this step to there as well?
https://github.com/libvips/libvips/blob/ea2ad5ffb03871a7747bb4ff3bf447c6bc32bf3a/.github/workflows/ci.yml#L103-L108

kleisauke:
The `pull_request` event segfaults with a stacktrace similar to the `push` event, the later CI build contains some more details:
```
OSError: cannot load library 'libvips.42.dylib': dlopen(libvips.42.dylib, 2): image not found.  Additionally, ctypes.util.find_library() did not manage to locate a library called 'libvips.42.dylib'
```

Is Apple's [System Integrity Protection](https://en.wikipedia.org/wiki/System_Integrity_Protection) (SIP) enabled on the M1 runner? If so, any dynamic linker (`dyld`) environment variables, such as `DYLD_LIBRARY_PATH`, are purged when launching protected processes.

(Commit https://github.com/kleisauke/libvips/commit/30feca305316940616873b3ee04857f2a1aa1c34 might help to debug this further)

jcupitt:
Could be I guess. Though SIP will only block env vars when launching the Apple's Python or Ruby, and I think we're using the homebrew one (I think?). We certainly ought to be using the homebrew one if we're not.

Let's merge your patch and try again. Yes, I wondered about using the `env:` section more.

kleisauke:
I think the latest CI log indicates that SIP is enabled (it can be verified with running `env | grep DYLD_LIBRARY_PATH`). 

How about using `/Users/Administrator` as a prefix instead of `/Users/Administrator/vips`? It looks like `ctypes` will search in `os.path.expanduser("~/lib")`, so I think that should work.
https://github.com/python/cpython/blob/ac811f9b5a68ce8756911ef2c8be83b46696018f/Lib/ctypes/macholib/dyld.py#L30
```bash
python3 -c "import ctypes.macholib.dyld; print(ctypes.macholib.dyld.DEFAULT_LIBRARY_FALLBACK)"
```

If that works, there's no need to set the `DYLD_LIBRARY_PATH` environment variable.

.github/workflows/ci.yml
```diff
@@ -13,25 +13,49 @@ jobs:
...
+
+    env:
+      CC: ${{ matrix.build.cc }}
+      CXX: ${{ matrix.build.cxx }}
+      LD: ${{ matrix.build.linker }}
+      CPPFLAGS: -Wall
+      PREFIX: ${{ matrix.prefix }}
+      SUDO: ${{ matrix.sudo }}
+      PKG_CONFIG_PATH: "$PREFIX/lib/pkgconfig"
+      DYLD_LIBRARY_PATH: "$PREFIX/lib"
```
kleisauke, line 58: I just noticed that shell variables are not working here, see for example this CI log:
https://github.com/kleisauke/libvips/runs/3351200368#step:17:19

Though, this seems to work properly:
```suggestion
      PKG_CONFIG_PATH: "${{ matrix.prefix }}/lib/pkgconfig"
      DYLD_LIBRARY_PATH: "${{ matrix.prefix }}/lib"
```
(see https://github.com/kleisauke/libvips/runs/3351365557#step:17:19)

kleisauke:
... ignore my earlier comments about SIP, I think it's because of GitHub Actions weird variable expansion in the `env:` section.

jcupitt:
I had a quick read and I think everything useful here has been merged in other PRs. I'll close.

PR #2367, opened 2021-07-29
pngsave: expose quantisation effort setting

lovell:
This is a rather useful setting for control over the creation of palette-based PNGs (and other palette-based "future" image formats) as the performance varies hugely between the slowest and fastest speeds.

```sh
$ time vips copy in.jpg out.png[palette]

real	0m2.641s
user	0m2.609s
sys	0m0.020s

$ time vips copy in.jpg out.png[palette,speed=1]

real	0m8.089s
user	0m8.072s
sys	0m0.017s

$ time vips copy in.jpg out.png[palette,speed=10]

real	0m0.676s
user	0m0.650s
sys	0m0.027s
```

### DIFF
--- a/libvips/foreign/pforeign.h
+++ b/libvips/foreign/pforeign.h
@@ -187,7 +187,7 @@ int vips__png_write_target( VipsImage *in, VipsTarget *target,
 	int compress, int interlace, const char *profile,
 	VipsForeignPngFilter filter, gboolean strip,
 	gboolean palette, int Q, double dither,
-	int bitdepth );
+	int bitdepth, int effort );
 
 /* Map WEBP metadata names to vips names.
  */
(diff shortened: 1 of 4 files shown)


jcupitt:
This is a tiny comment, but how about calling it `effort` instead, meaning the amount of effort it puts into finding the best palette? You'd need to reverse the range, of course, since `effort=10` would be the slowest.

The advantage would be that other savers use `effort` as well, so it's maybe more consistent.

I notice webpsave calls it `reduction-effort`, annoyingly. Perhaps we should change that too. heifsave uses `speed`, but calls it effort in the parameter description.

lovell:
Thanks, this makes sense, I've reversed the logic.

jcupitt:
Great!

PR #2381, opened 2021-08-06
Add gifsave operation, depends on cgif and libimagequant

lovell:
This PR adds support for GIF encoding via the MIT-licenced [cgif](https://github.com/dloebl/cgif), which is performant, well maintained and tested, plus supports streaming. The aim is that this new saver will be easier/simpler/faster than the *magick-based approach.

It relies on libimagequant to generate the (global) palette and quantise, as we already do for 8-bit PNGs. The palette is intentionally limited to a maximum of 255 colours, which leaves cgif one extra colour for it to optimise frames.

The cgif library is not yet "officially" available via the usual package managers but I've created https://github.com/lovell/cgif-packaging to help start that process.

TODO (there may be more):

- [x] cgif: Improved transparency / disposal method - see https://github.com/dloebl/cgif/pull/5
- [x] cgif: Make available via package managers - see https://github.com/lovell/cgif-packaging
- [x] libvips: Unit tests
- [x] libvips/oss-fuzz: Add cgif as a dependency and create a fuzz test for the encoder - see https://github.com/google/oss-fuzz/pull/6241
- [ ] libvips: support frame-local palettes (this might be better as a follow up PR)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,11 +43,12 @@ jobs:
           DEBIAN_FRONTEND: noninteractive
         run: sudo -E apt-get update -qq -o Acquire::Retries=3
 
-      - name: Add libheif PPA
+      - name: Add Ubuntu PPAs
         if: contains(matrix.os, 'ubuntu')
         run: |
           sudo add-apt-repository ppa:strukturag/libde265
           sudo add-apt-repository ppa:strukturag/libheif
+          sudo add-apt-repository ppa:lovell/cgif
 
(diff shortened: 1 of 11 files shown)


jcupitt:
LGTM.

Before this patch:

```
$ /usr/bin/time -f %M:%e vips copy neonatal.gif[n=-1] x.gif
185040:0.75
```

And after:

```
$ /usr/bin/time -f %M:%e vips copy neonatal.gif[n=-1] x.gif
50756:0.14
```

So about 5x faster and 4x less memory. Nice work!

PR #2419, opened 2021-08-27
Introduce unlimited flag, enabling unlimited text chunks in PNGs

joshuamsager:
Following this discussion: https://github.com/libvips/libvips/discussions/2418

Adds the flag: `unlimited`, which allows for unlimited-many text chunks when reading PNGs.

Open to feedback

### DIFF
--- a/libvips/foreign/pngload.c
+++ b/libvips/foreign/pngload.c
@@ -59,7 +59,7 @@ typedef struct _VipsForeignLoadPng {
 	 */
 	VipsSource *source;
 
-	/* Allow any number of text chunks.
+	/* remove all denial of service limits.
 	 */
 	gboolean unlimited;
 
@@ -166,7 +166,7 @@ vips_foreign_load_png_class_init( VipsForeignLoadPngClass *class )
 
 	VIPS_ARG_BOOL( class, "unlimited", 23,
 		_( "Unlimited" ),
-		_( "Allow any number of text chunks" ),
+		_( "Remove all denial of service limits" ),
 		VIPS_ARGUMENT_OPTIONAL_INPUT,
 		G_STRUCT_OFFSET( VipsForeignLoadPng, unlimited ),
 		FALSE );
@@ -409,13 +409,17 @@ vips_foreign_load_png_buffer_init( VipsForeignLoadPngBuffer *buffer )
  *
  * Optional arguments:
  *
- * * @unlimited: %gboolean, allow any number of text chunks
+ * * @unlimited: %gboolean, remove all denial of service limits
  *
  * Read a PNG file into a VIPS image. It can read all png images, including 8-
  * and 16-bit images, 1 and 3 channel, with and without an alpha channel.
  *
  * Any ICC profile is read and attached to the VIPS image. It also supports
  * XMP metadata.
+ *
+ * By default, the PNG loader limits the number of text and data chunks to 
+ * block some denial of service attacks. Set @unlimited to disable these 
+ * limits.
  *
  * See also: vips_image_new_from_file().
  *
@@ -443,7 +447,7 @@ vips_pngload( const char *filename, VipsImage **out, ... )
  *
  * Optional arguments:
  *
- * * @unlimited: %gboolean, allow any number of text chunks
+ * * @unlimited: %gboolean, Remove all denial of service limits
  *
  * Exactly as vips_pngload(), but read from a PNG-formatted memory block.
  *
@@ -482,7 +486,7 @@ vips_pngload_buffer( void *buf, size_t len, VipsImage **out, ... )
  *
  * Optional arguments:
  *
- * * @unlimited: %gboolean, allow any number of text chunks
+ * * @unlimited: %gboolean, Remove all denial of service limits
  *
  * Exactly as vips_pngload(), but read from a source. 
  *
--- a/libvips/foreign/pngload.c
+++ b/libvips/foreign/pngload.c
@@ -162,13 +162,6 @@ vips_foreign_load_png_class_init( VipsForeignLoadPngClass *class )
 	load_class->header = vips_foreign_load_png_header;
 	load_class->load = vips_foreign_load_png_load;
 
-	VIPS_ARG_BOOL( class, "unlimited", 23,
-		_( "Unlimited" ),
-		_( "Allow any number of text chunks" ),
-		VIPS_ARGUMENT_OPTIONAL_INPUT,
-		G_STRUCT_OFFSET( VipsForeignLoadPng, unlimited ),
-		FALSE );
-
 }
 
 static void
--- a/libvips/foreign/pngload.c
+++ b/libvips/foreign/pngload.c
@@ -85,9 +85,10 @@ static VipsForeignFlags
 vips_foreign_load_png_get_flags_source( VipsSource *source )
 {
 	VipsForeignFlags flags;
+	VipsForeignLoadPng *png = (VipsForeignLoadPng *) load;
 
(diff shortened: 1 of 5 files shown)


joshuamsager:
An alternative to this approach is to allow an optional argument `max-chunks` be provided instead of `unlimited`. -1 would behave the same as unlimited and all other values will replace the current value *10*

jcupitt:
Oh very nice. I'll have a look, thank you for doing this work!

jcupitt:
You're very close for `libspng` load. Just move the arg def for `unlimited` into the base class, rather than having three copies of it. The trick is to add:

```C
        gobject_class->dispose = vips_foreign_load_png_dispose;
        gobject_class->set_property = vips_object_set_property;
        gobject_class->get_property = vips_object_get_property;
```

to `vips_foreign_load_png_class_init()` to enable the libvips arg mechanism for that class.

It'd be a good idea to remove other limits too. This code:

```C
        /* Set limits to avoid decompression bombs. Set chunk limits to 60mb
         * -- we've seen 50mb XMP blocks in the wild.
         *
         * No need to test the decoded image size -- the user can do that if
         * they wish.
         */
        spng_set_image_limits( png->ctx, VIPS_MAX_COORD, VIPS_MAX_COORD );
        spng_set_chunk_limits( png->ctx, 60 * 1024 * 1024, 60 * 1024 * 1024 );
```

should probably be within an if as well.

Old fashioned libpng load is more complex, unfortunately. That has to support a function call interface as well (as part of vips7 compatibility), and that means the loader is split into two parts, with the OO interface layered on top of the function one.

1. `VIPS_ARG_BOOL` for `unlimited` needs to be in the base class (ie. `vips_foreign_load_png_class_init`).
2. You'll need to pass `png->unlimited` to `vips__png_header_source()` and `vips__png_read_source()` as an extra param.
3. Those funcs will need to pass `unlimited` to `read_new()`, and that'll need to save it into a new member of the `Read` struct.
4. Finally the load code in `vipspng.c` need to use the `unlimited` flag much as the libspng loader does.

joshuamsager:
Thanks for the help @jcupitt. When you have a chance, would you mind verifying that I did this correctly? I wasn't sure if:
```
gobject_class->set_property = vips_object_set_property;
gobject_class->get_property = vips_object_get_property;
```

was also necessary in `pngload.c` in `vips_foreign_load_png_class_init`.

libvips/foreign/pngload.c
```diff
@@ -81,9 +85,10 @@ static VipsForeignFlags
 vips_foreign_load_png_get_flags_source( VipsSource *source )
 {
 	VipsForeignFlags flags;
+	VipsForeignLoadPng *png = (VipsForeignLoadPng *) load;
```
joshuamsager, line 88: don't think this is right but i'm not sure how else to pass the unlimited flag down to `read_new` in `vips__png_isinterlaced_source`

libvips/foreign/vipspng.c
```diff
@@ -822,7 +825,7 @@ vips__png_isinterlaced_source( VipsSource *source )
 
 	image = vips_image_new();
 
-	if( !(read = read_new( source, image, TRUE )) ) { 
+	if( !(read = read_new( source, image, TRUE, FALSE )) ) {
```
joshuamsager, line 828: Is there a more appropriate alternative for this?

jcupitt: Unfortunately not :(

libvips/foreign/pngload.c
```diff
@@ -395,6 +407,10 @@ vips_foreign_load_png_buffer_init( VipsForeignLoadPngBuffer *buffer )
  * @out: (out): decompressed image
  * @...: %NULL-terminated list of optional named arguments
  *
+ * Optional arguments:
+ *
+ * * @unlimited: %gboolean, allow any number of text chunks
```
jcupitt, line 412: Perhaps "remove all denial of service limits".

libvips/foreign/pngload.c
```diff
@@ -395,6 +407,10 @@ vips_foreign_load_png_buffer_init( VipsForeignLoadPngBuffer *buffer )
...
+ *
  * Read a PNG file into a VIPS image. It can read all png images, including 8-
  * and 16-bit images, 1 and 3 channel, with and without an alpha channel.
  *
```
jcupitt, line 416: Add a para, eg. "By default, the PNG loader limits the number of text and data chunks to block some denial of service attacks. Set @unlimited to disable these limits."

libvips/foreign/spngload.c
```diff
@@ -63,6 +63,10 @@ typedef struct _VipsForeignLoadPng {
 	 */
 	VipsSource *source;
 
+	/* Allow any number of text chunks.
+	 */
```
jcupitt, line 67: Perhaps "Remove DoS limits."

libvips/foreign/spngload.c
```diff
@@ -249,7 +253,8 @@ vips_foreign_load_png_set_header( VipsForeignLoadPng *png, VipsImage *image )
 		/* Very large numbers of text chunks are used in DoS
 		 * attacks.
 		 */
-		if( n_text > 10 ) {
+
+		if( !png->unlimited && n_text > 10 ) {
```
jcupitt, line 257: Let's change this to 20. It seems 10 is too low for many people.

libvips/foreign/pngload.c
```diff
@@ -158,6 +164,12 @@ vips_foreign_load_png_class_init( VipsForeignLoadPngClass *class )
 	load_class->header = vips_foreign_load_png_header;
 	load_class->load = vips_foreign_load_png_load;
 
+	VIPS_ARG_BOOL( class, "unlimited", 23,
+		_( "Unlimited" ),
+		_( "Allow any number of text chunks" ),
```
jcupitt, line 169: Perhaps "Remove all denial of service limits".

libvips/foreign/vipspng.c
```diff
@@ -564,7 +567,7 @@ png2vips_header( Read *read, VipsImage *out )
 		/* Very large numbers of text chunks are used in DoS
 		 * attacks.
 		 */
-		if( num_text > 10 ) {
+		if( !read->unlimited && num_text > 10 ) {
```
jcupitt, line 570: 20 again. Maybe add a `#define MAX_PNG_TEXT_CHUNKS (20)` to `pforeign.h` (the library-private `foreign/` header).

libvips/foreign/spngload.c
```diff
@@ -662,6 +671,12 @@ vips_foreign_load_png_class_init( VipsForeignLoadPngClass *class )
 	load_class->header = vips_foreign_load_png_header;
 	load_class->load = vips_foreign_load_png_load;
 
+	VIPS_ARG_BOOL( class, "unlimited", 23,
+		_( "Unlimited" ),
+		_( "Allow any number of text chunks" ),
```
jcupitt, line 676: Matching change.

jcupitt (changes requested):
It looks great! You've got the hang of libvips very quickly.

I added some small notes on docs and naming.

jcupitt (approved):
LGTM!

jcupitt:
Merged. Thank you for doing this work, @joshuamsager!

We normally avoid API changes in released versions, but I think this needs to be backported to 8.11. I'll cherrypick and push there too.

jcupitt:
It'll be in 8.11.4. I made some very small formatting changes too, since libvips has a 80 col line limit, and added a note to the changelog crediting you.

PR #2420, opened 2021-08-29
Ensure VIPS_CODING_ERROR is handled for vips images

lovell:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=37820


lovell:
Another possible 8.11 backport?

PR #2421, opened 2021-08-29
Handle error values when guessing interpretation

lovell:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=37815 and https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=37812


jcupitt:
LGTM!

lovell:
Thanks, should we backport this to 8.11 also?

jcupitt:
Yes, it's a potential DoS I suppose. Let's do #2420 as well.

jcupitt:
OK, backported to 8.11.

PR #2424, opened 2021-09-04
heifload: switch demand to thin strip

lovell:
At the moment, unrecoverable errors that occur during HEIF decoding are presented as warnings and subsequent tiles are still requested.

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=38115 provides an example image that takes minutes to process before ultimately failing. I suspect a few of the other oss-fuzz timeout-related issues might be linked to this too.

By switching to thin strip, the decoding error is detected and handled as soon as possible, however I'm unsure if this is the right way to fix the problem e.g. why is small tile demand not propagating errors?

### DIFF
--- a/libvips/foreign/heifload.c
+++ b/libvips/foreign/heifload.c
@@ -621,7 +621,8 @@ vips_foreign_load_heif_set_header( VipsForeignLoadHeif *heif, VipsImage *out )
 	/* FIXME .. we always decode to RGB in generate. We should check for
 	 * all grey images, perhaps. 
 	 */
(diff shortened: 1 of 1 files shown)


jcupitt:
I had a look and it's because the loop in vips_tile_cache_gen() doesn't stop early on error, it'll wait for the end of the cache tile. In this case, the whole image fits into one tile, so processing only stops at the end of the image.

Let's switch to thinstrip, as you suggest, and I'll add an early term condition to the cache loop as well.

PR #2429, opened 2021-09-09
gifsave: expose bitdepth option

lovell:
This property provides control over the maximum number of colours in the output image palette, making the API more closely match that of `pngsave`.

Given the majority of CPU time when encoding GIF images is spent finding the nearest colour in the palette, lowering the bitdepth from 8 to 7 halves the maximum number of colours from 255 to 127 and reduces encoding time by ~20%.


jcupitt:
Nice!

PR #2445, opened 2021-09-24
Cgifsave per frame cmap

jcupitt:
This patch makes the GIF a frame at a time, and only redoes the palette optimisation if the new frame is significantly different from the frame it last ran the optimisation on.

It detects differences by simply summing pixels and checking of the sum has changed by more than 20%. I've no idea if this is good enough, we'll need some testing.

Pros:

- Lower memory use, esp. with huge GIFs, since it doesn't need to keep the whole image in memory
- The write part (palette, dither, compress) runs in a background thread, so it can overlap load and save
- It only reoptimises the palette occasionally, which helps performance too

Cons:

- Is the change detector good enough?
- GIFs are larger, since we don't use a global colourmap, but a local one for each frame

Benchmark with master:

```
$ /usr/bin/time -f %M:%e vips copy 3198.gif[n=-1] x.gif
94332:1.45
```

And with this branch:

```
$ /usr/bin/time -f %M:%e vips copy 3198.gif[n=-1] x.gif
49376:1.11
```

@lovell any thoughts?

### DIFF
--- a/libvips/foreign/cgifsave.c
+++ b/libvips/foreign/cgifsave.c
@@ -30,7 +30,9 @@
 
  */
 
+/*
 #define DEBUG_VERBOSE
+ */
 
 #ifdef HAVE_CONFIG_H
 #include <config.h>
@@ -60,6 +62,7 @@ typedef struct _VipsForeignSaveCgif {
 
 	/* Derived write params.
 	 */
+	VipsImage *in;				/* Not a reference */
 	gboolean has_transparency;
 	int *delay;
 	int delay_length;
@@ -79,6 +82,7 @@ typedef struct _VipsForeignSaveCgif {
 	liq_attr *attr;
 	liq_image *input_image;
 	liq_result *quantisation_result;
+	const liq_palette *lp;
 
 	/* The current colourmap, updated on a significant frame change.
 	 *
@@ -140,9 +144,13 @@ vips_foreign_save_cgif_write_frame( VipsForeignSaveCgif *cgif )
 {
 	VipsObjectClass *class = VIPS_OBJECT_GET_CLASS( cgif );
 	VipsRect *frame_rect = &cgif->frame->valid;
-	int page_index = frame_rect->top / cgif->frame->im->Ysize;
+	int page_index = frame_rect->top / frame_rect->height;
+	/* We know this fits in an int since we limit frame size.
+	 */
 	int n_pels = frame_rect->height * frame_rect->width;
 	guint max_sum = 256 * n_pels * 3;
+	VipsPel *frame_bytes = 
+		VIPS_REGION_ADDR( cgif->frame, 0, frame_rect->top );
 
 	VipsPel * restrict p;
 	guint sum;
@@ -157,19 +165,28 @@ vips_foreign_save_cgif_write_frame( VipsForeignSaveCgif *cgif )
 	 */
 	VIPS_FREEF( liq_image_destroy, cgif->input_image );
 	cgif->input_image = liq_image_create_rgba( cgif->attr,
-		VIPS_REGION_ADDR( cgif->frame, 0, 0 ), 
-		frame_rect->width, frame_rect->height, 0 );
+		frame_bytes, frame_rect->width, frame_rect->height, 0 );
+
+	/* Threshold alpha channel. It's safe to modify the region, since it's
+	 * a buffer we made.
+	 */
+	if( cgif->has_transparency ) {
+		p = frame_bytes;
+		for( i = 0; i < n_pels; i++ ) {
+			p[3] = p[3] > 128 ? 255 : 0;
+			p += 4;
+		}
+	}
 
 	/* Do we need to compute a new palette? Do it if the frame sum changes
 	 * by more than 20%.
+	 *
+	 * frame_sum 0 means no current colourmap.
 	 */
 	sum = 0;
-	p = VIPS_REGION_ADDR( cgif->frame, 0, 0 );
+	p = frame_bytes;
 	for( i = 0; i < n_pels; i++ )
 		sum += p[i]; 
-
-	/* frame_sum 0 means no current colourmap.
-	 */
 	if( cgif->frame_sum == 0 ||
 		fabs( ((double) sum / max_sum) - 
 			((double) cgif->frame_sum / max_sum)) > 0.2 ) {
@@ -177,7 +194,6 @@ vips_foreign_save_cgif_write_frame( VipsForeignSaveCgif *cgif )
 		printf( "vips_foreign_save_cgif_write_frame: new palette\n" );
 #endif/*DEBUG_VERBOSE*/
 
-		const liq_palette *lp;
 		VipsPel *rgb;
 
 		VIPS_FREEF( liq_result_destroy, cgif->quantisation_result );
@@ -188,16 +204,20 @@ vips_foreign_save_cgif_write_frame( VipsForeignSaveCgif *cgif )
 			return( -1 );
 		}
 
-		lp = liq_get_palette( cgif->quantisation_result );
+		cgif->lp = liq_get_palette( cgif->quantisation_result );
 		rgb = cgif->palette_rgb;
-		g_assert( lp->count <= 256 );
-		for( i = 0; i < lp->count; i++ ) {
-			rgb[0] = lp->entries[i].r;
-			rgb[1] = lp->entries[i].g;
-			rgb[2] = lp->entries[i].b;
+		g_assert( cgif->lp->count <= 256 );
+		for( i = 0; i < cgif->lp->count; i++ ) {
+			rgb[0] = cgif->lp->entries[i].r;
+			rgb[1] = cgif->lp->entries[i].g;
+			rgb[2] = cgif->lp->entries[i].b;
 
 			rgb += 3;
 		}
+
+#ifdef DEBUG_VERBOSE
+		printf( "  (generated %d item colormap)\n", cgif->lp->count );
+#endif/*DEBUG_VERBOSE*/
 
 		cgif->frame_sum = sum;
 	}
@@ -216,16 +236,14 @@ vips_foreign_save_cgif_write_frame( VipsForeignSaveCgif *cgif )
 	memset( &cgif_frame_config, 0, sizeof( CGIF_FrameConfig ) );
 	cgif_frame_config.pImageData = cgif->index;
 	cgif_frame_config.pLocalPalette = cgif->palette_rgb;
+	cgif_frame_config.attrFlags = cgif->cgif_config.attrFlags;
+	cgif_frame_config.genFlags = cgif->cgif_config.genFlags;
 	if( cgif->delay &&
 		page_index < cgif->delay_length )
 		cgif_frame_config.delay = 
 			VIPS_RINT( cgif->delay[page_index] / 10.0 );
-	if( !cgif->has_transparency ) 
-		/* Allow cgif to optimise by adding transparency
-		 */
-		cgif_frame_config.genFlags = 
-			CGIF_FRAME_GEN_USE_TRANSPARENCY |
-			CGIF_FRAME_GEN_USE_DIFF_WINDOW;
+	cgif_frame_config.numLocalPaletteEntries = cgif->lp->count;
+
 	cgif_addframe( cgif->cgif_context, &cgif_frame_config );
 
 	return( 0 );
@@ -284,7 +302,6 @@ vips_foreign_save_cgif_sink_disc( VipsRegion *region, VipsRect *area, void *a )
 			if( vips_foreign_save_cgif_write_frame( cgif ) ) 
 				return( -1 );
 
-			cgif->write_y += to->height;
 			frame_size.left = 0;
 			frame_size.top = cgif->write_y;
 			frame_size.width = to->width;
@@ -303,6 +320,9 @@ vips_foreign_save_cgif_build( VipsObject *object )
 	VipsForeignSave *save = (VipsForeignSave *) object;
 	VipsForeignSaveCgif *cgif = (VipsForeignSaveCgif *) object;
 	VipsObjectClass *class = VIPS_OBJECT_GET_CLASS( cgif );
+	VipsImage **t = (VipsImage **)
+		vips_object_local_array( VIPS_OBJECT( cgif ), 2 );
+
 
 	int page_height;
 	int loop;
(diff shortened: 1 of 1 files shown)


jcupitt:
We could maybe set the first frame palette as the global colourmap, and only start using local cmaps if we find we need a new one. 

The most common case is probably all frames pretty similar, so we'd often get a single cmap per file, as before.

jcupitt:
Regarding local vs global cmaps, I see:

```
-rw-r--r-- 1 john john 4044290 Sep 24 11:28 master.gif
-rw-r--r-- 1 john john 4248068 Sep 24 11:36 x-branch.gif
-rw-r--r-- 1 john john 4271972 Sep 24 11:23 8.11.gif
```

So it's ~5% worse than master because of this cmap difference, and 8.11 is 1% worse again, perhaps because of LZW effort differences.

jcupitt:
... I pushed something to use a global cmap for the start of processing, and only flip to local if we have to. 

The change detection isn't working very well, and it seems to leave transparent holes in some GIFs. I think this needs some work still :( 

This GIF ends up with holes:

vanity

kleisauke:
Here's another test GIF that seems to produce incorrect output after processing:
waves

This is probably because the transparency index of the image is set incorrectly.

jcupitt:
Yes, cgif likes index 0 to be transparent, but I don't think libimagequant knows that. I'll have a look.

jcupitt:
I fixed the issue with transparency and dropped the change threshold to 1%, it seems to work much better now.

kleisauke:
I noticed a case where this PR could introduce noticeable artifacts.
```bash
$ curl https://user-images.githubusercontent.com/12746591/134692488-e830184c-d801-4c29-9582-23786b1612f3.gif -o waves.gif
$ vips gaussblur waves.gif[n=-1] x.gif 10
```

<details>
  <summary>Details</summary>
 
`master`:
x

`cgifsave-per-frame-cmap`:
x2
</details>

Presumably this is due to the use of local palettes or the threshold of the change detector.

lovell:
Here's an example, the infamous 301 frame `sample.gif`, where the change detector isn't sensitive enough:

Before:

sample

After:

out

(This page is now a good test for CPU fans.)

kleisauke:
I tried to improve the initialization of  `attrFlags` and `genFlags` with commit https://github.com/kleisauke/libvips/commit/52501b316c524646438e65e1d0a4ea4917fc1388. While doing that I noticed 2 things in the cgif api:
- It looks like the [`st_gifconfig.genFlags`](https://github.com/dloebl/cgif/blob/38535cb4d4356b999c9223b4228b6247467e9742/cgif.h#L40) field is currently unused in cgif.
- The global transparency flag (`CGIF_ATTR_HAS_TRANSPARENCY`) seems to be used only in [`cgif_addframe`](https://github.com/dloebl/cgif/blob/38535cb4d4356b999c9223b4228b6247467e9742/cgif.c#L491), so maybe that should be converted to a frame attribute (e.g. `CGIF_FRAME_ATTR_HAS_TRANSPARENCY`)? But perhaps this is by design to avoid complexity.

/cc @dloebl (sorry for the ping)

jcupitt:
Wow, what a terrible GIF!

I added a `DEBUG_PERCENT` define and experimented with different values for the threshold. With 0% as the threshold (!!!) I see:

```
$ vips copy sample.gif[n=-1] x.gif
frame 0, 32.92% change, new 2 item colourmap
frame 20, 0.03862% change, new 44 item colourmap
frame 21, 0.2159% change, new 46 item colourmap
frame 22, 0.304% change, new 46 item colourmap
frame 23, 0.3596% change, new 47 item colourmap
frame 24, 0.3465% change, new 48 item colourmap
frame 25, 0.3237% change, new 46 item colourmap
frame 26, 0.324% change, new 49 item colourmap
frame 27, 0.297% change, new 47 item colourmap
frame 28, 0.3049% change, new 49 item colourmap
frame 29, 0.2666% change, new 50 item colourmap
frame 30, 0.2838% change, new 52 item colourmap
frame 31, 0.2687% change, new 53 item colourmap
frame 32, 0.2754% change, new 55 item colourmap
frame 33, 0.2587% change, new 52 item colourmap
frame 34, 0.2656% change, new 54 item colourmap
frame 35, 0.2312% change, new 56 item colourmap
frame 36, 0.1949% change, new 56 item colourmap
frame 37, 0.1292% change, new 57 item colourmap
frame 38, 0.0906% change, new 55 item colourmap
frame 39, 0.02581% change, new 54 item colourmap
frame 40, 0.006348% change, new 55 item colourmap
frame 41, 0.007392% change, new 54 item colourmap
frame 42, 0.00765% change, new 53 item colourmap
frame 43, 0.008321% change, new 55 item colourmap
frame 44, 0.00749% change, new 53 item colourmap
frame 45, 0.005529% change, new 55 item colourmap
frame 46, 0.008713% change, new 54 item colourmap
frame 47, 0.004312% change, new 55 item colourmap
frame 48, 0.0001019% change, new 55 item colourmap
frame 49, 0.001183% change, new 55 item colourmap
frame 50, 0.001708% change, new 55 item colourmap
frame 51, 0.0009327% change, new 54 item colourmap
frame 52, 0.0006004% change, new 55 item colourmap
frame 53, 0.0004685% change, new 54 item colourmap
frame 54, 0.0009247% change, new 55 item colourmap
frame 55, 0.0008867% change, new 54 item colourmap
frame 56, 0.0004751% change, new 55 item colourmap
frame 57, 0.0005171% change, new 54 item colourmap
frame 58, 0.0008545% change, new 55 item colourmap
frame 59, 0.0009071% change, new 54 item colourmap
frame 60, 0.0004941% change, new 55 item colourmap
frame 61, 0.0004751% change, new 54 item colourmap
frame 62, 0.0008629% change, new 55 item colourmap
frame 63, 0.0008356% change, new 54 item colourmap
frame 64, 0.0004685% change, new 55 item colourmap
frame 65, 0.0004996% change, new 54 item colourmap
frame 66, 0.0007954% change, new 55 item colourmap
frame 67, 0.0007899% change, new 54 item colourmap
frame 68, 0.0004616% change, new 55 item colourmap
frame 69, 0.0004389% change, new 54 item colourmap
frame 70, 0.0008359% change, new 55 item colourmap
frame 71, 0.0008286% change, new 54 item colourmap
frame 72, 0.0004441% change, new 55 item colourmap
frame 73, 0.0005036% change, new 54 item colourmap
frame 74, 0.0007859% change, new 55 item colourmap
frame 75, 0.0007764% change, new 54 item colourmap
frame 76, 0.0004616% change, new 55 item colourmap
frame 77, 0.0004214% change, new 54 item colourmap
frame 78, 0.0008279% change, new 55 item colourmap
frame 79, 0.000814% change, new 54 item colourmap
frame 80, 0.0004401% change, new 55 item colourmap
frame 81, 0.000486% change, new 54 item colourmap
frame 82, 0.0007779% change, new 55 item colourmap
frame 83, 0.0007549% change, new 54 item colourmap
frame 84, 0.0004576% change, new 55 item colourmap
frame 85, 0.0004539% change, new 54 item colourmap
frame 86, 0.006491% change, new 54 item colourmap
frame 87, 0.04426% change, new 53 item colourmap
frame 88, 0.0757% change, new 54 item colourmap
frame 89, 0.1202% change, new 53 item colourmap
frame 90, 0.1653% change, new 54 item colourmap
frame 91, 0.2094% change, new 53 item colourmap
frame 92, 0.2448% change, new 53 item colourmap
frame 93, 0.2837% change, new 52 item colourmap
frame 94, 0.3332% change, new 53 item colourmap
frame 95, 0.3187% change, new 52 item colourmap
frame 96, 0.3127% change, new 54 item colourmap
frame 97, 0.3055% change, new 51 item colourmap
frame 98, 0.2943% change, new 52 item colourmap
frame 99, 0.3011% change, new 52 item colourmap
frame 100, 0.3019% change, new 49 item colourmap
frame 101, 0.3086% change, new 47 item colourmap
frame 102, 0.3058% change, new 47 item colourmap
frame 103, 0.3094% change, new 47 item colourmap
frame 104, 0.2978% change, new 49 item colourmap
frame 105, 0.2135% change, new 47 item colourmap
frame 106, 0.05683% change, new 43 item colourmap
frame 244, 2.88e-05% change, new 18 item colourmap
301 frames
89 cmaps
```

So it's still able to reuse the previous map 2/3rds of the time. Here's the new image:

x

jcupitt:
> I tried to improve the initialization of `attrFlags` and `genFlags` with commit [kleisauke@52501b3](https://github.com/kleisauke/libvips/commit/52501b316c524646438e65e1d0a4ea4917fc1388). 

Cool! I'll just run some benchmarks, then merge this.

>     * It looks like the [`st_gifconfig.genFlags`](https://github.com/dloebl/cgif/blob/38535cb4d4356b999c9223b4228b6247467e9742/cgif.h#L40) field is currently unused in cgif.

Yes, I noticed that. I used the global config genflags field to hold frame gen options I didn't want to vary frame to frame.

>     * The global transparency flag (`CGIF_ATTR_HAS_TRANSPARENCY`) seems to be used only in [`cgif_addframe`](https://github.com/dloebl/cgif/blob/38535cb4d4356b999c9223b4228b6247467e9742/cgif.c#L491), so maybe that should be converted to a frame attribute (e.g. `CGIF_FRAME_ATTR_HAS_TRANSPARENCY`)? But perhaps this is by design to avoid complexity.

Ah I hadn't seen that. 

I found the names a little confusing --- it's not clear from reading the header which define is part of which flag. Perhaps the genFlags should be called `CGIF_GEN_` rather than `CGIF_FRAME_GEN_`, and maybe `CGIF_FRAME_ATTR_USE_LOCAL_TABLE` should be `CGIF_ATTR_USE_LOCAL_TABLE`?

jcupitt:
Benchmarks, after Kleis's fixes:

```
$ /usr/bin/time -f %M:%e vips copy sample.gif[n=-1] x.per-frame-cmap.gif
146456:24.37
$ /usr/bin/time -f %M:%e vips copy sample.gif[n=-1] x.master.gif
1777212:12.97
$ /usr/bin/time -f %M:%e vips copy sample.gif[n=-1] x.8.11.gif
3645856:48.93
$ ls -l sample.gif x.*
-rw-rw-r-- 1 john john  3449741 Sep 26 11:28 sample.gif
-rw-rw-r-- 1 john john 10197811 Sep 26 12:29 x.8.11.gif
-rw-r--r-- 1 john john  3355980 Sep 26 12:22 x.master.gif
-rw-r--r-- 1 john john 10123367 Sep 26 12:13 x.per-frame-cmap.gif
```

So this branch is half the speed of master on this GIF, but uses less memory. As you suggested Lovell, perhaps we can use lower quality cmaps, since we are optimising per frame rather than for the whole file?

The file size has really shot up, I'm not sure why. That 3x increase can't just be the extra cmaps.

jcupitt:
Another benchmark:

```
$ /usr/bin/time -f %M:%e vips copy 3198.gif[n=-1] x.per-frame-cmap.gif
139 frames
139 cmaps
48532:1.12
$ /usr/bin/time -f %M:%e vips copy 3198.gif[n=-1] x.master.gif
94292:1.37
john@banana ~/pics $ ls -l x.*.gif
-rw-r--r-- 1 john john  4044290 Sep 26 12:57 x.master.gif
-rw-r--r-- 1 john john  4271900 Sep 26 12:59 x.per-frame-cmap.gif
```

This is a short video clip, so there's a new cmap per page, but it's still quicker, I expect because the GIF write is done in the background. File size seems somewhat higher, as expected,

kleisauke:
I could fix the artifacts mentioned in comment https://github.com/libvips/libvips/pull/2445#issuecomment-927109824 with commit https://github.com/kleisauke/libvips/commit/d10a87903f8ef09720eb733432c70d39fbf71c0e.

| Before | After |
| :---: |  :---: |
| before | after |

jcupitt:
Oh nice!

Is this for `CGIF_FRAME_GEN_USE_TRANSPARENCY`? Is that why we don't need liq_image_add_fixed_color()` for the first cmap?

kleisauke:
I wasn't sure why that fixed it, I didn't debug it further. It was found by accident, when I was looking at how gifski uses the libimagequant API. See for example commit https://github.com/ImageOptim/gifski/commit/a3fb2ff4d02a6ea3b18601e538b5ea7e18018e54.

(It also uses the interesting [`liq_image_set_background`](https://pngquant.org/lib/#liq_image_set_background) and [`liq_image_set_importance_map`](https://pngquant.org/lib/#liq_image_set_importance_map) functions, but the [`liq_histogram_*`](https://pngquant.org/lib/#liq_histogram_create) API doesn't seem to be used by any public repo on GitHub)

dloebl:
> I tried to improve the initialization of `attrFlags` and `genFlags` with commit [kleisauke@52501b3](https://github.com/kleisauke/libvips/commit/52501b316c524646438e65e1d0a4ea4917fc1388). While doing that I noticed 2 things in the cgif api:
> 
>     * It looks like the [`st_gifconfig.genFlags`](https://github.com/dloebl/cgif/blob/38535cb4d4356b999c9223b4228b6247467e9742/cgif.h#L40) field is currently unused in cgif.
> 
>     * The global transparency flag (`CGIF_ATTR_HAS_TRANSPARENCY`) seems to be used only in [`cgif_addframe`](https://github.com/dloebl/cgif/blob/38535cb4d4356b999c9223b4228b6247467e9742/cgif.c#L491), so maybe that should be converted to a frame attribute (e.g. `CGIF_FRAME_ATTR_HAS_TRANSPARENCY`)? But perhaps this is by design to avoid complexity.
> 
> 
> /cc @dloebl (sorry for the ping)

Yes, `st_gifconfig.genFlags` is currently unused but we are planning to use it in future.
The second point is by design to keep the API consistent and simple. Furthermore, if the transparency could be set by a frame level, we would need to aggregate at least two frames before we could write them.
That's both something we could improve / point out in the documentation of cgif.

**Quick note on the size difference (local vs global color table):** If local color tables are used, the width / height + transparency optimization are not performed for the affected frame(s) at the moment (Flags CGIF_FRAME_GEN_USE_TRANSPARENCY and CGIF_FRAME_GEN_USE_DIFF_WINDOW are ignored in this case). That might be the reason for the size difference. Adding support for the optimizations in the case of local color tables is already on my ToDo for cgif. However, it will increase the complexity of the code a bit.

jcupitt:
Ah yes, that sounds like it. I had a look through the generated GIFs, and with per-frame tables most frames (but not all, curiously) are:

```
  redraw_x = 0
  redraw_y = 0
  redraw_width = 1080
  redraw_height = 1080
```

Which I suppose explains the size change and why experimenting with the transparency options did nothing.

Also: thank you for this nice library, @dloebl.

lovell:
Are we thinking the change in this PR should be optional and, for now at least, not the default?

jcupitt:
Sorry, I posted this in gitter, but perhaps it's better here for reference.

Here are a set of test GIFs plus a benchmark script and results for local-cmap, master and 8.11 (just for fun):

http://www.rollthepotato.net/~john/test-gifs

It's just 12 of the GIFs I had handy, but it's a mix of static images, video clips, and animated vector art with and without transparency. The test is making an animated thumbnail of each one.

Memory behaviour:

memory

master and local-cmaps are very similar, since we're downsizing to thumbnail size before doing the GIF save, so not keeping the whole GIF around makes little difference here. If the test was copying a large GIF, it'd look very different.

Time behaviour:

time

local-cmaps is generally a bit quicker, though slightly slower in one case, perhaps 10% saving overall. Again, we're only writing a small GIF, so time savings in cgifsave don't have a huge effect.

Perhaps copying a GIF should also be benchmarked.

jcupitt:
I added benchmarks for copy as well, but they are not very interesting.

- per-frame has low memuse for all GIFs in the test set (as expected), and master and 8.11 have roughly similar memory use.
- per-frame and master are sometimes faster and sometimes slower than each other depending on the GIF, but the differences are not usually very large
- 8.11 is much, much slower

kleisauke:
> I wasn't sure why that fixed it, I didn't debug it further. It was found by accident, when I was looking at how gifski uses the libimagequant API. See for example commit [ImageOptim/gifski@a3fb2ff](https://github.com/ImageOptim/gifski/commit/a3fb2ff4d02a6ea3b18601e538b5ea7e18018e54).

Looking into this further (after reading the docs), it seems that the `liq_image_add_fixed_color()` call isn't necessary (it should probably only be used in combination with `liq_image_set_background()`).

It looks like the issue with that test case is that the palette could be improved during remapping, so we must call `liq_get_palette()` after `liq_write_remapped_image()`, see for example commit https://github.com/kleisauke/libvips/commit/3e3d1624b9f6a70db40a597c95e5b742105d6fdb. This commit also fixes those artifacts for me.

Regarding `CGIF_FRAME_GEN_USE_TRANSPARENCY`, I think we could also do this optimization with `liq_image_set_background()` during the quantisation, but unfortunately this function is only available since version 2.11 of libimagequant.

jcupitt:
Wow, I wouldn't have guessed that `liq_write_remapped_image()` updated the palette. Nice find! I've merged your patch.

kleisauke:
PR https://github.com/dloebl/cgif/pull/23 would help here to reduce the file size, with that I see this (tested with the infamous `sample.gif`):
```bash
$ /usr/bin/time -f %M:%e vips copy sample.gif[n=-1] x.per-frame-cmap.gif
107196:17.82
$ /usr/bin/time -f %M:%e vips copy sample.gif[n=-1] x.master.gif
1757196:11.42
$ ls -l sample.gif x.*
-rw-r--r--. 1 kleisauke kleisauke 3449741 Sep 26 11:00 sample.gif
-rw-r--r--. 1 kleisauke kleisauke 3355980 Sep 29 10:06 x.master.gif
-rw-r--r--. 1 kleisauke kleisauke 3401174 Sep 29 10:04 x.per-frame-cmap.gif
```

jcupitt:
Nice! Yes, that PR seems to mostly fix the size problem, and helps performance a little too.

lovell:
https://github.com/dloebl/cgif/pull/23 has now been merged so I've updated the (temporary) cgif Ubuntu PPA and Homebrew tap to use the latest commit - see https://github.com/lovell/cgif-packaging

jcupitt:
Oh, that's great. I updated that benchmark:

memory

time

Summary: this branch keeps memory use low and constant, but save is up to 2x slower for some (not all) very large GIFs.

jcupitt:
I suppose my feeling is that the lower speed in some cases is unfortunate, but the much better memory behaviour is probably worth it. The lower speed is not that significant for thumbnailing since we only save small GIFs (the worst case is 2.36s vs 1.65s for `wave.gif`, and of course both are far, far better than 8.11).

Does anyone else have any thoughts?

PR #2476, opened 2021-10-12
jxlload: use memmove for overlapping input buffer

lovell:
This is a possible quick fix for https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=39481

It's likely there's an alternative/additional solution as we might want to guard against this occurring in the first place.


jcupitt:
I think you're right -- this is the correct fix. I always forget memcpy can't do overlaps :(

Issue #2438, opened 2021-09-14
Unable to process non-ASCII SVG images


kleisauke:
We recently deployed this IM security policy on our public service:
```xml
<policymap>
  <!--
  Use IM only to read certain image types. Assumes the rest is handled in libvips.
  Note: ICO data can be either a BMP image, or a complete PNG image.
  TODO: Switch to libvips' jp2kload (requires OpenJPEG >= 2.4).
  -->
  <policy domain="delegate" rights="none" pattern="*" />
  <policy domain="filter" rights="none" pattern="*" />
  <policy domain="coder" rights="none" pattern="*" />
  <policy domain="coder" rights="read" pattern="{ICO,BMP,PNG,JP2,PSD,AVI}" />
  <!--
  Remove when libvips 8.12 is released.
  -->
  <policy domain="coder" rights="write" pattern="GIF" />
</policymap>
```

And noticed that some SVG images were accidentally processed by IM as fallback because the first 24 characters of the SVG image were not plain ASCII.

https://github.com/libvips/libvips/blob/1caad7f228bc8b5ce3b3349c6f3f208162a31e42/libvips/foreign/svgload.c#L206
https://github.com/libvips/libvips/blob/1caad7f228bc8b5ce3b3349c6f3f208162a31e42/libvips/foreign/svgload.c#L214-L216

Some examples from the wild include:
```xml
<svg id="レイヤー_1のコピー" data-name="レイヤー 1のコピー" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100"></svg>
<svg id="Слой_1" data-name="Слой 1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100"></svg>
<svg id="图层_1" data-name="图层 1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100"></svg>
```

Perhaps this check should be removed? Or instead check whether the first 24 characters are valid Unicode code points (`0x00..0x10FFFF`)?

lovell:
We could use `g_utf8_validate` for this as it accepts a `maxlen`. We'd need to take care of the situation where the 24th byte is a partial unicode char, so perhaps call it initially with a `maxlen` of 24 then if that fails 23 then if that fails 22.

jcupitt:
I had a look at the utf8 functions and a substring find isn't too painful:

```suggestion
/* Does a string contain a substring within the first len characters? 
 * We compare utf8 strings.
 */
static gboolean
[...]

PR #2481, opened 2021-10-14
allow utf-8 header for svg detection

jcupitt:
We were checking that the first 24 chars of an SVG were plain ASCII,
but that's not always the case, for example:

	<svg id="レイヤー_1のコピー"
		data-name="レイヤー 1のコピー"
		xmlns="http://www.w3.org/2000/svg"
		viewBox="0 0 100 100">
	</svg>

We now test for the string "<svg" being in the first 1000 bytes, and
everything up to that being valid utf-8.

See https://github.com/libvips/libvips/issues/2438

### DIFF
--- a/libvips/foreign/webpload.c
+++ b/libvips/foreign/webpload.c
@@ -166,9 +166,9 @@ vips_foreign_load_webp_class_init( VipsForeignLoadWebpClass *class )
 	object_class->description = _( "load webp" );
 	object_class->build = vips_foreign_load_webp_build;
 
-	/* is_a() is not that quick ... lower the priority.
+	/* We are fast at is_a(), so high priority.
 	 */
-	foreign_class->priority = -50;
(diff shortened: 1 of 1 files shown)


kleisauke:
Note that OSS-Fuzz does not currently build against librsvg, so this change is not covered. I've looked into this in the past, but it requires adding these (sub-)dependencies as well:

<details>
  <summary>Dependency-graph librsvg</summary>

- gdk-pixbuf
- cairo
  - pixman
  - fontconfig (recommend)
- pango
  - cairo (recommend)
  - fontconfig
  - freetype
  - harfbuzz
  - fribidi
- libxml2
</details>

lovell:
@kleisauke Good spot, I can take a look at adding it. To start with we might be able to get away using apt to install some/all of these rather than build from source.

PR #2489, opened 2021-10-20
Prefer g_malloc over vips_malloc in vips_blob_copy

kleisauke:
The use of `vips_malloc` is not necessary here and could potentially
be slower since it uses `g_malloc0` (i.e. `calloc`).


jcupitt:
Ah true, I didn't realize it was so recent.

PR #2505, opened 2021-11-01
cplusplus: only generate doc comments for non-deprecated args

kleisauke:
I think this is clearer, other bindings may need something similar as well.


jcupitt:
Ooop, that's an oversight, they are certainly meant to have this, good find.

Yes, looks like pyvips and ruby-vips are both missing this. I'll have a look.

jcupitt:
OK, py and ruby done too.

PR #2508, opened 2021-11-01
Add hyperbolic functions

hroskes:
Added `sinh`, `cosh`, `tanh`, `asinh`, `acosh`, and `atanh` functions, in parallel with the normal trigonometric functions.

(The motivation is that a project I'm working on actually needs to take the `sinh` of an image.)


libvips/arithmetic/math.c
```diff
@@ -164,6 +164,12 @@ vips_math_buffer( VipsArithmetic *arithmetic,
 	case VIPS_OPERATION_MATH_ASIN: 	SWITCH( ADSIN ); break;
 	case VIPS_OPERATION_MATH_ACOS: 	SWITCH( ADCOS ); break;
 	case VIPS_OPERATION_MATH_ATAN: 	SWITCH( ADTAN ); break;
+	case VIPS_OPERATION_MATH_SINH: 	SWITCH( sinh ); break;
```
jcupitt, line 167: Like almost all image processing packages, libvips uses degrees everywhere, so I think it'd be very confusing to have these functions in radians. I would add:

```
#define DSINH( X ) (sinh( VIPS_RAD( X ) ))
#define DCOSH( X ) (cosh( VIPS_RAD( X ) ))
#define DTANH( X ) (tanh( VIPS_RAD( X ) ))
```

And the same for asinh() etc.

libvips/include/vips/arithmetic.h
```diff
@@ -46,6 +46,12 @@ extern "C" {
  * @VIPS_OPERATION_MATH_ASIN: asin(), angles in degrees
  * @VIPS_OPERATION_MATH_ACOS: acos(), angles in degrees
  * @VIPS_OPERATION_MATH_ATAN: atan(), angles in degrees
+ * @VIPS_OPERATION_MATH_SINH: sinh()
+ * @VIPS_OPERATION_MATH_COSH: cosh()
+ * @VIPS_OPERATION_MATH_TANH: tanh()
+ * @VIPS_OPERATION_MATH_ASINH: asinh()
+ * @VIPS_OPERATION_MATH_ACOSH: acosh()
+ * @VIPS_OPERATION_MATH_ATANH: atanh()
```
jcupitt, line 54: We should have a note about degrees here too.

test/test-suite/test_arithmetic.py
```diff
@@ -472,7 +472,88 @@ def my_atan(x):
...
+    # this requires pyvips 2.1.16 for sinh
+    @pytest.mark.skipif(versiontuple(pyvips.__version__) < 
+            versiontuple('2.1.16'),
+            reason='your pyvips is too old')
+    def test_sinh(self):
+        def my_sinh(x):
+            if isinstance(x, pyvips.Image):
+                return x.sinh()
+            else:
+                return math.sinh(math.radians(x))
```
jcupitt, line 484: And this should be degrees.

jcupitt (commented):
This looks good!

General notes: sinh/cosh/tanh are in C89 (the minimum C version for libvips), but the inverse asinh etc. are not. MSVC only added them in c++11 (I think?). Perhaps we don't need to support very old platforms.

We're in feature-freeze getting ready for 8.12 at the moment, but this is a minor and low-impact change, so let's get it in.

Could you also add a one-line note to the CHANGELOG?

libvips/arithmetic/math.c
hroskes: I've never heard of wanting to calculate hyperbolic functions of an angle in degrees (even though they're related to the trig functions by i).  For example the numpy docs specify that [`sin`](https://numpy.org/doc/stable/reference/generated/numpy.sin.html) uses radians, but don't even bother saying anything for [`sinh`](https://numpy.org/doc/stable/reference/generated/numpy.sinh.html).  The calculator on Windows has a switch for radians or degrees for the trig functions, but it doesn't affect the hyperbolic functions.

On the other hand the [C# docs](https://docs.microsoft.com/en-us/dotnet/api/system.math.sinh?view=net-5.0) do say that it's in radians, and I see a few [examples](https://docs.trifacta.com/display/SS/COSH+Function) of calculating hyperbolic functions of degree angles.

I see your point.  Personally I still think it's less confusing to keep the hyperbolic functions in radians, but I can switch it to degrees if you want me to.

jcupitt: You're right, I checked my elderly HP calculator and the rad/deg mode does not affect the hyperbolic functions. That's good enough for me -- let's leave this has you had it originally.

Could you add a note to the docs below saying these functions use radians?

jcupitt: (sorry for wasting your time)

cplusplus/include/vips/VImage8.h
```diff
@@ -1375,6 +1375,60 @@ class VImage : public VObject
...
+		return( math( VIPS_OPERATION_MATH_SINH, options ) );
+	}
+
+	/**
+	 * Find the hyperbolic cosine of each pixel. Angles are in degrees.
+	 */
+	VImage 
+	cosh( VOption *options = 0 ) const
+	{
+		return( math( VIPS_OPERATION_MATH_COS, options ) );
```
jcupitt, line 1393: Should be COSH

libvips/arithmetic/math.c
```diff
@@ -129,6 +129,28 @@ vips_math_build( VipsObject *object )
...
+#endif
+
+#if HAS_INVERSE_HYPERBOLICS
+  #define ASINH sinh
+  #define ACOSH cosh
+  #define ATANH tanh
+#else
+  #define ASINH( X ) log((X) + sqrt( (X)*(X)+1 ))
+  #define ACOSH( X ) log((X) + sqrt( (X)*(X)-1 ))
+  #define ATANH( X ) 0.5 * log( (1+(X)) / (1-(X)) )
```
jcupitt, line 151: I'd put an extra `()` around the whole expression to stop precedence messing this up if used in more complex expressions.

jcupitt (commented):
Great! I saw one typo and some possible extra brackets, but that's all.

jcupitt:
Great! Thanks again on this nice work.

hroskes:
Thank you!

PR #2515, opened 2021-11-03
pngsave: allow valid bitdepth of 16

lovell:
It should be possible to set a bitdepth of 16 for PNG output, avoiding the following warning.

```
(process:11191): GLib-GObject-WARNING **: 15:52:12.447: value "16" of type 'gint' is invalid or out of range for property 'bitdepth' of type 'gint'
```

Discovered whilst investigating https://github.com/lovell/sharp/issues/2934


jcupitt:
Ooop, that was a silly oversight. Nice!

Issue #2388, opened 2021-08-09
EXIF metadata strings containing multiple brackets have their values altered


lovell:
The following file contains EXIF Copyright metadata with multiple opening and closing brackets in the string value.

![test-1](https://user-images.githubusercontent.com/210965/128771922-17383495-3450-4151-b078-73aa01a401f1.jpeg)

```sh
$ exif -m test-1.jpeg | grep Copyright
Copyright	Arnold Jerocki/Divergence (Arnold Jerocki/Divergence (Photographer) - [None] (Photographer) - [None] (Editor)
```

When processed via vips this metadata is erroneously altered.

```sh
$ vips copy test-1.jpeg out.jpeg
$ exif -m out.jpeg | grep Copyright
Copyright	Arnold Jerocki/Divergence (Arnold Jerocki/Divergence (Photographer) - [None] (Arnold Jerocki/Divergence (Arnold Jerocki/Divergence (Photographer) - [None] (Photographer) - [None] (Photographer) - [None] (Editor)
```

Presumably this is due to brackets being considered "special" characters and `drop_tail` is not selecting the correct one.

https://github.com/libvips/libvips/blob/1ce301812d1c44ca6421ed93d6ba1ec4b1d14433/libvips/foreign/exif.c#L811-L818

This was originally reported at https://github.com/lovell/sharp/issues/2832

jcupitt:
Yes, this is a problem with the way libvips handles exif metadata.

I suppose we need to quote special characters to ensure we can always get the input back again.

lovell:
Thanks for confirming John, I wasn't sure how best to fix this. Could we use an alternative separator instead of an opening bracket? (The string is UTF-8 encoded at this point so there's plenty of choice.)

kleisauke:
I remembered a discussion some time ago about this, with a possible alternative solution: https://github.com/kleisauke/net-vips/issues/74#issuecomment-622465228.

jcupitt:
Fix merged to master for 8.12.

PR #2518, opened 2021-11-06
exif: separate entries into raw vs human-readable

lovell:
Here's a work-in-progress PR to fix https://github.com/libvips/libvips/issues/2388, which takes the approach suggested in https://github.com/kleisauke/net-vips/issues/74#issuecomment-622465228

It still needs unit tests (and more testing).

```
$ vipsheader -a test-1.jpeg | grep exif
exif-data: 5872 bytes of binary data
exif-ifd0-Orientation: 1
exif-ifd0-Orientation-human: Top-left, Short, 1 components, 2 bytes
exif-ifd0-XResolution: 72/1
exif-ifd0-XResolution-human: 72, Rational, 1 components, 8 bytes
exif-ifd0-YResolution: 72/1
exif-ifd0-YResolution-human: 72, Rational, 1 components, 8 bytes
exif-ifd0-ResolutionUnit: 2
exif-ifd0-ResolutionUnit-human: Inch, Short, 1 components, 2 bytes
exif-ifd0-Software: Windows Photo Editor 10.0.10011.16384
exif-ifd0-Software-human: Windows Photo Editor 10.0.10011.16384, ASCII, 38 components, 38 bytes
exif-ifd0-DateTime: 2021:08:06 17:17:03
exif-ifd0-DateTime-human: 2021:08:06 17:17:03, ASCII, 20 components, 20 bytes
exif-ifd0-Copyright: Arnold Jerocki/Divergence (Arnold Jerocki/Divergence (Photographer) - [None]
exif-ifd0-Copyright-human: Arnold Jerocki/Divergence (Arnold Jerocki/Divergence (Photographer) - [None] (Photographer) - [None] (Editor), ASCII, 77 components, 77 bytes
exif-ifd0-Padding: 2060 bytes undefined data
exif-ifd0-Padding-human: 2060 bytes undefined data, Undefined, 2060 components, 2060 bytes
exif-ifd1-Compression: 6
exif-ifd1-Compression-human: JPEG compression, Short, 1 components, 2 bytes
exif-ifd1-XResolution: 96/1
exif-ifd1-XResolution-human: 96, Rational, 1 components, 8 bytes
exif-ifd1-YResolution: 96/1
exif-ifd1-YResolution-human: 96, Rational, 1 components, 8 bytes
exif-ifd1-ResolutionUnit: 2
exif-ifd1-ResolutionUnit-human: Inch, Short, 1 components, 2 bytes
exif-ifd2-ExifVersion: Exif Version 2.1
exif-ifd2-ExifVersion-human: Exif Version 2.1, Undefined, 4 components, 4 bytes
exif-ifd2-DateTimeOriginal: 2021:08:05 12:37:28
exif-ifd2-DateTimeOriginal-human: 2021:08:05 12:37:28, ASCII, 20 components, 20 bytes
exif-ifd2-DateTimeDigitized: 2021:08:05 12:37:28
exif-ifd2-DateTimeDigitized-human: 2021:08:05 12:37:28, ASCII, 20 components, 20 bytes
exif-ifd2-SubSecTimeOriginal: 58
exif-ifd2-SubSecTimeOriginal-human: 58, ASCII, 3 components, 3 bytes
exif-ifd2-SubSecTimeDigitized: 58
exif-ifd2-SubSecTimeDigitized-human: 58, ASCII, 3 components, 3 bytes
exif-ifd2-FlashPixVersion: FlashPix Version 1.0
exif-ifd2-FlashPixVersion-human: FlashPix Version 1.0, Undefined, 4 components, 4 bytes
exif-ifd2-ColorSpace: 1
exif-ifd2-ColorSpace-human: sRGB, Short, 1 components, 2 bytes
exif-ifd2-PixelXDimension: 1440
exif-ifd2-PixelXDimension-human: 1440, Long, 1 components, 4 bytes
exif-ifd2-PixelYDimension: 960
exif-ifd2-PixelYDimension-human: 960, Long, 1 components, 4 bytes
exif-ifd2-Padding: 2060 bytes undefined data
exif-ifd2-Padding-human: 2060 bytes undefined data, Undefined, 2060 components, 2060 bytes
```

### DIFF
--- a/libvips/foreign/exif.c
+++ b/libvips/foreign/exif.c
@@ -267,6 +267,18 @@ vips_exif_get_double( ExifData *ed,
 	return( 0 );
 }
 
+static void
+vips_exif_to_s_human( ExifData *ed, ExifEntry *entry, VipsBuf *buf )
+{
+	char txt[256];
+
+	vips_buf_appendf( buf, "%s, %s, %lu components, %d bytes",
+		exif_entry_get_value( entry, txt, 256 ),
+		exif_format_get_name( entry->format ),
+		entry->components,
+		entry->size );
+}
+
 /* Save an exif value to a string in a way that we can restore. We only bother
  * for the simple formats (that a client might try to change) though.
  *
@@ -289,20 +301,20 @@ vips_exif_to_s( ExifData *ed, ExifEntry *entry, VipsBuf *buf )
 
 		memcpy( txt, entry->data, len );
 		txt[len] = '\0';
-		vips_buf_appendf( buf, "%s ", txt );
+		vips_buf_appendf( buf, "%s", txt );
 	}
 	else if( entry->components < 10 &&
 		!vips_exif_get_int( ed, entry, 0, &iv ) ) {
 		for( i = 0; i < entry->components; i++ ) {
 			vips_exif_get_int( ed, entry, i, &iv );
-			vips_buf_appendf( buf, "%d ", iv );
+			vips_buf_appendf( buf, "%d", iv );
 		}
 	}
 	else if( entry->components < 10 &&
 		!vips_exif_get_rational( ed, entry, 0, &rv ) ) {
 		for( i = 0; i < entry->components; i++ ) {
 			vips_exif_get_rational( ed, entry, i, &rv );
-			vips_buf_appendf( buf, "%u/%u ", 
+			vips_buf_appendf( buf, "%u/%u",
 				rv.numerator, rv.denominator );
 		}
 	}
@@ -310,19 +322,13 @@ vips_exif_to_s( ExifData *ed, ExifEntry *entry, VipsBuf *buf )
 		!vips_exif_get_srational( ed, entry, 0, &srv ) ) {
 		for( i = 0; i < entry->components; i++ ) {
 			vips_exif_get_srational( ed, entry, i, &srv );
-			vips_buf_appendf( buf, "%d/%d ", 
+			vips_buf_appendf( buf, "%d/%d",
 				srv.numerator, srv.denominator );
 		}
 	}
 	else 
-		vips_buf_appendf( buf, "%s ", 
+		vips_buf_appendf( buf, "%s",
 			exif_entry_get_value( entry, txt, 256 ) );
-
-	vips_buf_appendf( buf, "(%s, %s, %lu components, %d bytes)", 
-		exif_entry_get_value( entry, txt, 256 ),
-		exif_format_get_name( entry->format ),
-		entry->components,
-		entry->size );
 }
 
 typedef struct _VipsExifParams {
@@ -352,7 +358,9 @@ vips_exif_attach_entry( ExifEntry *entry, VipsExifParams *params )
 	char vips_name_txt[256];
 	VipsBuf vips_name = VIPS_BUF_STATIC( vips_name_txt );
 	char value_txt[256];
+	char value_txt_human[256];
 	VipsBuf value = VIPS_BUF_STATIC( value_txt );
+	VipsBuf value_human = VIPS_BUF_STATIC( value_txt_human );
 
 	if( !(tag_name = vips_exif_entry_get_name( entry )) )
 		return;
@@ -365,6 +373,11 @@ vips_exif_attach_entry( ExifEntry *entry, VipsExifParams *params )
 	 */
 	(void) vips_image_set_string( params->image, 
 		vips_buf_all( &vips_name ), vips_buf_all( &value ) );
+
+	vips_buf_appendf( &vips_name, "-human" );
+	vips_exif_to_s_human( params->ed, entry, &value_human );
+	(void) vips_image_set_string( params->image,
+		vips_buf_all( &vips_name ), vips_buf_all( &value_human ) );
 }
 
 static void
@@ -808,31 +821,6 @@ vips_exif_alloc_string( ExifEntry *entry, unsigned long components )
 	VIPS_FREEF( exif_mem_unref, mem );
 }
 
-/* The final " (xx, yy, zz, kk)" part of the string (if present) was
- * added by us in _to_s(), we must remove it before setting the string 
- * back again.
- *
- * It may not be there if the user has changed the string.
- */
-static char *
-drop_tail( const char *data )
-{
-	char *str;
-	char *p;
-
-	str = g_strdup( data );
-
-	p = str + strlen( str );
-	if( p > str &&
-		*g_utf8_prev_char( p ) == ')' &&
-		(p = g_utf8_strrchr( str, -1, (gunichar) '(')) &&
-		p > str &&
-		*(p = g_utf8_prev_char( p )) == ' ' )
-		*p = '\0';
-
-	return( str );
-}
-
 /* special header required for EXIF_TAG_USER_COMMENT.
  */
 #define ASCII_COMMENT "ASCII\0\0\0"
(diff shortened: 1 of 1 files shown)


jcupitt:
I had a thought about this --- how about leaving the current exif outpuf as is for compatibility, and (as fredprobidi suggests) adding a new set of fields using JSON? Something like:

```
exif-ifd2-ShutterSpeedValue: 253/50 (5.06 EV (1/33 sec.), SRational, 1 components, 8 bytes)
exif-json-ifd2-ShutterSpeedValue: {value: "253/50", humanValue: "5.06 EV (1/33 sec.)", dataType: "SRational", components: 1, length: 8}
```

Now we get compatibility, plus easy and reliable structured data handling in all language bindings. We could maybe hide the old exif tags in vipsheader output. We'd need to detect user edits to the JSON on save.

We'd need to add a dependency for a JSON read-write library though :-( An obvious choice might be:

https://gitlab.gnome.org/GNOME/json-glib

It's mature, well-maintained, integrates with glib, it's by ebassi (who is great), etc.

lovell:
If we're thinking of providing EXIF as JSON then perhaps this should be rolled up at a higher level as a single property, so the dirs and keys are also part of the JSON?
```json
{
  "ifd2": {
    "ShutterSpeed": {
      "value": "253/50",
      "humanValue": "5.06 EV (1/33 sec.)",
      "dataType": "SRational",
      "components": 1,
      "length": 8
    }
  }
}
```
Alternatively, for backwards compatibility, we could provide the non-human values in new `-raw` suffix properties:
```
exif-ifd2-ShutterSpeed: 253/50 (5.06 EV (1/33 sec.), SRational, 1 components, 8 bytes)
exif-ifd2-ShutterSpeed-raw: 253/50
```

jcupitt:
That's true, though you wonder what value we are adding then --- the user could simply parse the `exif-data` blob to make json themselves.

Off-topic, but I just reread `exif.c` to remind myself how this works and ... what a mess! libexif (and exif itself) is pretty bad.

> Alternatively, for backwards compatibility, we could provide the non-human values in new -raw suffix properties:

Maybe this is the best compromise. We keep back compat, and we don't need to add a json library. I'm not sure how useful the extra info in the json output would be anyway.

lovell:
A possible problem with the alternative `-raw` suffix approach is that whilst it keeps the read "API" backwards compatible, it either breaks the write "API" or makes the implementation more complex.

At the moment you can e.g. call `image.set("exif-ifd2-ShutterSpeed", "253/50")` and it overwrites the non-suffixed "human" variant. To continue to support this with two variants, where there is no longer any "tail dropping", we'll probably need to track which variant is to be written to the output (original raw vs updated human) for each field.

jcupitt:
Yes, write is harder. We could maybe do:

```
parse the exif-data block (contains the original values) to make `exif`

for each item in `old_exif`:
    is there a tag called `exif-$tag-raw`? If not:
        the user must have deleted the tag, so we must delete the tag from `exif` as well
    also test for `exif-$tag`. If that's missing:
        we must be dealing with old code. Also delete `$tag` from `exif`

for each metadata item on the image starting with `exif`:
    if it ends `-raw`:
        add the tag to `exif`, or update any existing tag
    if it doesn't end with `-raw`, AND there's no matching `-raw` tag:
        we are dealing with old API code, add or update `exif` from the `xxx (yyy)` value
```

ie. we add a `raw` case, and it takes priority over the old `xx (yy)` API. I think that code's all there now, except for the `raw` stuff.

jcupitt:
Thinking about this again, perhaps we could fix this another way?

Here's a tiny ruby program to show the issue:

```ruby
#!/usr/bin/ruby

require "vips"

x = Vips::Image.new_from_file ARGV[0]

x = x.mutate do |mutable| 
  mutable.set_type! GObject::GSTR_TYPE, "exif-ifd0-XPComment", "hello ( there"
end

x.write_to_file ARGV[1]
```

I can run like this and it works:

```
$ ~/try/badcomment.rb k2.jpg x.jpg
$ vipsheader -f exif-ifd0-XPComment x.jpg 
hello ( there (hello ( there, Byte, 28 components, 28 bytes)
```

So that's the correct `xx (xx, Byte, 28 components, 28 bytes)` value, it's just that `xx` contains an embedded `(` character.

When we later copy the image, it fails:

```
$ vips copy x.jpg y.jpg
$ vipsheader -f exif-ifd0-XPComment y.jpg 
hello ( there (hello (hello ( there (hello, Byte, 42 components, 42 bytes)
```

What's happened is:

1. The 28-byte value (it's in utf-16) `hello ( there` is read from `x.jpg`.
2. It is correctly coded as `hello ( there (hello ( there, Byte, 28 components, 28 bytes)` and attached to the image.
3. On jpegsave, `exif.c` takes the coded value from the image metadata, finds the trailing `( )` group, ie, `( there, Byte, 28 components, 28 bytes)`, and removes it, leaving `hello ( there (hello`.
4. This stripped value is compared to the value in the original exif data block (`hello ( there`) and is found to be different.
5. The exif data is updated with what we have erroneously identified as the new value, and the saved string grows to `hello ( there (hello`.

So for a fix, how about just changing steps 3 and 4. 

Don't try to transform the new image metadata value to match the original exif value (ie. reversing the stringification from step 2, a difficult problem), instead just re-stringify the exifdata block and look for differences in the coded value? 

If we find they are different, we simply attach the new value to the image, and assume the user didn't painstakingly create the trailing ` (yy)` stuff (very hard to get right anyway).

This removes the need to reverse the stringification and hopefully fixes the bug. What do you think?

It means we keep supporting this stupid and broken interface. Perhaps we should have a separate issue for replacing it with something better.

jcupitt:
I did a quick hack: https://github.com/libvips/libvips/commit/8195b67d83253a5abc8aee9273e8e4558e4c9973 It was easier than I thought -- you only need to add a change detector near the start. With this I see:

```
$ ~/try/badcomment.rb k2.jpg x.jpg
$ vipsheader -f exif-ifd0-XPComment x.jpg 
hello ( there (hello ( there, Byte, 28 components, 28 bytes)
$ vips copy x.jpg y.jpg
$ vipsheader -f exif-ifd0-XPComment y.jpg 
hello ( there (hello ( there, Byte, 28 components, 28 bytes)
```

ie. it's correctly seeing that the item hasn't changed and shouldn't  be regenerated.

What do you think?

lovell:
Looks good, and much simpler. Is it possible to remove an EXIF property with this approach?

kleisauke:
Looks good to me as well. I tried to cover this by a unit test, see commit https://github.com/kleisauke/libvips/commit/008fe2d2f57aace8f6f5aff56965d3afb32be4da.

jcupitt:
Yes, remove should work, I expanded the test prog:

```ruby
#!/usr/bin/ruby

require "vips"

# set the bad comment
x = Vips::Image.new_from_file ARGV[0]
x = x.mutate do |mutable|
  mutable.set_type! GObject::GSTR_TYPE, "exif-ifd0-XPComment", "hello ( there"
end
x.write_to_file "t1.jpg"

# copy the image with the bad comment
x = Vips::Image.new_from_file "t1.jpg"
comment = x.get("exif-ifd0-XPComment")
puts "comment after save, load:"
puts "#{comment}"
x.write_to_file "t2.jpg"

# read back again and check the comment is OK
x = Vips::Image.new_from_file "t2.jpg"
comment = x.get("exif-ifd0-XPComment")
puts "comment after save, load, save, load:"
puts "#{comment}"

# delete and save
x = x.mutate do |mutable|
  mutable.remove! "exif-ifd0-XPComment"
end
x.write_to_file "t3.jpg"

# read back again and check the comment is OK
x = Vips::Image.new_from_file "t3.jpg"
puts "try to get after remove, save, load:"
typeof = x.get_typeof("exif-ifd0-XPComment")
puts "typeof = #{typeof}" 
```

I see:

```
$ ./badcomment.rb ~/pics/k2.jpg
comment after save, load:
hello ( there (hello ( there, Byte, 28 components, 28 bytes)
comment after save, load, save, load:
hello ( there (hello ( there, Byte, 28 components, 28 bytes)
try to get after remove, save, load:
typeof = 0
```

So it's successfully saved and loaded again unchanged, and you can remove.

jcupitt:
Ah, @kleisauke 's unit test tested remove as well, that's easy.

OK, I've merged the change to master, I think we're done! Let's kill off the last few tiny things and make 8.12-rc1.

lovell:
Excellent, thanks John.

PR #2536, opened 2021-11-12
Add spngsave

jcupitt:
Save with libspng.

To do:

- [x] add metadata, eg. icc profiles, exif, xmp, comments, etc.
- [x] hook up options, eg. filtering, compression effort, etc.
- [x] add quantisation and dither
- [x] add interlaced write
- [x] 1, 2, 4 bit save
- [x] test 16-bit support (I think it should work?)
- [x] fix build against older libspng, or bump min required version to 0.7.2+


jcupitt:
Right now it's a bit slower than libpng, perhaps more things need hooking up before we can get a good benchmark.

libvips/foreign/spngsave.c
```diff
@@ -0,0 +1,837 @@
...
+vips_foreign_save_spng_file_class_init( VipsForeignSaveSpngFileClass *class )
+{
+	GObjectClass *gobject_class = G_OBJECT_CLASS( class );
+	VipsObjectClass *object_class = (VipsObjectClass *) class;
+
+	gobject_class->set_property = vips_object_set_property;
+	gobject_class->get_property = vips_object_get_property;
+
+	object_class->nickname = "pngsave";
+	object_class->description = _( "save image to files as PNG" );
```
kleisauke, line 761: nit:
```suggestion
	object_class->description = _( "save image to file as PNG" );
```
or perhaps:
```suggestion
	object_class->description = _( "save image to PNG file" );
```
(noticed this description change in commit https://github.com/libvips/libvips/commit/57a0bd82f91ec776911f6d3aa85fbe2b06547612)

jcupitt:
It's now about the same speed as libpng, though of course with a much cleaner and simpler API.

jcupitt:
It's all done except for low bit depth write, which is not yet directly supported by libspng.

libvips/foreign/spngsave.c
```diff
@@ -0,0 +1,848 @@
...
+		return( -1 );
+	}
+
+	spng_set_option( spng->ctx, 
+		SPNG_IMG_COMPRESSION_LEVEL, spng->compression );
+	spng_set_option( spng->ctx, 
+		SPNG_TEXT_COMPRESSION_LEVEL, spng->compression );
+	spng_set_option( spng->ctx, 
+		SPNG_FILTER_CHOICE, spng->filter );
+
```
kleisauke, line 426: I had a quick experiment with changing libspng default [`Z_FILTERED`](https://github.com/randy408/libspng/blob/8c74bfe96bcb17123d8e83de1edeffe443b75f24/spng/spng.c#L4869) strategy to `Z_RLE`, since I remembered [this discussion](https://github.com/lovell/sharp/commit/3917efdebd91ddaf3878a276a6c7b79edce2d2fd#r41720871) with Lovell.
```suggestion

	/* Use Z_RLE strategy by default as that is specifically suitable for PNG.
	 */
	spng_set_option( spng->ctx,
		SPNG_IMG_COMPRESSION_STRATEGY, 3/*Z_RLE*/ );

```

After testing this with the `large_rgb8.png` image from https://github.com/libspng/benchmark_images/, I see this:
```bash
$ curl -LO https://github.com/libspng/benchmark_images/raw/master/large_rgb8.png
$ LD_PRELOAD="/usr/lib64/libz-ng.so.2.0.5" /usr/bin/time -f %M:%e vips copy large_rgb8.png z_filterd.png
32516:3.04
# After applying the above Z_RLE patch
$ LD_PRELOAD="/usr/lib64/libz-ng.so.2.0.5" /usr/bin/time -f %M:%e vips copy large_rgb8.png z_rle.png
32532:0.74
$ ls -al z_*.png
-rw-r--r--. 1 kleisauke kleisauke 13160679 Nov 26 15:01 z_filterd.png
-rw-r--r--. 1 kleisauke kleisauke 13181100 Nov 26 15:02 z_rle.png
```

/me falls off chair.

jcupitt: wat

jcupitt: I tried with a 10k x 10k RGB image:

```
$ time vips copy wtc.jpg x.png
real	0m27.542s
user	0m27.767s
sys	0m0.256s
$ time vips copy wtc.jpg x-rle.png
real	0m13.399s
user	0m13.718s
sys	0m0.188s
$ ls -l x.png x-rle.png 
-rw-r--r-- 1 john john 135800382 Nov 26 16:45 x.png
-rw-r--r-- 1 john john 146518124 Nov 26 16:46 x-rle.png
```

deflate compression defaults to 6. I tried turning that down to try to match the compression speed:

```
$ time vips copy wtc.jpg x.png[compression=1]
real	0m14.923s
user	0m15.166s
sys	0m0.265s
$ ls -l x.png
-rw-r--r-- 1 john john 137881560 Nov 26 16:50 x.png
```

So maybe deflate level 1 might be another default to consider? Better compression than RLE, and almost as fast.

jcupitt:
Another benchmark to show off libspng. If you turn off compression, you take deflate encode out and see the relative speed of the PNG libraries more clearly.

Here's a 10k x 10k pixel RGB image with libpng:

```
$ time vips copy wtc.jpg x.png[compression=0]
real	0m2.431s
user	0m2.608s
sys	0m0.345s
```

And with libspng:

```
$ time vips copy wtc.jpg x.png[compression=0]
real	0m0.719s
user	0m0.737s
sys	0m0.383s
```

libvips/foreign/spngsave.c
lovell: I'd expect RLE to suit images with long runs of the same pixel value, e.g. icons, line art and perhaps palette-based images, which is typically what PNG is used for on the web, plus I guess it might also use a bit less memory. Interesting that the examples here are photographic in nature, which also seems to be a partial conclusion of http://optipng.sourceforge.net/pngtech/z_rle.html

libvips/foreign/spngsave.c
```diff
@@ -0,0 +1,914 @@
...
+#include <vips/intl.h>
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <vips/vips.h>
+
+#include "pforeign.h"
+
```
kleisauke, line 59: I think macOS CI  fails due to the missing `quantise.h` include.
```suggestion
#include "quantise.h"

```

jcupitt: Ooop, I rebased on master but didn't check it still built cleanly, argh. Thanks!

libvips/foreign/spngsave.c
```diff
@@ -0,0 +1,915 @@
...
+	case 2:
+		ihdr.color_type = SPNG_COLOR_TYPE_GRAYSCALE_ALPHA; 
+		break;
+
+	case 3:
+		ihdr.color_type = SPNG_COLOR_TYPE_TRUECOLOR;
+		break;
+
+	case 4: 
+		ihdr.color_type = SPNG_COLOR_TYPE_GRAYSCALE_ALPHA;
```
kleisauke, line 461: This change resolves a lot of test failures on the [images.weserv.nl test suite](https://github.com/weserv/images/tree/5.x/test).
```suggestion
		ihdr.color_type = SPNG_COLOR_TYPE_TRUECOLOR_ALPHA; 
```

But somehow it still produces transparent images on test cases where the alpha channel is temporarily separated, for example:
```bash
$ curl -LO https://github.com/weserv/images/raw/5.x/test/api/fixtures/alpha-layer-1-fill.png

# Separate alpha channel
$ vips extract_band alpha-layer-1-fill.png rgb.v 0 --n 3
$ vips extract_band alpha-layer-1-fill.png a.v 3

# Change brightness
$ vips linear rgb.v x.v 1 76

# Bandjoin
$ vips bandjoin "x.v a.v" x.png
```

kleisauke: Replacing the last command with:
```bash
# Cast float image to uchar
$ vips cast x.v x2.v uchar

# Bandjoin
$ vips bandjoin "x2.v a.v" x.png
```

Does produce a correct image. So I guess this cast is also needed in `spngsave.c`:
https://github.com/libvips/libvips/blob/de199fe59cde38290709c7eb0d25406daa1314d7/libvips/foreign/pngsave.c#L115-L129

```bash
$ vipsheader x.v
x.v: 2048x1536 float, 3 bands, srgb, pngload
```

libvips/foreign/spngsave.c
```diff
@@ -583,43 +583,60 @@ vips_foreign_save_spng_build( VipsObject *object )
...
+
+		if( vips_cast( in, &x, VIPS_FORMAT_UCHAR, NULL ) ) {
+			g_object_unref( in );
+			return( -1 );
+		}
+		g_object_unref( in );
+		in = x;
+	}
+
+	/* Filtering usually reduces the compression ratio for palette images,
```
randy408, line 626: spng already does this optimization https://github.com/randy408/libspng/blob/v0.7.1/spng/spng.c#L4727-L4730

libvips/foreign/spngsave.c
jcupitt: Oooof, that was a dumb error, good spot.

I revised the 8/16-bit save rules and the float-save case seems to work now. Thanks!

libvips/foreign/spngsave.c
jcupitt: Oh, nice. Fixed, thanks.

configure.ac
```diff
@@ -1302,37 +1302,33 @@ VIPS_INCLUDES="$VIPS_INCLUDES $TIFF_INCLUDES"
 VIPS_LIBS="$VIPS_LIBS $TIFF_LIBS"
 
 # Look for libspng first
-# 0.6.1 uses "libspng.pc", git master libspng uses "spng.pc"
```
kleisauke, line 1305: It seems that we still have to search for `libspng.pc` when we build with CMake (which is done in OSS-Fuzz). See for example:
https://github.com/randy408/libspng/blob/957d1ec1c19247767e245ab708903048f1220db8/CMakeLists.txt#L62
https://github.com/google/oss-fuzz/blob/c34c308faad86d154b52586ff66de8d77187cafd/projects/libvips/build.sh#L113-L114

(noticed while inspecting the CIFuzz build logs)

jcupitt: Ouch, OK, let's look for both. Thanks for pointing this out!

kleisauke: The `PACKAGES_USED="$PACKAGES_USED libspng"` thing is also needed when `libspng.pc` is used, unfortunately. Otherwise, dependencies that link to `vips.pc` or `vips-cpp.pc` may fail (since it cannot find `spng.pc`).

jcupitt: Ooop, good point. I added this too.

PR #2538, opened 2021-11-15
fails if library is requested and missing

remicollet:
Open for discussion, so only done for 1 library (jxl)

If a library is requested, ie when build using --with-libjxl, --with-libjxl=yes or --with-libjxl=module, and if the library is missing, the configure should fails and not silently (only warning) disable the requested feature

Examples, Auto mode

```
./configure
...
checking for LIBJXL... no
configure: WARNING: libjxl not found; disabling libjxl support
...
JXL load/save with libjxl:              no (dynamic module: no)

```

And, Explicitly required mode

```
./configure --with-libjxl=module
...
checking for LIBJXL... no
configure: error: libjxl not found

```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -832,7 +832,7 @@ AC_ARG_WITH([libjxl],
   [with_libjxl=test])
 
 # libjxl as a dynamically loadable module
-AS_IF([test x"$with_libjxl" = x"module" -o x"$with_libjxl" = x"tst"],
+AS_IF([test x"$with_libjxl" = x"module" -o x"$with_libjxl" = x"test"],
   [with_libjxl_module=$gmodule_supported_flag],
   [with_libjxl_module=no])
 
--- a/configure.ac
+++ b/configure.ac
@@ -829,10 +829,10 @@ VIPS_LIBS="$VIPS_LIBS $NIFTI_LIBS"
 AC_ARG_WITH([libjxl],
   AS_HELP_STRING([--without-libjxl], [build without libjxl (default: test)]),
   [with_libjxl=$withval],
-  [with_libjxl=$gmodule_with_flag])
+  [with_libjxl=test])
 
 # libjxl as a dynamically loadable module
-AS_IF([test x"$with_libjxl" = x"module"],
+AS_IF([test x"$with_libjxl" = x"module" -o x"$with_libjxl" = x"tst"],
   [with_libjxl_module=$gmodule_supported_flag],
   [with_libjxl_module=no])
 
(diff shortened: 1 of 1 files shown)


jcupitt:
I don't really have strong feelings about this, I guess it's maybe more expected behaviour.

Any thoughts @kleisauke?

remicollet:
For context, I was bitten by this issue when trying 8.12.0RC1, build --with-libjxl=module, and after a successful build, the jxl module was missing...

lovell:
I think this should make the behaviour more akin to Meson's concept of [feature options](https://mesonbuild.com/Build-options.html#features), which use `enable`, `auto` and `disable` so would map onto `--with-...`, no option and `--without-...` respectively. If so, +1 from me.

configure.ac
```diff
@@ -844,7 +844,9 @@ if test x"$with_libjxl" != x"no"; then
      AS_IF([test x"$with_libjxl_module" = x"no"],
        [PACKAGES_USED="$PACKAGES_USED libjxl"])
     ],
-    [AC_MSG_WARN([libjxl not found; disabling libjxl support])
+    [AS_IF([test x"$with_libjxl" = x"test"],
+       AC_MSG_WARN([libjxl not found; disabling libjxl support]),
+       AC_MSG_ERROR([libjxl not found]))
```
kleisauke, line 849: Perhaps we should mention what the reason for the error is?
```suggestion
       AC_MSG_ERROR([libjxl explicitly requested, but a suitable version was not found]))
```

configure.ac
```diff
@@ -829,10 +829,10 @@ VIPS_LIBS="$VIPS_LIBS $NIFTI_LIBS"
 AC_ARG_WITH([libjxl],
   AS_HELP_STRING([--without-libjxl], [build without libjxl (default: test)]),
   [with_libjxl=$withval],
-  [with_libjxl=$gmodule_with_flag])
+  [with_libjxl=test])
```
kleisauke, line 832: Nice, this means we can get rid of the `gmodule_with_flag` variable if we do this for all libraries.

jcupitt:
OK, let's do this.

It'll need lots of changes to configure to get consistent behaviour for every option, so let's delay merging until 8.12 is out of the door.

kleisauke (approved):
Looks good to me. It's a bit unfortunate that there is no autotools equivalent of Meson's ` --fatal-meson-warnings` option.

jcupitt:
We could even bite the bullet and switch to meson + ninja for 8.13.

remicollet:
> It'll need lots of changes to configure to get consistent behaviour for every option, so let's delay merging until 8.12 is out of the door.

Make sense, not urgent
Will try to take care of other lib in the next days

kleisauke:
> We could even bite the bullet and switch to meson + ninja for 8.13.

I'm fine with adding support for the Meson build system, but I'm not sure if we should remove the Autotools build system in the same release. Perhaps we should follow the [same strategy as GLib](https://mail.gnome.org/archives/gtk-devel-list/2018-June/msg00012.html) (i.e., keep the Autotools build in tree and available in the release archive; then, drop the Autotools build during the next development cycle)?

FWIW, I tried to improve the Autotools build system with commit https://github.com/kleisauke/libvips/commit/42228a7db9ace1add2e08b3ec43162c516f1ac4d last evening, but then I got tired.

jcupitt:
Agreed, a tick - tock strategy is probably more sensible.

> FWIW, I tried to improve the Autotools build system with commit kleisauke@42228a7 last evening, but then I got tired.

Working on build systems has that effect on me too :(

jcupitt:
Let's merge this and make matching changes for the other dependencies in more PRs during the 8.13 cycle. Thank you for this, @remicollet.

PR #2539, opened 2021-11-15
allow libjxl 0.5

remicollet:
See https://github.com/libvips/libvips/commit/21cdf22535e2ff6f070dc54c1e172eca02100d10#commitcomment-60291444

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -837,9 +837,9 @@ AS_IF([test x"$with_libjxl" = x"module"],
   [with_libjxl_module=no])
 
 if test x"$with_libjxl" != x"no"; then
-  PKG_CHECK_MODULES(LIBJXL, libjxl_threads >= 0.6 libjxl >= 0.6,
+  PKG_CHECK_MODULES(LIBJXL, libjxl_threads >= 0.5 libjxl >= 0.5,
     [AC_DEFINE(HAVE_LIBJXL,1,
-       [define if you have libjxl >= 0.6 installed.])
+       [define if you have libjxl >= 0.5 installed.])
      with_libjxl=yes
      AS_IF([test x"$with_libjxl_module" = x"no"],
        [PACKAGES_USED="$PACKAGES_USED libjxl"])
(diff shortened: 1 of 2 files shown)


jcupitt:
Hi again @remicollet 

I was trying to avoid adding tests and ifdefs for an old library API that all distros will surely replace very soon. Do you think this is an essential change?

remicollet:
IIUC, the is API changein 0.6.

Update policy usually forbid such change in stable branches, ex nothing will probably happen in Fedora before F36, but @eclipseo which maintain the package may help here

So all users of vips 8.11 using the jxl module will lost features when upgrading to 8.12

lovell:
Given there are a couple of CVEs present in most versions of libjxl prior to the latest 0.6.1, is increasing the priority of https://bugzilla.redhat.com/show_bug.cgi?id=2018648 something we're able to do?

There's a bit of a coincidence here in that libjxl [CVE-2021-22563](https://nvd.nist.gov/vuln/detail/CVE-2021-22563), which can cause a process to segfault when decoding a JPEG-XL image, was found by the libvips fuzzers. I believe Chrome had to issue an emergency patch for this.

I'd be OK with temporarily removing libjxl as a dependency of libvips if only versions prior to v0.6.1 are available.

Le plus important, merci beaucoup Remi for everything you do to help maintain the RPM ecosystem!

remicollet:
> Given there are a couple of CVEs present in most versions of libjxl prior to the latest 0.6.1

So is irrelevant in packages, CVE fixes can be backported, so you can use a 0.5 and be safe ;)
This is how enterprise distributions work, ex https://access.redhat.com/security/updates/backporting/

remicollet:
> I'd be OK with temporarily removing libjxl as a dependency of libvips if only versions prior to v0.6.1 are available.

This is what I have (tried to) explained, users updating from 8.11 to 8.12 will lose the feature (and this will block the update, as the vips-jxl sub-package is no more available)

jcupitt:
OK, that's reasonable. Let's do this painful thing.

I suppose it's just the downside of us adding support for a very young and rapidly developing library.

configure.ac
```diff
@@ -851,6 +851,19 @@ if test x"$with_libjxl" != x"no"; then
   )
 fi
 
+if test x"$with_libjxl" = x"yes"; then
+  save_LIBS="$LIBS"
+  LIBS="$LIBS $LIBJXL_LIBS"
+  # Introduced in 0.6
+  AC_CHECK_FUNCS(JxlEncoderInitBasicInfo,[
+     AC_DEFINE(HAVE_LIBJXL_JXLENCODERINITBASICINFO,1,
+               [define if you have heif_image_handle_get_raw_color_profile.])
```
kleisauke, line 860:
```suggestion
               [define if you have JxlEncoderInitBasicInfo.])
```

jcupitt: Heh yes I noticed after merge, sigh. I'll fix it.

remicollet:
Thanks

PR #2547, opened 2021-11-20
libnsgif: Update to the latest upstream master

tlsa:
This removes the libvips patch to stop libnsgif modifying the source data.  Upstream libnsgif now treats the source data as `const`.

It also updates for a minor API change.

I'm opening this now because the next set of changes I'm planning for libnsgif will be a full overhaul of the library API.


jcupitt:
Thanks @tlsa -- merged for 8.13.

PR #2553, opened 2021-11-24
Set memory (image volume) limit for fuzzing.

eustas:
This should resolve OOMs / timeouts in JXL decoder.


eustas:
Nota bene: has not yet checked if it will compile.

eustas:
As suspected, private API is not visible. Going to see what could be done on JXL / VIPS side to fix that.

eustas:
closing for now: likely will be done on jxl side

Issue #2554, opened 2021-11-24
autogen error: HAVE_INTROSPECTION does not appear in AM_CONDITIONAL


eustas:
## Bug report

**To Reproduce**
1. run `autogen.sh`
3. See `error: HAVE_INTROSPECTION does not appear in AM_CONDITIONAL`

**Expected behavior**
Autogen should generate configure, etc.

**Environment**
 - OS: Debian
 - Vips: master

eustas:
Executing (incomplete?) configure reports:
```
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a race-free mkdir -p... /usr/bin/mkdir -p
checking for gawk... gawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
./configure: line 4318: syntax error near unexpected token `1.30.0'
./configure: line 4318: `GOBJECT_INTROSPECTION_CHECK(1.30.0)'
```

lovell:
Hi, did you see #1752 ?

eustas:
Indeed, this seems to be the same issue. I believe it is worth adding a check to `autogen.sh`, so that users will not be confused anymore and get the problem reported in a human-readable way.

Same for missing `gtk-doc-tools` dependency.

jcupitt:
Perhaps something like this at the top of autogen.sh?

```
if ! gtkdocize; then
        echo "*** You must have gtkdocize to be able to reconfigure"
        exit 1
fi      

# this will only detect the most basic kind of missing gio error,
# hopefully that's enough to catch this most times
if ! pkg-config gobject-introspection-1.0 --exists; then
        echo "*** You must have gobject-introspection to be able to reconfigure"
        exit 1
fi
```

Very basic, but it would catch most errors.

eustas:
Looks good!

jcupitt:
OK, committed to master. Thanks for the suggestion @eustas.

PR #2555, opened 2021-11-25
Make GOI dependency optional

kleisauke:
See: #2554.


kleisauke:
Note that this was inspired by HarfBuzz: https://github.com/harfbuzz/harfbuzz/blob/3160789701463bcd822c26010da4de33c92c85d9/configure.ac#L175-L186

jcupitt:
Ooof, thanks for fixing my mess.

PR #2559, opened 2021-11-26
ci: add CIFuzz integration

DavidKorczynski:
This adds CI integration with CIFuzz. CIFuzz will run the libvips OSS-Fuzz fuzzers for a short period of time (600 sec) for each PR. The benefit is that regressions and bugs has a chance of being captured early in the development process. Currently the action is triggered on pull requests, but feel free to modify however you like both in terms of when it's triggered and the duration.

Signed-off-by: David Korczynski <david@adalogics.com>

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,24 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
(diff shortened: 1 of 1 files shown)


jcupitt:
Hi @DavidKorczynski,

This is nice! Sure, let's try it out. Thank you very much for doing this work.

jcupitt:
See fuzz results here, for reference: https://github.com/libvips/libvips/actions/workflows/cifuzz.yml?query=

jcupitt:
See fuzz results here, for reference: https://github.com/libvips/libvips/actions/workflows/cifuzz.yml?query=

PR #2563, opened 2021-11-29
cplusplus: silence missing sentinel warning

lovell:
Spotted this warning when compiling the latest 8.12.

```
VImage.cpp: In member function 'void vips::VImage::write_to_buffer(const char*, void**, size_t*, vips::VOption*) const':
VImage.cpp:736:58: warning: missing sentinel in function call [-Wformat=]
  736 |   g_object_get( target.get_target(), "blob", &blob, NULL );
      |                                                          ^
```
Low priority, and same fix as https://github.com/libvips/libvips/pull/668 etc.

### DIFF
--- a/cplusplus/VImage.cpp
+++ b/cplusplus/VImage.cpp
@@ -733,7 +733,7 @@ VImage::write_to_buffer( const char *suffix, void **buf, size_t *size,
 				set( "in", *this )->
 				set( "target", target ) );
 
-		g_object_get( target.get_target(), "blob", &blob, NULL );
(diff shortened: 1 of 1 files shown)


jcupitt:
C++ rules around NULL are so confusing :( Would `0` also work?

lovell:
C++11 onwards provides `nullptr` for this scenario, but versions prior to that can use either `(void *) NULL` or `(void *) 0`

jcupitt:
Ah got it, thanks. OK, let's mereg!

PR #2571, opened 2021-12-04
Windows: don't set create time on invalid/special file descriptors

lovell:
Calling `vips__set_create_time` with a file descriptor value less than 3 appears to end the process when run on Windows.

Having read https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/get-osfhandle?view=msvc-170 my understanding is:

* If you pass -1 to `_get_osfhandle` it can crash.
* If you pass 0, 1 or 2 to `_get_osfhandle` it can return a "valid" handle of -2 (note this is not the same as an `INVALID_HANDLE_VALUE` of -1) then `SetFileTime` can crash.

Originally reported at https://github.com/lovell/sharp/issues/2999

### DIFF
--- a/libvips/iofuncs/util.c
+++ b/libvips/iofuncs/util.c
@@ -586,6 +586,12 @@ vips__set_create_time( int fd )
 	SYSTEMTIME st;
 	FILETIME ft;
 
+	/* Create time cannot be set on invalid or stream
+	 * (stdin, stdout, stderr) file descriptors.
+	 */
+	if( fd < 3 )
+		return;
+
 	if( (handle = (HANDLE) _get_osfhandle( fd )) == INVALID_HANDLE_VALUE )
 		return;
 	GetSystemTime( &st );
@@ -635,7 +641,7 @@ vips__fopen( const char *filename, const char *mode )
 	fp = g_fopen( filename, mode );
 
 #ifdef G_OS_WIN32
-	if( mode[0] == 'w' )
+	if( fp && mode[0] == 'w' )
 		vips__set_create_time( _fileno( fp ) ); 
 #endif /*G_OS_WIN32*/
 


libvips/iofuncs/util.c
```diff
@@ -612,7 +612,7 @@ vips__open( const char *filename, int flags, int mode )
 	fd = g_open( filename, flags, mode );
 
 #ifdef G_OS_WIN32
-	if( mode & O_CREAT )
+	if( fd > 2 && mode & O_CREAT )
```
kleisauke, line 615: Perhaps we could should move this check to `vips__set_create_time`, since it's also used by `vips__fopen`?
```diff
--- a/libvips/iofuncs/util.c
+++ b/libvips/iofuncs/util.c
@@ -586,7 +586,8 @@ vips__set_create_time( int fd )
 	SYSTEMTIME st;
 	FILETIME ft;
 
-	if( (handle = (HANDLE) _get_osfhandle( fd )) == INVALID_HANDLE_VALUE )
+	if( fd < 3 ||
+		(handle = (HANDLE) _get_osfhandle( fd )) == INVALID_HANDLE_VALUE )
 		return;
 	GetSystemTime( &st );
 	SystemTimeToFileTime( &st, &ft );
```

After checking `vips__fopen`, I think that also needs this patch:
```diff
--- a/libvips/iofuncs/util.c
+++ b/libvips/iofuncs/util.c
@@ -635,7 +635,7 @@ vips__fopen( const char *filename, const char *mode )
 	fp = g_fopen( filename, mode );
 
 #ifdef G_OS_WIN32
-	if( mode[0] == 'w' )
+	if( fp && mode[0] == 'w' )
 		vips__set_create_time( _fileno( fp ) ); 
 #endif /*G_OS_WIN32*/
```
(since `_fileno( NULL );` can crash too, arghh)

kleisauke:
For context; this crash is a [security feature of UCRT](https://docs.microsoft.com/en-us/cpp/c-runtime-library/security-features-in-the-crt#additional-security-features) (Universal C Runtime in Windows), similar to https://github.com/lovell/sharp/issues/2494#issuecomment-749498530. Since v8.10.6 the Windows binaries have switched from `msvcrt` to UCRT by default (by using the [llvm-mingw](https://github.com/mstorsjo/llvm-mingw) toolchain).

libvips/iofuncs/util.c
lovell: Done, thanks Kleis.

kleisauke (approved):
LGTM! I can confirm that this resolves the crash with the test added via commit https://github.com/lovell/sharp/commit/9eada2ac5a9d55574fe85e8e958087b467fad005.

Should I make a `v8.12.1-build2` release with this patch incorporated?

kleisauke:
Thinking about this more, we can also consider making these errors recoverable by using [`_set_invalid_parameter_handler`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/set-invalid-parameter-handler-set-thread-local-invalid-parameter-handler) in sharp. See Microsoft's documentation regarding [Parameter Validation](https://docs.microsoft.com/en-us/cpp/c-runtime-library/parameter-validation).

> **Invalid Parameter Handler Routine**
> [...] **By default, the invalid parameter calls `_invoke_watson`, which causes the application to close and generate a mini-dump.** [...] You can change this behavior by using the functions `_set_invalid_parameter_handler` or `_set_thread_local_invalid_parameter_handler` to set the invalid parameter handler to your own function. [...]

mingw-w64 makes these errors recoverable by default for executables.
https://github.com/mingw-w64/mingw-w64/blob/eac8c38c10ca4e22feb2f1e5e530b1e2daedc20f/mingw-w64-crt/crt/crtexe.c#L274

jcupitt:
Urrrgh what a horrible mess, well done for finding all these things.

Yes, maybe disabling these extra checks is the best solution? It seems to make libc behave in a more expected way (returning an error code rather than terminating the process), from reading the docs.

Presumably UCRT also has this feature?

lovell:
Node.js already uses `_set_invalid_parameter_handler` to set a no-op handler via libuv.

https://github.com/nodejs/node/blob/e46c680bf2b211bbd52cf959ca17ee98c7f657f5/deps/uv/src/win/core.c#L181-L194

Perhaps something else in the chain of dependencies is erroneously setting this to something else, or overriding via a thread-local handler?

kleisauke:
It seems that the `_set_invalid_parameter_handler` state is local to the instance of the DLL or application, so it is not shared across DLL boundaries. See e.g. [this SO answer](https://stackoverflow.com/a/13415461).

`msvcrt` did not had this issue as it ignored invalid parameters, but [UCRT will abort by default](https://docs.microsoft.com/en-us/cpp/porting/visual-cpp-change-history-2003-2015#crt-2).
> Many functions now validate their parameters, halting execution if given invalid parameters. This validation may break code that passes invalid parameters and relies on the function ignoring them or just returning an error code.

Given this, would it make sense to initialize an empty invalid parameter handler during `vips_init`? I tried this patch:
<details>
  <summary>Details</summary>
  
```diff
--- a/libvips/iofuncs/init.c
+++ b/libvips/iofuncs/init.c
@@ -322,6 +322,20 @@ empty_log_handler( const gchar *log_domain, GLogLevelFlags log_level,
 {       
 }
 
+#ifdef G_OS_WIN32
+/*
+ * This is a (empty) invalid parameter handler
+ * used on Windows to ensure that the process does
+ * not abort automatically.
+ */
+static void
+empty_invalid_parameter_handler( const wchar_t *expression,
+	const wchar_t *function, const wchar_t *file, unsigned int line,
+	uintptr_t pReserved )
+{
+}
+#endif
+
 /* Attempt to set a minimum stacksize. This can be important on systems with a
  * very low default, like musl.
  */
@@ -684,6 +698,13 @@ vips_init( const char *argv0 )
 		g_log_set_handler( G_LOG_DOMAIN, G_LOG_LEVEL_WARNING, 
 			empty_log_handler, NULL );
 
+	/* Tell the CRT to not exit the application when an invalid parameter is
+	 * passed. The main issue is that invalid FDs will trigger this behaviour.
+	 */
+#ifdef G_OS_WIN32
+	_set_invalid_parameter_handler( empty_invalid_parameter_handler );
+#endif
+
 	/* Set a minimum stacksize, if we can.
 	 */
         if( (vips_min_stack_size = g_getenv( "VIPS_MIN_STACK_SIZE" )) )
```
</details>

Which resolves this crash too.

kleisauke:
... we could also consider switching the Windows binaries back to `msvcrt`, but nowadays UCRT is recommended because `msvcrt` is not compatible with C99 and lacks some features. See: https://www.msys2.org/docs/environments/#msvcrt-vs-ucrt.

jcupitt:
Do we need to test for the availability of `_set_invalid_parameter_handler()` somehow? Perhaps we can assume it.

If this setting is per DLL, does that mean that all out dependencies potentially also need patching?

kleisauke:
Oh, my assumption that `_set_invalid_parameter_handler` state is not shared across DLL boundaries was incorrect. I had a go to test this with this gist: https://gist.github.com/kleisauke/6526f33f6ea77a1216ca3a1222417799.

The trick is to _not_ statically link the UCRT (which is the default, I think), but link this dynamically instead (i.e. with the `/MD` option). See commit https://github.com/kleisauke/sharp/commit/7b49beb79771e246cf94a3b6dbb0f96238485dd9 and the associated test suite run: https://github.com/kleisauke/sharp/runs/4423202097.

lovell:
@kleisauke A PR for sharp with that change would be very welcome.

I guess this is a case of _why not both_, as other consumers of libvips could run into this at some point, and there's no point trying to set creation timestamps on streams anyway.

kleisauke:
I just opened PR https://github.com/lovell/sharp/pull/3002. I agree, the `_set_invalid_parameter_handler()` workaround is only there to prevent further regressions. Calling Windows functions with invalid parameters should be avoided whenever possible.

The only thing that worries me is that this hides further possible bugs, maybe we should log an error in this handler instead of doing nothing.

jcupitt:
OK, let's do it. What do you think about adding something like that sharp patch to `vips_init()` as well?

Issue #2576, opened 2021-12-10
cgifsave file size larger than expected for some gifs


jmreid:
I have a few example gifs that seem to really increase in filesize when run through libvips. This doesn't seem to be the case for _all_ gifs, but i haven't been able to figure out why it's seemingly much worse for some files.

Given this example input gif (~897 KB):
![example](https://user-images.githubusercontent.com/76880/145625646-e5dfc67a-3c16-4ab4-aadc-14652d7b2698.gif)

`vipsthumbnail` (using cgif) returns a 3.5MB image:
`vipsthumbnail "example.gif[n=-1]" --size 200 -o example_vips_out.gif`
![example_vips_out](https://user-images.githubusercontent.com/76880/145625858-0c6245af-3f61-41bf-98c0-d9da2df3f18b.gif)

Gifsicle returns a less nicer-looking image, but it's ~1MB:
`gifsicle --resize-fit 200x_ -o example_gifsicle_out.gif example.gif `
![example_gifsicle_out](https://user-images.githubusercontent.com/76880/145625952-0eec2aee-ca2b-49b1-a6d9-ecbb6a772451.gif)

Here's another example I found online:
![test](https://user-images.githubusercontent.com/76880/145632813-5fe13f32-dea8-4bf0-a0da-5e8396c723cf.gif)

This is a 570KB file, and if I run it through vips to resize to 1400px wide:
```
vipsthumbnail "test.gif[n=-1]" --size 1400 -o test_vips_out.gif

du -h test_vips_out.gif
 24M	test_vips_out.gif
```

Using gifsicle:
```
gifsicle --resize-fit 1400x_ -o test_gifsicle_out.gif test.gif  

du -h test_gifsicle_out.gif
344K	test_gifsicle_out.gif
```

Again, Gifsicle's image is not as nice looking. But that filesize difference is quite a bit, especially because the source is only 570KB.

I'm running:
```
vips --vips-version
libvips 8.12.1-Wed Nov 24 15:41:04 UTC 2021

gifsicle --version
LCDF Gifsicle 1.93
```

@dloebl @MCLoebl Found you a couple examples! 😄

dloebl:
Thanks!
I had a quick look:
`example.gif:` has only a global colour table with 128 entries
`example_vips_out.gif:` almost every frame has its own local colour table with 256 entries
`example_gifsicle_out.gif:` has only a global colour table with 128 entries

`test.gif:` has only a global colour table with 64 entries
[...]

PR #2628, opened 2022-01-19
cgifsave: avoid size issue

dloebl:
Avoid size issues when resizing GIF animations with a static background. The trick is to restore the original transparency setting after quantization and dithering.
**Background:** Keeping many pixels at the transparency index can reduce the size a lot and is the reason why the original animations in (#2576) are quite small in terms of bytes despite their huge width/height. The current resizing and subsequent color quantization method puts the focus on image quality but can make this transparency optimization less effective.

The suggested fix depends on dloebl/cgif#38 and cgif 0.2.0 to be released:

**Results:**
`$ vipsthumbnail --size 200 city.gif[n=-1] -o vips_city.gif`
`$ vipsthumbnail --size 1400 watch.gif[n=-1] -o vips_watch.gif`
`$ du -h vips_*`
**`2.2M vips_city.gif`
`812K vips_watch.gif`**

The size is still larger than before, as the transparency setting is changed by the resizing. However, the two animations are probably extreme cases and the size problem is mitigated quite a bit with this change.
I've visualized the transparency setting for the resized city GIF and the original (green means that a pixel is identical to the frame before):
**Resized:**
out
**Original:**
sorg

One can see some "kernels" (in the resized image). I guess these come from the interpolation method and illustrate where the remaining increase in size comes from. To further improve the size it could be interesting to have a "transparency-friendly" resize method (which keeps identical areas identical). For instance, nearest-neighbor resizing would be "transparency-friendly" but is of course not ideal in terms of quality. An alternative would be to determine transparent areas in a lossy way in cgifsave (I think @jcupitt once mentioned such a low-pass filter). So the problem is finding the right balance between quality and size.

Please let me know if there are changes required on this PR.

### DIFF
--- a/libvips/foreign/cgifsave.c
+++ b/libvips/foreign/cgifsave.c
@@ -100,6 +100,10 @@ typedef struct _VipsForeignSaveCgif {
 	 */
 	VipsPel *index;
 
+	/* frame_bytes before (needed for transparency trick).
+	*/
+	VipsPel *frame_bytes_bef;
+
 	/* The frame as written by libcgif.
 	 */
 	CGIF *cgif_context;
@@ -138,6 +142,7 @@ vips_foreign_save_cgif_dispose( GObject *gobject )
 
 	VIPS_FREE( cgif->palette_rgb );
 	VIPS_FREE( cgif->index );
+	VIPS_FREE( cgif->frame_bytes_bef)
 
 	VIPS_FREEF( cgif_close, cgif->cgif_context );
 
@@ -165,6 +170,7 @@ vips_foreign_save_cgif_write_frame( VipsForeignSaveCgif *cgif )
 	 */
 	int n_pels = frame_rect->height * frame_rect->width;
 	guint max_sum = 256 * n_pels * 4;
+	VipsPel *frame_bytes_cur;
 	VipsPel *frame_bytes = 
 		VIPS_REGION_ADDR( cgif->frame, 0, frame_rect->top );
 
@@ -193,6 +199,12 @@ vips_foreign_save_cgif_write_frame( VipsForeignSaveCgif *cgif )
 		p[3] = p[3] >= 128 ? 255 : 0;
 		p += 4;
 	}
+
+	/* Keep last frame_bytes in memory for transparency trick
+	*  which avoids the size explosion (#2576).
+	*/
+	frame_bytes_cur = g_malloc0(4 * frame_rect->width * frame_rect->height);
+	memcpy(frame_bytes_cur, frame_bytes, 4 * frame_rect->width * frame_rect->height);
 
 	/* Do we need to compute a new palette? Do it if the frame sum
 	 * changes.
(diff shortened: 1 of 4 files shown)


jcupitt:
> @jcupitt: Do you think such a modification would make sense? How would you weight the different color channels?

Yes, that's a good idea. CIELAB is the simplest perceptually uniform colour space (mostly uniform anyway!) so I would use that.

https://en.wikipedia.org/wiki/CIELAB_color_space#From_CIEXYZ_to_CIELAB

sRGB -> XYZ is this:

https://en.wikipedia.org/wiki/SRGB#From_CIE_XYZ_to_sRGB

You only need 8 bit data and you don't care about small errors, so you can cut a lot of corners. You'll find you end up with something like:

- Look up table for each of RGB to get linear data
- 3x3 matmul (so R' = a * R + b * G + c * B etc.)
- Look up table for cuberoot term in LAB
- L = G, a = 500 * (R - G), b = 200 * (G - B)

Not too painful.

libvips/foreign/cgifsave.c
```diff
@@ -303,6 +335,36 @@ vips_foreign_save_cgif_write_frame( VipsForeignSaveCgif *cgif )
...
+	if( cgif->frame_bytes_head ) {
+		if( !cgif->has_transparency ) {
+			const uint8_t transIndex = cgif->lp->count;
+			for(uint32_t i = 0; i < frame_rect->width * frame_rect->height; i++) {
+				if(cgif_cmppixel(frame_bytes + 4 * i, cgif->frame_bytes_head + 4 * i, cgif->maxerror) == 0) {
+					cgif->index[i] = transIndex;
+				} else {
+					memcpy(cgif->frame_bytes_head + 4 * i, frame_bytes + 4 * i, 4);
+				}
+			}
```
jcupitt, line 359: The `*` inside the inner loop makes me uncomfortable, though the compiler ought to optimize this away. I think I'd still rework this as:

```C
                        const uint8_t trans_index = cgif->lp->count;
                        int n_pixels = frame_rect->width * frame_rect->height;
                        double maxerror = cgif->maxerror;

                        VipsPel *cur, *bef;
                        int i;

                        cur = frame_bytes;
                        bef = cgif->frame_bytes_head;
                        for( i = 0; i < n_pixels; i++ ) {
                                if( cgif_pixels_are_equal( cur, bef, maxerror ) )   
                                        cgif->index[i] = trans_index;
                                else {
                                        bef[0] = cur[0];
                                        bef[1] = cur[1];
                                        bef[2] = cur[2];
                                        bef[3] = cur[3];
                                }

                                cur += 4;
                                bef += 4;
                        }
```

Not all compilers will take out the very slow `memcpy()` for you.

You might as well make `maxerror` a double in the operation args as well.

jcupitt (commented):
The C programmer in me couldn't resist making some unnecessary optimisation, of course :( gif write is about 10% faster on my PC with the suggested changes. Maybe not a useful amount.

I noticed some small layout and naming things too.

dloebl:
Makes all perfect sense. Thanks!
I added your recommendations accordingly.
Additionally, I switched the lossy transparency off by default (maxerror = 0).
I think it makes sense to be a bit conservative before the pixel compare function has been improved (RGB to CIELAB).

PR #2630, opened 2022-01-23
Fix jpegload pipe

jcupitt:
We could sometimes free the VipsSource header cache with unused bytes. This patch delays freeing the header cache until we are certain it is exhausted.

See https://github.com/kleisauke/net-vips/issues/155


kleisauke (approved):
LGTM, I can confirm that this fixes the mentioned issue. This should probably be cherry-picked in 8.12 as well.

PR #2637, opened 2022-01-31
Add meson build system

tintou:
Allows to use the Meson build system to build the project.

Note: This PR isn't complete yet but is opened to show the interest in using the meson build system.


meson.build
```diff
@@ -0,0 +1,171 @@
...
+    cfg_var.set('ENABLE_DEPRECATED', '1')
+endif
+
+#AC_CHECK_HEADERS([])
+cfg_var.set_quoted('GETTEXT_PACKAGE', 'vips8.13')
+cfg_var.set_quoted('VIPS_PREFIX', get_option('prefix'))
+cfg_var.set_quoted('VIPS_LIBDIR', get_option('prefix') / get_option('libdir'))
+libintl_dep = cc.find_library('intl', required : false)
+if libintl_dep.found()
+    cfg_var.set('ENABLE_NLS', 1)
```
eli-schwartz, line 145: This will not be found on platforms where the gettext family of functions is built into the libc. That means it won't work on Linux with either the GNU libc or musl libc, and won't work on any of the BSDs by default.

It will be found on Windows using MSYS2, if GNU libintl is installed, or on the BSDs in the event that the user manually installs the GNU libintl port.

Instead, do one of two things:
- check if `cc.has_function('ngettext')` is successfully found, and if it is not found, fallback to checking if the `intl` library can be found
- Use https://mesonbuild.com/Dependencies.html#intl, which will just report "not found" on versions of meson before 0.59

libvips/meson.build
```diff
@@ -0,0 +1,70 @@
...
+subdir('morphology')
+subdir('module')
+subdir('mosaicing')
+subdir('create')
+
+libvips_lib = library('vips',
+    enumtypes,
+    link_with: [
+        deprecated_lib,
+        foreign_lib,
```
eli-schwartz, line 34: I think you probably want link_whole here. The static libraries need to be merged into a shared library, which is the purpose of `-Wl,--whole-archive`, otherwise with the default library being `shared` libvips.so will discard all symbols that only appear in archives to `link_with` but aren't referenced in the library's direct object files.

Also you could append each library to an array and avoid some use of placeholder arrays for optional components. :) (Also, it is more easily extended from a single location.)

eli-schwartz (changes requested):

libvips/meson.build
tintou: Thanks for the explanation, i've now adapted it to your suggestion

meson.build
tintou: Thanks for the hint, I've adapted it to the first suggestion, I hope that we could just use `glib/gi18n.h` at some point as we already have GLib available

jcupitt:
Hi @tintou,

Thank you for this PR.

Yes, we've been talking about switching build systems on libvips/devchat and meson is one of the front runners. If we can make this work well, we'd like to get rid of autotools completely.

Anyway, thumbs up from me and I'll try to find time to try this out.

meson.build
```diff
@@ -0,0 +1,201 @@
...
+
+# rules:
+# sources changed: increment revision
+# binary interface changed: increment current, reset revision to 0
+#   binary interface changes backwards compatible?: increment age
+#   binary interface changes not backwards compatible?: reset age to 0
+library_current = '57'
+library_revision = '0'
+library_age = '15'
+library_version = '.'.join ([library_current, library_revision, library_age])
```
kleisauke, line 22: I think this causes the soname to be generated incorrectly as `libvips{-cpp}.so.57` instead of `libvips{-cpp}.so.42`. Also, the `library_revision` and `library_age` should probably be reversed.
```suggestion
library_current = 57
library_age = 15
library_revision = 0
library_version = '@0@.@1@.@2@'.format(library_current - library_age, library_age, library_revision)
```

tintou: Oops, applied, thanks :)

fuzz/meson.build
```diff
@@ -0,0 +1,40 @@
...
+        fuzz_prog + '.cc',
+        dependencies: libvips_dep,
+        link_with: standaloneengine
+    )
+endforeach
+
+test_fuzz = configure_file(
+    input: 'test_fuzz.sh',
+    output: 'test_fuzz.sh',
+    copy: true,
```
eli-schwartz, line 30: Do you actually need this file? It does nothing other than define $ASAN_OPTIONS $UBSAN_OPTIONS $VIPS_WARNING and then run each of fuzz_execs in turn. You could easily implement that directly in meson. There's an `env` kwarg:

https://mesonbuild.com/Reference-manual_functions.html#test

meson.build
```diff
@@ -0,0 +1,440 @@
...
+    if cc.has_function('jpeg_c_bool_param_supported', prefix: '#include <jpeglib.h>', dependencies: libjpeg_dep)
+        cfg_var.set('HAVE_JPEG_EXT_PARAMS', '1')
+    endif
+endif
+
+# Look for libspng first
+# 0.6.1 uses "libspng.pc", git master libspng uses "spng.pc"
+spng_dep = dependency('libspng', version: '>=0.6', required: get_option('spng'))
+if not spng_dep.found()
+    spng_dep = dependency('spng', version: '>=0.6', required: get_option('spng'))
```
eli-schwartz, line 176: https://mesonbuild.com/Reference-manual_functions.html#dependency

> *Since 0.60.0* more than one name can be provided, they will be tried in order and the first name to be found will be used.

Currently you're listing 0.56.0 as the minimum, so you may just wish to add a comment as a reminder to upgrade in the future.

It is a bit ugly in the current state because if `-Dspng=enabled` then the first lookup will error out if it cannot be found, and will not proceed to check the new name. You can script around this with `if not get_option('spng').disabled()`...

tintou: Yeah I think that meson 0.60.0 might be too new for now, thanks for catching this though!

fuzz/meson.build
tintou: it might not be needed but I'm starting with a 1:1 port and we can simplify it once it is done :smiley:

meson.build
eli-schwartz: The `if not X.disabled() and` check for spng should probably be copied over to imagemagick too (which I didn't notice has the same issue until you changed the `required: ` kwarg there.)

libvips/module/meson.build
```diff
@@ -0,0 +1,16 @@
...
+    'jxl.c',
+    'heif.c',
+    'magick.c',
+    'openslide.c',
+    'poppler.c',
+)
+
+libvips_sources += module_sources
+
+module_lib = static_library('module',
```
kleisauke, line 11: This probably needs to be `shared_module` instead of `static_library`. It also needs a few other changes (such as depending on `libvips_dep`, etc.) and including the `name_suffix: module_suffix` workaround:
```meson
module_suffix = []
# Keep the autotools convention for shared module suffix because GModule
# depends on it: https://gitlab.gnome.org/GNOME/glib/issues/1413
if ['darwin', 'ios'].contains(host_machine.system())
  module_suffix = 'so'
endif
```

tintou: I've now implemented the module handling

doc/meson.build
```diff
@@ -0,0 +1,132 @@
...
+)
+
+markdown_content_files_docbook = []
+foreach markdown_content_file : markdown_content_files
+    # we have some files in markdown ... convert to docbook for gtk-doc
+    # pandoc makes section headers, we want refsect3 for gtk-doc
+    markdown_content_files_docbook += configure_file(
+        input: markdown_content_file,
+        command: [
+            find_program('pandoc'),
```
eli-schwartz, line 68: I think you probably want to put that in the same get_option() guard as your `gnome.gtkdoc()` invocation.

With the current implementation, you always build this pandoc output, which also means you always require pandoc installed as a build dependency, even when configuring with gtk_doc disabled.

eli-schwartz: Also slightly interested to see you using configure_file there... this triggers buildsystem regeneration when the inputs are modified. Although maybe they aren't frequently modified...

tintou: Ah right, I've added a guard around the folder inclusion then

tintou:
I think that I got almost everything covered, I only have to add the Doxygen support and it'll be ready for review

tintou:
It is now ready for review

lovell:
Thank you Corentin, I'm +1 for having to maintain configuration for only one build tool and am happy with meson as the replacement for autotools.

To test this, I think we should ensure at least the following scripts/repos can successfully build libvips with meson/ninja.

* https://github.com/libvips/build-win64-mxe/
* https://github.com/lovell/sharp-libvips

These scripts generate various combinations of binaries, shared and static libraries, using both native and cross compilation with gcc and clang on x64 and ARM64 machines, so I'd expect them to catch most things that might be missing.

Do we want to remove autotools as part of this PR, or in a follow up?

jcupitt:
> Do we want to remove autotools as part of this PR, or in a follow up?

I don't mind. Maybe in a follow-up? It would make this PR easier to work with if it's a bit smaller.

kleisauke:
> * https://github.com/libvips/build-win64-mxe/

I'll test this today.

> Do we want to remove autotools as part of this PR, or in a follow up?

I don't mind either. A tick–tock strategy (as noted in https://github.com/libvips/libvips/pull/2538#issuecomment-970180738) is probably more appropriate for downstream packagers.

jcupitt:
It's failing for me:

```
$ meson setup build --prefix=~/vips
Using 'PKG_CONFIG_PATH' from environment with value: '/home/john/vips/lib/pkgconfig:'
Using 'PKG_CONFIG_PATH' from environment with value: '/home/john/vips/lib/pkgconfig:'
The Meson build system
Version: 0.56.2
Source dir: /home/john/GIT/libvips
Build dir: /home/john/GIT/libvips/build
...
Run-time dependency libheif found: YES 1.11.0

meson.build:322:19: ERROR: Function does not take positional arguments.
```

Something odd about `get_variable` here?

eli-schwartz:
The specific manner in which it succeeds for me is instructive.


```
WARNING: Project specifies a minimum meson_version '>=0.56' but uses features which were added in newer versions:
 * 0.58.0: {'Positional argument to dep.get_variable()'}
```

jcupitt:
Ah OK, duh. Perhaps `meson.build` should have 0.58 as the min version?

jcupitt:
I upgraded my meson and it's failing for nifti -- it doesn't seem to add `-I/usr/include/nifti` to cflags.

Maybe:

```
libnifti_dep = cc.find_library('nifti2', has_headers: 'nifti1_io.h', required: false)
```

jcupitt:
`libvips/foreign/meson.build` has a typo:

```
openslide_module_sources = files(
    'popplerload.c',
)
```

tintou:
I've fixed the typo in `libvips/foreign/meson.build`
Regarding the `nifti2` library, the best would really be to add a pkg-config file upstream
I've also changed the `get_variable` to use Eli's solution

eli-schwartz:
```diff
-glib_prefix = glib_dep.get_variable(pkgconfig: 'prefix')
+glib_prefix = glib_dep.get_variable('prefix')
```

Seems like you accidentally broke one working use of it too, though.

meson.build
```diff
@@ -0,0 +1,468 @@
...
+if get_option('nsgif')
+    cfg_var.set('HAVE_NSGIF', '1')
+endif
+if get_option('ppm')
+    cfg_var.set('HAVE_PPM', '1')
+endif
+if get_option('analyze')
+    cfg_var.set('HAVE_ANALYZE', '1')
+endif
+if get_option('deprecated')
```
jcupitt, line 417: Looks like a copy/paste error, this is on lines 405/6/7 as well.

meson_options.txt
```diff
@@ -0,0 +1,46 @@
+option('deprecated', type : 'boolean', value : false, description: 'Build deprecated components')
```
jcupitt, line 1: I think this needs to default to `true` --- we still have some large projects which use the old vips7 API.

meson.build
```diff
@@ -0,0 +1,468 @@
...
+    cfg_var.set('ENABLE_DEPRECATED', '1')
+endif
+
```
jcupitt, line 420: I think we need an option for radiance as well, perhaps:

```
if get_option('radiance')
    cfg_var.set('HAVE_RADIANCE', '1')
endif
```

`radiance` is already in `meson_options.txt`.

libvips/include/vips/meson.build
```diff
@@ -0,0 +1,169 @@
...
+        'morphology.h',
+        'draw.h',
+        'basic.h',
+        'object.h',
+        'region.h'
+    ),
+    h_template: 'enumtypes.h.in',
+    c_template: 'enumtypes.c.in',
+    install_header: true,
+    install_dir: 'vips'
```
jcupitt, line 76: I think this needs to be:

```
    install_dir: 'include/vips'
```

jcupitt (commented):
I found a few small things. With these changes I was able to successfully build a working nip2 (an old vips7 application), and `ninja test` passed.

I configured with:

```
meson setup build --prefix=~/vips --libdir=/home/john/vips/lib
```

I had to set libdir, since meson puts the arch in there by default. This is probably better, but it is a change from the autotools behaviour.

Curiously you can use `~` in the prefix argument, but not in libdir.

jcupitt:
Build speed isn't the main benefit of switching, of course, but for fun I did try a quick benchmark:

autotools:
- autogen: 17s
- make install: 16s

meson:
- meson setup: 2.8s
- ninja install: 4.6s

So 3x or maybe 4x faster, depending. That's a very nice improvement, and more than I was expecting.

meson.build
```diff
@@ -0,0 +1,468 @@
...
+    gio_dep,
+    gobject_dep,
+    gmodule_dep,
+    expat_dep,
+]
+
+prefix_dir = get_option('prefix')
+lib_dir = prefix_dir / get_option('libdir')
+module_dir = lib_dir / 'vips-modules-@0@.@1@'.format(version_major, version_minor)
+
```
jcupitt, line 49: I noticed there's no support for i18n. I don't know how easy it is to move the `po/` machinery to meson.

eli-schwartz: `import('i18n').gettext('localename')` and you're done.

tintou: Ah, I indeed forgot this part, I've just re-added it, the handling of the po/ machinery is very easy in meson

eli-schwartz:
> I had to set libdir, since meson puts the arch in there by default. This is probably better, but it is a change from the autotools behaviour.

It's a debian quirk, meson runs dpkg to get the architecture directory (otherwise it would be wrong).

> Curiously you can use ~ in the prefix argument, but not in libdir.

Use `--libdir=lib`, it's calculated relative to the prefix anyway (as in, `get_option('libdir')` will remove the prefix and convert it into a relative path).

meson.build
```diff
@@ -0,0 +1,468 @@
+project('vips', 'c', 'cpp',
+    version: '8.13.0',
+    meson_version: '>=0.56',
+    default_options: [
+        'c_std=c99'
```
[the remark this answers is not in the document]
tintou: The `restrict` argument is used in the project which is a c99 addition

libvips/include/vips/meson.build
eli-schwartz: Well, `get_option('includedir') / 'vips'` anyway.

meson.build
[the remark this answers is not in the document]
jcupitt: True, but the old configure tested for restrict, so it still built on vanilla C89.

meson.build
```diff
@@ -0,0 +1,471 @@
...
+
+
+libjpeg_dep = dependency('libjpeg', required: get_option('jpeg'))
+if libjpeg_dep.found()
+    libvips_deps += libjpeg_dep
+    cfg_var.set('HAVE_JPEG', '1')
+    # features like trellis quant are exposed as extension parameters ...
+    # mozjpeg 3.2 and later have #define JPEG_C_PARAM_SUPPORTED, but we must 
+    # work with earlier versions
+    if cc.has_function('jpeg_c_bool_param_supported', prefix: '#include <jpeglib.h>', dependencies: libjpeg_dep)
```
kleisauke, line 179: `stdio.h` needs to be included before `jpeglib.h`. See: https://github.com/libjpeg-turbo/libjpeg-turbo/issues/17#issuecomment-139620135. Otherwise, `HAVE_JPEG_EXT_PARAMS` would never be defined when building with MozJPEG.
```suggestion
    if cc.has_function('jpeg_c_bool_param_supported',
                        prefix: '''#include <stdio.h>
                                   #include <jpeglib.h>''',
                        dependencies: libjpeg_dep)
```

kleisauke:
It looks like there are some differences in `include/vips` when cross-compiling libvips for Windows with Meson (also `include/vips/enumtypes.h` is missing, but I'm not sure if that would be a problem).

<details>
  <summary>Details</summary>

```diff
diff --git a/include/vips/soname.h b/include/vips/soname.h
index 1111111..2222222 100644
--- a/include/vips/soname.h
+++ b/include/vips/soname.h
@@ -1,2 +1,9 @@
-/* This file is autogenerated, do not edit. */
-#define VIPS_SONAME "libvips-42.dll"
+/*
+ * Autogenerated by the Meson build system.
+ * Do not edit, your changes will be lost.
+ */
+
+#pragma once
+
+#define VIPS_SONAME "libvips"
+
diff --git a/include/vips/version.h b/include/vips/version.h
index 1111111..2222222 100644
--- a/include/vips/version.h
+++ b/include/vips/version.h
@@ -16,7 +16,43 @@
 #define VIPS_LIBRARY_REVISION	(0)
 #define VIPS_LIBRARY_AGE	(15)
 
-#define VIPS_CONFIG		"enable debug: no, enable deprecated library components: no, enable modules: yes, use fftw3 for FFT: yes, accelerate loops with orc: yes, ICC profile support with lcms: yes (lcms2), zlib: yes, text rendering with pangocairo: yes, font file support with fontconfig: yes, RAD load/save: yes, Analyze7 load/save: yes, PPM load/save: yes, GIF load:  yes, GIF save with cgif: yes, EXIF metadata support with libexif: yes, JPEG load/save with libjpeg: yes (pkg-config), JXL load/save with libjxl: yes (dynamic module: yes), JPEG2000 load/save with libopenjp2: yes, PNG load with libspng: yes, PNG load/save with libpng: yes (pkg-config libpng >= 1.2.9), quantisation to 8 bit: yes, TIFF load/save with libtiff: yes (pkg-config libtiff-4), image pyramid save: yes, HEIC/AVIF load/save with libheif: yes (dynamic module: no), WebP load/save with libwebp: yes, PDF load with PDFium:  no, PDF load with poppler-glib: yes (dynamic module: yes), SVG load with librsvg-2.0: yes, EXR load with OpenEXR: yes, OpenSlide load: yes (dynamic module: yes), Matlab load with matio: yes, NIfTI load/save with niftiio: yes, FITS load/save with cfitsio: yes, Magick package: MagickCore (dynamic module: yes), Magick API version: magick6, load with libMagickCore: yes, save with libMagickCore: yes"
+#define VIPS_CONFIG		"enable debug: $enable_debug \
+enable deprecated library components: $enable_deprecated \
+enable modules: $gmodule_supported_flag \
+use fftw3 for FFT: $with_fftw \
+accelerate loops with orc: $with_orc \
+ICC profile support with lcms: $with_lcms \
+zlib: $with_zlib \
+text rendering with pangocairo: $with_pangocairo \
+font file support with fontconfig: $with_fontconfig \
+RAD load/save: $with_radiance \
+Analyze7 load/save: $with_analyze \
+PPM load/save: $with_ppm \
+GIF load:  $with_nsgif \
+GIF save with cgif: $with_cgif \
+EXIF metadata support with libexif: $with_libexif \
+JPEG load/save with libjpeg: $with_jpeg \
+JXL load/save with libjxl: $with_libjxl (dynamic module: $with_libjxl_module) \
+JPEG2000 load/save with libopenjp2: $with_libopenjp2 \
+PNG load with libspng: $with_libspng \
+PNG load/save with libpng: $with_png \
+selected quantisation package: $quantisation_package \
+TIFF load/save with libtiff: $with_tiff \
+image pyramid save: $with_gsf \
+HEIC/AVIF load/save with libheif: $with_heif (dynamic module: $with_heif_module) \
+WebP load/save with libwebp: $with_libwebp \
+PDF load with PDFium:  $with_pdfium \
+PDF load with poppler-glib: $with_poppler (dynamic module: $with_poppler_module) \
+SVG load with librsvg-2.0: $with_rsvg \
+EXR load with OpenEXR: $with_OpenEXR \
+OpenSlide load: $with_openslide (dynamic module: $with_openslide_module) \
+Matlab load with matio: $with_matio \
+NIfTI load/save with niftiio: $with_nifti \
+FITS load/save with cfitsio: $with_cfitsio \
+Magick package: $with_magickpackage (dynamic module: $with_magick_module) \
+Magick API version: $magick_version \
+load with libMagickCore: $enable_magickload \
+save with libMagickCore: $enable_magicksave"
 
 /** 
  * VIPS_SONAME:
@@ -30,7 +66,7 @@
 /* Not really anything to do with versions, but this is a handy place to put
  * it.
  */
-#define VIPS_EXEEXT ".exe"
+#define VIPS_EXEEXT ""
 #define VIPS_ENABLE_DEPRECATED 0
 
 #endif /*VIPS_VERSION_H*/
```
</details>

jcupitt:
Yes, I noticed a small problem with `VIPS_CONFIG` as well. With the autotools build system you get:

```
$ vips --vips-config
enable debug: no
enable deprecated library components: yes
enable modules: no
use fftw3 for FFT: yes
...
```

But with meson it's:

```
$ vips --vips-config
enable debug: $enable_debug enable deprecated library components: $enable_deprecated ...
```

ie. config variables are not expanded (and perhaps need updating?), newlines are missing.

tintou:
I've fixed the issue with `VIPS_CONFIG`, note that meson is outputing `true` and `false` instead of `yes` and `no`, that's something we can fine-tune if we want to.
I've naively reimplemented `VIPS_EXEEXT` and `VIPS_SONAME` but in the future it would be better to remove them (I haven't seen another project providing such values)

jcupitt:
> I've naively reimplemented VIPS_EXEEXT and VIPS_SONAME but in the future it would be better to remove them (I haven't seen another project providing such values)

I agree, I think they were used in an old win32 build system we had, but are not needed now.

libvips/meson.build
```diff
@@ -0,0 +1,126 @@
...
+
+pkg.generate(
+    libvips_lib,
+    requires: [ glib_dep, gio_dep, gobject_dep ],
+    name: 'vips',
+    description: 'Image processing library',
+)
+
+if get_option('introspection')
+    vips_gir = gnome.generate_gir(
```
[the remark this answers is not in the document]
jcupitt: fixed! thanks!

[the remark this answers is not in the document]
jcupitt: Actually, now I try it, the `--quiet` option makes the scanner less chatty, but does not hide warnings. Maybe there's no simple way to fix this since it's caused (I think) by the scanner not understanding some C syntax.

How about setting introspection to default off? At the moment the deluge of spurious scanner warnings is making genuine compiler warnings harder to spot.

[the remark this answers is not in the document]
tintou: I've set it to false by default, we can revise this once the warnings are fixed :)

meson.build
tintou: thanks, fixed

meson.build
```diff
@@ -0,0 +1,491 @@
...
+    gio_dep,
+    gobject_dep,
+    gmodule_dep,
+    expat_dep,
+    m_dep,
+]
+
+prefix_dir = get_option('prefix')
+lib_dir = prefix_dir / get_option('libdir')
+modules_enabled = get_option('modules').require(
```
jcupitt, line 51: I see an error here with meson 0.56.2 on ubuntu 21.10:

```
meson.build:51:0: ERROR: Unknown method "require" in object.
```

Is this a later feature?

tintou: Ah indeed, this is a 0.59 feature, I've reworked it to be able to use 0.56, thanks for trying it

meson.build
```diff
@@ -0,0 +1,494 @@
...
+        libvips_deps += quantizr_depquantizr_dep
+        cfg_var.set('HAVE_QUANTIZR', '1')
+        quantisation_package = quantizr_dep
+    endif
+endif
+
+cgif_dep = disabler()
+if quantisation_package.found()
+    cgif_dep = dependency('cgif', required: get_option('cgif'))
+    if quantizr_dep.found()
```
kleisauke, line 179: Looks like a copy/paste error.
```suggestion
    if cgif_dep.found()
```

tintou: Thanks, looks like I made many :sweat:

jcupitt:
I think all my personal use-cases are now covered, so this all done, from my point of view. This is a great! I tested:

- `ls -R` on the install area matches autotools in terms of number of files and filenames
- I can build vips7 projects successfully (nip2)
- downstream python projects with native components all build correctly
- i18n works (eg. try `vips measure` with en_GB locale)
- I can build C++ projects that use the libvips C++ API
- the generated .pc files look reasonable
- I fixed the compiler warnings in git master

I've not tested cross-compiling to Windows, I've not tried all of the options, and I've not looked at eg. packaging for homebrew. We should get sign-off from Kleis and Lovell too.

kleisauke:
Windows cross-complication builds seems to work. However, it did require a large patch for symbol visibility, as there is no equivalent of  CMake's `WINDOWS_EXPORT_ALL_SYMBOLS` (or libtool's `always_export_symbols` option) in Meson. See:
https://github.com/tintou/libvips/compare/tintou/meson...kleisauke:tintou/meson
(I think it would be better to resolve that as a follow-up)

It also required a smaller patch regarding the dynamic modules (i.e. avoid linking module dependencies, built with `name_prefix: ''` to remove the lib-prefix, and make modules a `type: 'feature'` option). See commit https://github.com/kleisauke/libvips/commit/9ba6a36bafffca83981c5d911bdf652ddb9e1f5c.

eli-schwartz:
Do you actually want all symbols unconditionally? This is the default with GCC/clang anyway, and for MSVC all cmake does here is the equivalent of `dumpbin` and autogenerate what meson calls `vs_module_defs: 'libvips.def'`. You can just check this file into git, though you'd need to manually update it when new symbols are added.

Limiting the scope of your symbol table isn't a bad thing though.

kleisauke:
Indeed, those changes are not bad and helps to reduce the symbol table. My purpose is to expose only the symbols of libvips within the semi-static DLL (i.e. all dependencies are statically-linked, except GLib). I think libtool internally did `-Wl,--exclude-libs=ALL` to prevent exporting the symbols from the static libraries.

I actually implemented those `VIPS_API` definitions with symbol visibility.
https://github.com/kleisauke/libvips/blob/fba4a76c3855a399a91ea771d8eaa9f6606b41d3/meson.build#L76-L94
https://github.com/kleisauke/libvips/blob/fba4a76c3855a399a91ea771d8eaa9f6606b41d3/libvips/include/vips/vips.h#L92-L102

meson.build
```diff
@@ -0,0 +1,494 @@
...
+endif
+if (glib_dep.version().version_compare('>=2.62'))
+    cfg_var.set('HAVE_DATE_TIME_FORMAT_ISO8601', '1')
+endif
+
+if modules_enabled
+    cfg_var.set('ENABLE_MODULES', '1')
+endif
+
+func_names = [ 'getcwd', 'gettimeofday', 'getwd', 'memset', 'munmap', 'putenv', 'realpath', 'strcasecmp', 'strchr', 'strcspn', 'strdup', 'strerror', 'strrchr', 'strspn', 'vsnprintf', 'realpath', 'mkstemp', 'mktemp', 'random', 'rand', 'sysconf', 'atexit', '_aligned_malloc', 'posix_memalign', 'memalign', 'cbrt', 'hypot', 'atan2', 'asinh' ]
```
kleisauke, line 75: Note to myself: most checks are currently unused, this can be simplified to:
```suggestion
func_names = [ 'vsnprintf', 'atexit', '_aligned_malloc', 'posix_memalign', 'memalign', 'cbrt', 'hypot', 'atan2', 'asinh' ]
```

eli-schwartz:
Ah, I probably misread that then.

Note the meson.build portion can be partly simplified by using the meson kwarg `gnu_symbol_visibility: 'hidden'` which sets that compiler option only when the compiler supports it. Also, in general you can just assume setting GNU symbol visibility attributes is either "correct" or "ignored when not building a shared library" and just set them unconditionally all the time. It is only MSVC where this behaves annoyingly.

kleisauke:
I noticed that these 2 definitions are never defined in `config.h` when building with Meson:
```c
/* have pthread_setattr_default_np(). */
#define HAVE_PTHREAD_DEFAULT_NP 1

/* define if your C++ can mix vector and scalar arithmetic */
#define HAVE_VECTOR_ARITH 1
```

tintou:
Added the two missing definitions, checked for the "thread" dependency

meson.build
```diff
@@ -0,0 +1,533 @@
...
+    cfg_var.set('HAVE_TIFF', '1')
+    # ZSTD and WEBP in TIFF added in libtiff 4.0.10
+    if cc.get_define('COMPRESSION_WEBP', prefix: '#include <tiff.h>', dependencies: libtiff_dep) != ''
+        cfg_var.set('HAVE_TIFF_COMPRESSION_WEBP', '1')
+    endif
+endif
+
+# 2.40.3 so we get the UNLIMITED open flag
+librsvg_dep = dependency('librsvg-2.0', version: '>=2.40.3', required: get_option('rsvg'))
+cairo_dep = dependency('cairo', version: '>=1.2', required: get_option('rsvg'))
```
kleisauke, line 303: Oh, it looks like I mistakenly assumed that `popplerload` didn't use cairo in one of my patches. Perhaps we could do something like this?:
```suggestion
cairo_dep = dependency('cairo', version: '>=1.2', required: get_option('rsvg') or get_option('poppler'))
```
(untested)

and change
```meson
elif libpoppler_dep.found()
```
below to
```meson
elif libpoppler_dep.found() and cairo_dep.found()
```

BTW, I suppose that it's safe to add the same dependency twice to `libvips_deps`? Since
```meson
        libvips_deps += libpoppler_dep
```
needs probably to be changed to:
```meson
        libvips_deps += libpoppler_dep
        libvips_deps += cairo_dep
```

meson.build
```diff
@@ -0,0 +1,533 @@
...
+    if librsvg_dep.version().version_compare('>=2.46')
+        cfg_var.set('HAVE_RSVG_HANDLE_RENDER_DOCUMENT', '1')
+    endif
+endif
+
+zlib_dep = dependency('zlib', version: '>=0.4', required: get_option('zlib'))
+if zlib_dep.found()
+    libvips_deps += zlib_dep
+    cfg_var.set('HAVE_ZLIB', '1')
+endif
```
kleisauke, line 318: This can be merged into the if statement above. Since libvips only uses zlib to handle svgz images.

meson.build
```diff
@@ -0,0 +1,533 @@
...
+    cfg_var.set('HAVE_IMAGEQUANT', '1')
+    quantisation_package = imagequant_dep
+endif
+
+# only if libimagequant not found
+quantizr_dep = disabler()
+if not quantisation_package.found()
+    quantizr_dep = dependency('quantizr', required: get_option('quantizr'))
+    if quantizr_dep.found()
+        libvips_deps += quantizr_depquantizr_dep
```
kleisauke, line 205: Typo: 
```suggestion
        libvips_deps += quantizr_dep
```

libvips/foreign/meson.build
```diff
@@ -0,0 +1,133 @@
...
+    foreign_sources += openslide_module_sources
+endif
+
+libvips_sources += foreign_sources
+
+foreign_lib = static_library('foreign',
+    foreign_sources,
+    foreign_headers,
+    dependencies: libvips_deps,
+    link_args: '-no-undefined',
```
kleisauke, line 130: Ah, this was probably the reason for the warnings that were fixed with commit https://github.com/libvips/libvips/commit/0d8056b8c59818e0f57c379d64db8e1ffaa00a03. I wonder if we should add this linker option to all libraries (perhaps with `add_project_link_arguments`?).

meson.build
```diff
@@ -0,0 +1,533 @@
...
+elif libpoppler_dep.found()
+    if get_option('poppler-module') and modules_enabled
+        cfg_var.set('POPPLER_MODULE', '1')
+    else
+        libvips_deps += libpoppler_dep
+    endif
+    cfg_var.set('HAVE_POPPLER', '1')
+endif
+
+libnifti_dep = cc.find_library('nifti2', has_headers: 'nifti1_io.h', required: false)
```
kleisauke, line 438: AFAIK, libvips NIfTI loader/saver has only been tested with the NIfTI-1 library.
```suggestion
libnifti_dep = cc.find_library('niftiio', has_headers: 'nifti1_io.h', required: false)
```

meson.build
tintou: It's safe to add the dependency several times :), changed

meson.build
```diff
@@ -0,0 +1,544 @@
...
+)
+
+libvips_deps += config_dep
+
+subdir('libvips')
+if get_option('gtk_doc')
+    subdir('doc')
+endif
+subdir('cplusplus')
+subdir('man')
```
lovell, line 540: Would it be possible to wrap each of these last 5 `subdir` directives ("man" through to "fuzz") with their own boolean options, defaulting to true, to allow building only the shared/static libraries without tools and tests? This is based on a discussion at https://github.com/lovell/sharp-libvips/issues/127#issuecomment-1030901615

eli-schwartz: I don't really see the point e.g. "man" just creates installation rules which have no build time effect, only an install-time effect. You can use `meson install --tags runtime` to avoid installing any files tagged as `man`, see https://mesonbuild.com/Installing.html#installation-tags for details.

For the other directories, they actually do create build rules, but those could for the most part be defined with `build_by_default: false` if you don't want to make everyone automatically build them (they could be built by name if desired).

Granted Meson currently doesn't respect build_by_default for test executables, ideally they should not be built by `ninja all` in that case but still get built by e.g. `ninja test-prereq` and pulled in as a dependency of `ninja test`. However, this theoretical `test-prereq` is instead merged into `all` and that's a bit annoying. We do intend to fix it, there's a WIP PR for it somewhere... in the meantime many people do have *one* project option "tests=false" which disables descending into `tests/meson.build` or what have you.

Adding 5 different options seems like a lot of cognitive overhead for very little gain.

lovell: Thank you, I agree, options fatigue is a real thing. Meson's use of a human-readable `meson_options.txt` file and its ability to fail a build when it sees unknown options is a huge improvement. I wasn't aware of the predefined `--tags`, very useful, thanks. I'd be happy with a single `tests` boolean option, which I guess would prevent descent into the `test` and `fuzz` subdirs.

meson.build
```diff
@@ -0,0 +1,544 @@
...
+cairo_dep = dependency('cairo', version: '>=1.2', required: get_option('rsvg'))
+if librsvg_dep.found() and cairo_dep.found()
+    libvips_deps += librsvg_dep
+    libvips_deps += cairo_dep
+    cfg_var.set('HAVE_RSVG', '1')
+    # 2.46 for rsvg_handle_render_document
+    if librsvg_dep.version().version_compare('>=2.46')
+        cfg_var.set('HAVE_RSVG_HANDLE_RENDER_DOCUMENT', '1')
+    endif
+    zlib_dep = dependency('zlib', version: '>=0.4', required: get_option('zlib'))
```
kleisauke, line 316: I think we need to add `zlib_dep = disabler()` outside this if-statement as well. Otherwise it would fail when configuring with `-Drsvg=disabled` (noticed this on the WebAssembly build of libvips).
```
libvips/include/vips/meson.build:86:0: ERROR: Unknown variable "zlib_dep".
```

tools/meson.build
```diff
@@ -0,0 +1,42 @@
+tools = [
+    'vips',
+    'vipsedit',
+    'vipsthumbnail',
+    'vipsheader'
+]
+
+foreach tool : tools
+executable(tool,
```
kleisauke, line 9: Indentation is missing here.

tools/meson.build
```diff
@@ -0,0 +1,42 @@
...
+    'shrink_width.in',
+    'batch_image_convert.in',
+    'batch_rubber_sheet.in',
+    'batch_crop.in',
+]
+
+script_data = configuration_data()
+script_data.set('prefix', get_option('prefix'))
+foreach script : scripts
+configure_file(
```
kleisauke, line 27: ditto.

meson.build
eli-schwartz: A disabler is functionally identical to `dependency('', required: false)` except for the special feature of disabling any meson.build statement that tries to use it directly rather than via the `.found()` method.

This doesn't directly matter in the current PR, but may or may not be desirable in the general sense.

eli-schwartz: In meson 0.59.0, you can use this: https://mesonbuild.com/Reference-manual_returned_feature.html#featurerequire

```meson
want_zlib = get_option('zlib').require(
    librsvg_dep.found() and cairo_dep.found(),
    error_message: 'librsvg and cairo not found'
)
```

This would then error out if `-Dzlib=enabled` with the error message:
```
Feature zlib cannot be enabled: librsvg and cairo not found
```

Currently if you define zlib as enabled but librsvg and/or cairo are optional + they aren't found, zlib simply gets ignored. Alas, that's the best you can do while still building with the version of Meson available on Debian stable. Perhaps add a comment to migrate to `.require()` once the minimum version of Meson is updated?

meson.build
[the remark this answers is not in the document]
kleisauke: How about moving to `gnu99` (i.e. the same as GLib)? That would fix these errors (noticed on the WebAssembly build):
```
../libvips/foreign/analyze2vips.c:252:8: error: implicit declaration of function 'isascii' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
                if( !isascii( buf[i] ) || buf[i] < 32 )
                     ^
../libvips/iofuncs/util.c:746:26: error: implicit declaration of function 'fileno' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
        len = vips_file_length( fileno( fp ) ); 
                                ^
../libvips/iofuncs/util.c:1125:13: error: implicit declaration of function 'ftruncate' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
        if( ftruncate( fd, pos ) ) {
            ^
```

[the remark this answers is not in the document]
jcupitt: Good idea, why not!

Let's get this PR merged first though.

meson.build
kleisauke: +1 for a comment to migrate to `.require()` once we upgraded our minimum version to Meson 0.59.

Note to myself: the module logic here:
```meson
if gmodule_dep.get_variable(pkgconfig: 'gmodule_supported') != 'false'
    modules_enabled = get_option('modules').enabled() or get_option('modules').auto()
elif get_option('modules').enabled()
    error('GModule is not supported on your system, please reconfigure with -Dmodules=disabled')
else
    modules_enabled = false
endif
```

could also be replaced with:
```meson
modules_enabled = get_option('modules').require(
    gmodule_dep.get_variable(pkgconfig: 'gmodule_supported') != 'false',
    error_message: 'GModule is not supported on your system, please reconfigure with -Dmodules=disabled
).allowed()
```
by then.

eli-schwartz: Oh ha, that code is even more gnarly, fortunately Meson 0.59 can save us from complicated scripting logic.

meson.build
```diff
@@ -0,0 +1,544 @@
...
+add_project_arguments('-Drestrict=__restrict', language: 'cpp')
+add_project_link_arguments(
+    '-no-undefined',
+    language: 'c',
+)
+
+cc = meson.get_compiler('c')
+cpp = meson.get_compiler('cpp')
+
+glib_dep = dependency('glib-2.0')
```
kleisauke, line 38: Note to myself: minimum required GLib version is missing here.
```suggestion
glib_dep = dependency('glib-2.0', version: '>=2.40')
```

kleisauke (approved):
LGTM! The final comments can also be incorporated later.

jcupitt (approved):
Sorry, I didn't approve earlier. Looks good for my uses.

lovell (approved):
Thank you!

jcupitt:
I made an issue for tracking any small snags we find with the new build system here: https://github.com/libvips/libvips/issues/2650

Let's shift any discussion to that issue (this PR is large enough already).

eli-schwartz:
It is my pleasure to offer advice and reviews to people migrating to / updating their usage of meson. :)

(Seriously. Feel free to `@` me in issues/PRs with your meson-related questions.)

yisibl:
Does this mean we can more easily compile statically linked libvips?

Does this mean we can more easily compile statically linked libvips?

When I try to cross-compile sharp with zig in centos7, I get the error:

```bash
CC="zig cc -target x86_64-macos" CXX="zig c++ -target x86_64-macos" npx node-gyp rebuild

COPY Release/nothing.a
  CXX(target) Release/obj.target/sharp-linux-x64/src/common.o
../src/common.cc:24:10: fatal error: 'vips/vips8' file not found
#include <vips/vips8>
         ^~~~~~~~~~~~
1 error generated.
make: *** [Release/obj.target/sharp-linux-x64/src/common.o] Error 1
``

**Comment by jcupitt** [comment]
You can make a static library very simply with most build systems. With autotools try:

```
./configure --enable-static --disable-shared
```

With meson it's:

```
meson setup build --default-library=static
```

Then compile a program with eg.:

```
gcc -g -Wall myprog.c -Wl,-Bstatic `pkg-config vips --cflags --libs`
```

I think you'll find the difficulty comes in making static versions of all the things that libvips depends on.

**Comment by yisibl** [comment]
@jcupitt Thanks for your reply, I found someone using staticx to do this, can you see if this way can make libvips dependencies static?

https://github.com/whalehub/libvips-static/blob/31a21dadca84f3f9e3c568650f3844feaea5cb29/Dockerfile#L103

**Comment by jcupitt** [comment]
staticx doesn't use static libraries -- it's more like an executable zip file. You should be able to use it on any vips binary.

That dockerfile is interesting, I'll try it here.

**Comment by eli-schwartz** [comment]
I alluded to it earlier in this thread, but you can make static versions of all the things libvips depends on, using Meson subprojects:

```
for i in expat glib lcms2 libexif libjpeg-turbo libpng libtiff spng zlib; do 
    meson wrap install "$i"
done
```

And while this doesn't cover every one of libvips' dependencies, it covers a bunch of them and more could be added to the wrapdb.

There is one slight flaw: checking for gmodule_supported doesn't (currently) work from a glib subproject, since the latest stable release of glib has to support versions of Meson too old to have support for defining variables in a declare_dependency. And this project is *slightly* too backwards compatible to support getting that variable anyway.

The meson.build could be slightly tweaked to not actually check this in the `-Dmodules=disabled` case. Basically change:
```diff
-gmodule_dep = dependency('gmodule-no-export-2.0')
+gmodule_dep = dependency('gmodule-no-export-2.0', required: get_option('modules'))
```

And change:
```meson
if gmodule_dep.get_variable('gmodule_supported') != 'false'
    modules_enabled = get_option('modules').enabled() or get_option('modules').auto()
elif get_option('modules').enabled()
    error('GModule is not supported on your system, please reconfigure with -Dmodules=disabled')
else
    modules_enabled = false
endif
```
to (first two lines changed)
```meson
if gmodule_dep.found() and gmodule_dep.get_variable('gmodule_supported') != 'false'
    modules_enabled = true
elif get_option('modules').enabled()
    error('GModule is not supported on your system, please reconfigure with -Dmodules=disabled')
else
    modules_enabled = false
endif
```

It's normally guaranteed to be found if glib is found at all. But if you explicitly disable it, then you:
1) don't need to bother linking to it when it isn't used (ENABLE_MODULES=0)
2) don't need to check its variables

jcupitt:
@eli-schwartz huh that's interesting. It sounds like it'd be useful a little way down the line.

PR #2654, opened 2022-02-14
Refine header inclusion to allow better introspection generation

tintou:
The headers need to be self-contained for the introspection parser
to work correctly. The entry for external users remains vips/vips.h

### DIFF
--- a/libvips/Vips-8.0.metadata
+++ b/libvips/Vips-8.0.metadata
@@ -1,0 +1,7 @@
+Object.new_from_string skip
+Object.print_summary_class skip
+ArgumentClass.object_class skip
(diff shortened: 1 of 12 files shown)


jcupitt:
Oh is that all it was? That's much better, it's just two pages of warnings now. Thanks!

Is this ready to merge, or would you like to make some more changes?

tintou:
Yeah that what was just it 🙂 and the resulting introspection is really different (and actually usable). It is ready to merge, I'll try to do some follow up MRs to fix the remaining warnings.

jcupitt:
OK, merged. Nice!

PR #2655, opened 2022-02-15
Enhance the introspection support by fixing annotations

tintou:
- Ensure that double asterisk characters are only used for gtk-doc comments
- Also refine the .metadata file

### DIFF
--- a/libvips/conversion/conversion.c
+++ b/libvips/conversion/conversion.c
@@ -70,7 +70,7 @@
  */
 
 /**
- * vips_composite: (method)
+ * vips_composite:
  * @in: (array length=n) (transfer none): array of input images
  * @out: (out): output image
  * @n: number of input images
@@ -79,10 +79,10 @@
  *
  * Optional arguments:
  *
- * * @compositing_space: #VipsInterpretation to composite in
- * * @premultiplied: %gboolean, images are already premultiplied
- * * @x: #VipsArrayInt, position of subimages
- * * @y: #VipsArrayInt, position of subimages
+ * * `compositing_space`: #VipsInterpretation to composite in
+ * * `premultiplied`: %gboolean, images are already premultiplied
+ * * `x`: #VipsArrayInt, position of subimages
+ * * `y`: #VipsArrayInt, position of subimages
  *
  * Composite an array of images together. 
  *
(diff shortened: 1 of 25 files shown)


jcupitt:
Ooop, sorry, I was bored and I had a quick go at this as well. I'll fix the conflicts and merge.

libvips/iofuncs/header.c
```diff
@@ -367,7 +367,7 @@ meta_init( VipsImage *im )
 }
 
 /**
- * vips_image_get_width: (method)
```
jcupitt, line 370: Have you removed this because width is also available as a property? Or have I misunderstood the method annotation (quite likely).

libvips/iofuncs/header.c
```diff
@@ -1451,7 +1451,7 @@ add_fields( VipsImage *image, const char *field, GValue *value, void *a )
 }
 
 /**
- * vips_image_get_fields: (method)
+ * vips_image_get_fields:
```
jcupitt, line 1454: Isn't this a method? I'm not sure I understand.

jcupitt:
Sorry, reverted since I had some questions. Too much coffee.

libvips/iofuncs/header.c
tintou: having a `(method)` forces it to be a method, here as it is already prefixed by `vips_image_` and has `VipsImage *` as first argument there is no ambiguity.
Forcing it to be a method actually makes the method to be `Vips.Image.image_get_fields` when not forcing it makes the method name be `Vips.Image.get_fields`.

libvips/conversion/conversion.c
```diff
@@ -79,10 +79,10 @@
...
  * Optional arguments:
  *
- * * @compositing_space: #VipsInterpretation to composite in
- * * @premultiplied: %gboolean, images are already premultiplied
- * * @x: #VipsArrayInt, position of subimages
- * * @y: #VipsArrayInt, position of subimages
+ * * `compositing_space`: #VipsInterpretation to composite in
+ * * `premultiplied`: %gboolean, images are already premultiplied
+ * * `x`: #VipsArrayInt, position of subimages
+ * * `y`: #VipsArrayInt, position of subimages
```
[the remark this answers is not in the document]
tintou: This change is not really necessary for the introspection, I can revert that change.

libvips/iofuncs/header.c
jcupitt: Ah gotcha. Thanks for the explanation.

PR #2657, opened 2022-02-15
Enhance the introspection support by fixing annotations

tintou:
Also refine the .metadata file


jcupitt:
Look great! I'll add a note to the changelog and credit you.

Thanks for fixing this mess.

jcupitt:
I defaulted introspection on as well, since there are so few warnings now.

PR #2672, opened 2022-02-18
LSan improvements and fixes

kleisauke:
As discussed in https://gitter.im/libvips/devchat?at=61f6c5059a335454064ac2e0, this PR fixes a small mem leak when the `VIPSHOME` env is not set and the `$libdir/vips-modules-8.13` directory does not exits. It also improves the sanitizer options and suppressions.

This `g_malloc` leak (see CI result of commit f78ce58957cae14e0f2f40e388b4296f213a19d5) was suppressed due to the `libglib-2.0.so` presence in the LSan suppressions. The `fast_unwind_on_malloc=0` wasn't added to the `LSAN_OPTIONS` which caused LSan to fail when unwinding with frame pointers (GLib isn't compiled with `-fno-omit-frame-pointer`) and thus resulted in the error being suppressed.

The last commit (b9d60c196fe66e757542dff48f38b8139d9c24b0) fixes this small leak by avoiding the `vips_strdup` call altogether.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ of the libvips optional dependencies.
 cd libvips-x.y.x
 meson setup build-dir --prefix=/aaa/bbb/ccc
 cd build-dir
-meson build
+meson compile
 meson test
 meson install
 ```
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,15 +137,15 @@ jobs:
       - name: Build libvips
         run: ninja -C build
 
+      - name: Check libvips
+        run: ninja -C build test
+
       - name: Install libvips
         run: sudo ninja -C build install
 
       - name: Rebuild the shared library cache
         if: runner.os == 'Linux'
         run: sudo ldconfig
-
-      - name: Check libvips
-        run: ninja -C build test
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -77,7 +77,7 @@ of the libvips optional dependencies.
 cd libvips-x.y.x
 meson setup build-dir --prefix=/aaa/bbb/ccc
 cd build-dir
-meson build
+meson compile
```
jcupitt, line 80: oop!

jcupitt:
nice!

kleisauke:
It looks like we need to increase this timeout even more.
https://github.com/libvips/libvips/blob/3da7caee6c270f0881a8cf1ceac9b5e30ead17f3/test/meson.build#L24-L26

See for example this CI result:
https://github.com/libvips/libvips/runs/5259338474#step:14:65

Probably this slowdown is due to the GLib suppression removal and/or the `fast_unwind_on_malloc=0` setting. Perhaps 120 seconds is a sane limit?

jcupitt:
Wow, that's a really long time. Sure, I'll make it 120.

PR #2676, opened 2022-02-22
Fix building with -Dmodules=disabled

kleisauke:
See: https://github.com/libvips/libvips/commit/d80185060dd26eb1151b965f6f893ec076a480d0#r67247589.

### DIFF
--- a/libvips/deprecated/package.c
+++ b/libvips/deprecated/package.c
@@ -613,6 +613,8 @@ static im_package *built_in[] = {
 	&im__resample,
 	&im__video
 };
+
+#ifdef ENABLE_MODULES
 /* How we represent a loaded plugin.
  */
 typedef struct _Plugin {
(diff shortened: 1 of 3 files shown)


lovell:
Thanks Kleis, it looks like the fuzz tests are failing to link - see https://github.com/libvips/libvips/runs/5293376715?check_suite_focus=true
```
/usr/bin/ld: /lib/x86_64-linux-gnu/libgio-2.0.a(giomodule.c.o): in function `g_io_module_unload_module':
(.text+0x3db): undefined reference to `g_module_close'
```

kleisauke:
It looks like the linker flags need to be re-ordered when libvips does not depend on GModule symbols. PR https://github.com/google/oss-fuzz/pull/7317 should fix that.

lovell (approved):

PR #2677, opened 2022-02-22
heifload: add unlimited option, default to false

lovell:
Prior to v1.6.0, libheif used to limit input images to 16384 pixels along either dimension. This was increased to 32768 in libheif v1.6.0, which also added the `heif_context_set_maximum_image_size_limit` API to set the maximum - see https://github.com/strukturag/libheif/issues/150

This PR resets the default maximum to the sensible 16384 and adds a new `unlimited` flag, the same name/approach as the SVG and PNG loaders, to optionally remove the limit for trusted input (HEIF supports a max image size of 65535x65535). This should keep the default max memory consumption for decompressing HEIF images below ~1.6GB (16384 * 16384 * 3 * 2).

Attempting to load an image above the limits results in an error:
```
Security limit exceeded: Image size 16x17000 exceeds the maximum image size 16384x16384
```
(Note with slight horror the compression efficacy/bombliness of the AVIF format such that a 16x17000 image, albeit blank, occupies only 363 bytes.)


jcupitt:
Nice! I added a note to the changelog as well.

PR #2678, opened 2022-02-23
Add missing annotations to #endif directives

kleisauke:
This PR is to check whether CIFuzz is now working.

### DIFF
--- a/libvips/deprecated/package.c
+++ b/libvips/deprecated/package.c
@@ -652,7 +652,7 @@ plugin_free( Plugin *plug )
 
 	return( 0 );
 }
-#endif
+#endif /*ENABLE_MODULES*/
 
 /* Load a plugin.
  */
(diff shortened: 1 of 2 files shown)


kleisauke:
Seems to work! Only noticed that libjxl uses a bundled lcms2 library. I tried to pass `-DJPEGXL_FORCE_SYSTEM_LCMS2=1` locally to no avail.
```
[ 38%] Building CXX object lib/CMakeFiles/jxl_enc-obj.dir/jxl/enc_color_management.cc.o
In file included from /src/libjxl/lib/jxl/enc_color_management.cc:27:
In file included from /src/libjxl/third_party/highway/hwy/foreach_target.h:113:
/src/libjxl/lib/jxl/enc_color_management.cc:40:10: fatal error: 'lcms2.h' file not found
#include "lcms2.h"
         ^~~~~~~~~
1 error generated.
make[2]: *** [lib/CMakeFiles/jxl_enc-obj.dir/build.make:258: lib/CMakeFiles/jxl_enc-obj.dir/jxl/enc_color_management.cc.o] Error 1
make[2]: *** Waiting for unfinished jobs....
```

That include directive should probably be changed to `#include <lcms2.h>`.

jcupitt:
> That include directive should probably be changed to `#include <lcms2.h>`.

I agree, looks likely.

PR #2682, opened 2022-02-24
set some -Ds depending on the release type

jcupitt:
libvips uses g_assert() quite a bit to check things like pointer bounds,
and this can have a large performance impact.

This PR turns off g_assert() (plus some other things) in release mode,
and enables leak checking (and some other things) in debug mode.

I'm not sure if this is the best way to implement this kind of thing.
Does meson have a more offical path for this?

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -3,7 +3,9 @@ project('vips', 'c', 'cpp',
     meson_version: '>=0.56',
     default_options: [
         # this is what glib uses (one of our required deps), so we use it too
-        'c_std=gnu99'
+        'c_std=gnu99',
+        # turn off asserts etc. in release mode
+        'b_ndebug=if-release'
     ]
 )
 
@@ -33,16 +35,16 @@ add_project_link_arguments(
     language: 'c',
 )
 
-# in release mode, we automatically turn off cast checks and asserts
+# in release mode, we automatically turn off cast checks and g_assert
 if get_option('buildtype').startswith('release')
-    add_project_arguments('-DNDEBUG', language : ['cpp', 'c'])
     add_project_arguments('-DG_DISABLE_CAST_CHECKS', language : ['cpp', 'c'])
     add_project_arguments('-DG_DISABLE_CHECKS', language : ['cpp', 'c'])
     add_project_arguments('-DG_DISABLE_ASSERT', language : ['cpp', 'c'])
 endif
 
 # in debug mode mode, we automatically enable leak checks and fatal warnings
-if get_option('buildtype').startswith('debug')
+# also true for 'debugoptimized'
+if get_option('debug')
     add_project_arguments('-DDEBUG_FATAL', language : ['cpp', 'c'])
     add_project_arguments('-DDEBUG_LEAK', language : ['cpp', 'c'])
 endif
--- a/meson.build
+++ b/meson.build
@@ -35,14 +35,15 @@ add_project_link_arguments(
     language: 'c',
 )
 
-# in release mode, we automatically turn off cast checks and g_assert
-if get_option('buildtype').startswith('release')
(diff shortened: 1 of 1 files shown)


jcupitt:
Could I ping you on this @eli-schwartz ? Does this look sensible?

meson.build
```diff
@@ -33,6 +33,20 @@ add_project_link_arguments(
     language: 'c',
 )
 
+# in release mode, we automatically turn off cast checks and asserts
+if get_option('buildtype').startswith('release')
+    add_project_arguments('-DNDEBUG', language : ['cpp', 'c'])
```
eli-schwartz, line 38: https://mesonbuild.com/Builtin-options.html#base-options

You can set `default_options: [''b_ndebug=if-release']` in the `project()` defaults.

eli-schwartz (commented):
In general, checking the buildtype to start with `debug` might be better off checking the boolean state of `get_option('debug')`. It will be true when the buildtype is debug or debugoptimized, as well as a few cases where it is "custom".

IIRC upstream GLib itself has separate options for this and doesn't consider it strictly tied to the buildtype (i.e. they believe it is reasonable to build in release mode with checks nevertheless enabled).

meson.build
jcupitt: thanks! I think I've followed your suggestions

meson.build
```diff
@@ -33,6 +35,20 @@ add_project_link_arguments(
     language: 'c',
 )
 
+# in release mode, we automatically turn off cast checks and g_assert
+if get_option('buildtype').startswith('release')
```
kleisauke, line 39: Perhaps we should follow GLib's convention here?
```suggestion
if get_option('optimization') in ['2', '3', 's']
```
https://github.com/GNOME/glib/blob/b3311e82ee04c90136ff231358d356e8e587f4da/meson.build#L279-L282

That would also turn off cast checks and make `g_assert` no-op when building with `--buildtype=debugoptimized` (not sure if that's desired) or `--buildtype=minsize`, see:
https://mesonbuild.com/Builtin-options.html#build-type-options

kleisauke: We could also do this:
```diff
--- a/meson.build
+++ b/meson.build
@@ -35,18 +35,16 @@ add_project_link_arguments(
     language: 'c',
 )
 
-# in release mode, we automatically turn off cast checks and g_assert
-if get_option('buildtype').startswith('release')
-    add_project_arguments('-DG_DISABLE_CAST_CHECKS', language : ['cpp', 'c'])
-    add_project_arguments('-DG_DISABLE_CHECKS', language : ['cpp', 'c'])
-    add_project_arguments('-DG_DISABLE_ASSERT', language : ['cpp', 'c'])
-endif
-
 # in debug mode mode, we automatically enable leak checks and fatal warnings
 # also true for 'debugoptimized'
 if get_option('debug')
     add_project_arguments('-DDEBUG_FATAL', language : ['cpp', 'c'])
     add_project_arguments('-DDEBUG_LEAK', language : ['cpp', 'c'])
+elif get_option('optimization') in ['2', '3', 's']
+    # in release mode, we automatically turn off cast checks and g_assert
+    add_project_arguments('-DG_DISABLE_CAST_CHECKS', language : ['cpp', 'c'])
+    add_project_arguments('-DG_DISABLE_CHECKS', language : ['cpp', 'c'])
+    add_project_arguments('-DG_DISABLE_ASSERT', language : ['cpp', 'c'])
 endif
 
 cc = meson.get_compiler('c')
```
to ensure that building with `--buildtype=debugoptimized` or `--buildtype=minsize` would still enable debug facility.

jcupitt: Ah good idea, I'll update.

kleisauke (approved):
LGTM!

PR #1593, opened 2020-03-24
Fix for --disable-deprecated (#1273)

kleisauke:
~~Marked this PR as draft since the whole mosaicing folder has yet to be converted to vips8.~~

This PR resolves #1273 and #798.

PR #2688, opened 2022-02-26
Improve symbol visibility

kleisauke:
As discussed in https://github.com/libvips/libvips/pull/2637#issuecomment-1029063280, this improves the symbol visibility of libvips when building with Meson. This allows the compiler freely optimize functions and in many cases avoid symbols at all.

As noticed in https://github.com/lovell/sharp-libvips/issues/127#issuecomment-1031800398, this would reduce the shared library size by ~1.8 MB when all the dependencies are statically linked. Though, perhaps that's not a fair comparison because previously the symbols of all these dependencies were also exported.

Marked this PR as draft due to these TODO items:
- [x] ~Check `abi-compliance-checker` result.~ _See compat reports below._
- [x] ~Split out gettext cleanup as it affects many files.~
  #2695
- [x] ~Split out the last two commits.~ _Split into individual PRs, see:
  #2691
  #2692_
- [x] ~What about the private `*_get_type` symbols? Do we still need to export those? See:
  https://gist.github.com/kleisauke/4ef71366856f1f36535558b69fd0e946~ _These should not be exported._
- [ ] Export all symbols from the `vips/internal.h`, `vips/private.h` and `vips/util.h` headers?
  - [ ] Or move required internal symbols to public API?
- [x] Check whether dependencies/bindings of libvips doesn't depend on internal symbols
  - [x] ~nip2~ See #2704.
  - [x] ~pyvips~ No additional exports are needed.
  - [x] ~ruby-vips~ No additional exports are needed.
  - [x] ~net-vips~ No additional exports are needed.
  - [x] ~vipsdisp~ See commit 2a840f15fbdeccc7acd76e8082050a81581507a1.
  - [x] ~sharp~ No additional exports are needed.
  - [x] ~php-vips-ext~ No additional exports are needed.
  - [x] ~wasm-vips~ See commit https://github.com/kleisauke/wasm-vips/commit/5a92a23a7689396dad2280de32b7d9ce9088ee56.
  - [x] Anything else that needs to be checked?
- [x] ~Support autotools build as well?~ _Only Meson for now, to avoid unnecessary extra work._
- [ ] Inform downstream packagers.


jcupitt:
I'm happy to skip autotools support -- I think the whole autotools system should be removed once 8.13 is out.

meson.build
```diff
@@ -91,6 +92,20 @@ if modules_enabled
...
+    if host_os == 'windows' or host_os == 'cygwin'
+        cfg_var.set('DLL_EXPORT', true)
+        if cc.get_id() == 'msvc' or cc.get_id() == 'clang-cl'
+            cfg_var.set('_VIPS_PUBLIC', '__declspec(dllexport)')
+        elif cc.has_argument('-fvisibility=hidden')
+            cfg_var.set('_VIPS_PUBLIC', '__attribute__((visibility("default"))) __declspec(dllexport)')
+        endif
+    elif cc.has_argument('-fvisibility=hidden')
+        cfg_var.set('_VIPS_PUBLIC', '__attribute__((visibility("default")))')
+    endif
```
eli-schwartz, line 106: I would set this regardless of default_library, as long as the host_os is some kind of non-Windows platform. It will never hurt, and may help the compiler optimize out some things even when building statically (I think).

Instead of cc.has_argument, you should use `cc.has_function_attribute('visibility:hidden')`.

eli-schwartz (changes requested):

jcupitt:
> What about the *_get_type symbols?

I think at least some are necessary -- the type cast macros like eg.`VIPS_IMAGE()` will use `vips_image_get_type()`.

kleisauke:
Indeed, all enumerations `_get_type()`'s and those defined in the public headers are already exported. 
https://github.com/libvips/libvips/blob/a7a4afbca7173d2f9857008b818a011b66fb2669/libvips/include/vips/enumtemplate#L14-L15
https://github.com/libvips/libvips/blob/a7a4afbca7173d2f9857008b818a011b66fb2669/libvips/include/vips/image.h#L367-L368

It's just those defined with `G_DEFINE_TYPE` that causes a lot of noise in the compat reports. https://github.com/libvips/libvips/blob/a7a4afbca7173d2f9857008b818a011b66fb2669/libvips/arithmetic/abs.c#L81

The `abi-compliance-checker` results can be found here:
https://kleisauke.nl/compat_reports/vips/8.12.2_to_master/compat_report.html
https://kleisauke.nl/compat_reports/vips/master_to_symbol-visibility/compat_report.html

meson.build
kleisauke: This was somewhat inspired by GLib:
https://github.com/GNOME/glib/blob/888ca6e36fddaf67b7a02e5ffabecec80f979d06/meson.build#L216-L234

I think(?) they excluded symbol visibility for static libraries because of the Windows-specific `DLL_EXPORT` definition (which controls whether the `DllMain` entry point should be present) and the `__declspec(dllexport)` thing (which is usually avoided for static libraries). But indeed, for non-Windows platforms, I think symbol visibility would also be useful for static libraries. (This made me wonder about [this hack](https://github.com/kleisauke/libvips-packaging/blob/meson/build/lin.sh#L471-L473), that is probably no longer necessary if symbol visibility is also set for static libraries in GLib)

Commit https://github.com/libvips/libvips/pull/2688/commits/dd5652f4033fb21c4d998c6042f1e82a24d961a4 resolves this.

kleisauke:
I've updated the compat reports above. It seems that I missed the `-public-headers` argument of `abi-dumper`, which caused private symbols (such as `vips_abs_get_type`, etc.) to be marked as public.

kleisauke:
Looking at the latest compat report, these symbols needs some further investigation:
https://github.com/libvips/libvips/blob/357fb057aaa4479049d88c069890d42f97485b29/libvips/include/vips/private.h#L204-L206
https://github.com/libvips/libvips/blob/357fb057aaa4479049d88c069890d42f97485b29/libvips/include/vips/internal.h#L242-L243

I think the `_vips__argument_id` still needs to be exported(?), although it is preceded by an underscore.
https://github.com/libvips/libvips/blob/357fb057aaa4479049d88c069890d42f97485b29/libvips/iofuncs/object.c#L215-L217

However, the `vips_image_eval` and `vips_image_posteval` are only visible internally in `vips7compat.h` (as a double underscore definition).
https://github.com/libvips/libvips/blob/357fb057aaa4479049d88c069890d42f97485b29/libvips/include/vips/vips7compat.h#L230-L232

kleisauke:
With commit 02ac949d7b9a481b30eda65184f00107b05b4710 I'm able to build and run nip2. All builtin examples of nip2 seems to work expect for `2_point_mosaic`; which segfaults here. Let me investigate that further.

kleisauke:
Commit 916c90f4ae3bd2d96d6705503642a742cfeb5526 and 14c5f8812fe61854225f170f1517023e9f379ac1 fixes the segfault and those `g_object_unref: assertion 'G_IS_OBJECT (object)' failed` warnings in the nip2 `2_point_mosaic` workspace. These regressions were erroneously introduced in PR #1593, which converted the whole mosaicing directory to vips8.

Unfortunately, it still seems to produce a incorrectly sized image, see:
<details>
  <summary>Details</summary>

2_point_mosaic
</details>

@jcupitt Do you have any idea what is causing this? It looks like some sort of overflow, but I couldn't find anything obvious in the `mosaic1.c` vips7 -> vips8 diff.

kleisauke:
Commit 6b69782450a0d6de88c402998daf37849f50d0f2 fixes the incorrectly sized image bug for me. I now see this:
<details>
  <summary>Details</summary>

2_point_mosaic
</details>

I'll move these mosaicing fixes to a separate PR.

kleisauke:
I think this is ready for review now. Here are the updated compat reports:
https://kleisauke.nl/compat_reports/vips/8.12.2_to_master/compat_report.html
https://kleisauke.nl/compat_reports/vips/master_to_symbol-visibility/compat_report.html

To minimize the diff, the remaining TODO item should probably be addressed after this PR. I've annotated the required internal symbols with `TODO(kleisauke): VIPS_API is required by ...`.

Note that this PR intentionally does not expose the `vips_image_eval` and `vips_image_posteval` functions defined in `internal.h`.
https://github.com/libvips/libvips/blob/97a1d6f8dc5e39b012428aff8d9377efc1597f04/libvips/include/vips/internal.h#L238-L239

As these are only visible internally in `vips7compat.h` (as a double underscore definition).
https://github.com/libvips/libvips/blob/97a1d6f8dc5e39b012428aff8d9377efc1597f04/libvips/include/vips/vips7compat.h#L230-L232

(If required, I could still expose these functions)

libvips/iofuncs/type.c
```diff
@@ -698,7 +698,8 @@ vips_blob_get( VipsBlob *blob, size_t *length )
 		length, NULL, NULL, NULL ) ); 
 }
 
-/* vips_blob_set:
+/**
+ * vips_blob_set:
```
jcupitt, line 702: Ahhhh so *that's* why we were getting the complaints, sigh. Nice spot.

jcupitt:
This is really nice cleanup operation! And a lot of work, good job.

PR #2689, opened 2022-02-26
Meson: ensure NIfTI can be found

kleisauke:
```
Dependency NIFTI not found: CMake: invalid module NIFTI for NIFTI.
Try to explicitly specify one or more targets with the "modules" property.
Valid targets are:
['NIFTI::znz', 'NIFTI::niftiio', 'NIFTI::nifti1_tool']
Run-time dependency nifti found: NO 
```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -465,11 +465,12 @@ elif libpoppler_dep.found() and cairo_dep.found()
     cfg_var.set('HAVE_POPPLER', '1')
 endif
 
-libnifti_dep = cc.find_library('niftiio', has_headers: 'nifti1_io.h', required: false)
+# niftiio.pc is not always present, so fallback to CMake's find_package() functionally
+libnifti_dep = dependency('niftiio', required: false)
 if not libnifti_dep.found()
-    libnifti_dep = dependency('NIFTI', method: 'cmake', modules: ['NIFTI'], required: get_option('nifti'))
+    libnifti_dep = dependency('NIFTI', method: 'cmake', modules: ['NIFTI::niftiio'], required: get_option('nifti'))
 endif
-if libnifti_dep.found()
+if not get_option('nifti').disabled() and libnifti_dep.found()
     libvips_deps += libnifti_dep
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -465,11 +465,12 @@ elif libpoppler_dep.found() and cairo_dep.found()
     cfg_var.set('HAVE_POPPLER', '1')
 endif
 
-libnifti_dep = cc.find_library('niftiio', has_headers: 'nifti1_io.h', required: false)
+# niftiio.pc is not always present, so fallback to CMake's find_package() functionally
+libnifti_dep = dependency('niftiio', required: false)
```
eli-schwartz, line 469: The `required: false` and later explicit check for whether the option is .disabled() constitutes an annoying hack because of the need to manually, fragilely, account for different lookup methods.

Fortunately, Meson 0.60 can help out here! Unfortunately, you cannot assume the availability of Meson 0.60. So what you *can* do, is leave a TODO / FIXME comment here. Once you bump your minimum version of Meson, you can use:

```meson
dependency('niftiio', 'NIFTI', modules: ['NIFTI::niftiio'], required: get_option('nifti'))
```

It will check 4 things:
- niftiio as a pkg-config dependency (ignoring modules kwarg that pkg-config doesn't use)
- niftiio as a cmake dependency (trying to find a NIFTI::niftiio module)
- NIFTI as a pkg-config dependency (ignoring modules kwarg that pkg-config doesn't use)
- NIFTI as a cmake dependency (trying to find a NIFTI::niftiio module)

You can then rely on libnifti_dep to always have what you want -- it will never find the dependency when it is supposed to be disabled.

kleisauke: Indeed, Meson 0.60 handles dependencies with multiple names much neater. I've added a TODO comment for this with commit 9f8fab9a104722476904e6542b6b1d6febcb5d8d.

FWIW, I've made all these Meson >= 0.60 changes in a seperate branch, see:
https://github.com/libvips/libvips/compare/master...kleisauke:meson-0.60

PR #2695, opened 2022-02-27
Cleanup gettext handling

kleisauke:
Use GLib's i18n support instead of copying and pasting that
logic into its own header. This deprecates the `vips/intl.h`
header in favour of `glib/gi18n.h`.

_Split out from #2688._


jcupitt:
I'm seeing a crash with this commit during introspection:

```
$ meson setup build --prefix=~/vips --libdir=lib
$ cd build
$ ninja -v
...
Command '['/home/john/GIT/libvips/build/tmp-introspectn561ddf0/Vips-8.0', '--introspect-dump=/home/john/GIT/libvips/build/tmp-introspectn561ddf0/functions.txt,/home/john/GIT/libvips/build/tmp-introspectn561ddf0/dump.xml']' died with <Signals.SIGSEGV: 11>.
```

Ubuntu 21.10, if that makes a difference.

jcupitt:
I got a stacktrace for the crash:

```
#0  0x00007ffff7996d59 in g_mutex_lock ()
    at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#1  0x00007ffff7c96b4c in vips_object_class_install_argument
    (object_class=0x5555555c8a70, pspec=0x5555555abce0, flags=VIPS_ARGUMENT_SET_ONCE, priority=1, offset=40) at ../libvips/iofuncs/object.c:1732
#2  0x00007ffff7c9682a in vips_object_class_init (class=0x5555555c8a70)
    at ../libvips/iofuncs/object.c:1599
#3  0x00007ffff7c93078 in vips_object_class_intern_init (klass=0x5555555c8a70)
    at ../libvips/iofuncs/object.c:223
#4  0x00007ffff7704e78 in g_type_class_ref ()
    at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#5  0x0000555555557bd6 in dump_properties ()
#6  0x000055555555811c in dump_object_type ()
#7  0x0000555555558767 in dump_type ()
#8  0x0000555555558ba0 in dump_irepository ()
#9  0x0000555555558e10 in main ()
```

So it's the global lock (used to serialize arg addition) not being created. I put a breakpoint on `vips_init()` and it's never called. It looks like we're introspecting without starting up libvips.

autotools did this:

```
Vips_8_0_gir_SCANNERFLAGS = \
        --program=./introspect$(EXEEXT) \
        --identifier-prefix=Vips \
        --identifier-prefix=vips \
        --symbol-prefix=vips \
        --c-include=vips/vips.h
```

ie. used the `--program` flag to set a custom `main()` for introspection, and that custom main (`introspect.c`) did the init.

Looking at the meson wrapper for the scanner:

https://mesonbuild.com/Gnome-module.html#gnomegenerate_gir

I'm not sure it supports the `program` flag. Maybe it should?

I've no idea why this worked with the previous commit, it seems unrelated to the changes here. Perhaps the changing symbol table causes init to happen in a different order?

kleisauke:
Ouch, I'm seeing the same segv on my Fedora PC with introspection enabled. Here's the GDB backtrace (made possible by using `GI_SCANNER_DEBUG=save-temps`):
```c
Thread 1 "Vips-8.0" received signal SIGSEGV, Segmentation fault.
0x00007ffff7a832a9 in g_mutex_lock (mutex=0x0) at ../glib/gthread-posix.c:1516
1516	  if G_UNLIKELY (g_atomic_int_add (&mutex->i[0], 1) != 0)
Missing separate debuginfos, use: dnf debuginfo-install libcgif-0.2.0-1.fc35.remi.x86_64
(gdb) bt
#0  0x00007ffff7a832a9 in g_mutex_lock (mutex=0x0) at ../glib/gthread-posix.c:1516
#1  0x00007ffff7cda5cb in vips_object_class_install_argument (object_class=0x458390, pspec=0x43dce0, flags=VIPS_ARGUMENT_SET_ONCE, priority=1, offset=40) at ../libvips/iofuncs/object.c:1734
#2  0x00007ffff7cda973 in vips_object_class_init (class=0x458390) at ../libvips/iofuncs/object.c:1599
#3  vips_object_class_intern_init (klass=0x458390) at ../libvips/iofuncs/object.c:223
#4  0x00007ffff77e2fb8 in type_class_init_Wm (pclass=<optimized out>, node=<optimized out>) at ../gobject/gtype.c:2297
#5  g_type_class_ref (type=<optimized out>) at ../gobject/gtype.c:3012
#6  0x0000000000403785 in dump_properties ()
#7  0x0000000000403c52 in dump_object_type ()
#8  0x00000000004041f5 in dump_type ()
#9  0x000000000040458c in dump_irepository ()
#10 0x000000000040474a in main ()
```

It looks like it's attempting to call `g_mutex_lock` on a `NULL` pointer.
https://github.com/libvips/libvips/blob/f11b58ff6c602549d553c57e506142de5b6d28e9/libvips/iofuncs/object.c#L1732

Adding `vips_check_init();` above that call seems to fix it, which might suggest that this previous call in `vips__gettext`:
https://github.com/libvips/libvips/blob/517089fb4b159e2e1f79cd71101743ff71d21f33/libvips/iofuncs/init.c#L782

Ensured that vips was always initialized before it initialized these arguments:
https://github.com/libvips/libvips/blob/f11b58ff6c602549d553c57e506142de5b6d28e9/libvips/iofuncs/object.c#L1599-L1611

jcupitt:
Another possible fix might be this change to `vips_object_class_install_argument`:

```C
        if( vips__global_lock )
                g_mutex_lock( vips__global_lock );
```

and the same for unlock. The scanner isn't threaded, so it won't be doing class init for several gobjects at the same time, so this is safe.

It does look rather ugly and hard to explain though !

jcupitt:
On the plus side, we could get rid of `introspect.c` once autotools goes.

jcupitt:
Ah you're right, it must have been the init in `vips__gettext()`. Yuk!

kleisauke:
Another idea is to call `vips_check_init()` during [`vips_object_class_init`](https://github.com/libvips/libvips/blob/f11b58ff6c602549d553c57e506142de5b6d28e9/libvips/iofuncs/object.c#L1561). I noticed that `vips_image_class_init` has a similar approach.
https://github.com/libvips/libvips/blob/f11b58ff6c602549d553c57e506142de5b6d28e9/libvips/iofuncs/image.c#L1069-L1071

jcupitt:
Oh that's nice. We could just move the check from `image_class_init` to `object_class_init` (the base class).  No need to have both.

jcupitt:
.... I'll push that change, it's clearly the best fix.

PR #2699, opened 2022-03-01
WIP: Update libnsgif

tlsa:
There are some commented out sections where I don't really understand the VIPS code:

* Background color handling in `vips_foreign_load_nsgif_set_header`
* Handling the frame transparency flag in `vips_foreign_load_nsgif_header`
* Pages handling in: `vips_foreign_load_nsgif_generate`

And the `VERBOSE` print_frame / print_animation has a few lines commented out where the information isn't currently exposed.


libvips/foreign/nsgifload.c
```diff
@@ -457,30 +421,34 @@ vips_foreign_load_nsgif_generate( VipsRegion *or,
...
 
-		g_assert( line >= 0 && line < gif->anim->height );
+		g_assert( line >= 0 && line < info->height );
 		g_assert( page >= 0 && page < gif->frame_count_displayable );
 
-		if( gif->anim->decoded_frame != page ) {
-			result = gif_decode_frame( gif->anim, page ); 
-			VIPS_DEBUG_MSG( "  gif_decode_frame(%d) = %d\n", 
+//
+// I don't really understand the concept of pages
```
jcupitt, line 436: GIFs are represented as very tall, thin images, with the frames laid out vertically like a strip of film.

The `_generate()` function has to fill the pixels in `or->valid`. The coordinates are in the output image space, so `or->valid.top / page_height` is the frame number, and `% page_height` is the y in this frame.

Calls to `_generate()` are guaranteed to be non-overlapping and to be in strictly sequential order (the `sequential` that `_nsgif_load()` runs upstream of generate makes this guarantee). The height of each request is unknown, so requests can potentially need several frames to fill.

The algorithm in generate is therefore:

1. For each line in the output:
2. Find the frame number and line number within that frame.
3. Is the currently loaded GIF frame the right one? If not, load the next frame.
4. Copy one line from the decoded frame into the output region.

You could do the loop in rects rather than a line at a time, but line at a time is very simple and I doubt if the speed difference would be measureable.

Looking at `nsgif_frame_decode()`, you have step 3. above in there already, so I agree, this `if()` could go, but it's handy for debugging (eg. the `print_animation()` call below) to be able to see when new frames are decoded, so I'd be inclined to keep it.

This means the `bitmap` return must be kept between `_generate()` calls, so it'll need to be a member of `VipsForeignLoadNsgif`.

Since `anim` is now opaque, `VipsForeignLoadNsgif` will need `int decoded_frame;`, and it'll need updating after the call to `nsgif_frame_decode`.

Alternatively, just call `nsgif_frame_decode()` for every scanline and get rid of `print_animation`. It's no longer very useful.

Perhaps libnsgif could have a debug method that pretty-printed the decoder state? I'd find it useful for troubleshooting tricky GIFs.

jcupitt:
Ah I think the test failures are because the saver tries to distinguish between RGB and RGBA GIFs. At the moment I see:

```
$ vipsheader silver-gear-cogs-animation-5.gif 
silver-gear-cogs-animation-5.gif: 281x320 uchar, 3 bands, srgb, gifload
```

So it's dropping the alpha.

The fix would be to give up trying to detect alpha and just save everything as RGBA, or alternatively to add some API to libnsgif to allow alpha detection. We could possibly add monochrome detection too -- just check the colortable for each frame for R = G = B.

tlsa:
I've added the transparency flag to the exposed frame info. Note that this is not saying that the frame definitely has transparency, only that it might. (The LZW data might not specify any pixels with the transparency index.) That's always been the case. That's why I was initially hesitant to expose it; it's really an internal consideration for what should happen when we are decoding the data and we hit a pixel with the transparency index. I'm in two minds about reverting that.

The correct way would be to implement the bitmap table callbacks for get and set opacity, however there is more computational overhead to do this. It involves scanning every pixel in the bitmap for a transparent pixel. The result is cached per frame. This is useful for a client app that runs looping animations, but it might be too much of an overhead for a use case like VIPS?

I've also exposed the background colour in the gif info structure.

jcupitt:
That's great, it seems to spot transparency and no transparency sensibly for me now.

One test is still failing because of a change in how delay is being reported.  For this test image:

cogs

I now see:

```
$ vipsheader -a cogs.gif 
cogs.gif: 85x77 uchar, 4 bands, srgb, gifload
...
loop: 0
delay: 100 50 50 50 50 
background: 255 255 255 
```

Previously it would report `delay: 0 50 50 50 50` and `loop: 32760`. I think if you change these line in the test it should fix it:

https://github.com/libvips/libvips/blob/master/test/test-suite/test_foreign.py#L911

tlsa:
Thanks, I've changed that in [upstream](http://git.netsurf-browser.org/libnsgif.git/commit/?id=030c31ccd5287813cdd60ff299a577a36900fdda), so that now the actual delay in the GIF file is available in the frame info. The minimum delay is now only applied in `nsgif_frame_prepare()` which is a sort of helper to manage animations, which libvips doesn't need to do.

tlsa:
Inside `_generate` the way I did it will call `nsgif_frame_decode` at least once per call to `_generate`. So if every line is a new call to `_generate`, there is a function all overhead per line, but if the page hasn't change since the last `_generate` call, then `nsgif_frame_decode` won't have to do anything.

libvips/foreign/nsgifload.c
```diff
@@ -116,6 +116,10 @@ typedef struct _VipsForeignLoadNsgif {
 	 */
 	int frame_count_displayable;
 
+	/* The currently decoded frame.
+	 */
+	unsigned int decoded_frame;
```
jcupitt, line 121: This can go, I think.

libvips/foreign/nsgifload.c
```diff
@@ -116,6 +116,10 @@ typedef struct _VipsForeignLoadNsgif {
 	 */
 	int frame_count_displayable;
```
jcupitt, line 117: This is just equal to `frame_count` now, let's remove this.

How about putting `nsgif_info_t *info;` here instead and setting it just after `nsgif_data_scan()`? You could use it in a lot of places.

libvips/foreign/nsgifload.c
```diff
@@ -444,7 +399,9 @@ static int
 vips_foreign_load_nsgif_generate( VipsRegion *or,
 	void *seq, void *a, void *b, gboolean *stop )
 {
-        VipsRect *r = &or->valid;
+	int prev_page = 0;
```
jcupitt, line 402: If you get prev_page = -1 here you could remove the `bitmap == NULL` from the if() around frame decode below.

libvips/foreign/nsgifload.c
```diff
@@ -496,6 +454,8 @@ vips_foreign_load_nsgif_generate( VipsRegion *or,
 				p += 4;
 			}
 		}
+
+		prev_page = page;
```
jcupitt, line 458: It'd be clearer if this was inside the `if(prev_page != page)` block above.

jcupitt (commented):
I made a few small comments. I can do it if you like.

Thanks for doing this work Michael!

libvips/foreign/nsgifload.c
tlsa: Good spot, done!

libvips/foreign/nsgifload.c
tlsa: Done!

libvips/foreign/nsgifload.c
tlsa: Done!

libvips/foreign/nsgifload.c
tlsa: Done!

jcupitt:
... I made some very minor revisions and added a note to the main changelog with a credit to you (hope that's OK).

PR #2702, opened 2022-03-03
nsgifload: Fix gif decoding issue.

tlsa:
The nsgif_data_scan() call returns an error code that says
if anything was wrong with the source data.

It is not unusual for there to be bad stuff in a GIF, so we
only need to worry if we failed to find any frames.

Fixes #2698.

### DIFF
--- a/libvips/foreign/nsgifload.c
+++ b/libvips/foreign/nsgifload.c
@@ -324,7 +324,8 @@ vips_foreign_load_nsgif_header( VipsForeignLoad *load )
 #ifdef VERBOSE
 	print_animation( gif->anim, gif->info );
 #endif /*VERBOSE*/
-	if( result != NSGIF_OK ) {
+	if( result != NSGIF_OK &&
+		load->fail_on >= VIPS_FAIL_ON_WARNING ) {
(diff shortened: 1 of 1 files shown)


jcupitt:
Huh that's nice and simple.

libvips has a `fail_on` setting, so users can ask loaders to be strict. We could change it to be:

```C
        result = nsgif_data_scan( gif->anim, size, (void *) data );
        gif->info = nsgif_get_info(gif->anim);
#ifdef VERBOSE
        print_animation( gif->anim, gif->info );
#endif /*VERBOSE*/
        if( result != NSGIF_OK &&
                load->fail_on >= VIPS_FAIL_ON_WARNING ) {
                vips_foreign_load_nsgif_error( gif, result );
                return( -1 );
        }         
```

Now this will work:

```
$ vips copy garden.gif[n=-1] x.gif
```

And this will fail:

```
$ vips copy garden.gif[n=-1,fail-on=warning] x.gif
gifload: Invalid frame data
$
```

tlsa:
Oh, that's cool. Yes, I'll do it like that, thanks!

tlsa:
Done!

PR #2709, opened 2022-03-03
cgifsave: Make loop counts more accurate

TheEssem:
See dloebl/cgif#46 and dloebl/cgif#48.

This fixes cgifsave to make the libvips `loop` property specify the number of iterations instead of the number of times an animation loops. Depends on (the currently unreleased) CGIF version 0.3.0.

### DIFF
--- a/libvips/foreign/cgifsave.c
+++ b/libvips/foreign/cgifsave.c
@@ -63,6 +63,7 @@ typedef struct _VipsForeignSaveCgif {
 	int effort;
 	int bitdepth;
 	double maxerror;
+	gboolean noloop;
 	VipsTarget *target;
 
 	/* Derived write params.
@@ -309,7 +310,7 @@ vips_foreign_save_cgif_write_frame( VipsForeignSaveCgif *cgif )
 	 */
 	if( !cgif->cgif_context ) {
 		cgif->cgif_config.pGlobalPalette = cgif->palette_rgb;
-		cgif->cgif_config.attrFlags = CGIF_ATTR_IS_ANIMATED;
+		cgif->cgif_config.attrFlags = CGIF_ATTR_IS_ANIMATED | ( cgif->noloop ? CGIF_ATTR_NO_LOOP : 0 );
 		cgif->cgif_config.width = frame_rect->width;
 		cgif->cgif_config.height = frame_rect->height;
 		cgif->cgif_config.numGlobalPaletteEntries = cgif->lp->count;
@@ -612,6 +613,13 @@ vips_foreign_save_cgif_class_init( VipsForeignSaveCgifClass *class )
 		VIPS_ARGUMENT_OPTIONAL_INPUT,
 		G_STRUCT_OFFSET( VipsForeignSaveCgif, maxerror ),
 		0, 32, 0.0 );
+
+	VIPS_ARG_BOOL( class, "noloop", 14,
+		_( "No loop" ),
+		_( "Disables looping" ),
+		VIPS_ARGUMENT_OPTIONAL_INPUT,
+		G_STRUCT_OFFSET( VipsForeignSaveCgif, noloop ),
+		FALSE );
 }
 
 static void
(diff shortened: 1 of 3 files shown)


jcupitt:
Hi @TheEssem,

The libvips `loop` property is used by eg. webp and others as well, so it doesn't have the usual GIF meaning (there's a hidden field called `gif-loop` with the bizarre GIF interpretation). It really does mean the total number of iterations.

Instead of a flag, how about automatically setting `CGIF_ATTR_NO_LOOP` if loop == 1? I suppose we might need to subtract 1 from loop for values > 1.

TheEssem:
Ah, I see. Fixed.

jcupitt:
I tested it a bit. It looks like the GIF loader needs a matching change :/ I thought it had this logic in already.

TheEssem:
Patched nsgif to (hopefully) fix it, let me know if there's a more elegant solution.

jcupitt:
I don't think we should patch libnsgif unless there's really no choice. I'll see if there's some way to do it within the libnsgif API.

tlsa:
Good spot, thanks @TheEssem, @jcupitt. I'll fix LibNSGIF.

tlsa:
Done upstream:

* [Fix handling of loop count value in application extension](http://git.netsurf-browser.org/libnsgif.git/commit/?id=c77c518aa14ed14ca1424cfa635ca8bf95cb0814)
* [Support AnimExts Looping Application Extension](http://git.netsurf-browser.org/libnsgif.git/commit/?id=61934ff70402ffff91fbea9238cb1833ffa6cec6)

jcupitt:
The libnsgif change has helped a lot, now we need to make it work with cgif 0.2 (the version that's used for testing, since 0.3 isn't out yet).

We'll need a test in `libvips/meson.build`, if you're able, or I could have a go if you like.

jcupitt:
Eg. something like:

```meson
cgif_dep = disabler()
if quantisation_package.found()
    cgif_dep = dependency('cgif', version: '>=0.2.0', required: get_option('cgif'))
    if cgif_dep.found()
        libvips_deps += cgif_dep
        cfg_var.set('HAVE_CGIF', '1')
        if cc.compiles('#include <cgif.h>\nint i = CGIF_ATTR_NO_LOOP;', name: 'Has CGIF_ATTR_NO_LOOP', dependencies: cgif_dep)
            cfg_var.set('HAVE_CGIF_ATTR_NO_LOOP', '1')
        endif
    endif
endif
```

Then you can add an `#ifdef HAVE_CGIF_ATTR_NO_LOOP` to your C.

TheEssem:
How exactly would the `HAVE_CGIF_ATTR_NO_LOOP` fallback work? Would it just revert to the old behavior?

jcupitt:
Yes, I suppose so, unless you can think of something better.

eli-schwartz:
I think you don't need a `cc.compiles()` here, because it is gated on whether you have 0.3, right?

So you can just check `cgif_dep.version().version_compare('>=0.3')` to see if it is new enough -- this runs entirely as an in process string version comparison, rather than forking to the compiler and running a compile check.

jcupitt:
My code scrap was gated on 0.2+ I think, so you would need to test for `CGIF_ATTR_NO_LOOP` as well.

The usual principle is to test for features rather than version numbers where possible. Version number tests can fail if features get backported, for example.

jcupitt:
Oh, I misread your comment, sorry. Sure, we could use `version_compare` too, I don't really mind.

jcupitt:
OK, tests are passing, let's merge! Thank you for doing this work, @TheEssem.

jcupitt:
I think the example you see most often is ImageMagick :( They have a sad history of adding and removing API on minor version number changes, and also backporting. A single scalar isn't enough to represent an IM version number, you need at least (x + iy) hehe. I think they are trying harder with im7 to provide a supported API, so perhaps it's not the issue it was.

You know a lot more than me about this of course, perhaps I'm being overly cautious.

PR #2718, opened 2022-03-11
add dzsave to a target

jcupitt:
made a thing for gsf to write to a target


jcupitt:
next: sort out the temp file mess in dzsave

PR #2742, opened 2022-04-02
vipsload: sanity check resolution

lovell:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=46009 (when compiled with debug assertions).

Before:
```
$ vipsheader -a fail.v

(vipsheader:15655): VIPS-WARNING **: 10:35:38.096: error reading vips image metadata: VipsImage: XML parse error

sanity failure: VipsImage (0x558a4f1da330) bad resolution

**
VIPS:ERROR:../libvips/iofuncs/image.c:3619:vips_image_pio_input: assertion failed: (vips_object_sanity( VIPS_OBJECT( image ) ))
Bail out! VIPS:ERROR:../libvips/iofuncs/image.c:3619:vips_image_pio_input: assertion failed: (vips_object_sanity( VIPS_OBJECT( image ) ))
Aborted (core dumped)
```

After:
```
$ vipsheader -a fail.v
vipsheader: VipsImage: invalid resolution
VipsImage: unable to read header for "fail.v"
```

### DIFF
--- a/libvips/iofuncs/vips.c
+++ b/libvips/iofuncs/vips.c
@@ -358,8 +358,8 @@ vips__read_header_bytes( VipsImage *im, unsigned char *from )
 	/* We read xres/yres as floats to a staging area, then copy to double
 	 * in the main fields.
 	 */
-	im->Xres = im->Xres_float;
-	im->Yres = im->Yres_float;
+	im->Xres = VIPS_MAX( 0, im->Xres_float );
+	im->Yres = VIPS_MAX( 0, im->Yres_float );
 
 	/* Some protection against malicious files. We also check predicted
 	 * (based on these values) against real file length, see below. 


lovell:
Looks like the CIFuzz job failed for an unrelated reason (perhaps a build failure that needs fixing downstream in OSS Fuzz?).

kleisauke:
>  [100%] Linking CXX shared library libjxl_extras_dec.so

It seems that libjxl builds shared libraries despite passing the `-DBUILD_SHARED_LIBS=0` CMake option (since commit https://github.com/libjxl/libjxl/commit/775d010700bc8bb752f8db3753e9e7246093758f).
https://github.com/libjxl/libjxl/blob/79d85890e7b409b4f76db019e2d0f8d21a354cb5/lib/jxl_extras.cmake#L176-L178

I'm preparing a PR to resolve this in libjxl. The latest libvips' OSS-Fuzz build also seems to fail due to this.
https://oss-fuzz-build-logs.storage.googleapis.com/index.html#libvips

kleisauke:
PR https://github.com/libjxl/libjxl/pull/1315 should fix the failing CIFuzz and OSS-Fuzz builds.

jcupitt:
In exif load we do:

```C
        image->Xres = VIPS_MAX( 0, xres );
        image->Yres = VIPS_MAX( 0, yres );
```

Perhaps the behaviour for vips load should be the same?

lovell:
Thanks John, makes sense, I've updated to do so. This approach keeps things consistent with `vips_image_init_fields` too.

jcupitt:
Great!

PR #2747, opened 2022-04-04
Update to latest libnsgif

tlsa:
The main changes are that now clients can control the pixel format that LibNSGIF decodes to, and some internal state has been removed from the GIF info structure.

### DIFF
--- a/libvips/foreign/libnsgif/README-ns.md
+++ b/libvips/foreign/libnsgif/README-ns.md
@@ -22,14 +22,17 @@ Using
 LibNSGIF allows the client to allocate the bitmap into which the GIF is
 decoded. The client can have an arbitrary bitmap structure, that is simply
 a void pointer to LibNSGIF. The client must provide a callback table for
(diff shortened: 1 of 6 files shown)


tlsa:
Note: A list of available pixel formats is [in the header](http://source.netsurf-browser.org/libnsgif.git/tree/include/nsgif.h#n101).

I just used the one that gave the same behavior as LibNSGIF has had in the past. However, it would now be possible to change the ordering, if that's something VIPS would want.

jcupitt:
Nice! Thanks Michael.

Re. formats, libvips supports reading packed RGBRGBRGB from GIF if the GIF has no transparency (a very common case for eg. short video clips). I imagine this would be awkward to add to libnsgif though.

https://github.com/libvips/libvips/blob/master/libvips/foreign/nsgifload.c#L429-L442

tlsa:
Yeah, internally, libnsgif is always 32 bits per pixel.

Issue #2749, opened 2022-04-05
Stop libjpeg load after too many warnings


jcupitt:
libvips can be slow if a JPEG file contains many warnings, example here:

https://storage.googleapis.com/clusterfuzz-external-blobs/6550c970-300a-491b-8ea1-9f6a56cd765b?GoogleAccessId=clusterfuzz-external%40appspot.gserviceaccount.com&Expires=1649089629&Signature=Cu%2BhPCEPy0hrpmpdpgY3vjbgeYBrtLgBx%2BYwMTqt3SX5awSi2e5yCGTh45wQGTJxOQKYuxr8u8ygVrUbZvHo%2BBSlxF3xD%2F5lwN0ZSn0ASFS57NV6jBFSB4rnqjO2R5L6P3HBJflfsjEb8kR2aWidE8QWXvBM88ewAvk41tQmlu6rxapNs%2B3ckGAXbjRupYOSMSAWrje7RS4KY7dTxIc34g41XT2yCOalyanvNGz%2Bgu%2FGx4zwBho7N9mWhYCsexTUWvOZMloD6iiwxy6s4USWaKMlLZ8qglcMcJg6VjxdVnDj2CcUdm%2FuZk27AhTPyZscl39pDVhQrwxLq8uYQcHx9Q%3D%3D&response-content-disposition=attachment%3B+filename%3Dclusterfuzz-testcase-minimized-pngsave_buffer_fuzzer-4635059014991872

That 1mb file has 80,000 malformed APP sections and takes 5s to run `vipsheader`.

We need some way to stop jpeg_read_header() half-way.

See https://github.com/libvips/libvips/commit/89bd46d1c4e0465974b886ef5b5021e988f111a2 for an earlier, failed attempt.

kleisauke:
Commit https://github.com/kleisauke/libvips/commit/df26bd1e46c67ed7745996627ca0d83e5a6d678d seems to resolve this for me locally, but I have not yet tested the regressions caused by the previous commits.

kleisauke:
Note that this implementation was based on:
https://github.com/libjpeg-turbo/libjpeg-turbo/blob/df3c3dcb9b0b24d78f9e58e28c8b509a49967ba8/jerror.c#L115-L145
https://github.com/libjpeg-turbo/libjpeg-turbo/blob/df3c3dcb9b0b24d78f9e58e28c8b509a49967ba8/cjpeg.c#L538-L548

jcupitt:
[...]

PR #2751, opened 2022-04-06
Add comments to `setjmp` invocations

kleisauke:
Also, pre-increment `num_warnings` to ensure that we really stop
at 100 warnings.

Context: #2749.

### DIFF
--- a/libvips/foreign/jpeg2vips.c
+++ b/libvips/foreign/jpeg2vips.c
@@ -322,12 +322,12 @@ readjpeg_emit_message( j_common_ptr cinfo, int msg_level )
 	if( msg_level < 0 ) {
 		/* Always count warnings in num_warnings.
 		 */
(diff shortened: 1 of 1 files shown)


kleisauke:
Oh, perhaps this is redundant, it seems that the above OSS-Fuzz issue was last tested on commit https://github.com/libvips/libvips/commit/89bd46d1c4e0465974b886ef5b5021e988f111a2. I'll mark this PR as draft.

kleisauke:
I've dropped commit https://github.com/libvips/libvips/commit/50aa554144cdf5f3baedaa9fc1ad50a55581a0ca for now, this should already be handled by the `setjmp` in `read_jpeg_image` or `vips__jpeg_read_source`. I've added a couple of comments to clarify this.

jcupitt:
The libjpeg API has aged very well, mostly, except for setjmp() / longjmp() :-( argh.

PR #2753, opened 2022-04-11
add "configure" operation

jcupitt:
This operation can be used to set various library config options, such
as vips_block_untrusted_set().

Having these options available as an operation makes like simpler for
bindings: they can expose them all via the usual introspection system.

Example:

```python
pyvips.Image.configure(untrusted_block=True)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 compile
+build
 .pytest_cache
 .dirstamp
 a.out
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,8 +1,6 @@
 21/11/21 started 8.13
 - configure fails for requested but unmet dependencies [remicollet]
 - add support for another quantiser [DarthSim]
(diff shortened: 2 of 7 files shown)


jcupitt:
@kleisauke I thought this might be handy for bindings like NetVips. Rather than having to add API for every libvips configure option, we can just have a single vips operation which lets you set the config options in one place.

With this PR (for example), pyvips will instantly let you do:

```python
pyvips.Image.configure(untrusted_block=True)
```

Same for lua-vips, ruby-vips, NetVips etc. And we can use the libvips arg system to safely deprecate, rename, add, extend, etc. options in future.

What do you think?

We should all the other config options too, eg concurrency, leak, etc. etc.

jcupitt:
- Maybe this should be called `options`? The name `configure` sounds too much like autotools and compile time configuration.
- We could allow read of settings as well as write.
- We could deprecate all the annoying `vips_concurrency_set()` etc. API.

PR #2754, opened 2022-04-11
nsgif: try to recover after LZW_EOI_CODE

DarthSim:
Here's an issue with a GIF sample that can't be loaded because of the `NSGIF_ERR_DATA_FRAME` error: https://github.com/imgproxy/imgproxy/issues/842. I found that it is caused by `LZW_EOI_CODE`, and it seems like we can try to recover after it as we do with `LZW_OK_EOD`. I've tried to process the sample from the issue with this patch applied and everything worked fine.

### DIFF
--- a/libvips/foreign/libnsgif/gif.c
+++ b/libvips/foreign/libnsgif/gif.c
@@ -471,7 +471,7 @@ static nsgif_error nsgif__decode_complex(
 			while (available == 0) {
 				if (res != LZW_OK) {
 					/* Unexpected end of frame, try to recover */
-					if (res == LZW_OK_EOD) {
+					if (res == LZW_OK_EOD || res == LZW_EOI_CODE) {
 						ret = NSGIF_OK;
 					} else {
 						ret = nsgif__error_from_lzw(res);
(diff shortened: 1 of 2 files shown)


jcupitt:
Oh, nice idea. 

@tlsa this PR from DarthSim looks interesting, what do you think?

tlsa:
At first glance it looks reasonable. I'll experiment properly on Friday.

tlsa:
@DarthSim Thanks for this, it has been [applied upstream](http://git.netsurf-browser.org/libnsgif.git/commit/?id=aa6e2af43ebb898167f6dc0bb8215eacf0a17389).

@jcupitt I opend https://github.com/libvips/libvips/pull/2766 which includes this and also [this fix to the uncommon case of frame clipping](http://git.netsurf-browser.org/libnsgif.git/commit/?id=75ed38539447571ec961c985c8a1e9cbd76573e2).

jcupitt:
Ok, this PR has gone into libnsgif, and from there into libvips. Thank you for working on this, @DarthSim!

PR #2778, opened 2022-04-26
cgifsave: allow to apply maxerror optimization to frames with alpha channel

dloebl:
As of now, the `maxerror` parameter cannot be applied to animated GIFs with an active alpha channel.
Make it possible to detect cases where the alpha channel component of the current frame matches with the one from the frame before.
In these cases it is possible to use the transparency index for the `maxerror` optimization instead of for the alpha channel.
This pull request helps to further improve the file size of animated GIFs (for reference: #2576).

**Example:**
`$ vips copy test.gif[n=-1] old.gif[maxerror=16]`
`$ vips copy test.gif[n=-1] new.gif[maxerror=16]`
`$ gifsicle -O3 test.gif -o gifsicle.gif`
**Results:**
`du -h *.gif`
` 1.0M	new.gif`
` 5.8M	gifsicle.gif`
` 6.1M	test.gif`
` 9.8M	old.gif`
150592639-fcf2489c-c43e-478e-a66f-83fb3a9a98aa

### DIFF
--- a/libvips/foreign/cgifsave.c
+++ b/libvips/foreign/cgifsave.c
@@ -166,10 +166,14 @@ static gboolean
 vips_foreign_save_cgif_pixels_are_equal( const VipsPel *cur, const VipsPel *bef,
 	double maxerror )
 {
-	if( bef[3] != 0xFF )
-		/* Solid pixels only.
+	if( bef[3] != cur[3] )
+		/* Alpha channel must be identical.
 		 */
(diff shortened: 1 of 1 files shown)


jcupitt:
This looks great @dloebl, let's merge. Thank you for doing this, and sorry it took me a while :(

I think the asan failure is a temporary clang issue.

PR #2791, opened 2022-05-08
Update LibNSGIF to latest upstream

tlsa:
This adds access to palette info, as required by #2576.

The upstream test decoder also has a new `-p` option to save images of the palettes.


jcupitt:
Wow thanks for doing this so quickly Michael. I'll try prototyping a GIF -> GIF resizer that uses this feature.

PR #2794, opened 2022-05-09
heifsave: ensure added EXIF metadata is persisted

lovell:
Currently an output HEIF image can only have EXIF metadata if the input image also has EXIF metadata. The change in this PR ensures newly-set EXIF is also persisted, bringing the behaviour in line with jpegsave et al.


jcupitt:
Oh, good spot, I hadn't thought of that.

PR #2795, opened 2022-05-09
Allow to modify/create EXIF 2.3 ASCII tags

kleisauke:
Make `tag_is_ascii` aware of the EXIF 2.3 ASCII tags that are
available since libexif 0.6.22.

See: https://github.com/kleisauke/net-vips/issues/167


jcupitt:
👍

PR #2796, opened 2022-05-10
Allow to modify/create EXIF OffsetTime* tags

kleisauke:
Available since libexif 0.6.23.

Noticed this while working on PR https://github.com/libexif/libexif/pull/85.

Context: https://github.com/libvips/libvips/pull/2795.


jcupitt:
Heh libexif is a masterclass in how to not create an extensible API.

PR #2797, opened 2022-05-10
cgifsave: reuse global palette, if possible

dloebl:
For reference: [2576#comment](https://github.com/libvips/libvips/issues/2576#issuecomment-1121206055)
Reuse the global palette in `gifsave` if it has been attached during `gifload`.
This should both improve the file size and processing time for many GIF -> GIF operations.

@jcupitt **Notes / Questions:**
1. [Quantizing the attached global palette](https://github.com/dloebl/libvips/blob/75b08f3bce9f7a42a60e7859824cabb3c483d549/libvips/foreign/cgifsave.c#L256) to get a `VipsQuantiseResult` feels a bit hacky. Maybe there is a better way to do it?
2. As we probably don't want to enable this new behaviour for any kind of GIF -> GIF operation, should we add a parameter to `gifsave` to enable/disable global palette reuse?
3. Dithering seems to cause colour artifacts with certain GIFs:
$ vipsthumbnail --size 200x [phone.gif](https://user-images.githubusercontent.com/34104349/165298655-f26d5afe-7179-42a5-8029-56490bf6fecf.gif)[n=-1] -o out.gif

**Results:**
GIF used: [earth.gif](https://upload.wikimedia.org/wikipedia/commons/2/2c/Rotating_earth_%28large%29.gif)

**Before:**
`$ time vipsthumbnail --size 200x earth.gif[n=-1] -o old.gif`
`518ms total`

**After:**
`$ time vipsthumbnail --size 200x earth.gif[n=-1] -o new.gif`
`219ms total`

`$ du -h old.gif new.gif` 
`580K	old.gif`
`456K	new.gif`

### DIFF
--- a/libvips/foreign/cgifsave.c
+++ b/libvips/foreign/cgifsave.c
@@ -615,8 +615,12 @@ vips_foreign_save_cgif_build( VipsObject *object )
 		tmp_image = vips__quantise_image_create_rgba( cgif->attr,
 			tmp_gct, gct_length + 1, 1, 0 );
 
-		vips__quantise_image_quantize( tmp_image,
-			cgif->attr, &cgif->quantisation_result );
+		if( vips__quantise_image_quantize( tmp_image,
+		       cgif->attr, &cgif->quantisation_result ) ) {
+		       vips_error( class->nickname,
+		       	"%s", _( "quantisation failed" ) );
+		       return( -1 );
+		}
 		VIPS_FREE( tmp_gct );
 		VIPS_FREEF( vips__quantise_image_destroy, tmp_image );
 	}
--- a/libvips/foreign/cgifsave.c
+++ b/libvips/foreign/cgifsave.c
@@ -894,6 +894,7 @@ vips_foreign_save_cgif_buffer_init( VipsForeignSaveCgifBuffer *buffer )
  * * @effort: %gint, quantisation CPU effort
  * * @bitdepth: %gint, number of bits per pixel
  * * @maxerror: %gdouble, maximum inter-frame error for transparency
+ * * @reoptimise: %gboolean, reoptimise colour palettes
  *
  * Write to a file in GIF format.
  *
@@ -940,6 +941,7 @@ vips_gifsave( VipsImage *in, const char *filename, ... )
  * * @effort: %gint, quantisation CPU effort
  * * @bitdepth: %gint, number of bits per pixel
  * * @maxerror: %gdouble, maximum inter-frame error for transparency
+ * * @reoptimise: %gboolean, reoptimise colour palettes
  *
  * As vips_gifsave(), but save to a memory buffer.
  *
@@ -991,6 +993,7 @@ vips_gifsave_buffer( VipsImage *in, void **buf, size_t *len, ... )
  * * @effort: %gint, quantisation CPU effort
  * * @bitdepth: %gint, number of bits per pixel
  * * @maxerror: %gdouble, maximum inter-frame error for transparency
+ * * @reoptimise: %gboolean, reoptimise colour palettes
  *
  * As vips_gifsave(), but save to a target.
  *
--- a/libvips/foreign/cgifsave.c
+++ b/libvips/foreign/cgifsave.c
@@ -63,6 +63,7 @@ typedef struct _VipsForeignSaveCgif {
 	int effort;
 	int bitdepth;
 	double maxerror;
+	gboolean reuse_palette;
 	VipsTarget *target;
 
 	/* Derived write params.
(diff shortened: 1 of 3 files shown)


jcupitt:
Oh very cool, I hadn't thought of using the palette as an image source. @DarthSim, do you have an opinion on the best way to initialise the quantisation result from an existing palette?

On new switches to control behaviour:

1. We could add an argument to not use the global palette from the metadata (perhaps `reoptimise`?), or alternatively users could delete `gif-palette` before saving. I suppose having a switch would be more convenient.
2. We could add a switch to enable dynamic palette generation (perhaps `local_palette`?). Or I suppose we could just delete all that code and only support a single global palette.

DarthSim:
> I hadn't thought of using the palette as an image source. @DarthSim, do you have an opinion on the best way to initialise the quantisation result from an existing palette?

That's exactly want I'd do :)

jcupitt:
I think we should reuse by default. The most common case must be GIF -> GIF resize, and that should reuse with no extra parameters.

How about something like `reoptimise` (or maybe `optimise_palette`?) to generate a new global palette.

dloebl:
@jcupitt Good point. I'll introduce a `reoptimise` parameter.

> generate a new global palette

For now, I would prefer to keep the current method of determining palettes (in case no palette has been attached or `reoptimise` == `TRUE`), because that keeps the possibility to output chunks of GIF-data to the target (streaming) in `gifsave`. Maybe quantizing to a single global palette is something for a different PR?

jcupitt:
> Maybe quantizing to a single global palette is something for a different PR?

OK, sounds reasonable.

dloebl:
Ready for review from my side.
One additional improvement we could do, is to attach local palettes on `gifload` as well. This should both improve file sizes and processing time for the affected GIFs again. @jcupitt Do you think that would make sense for a follow-up PR?

**Additional notes:**
I noticed artifacts with certain GIFs depending on the `quantizr` / `libimagequant` version:
[original image](https://user-images.githubusercontent.com/34104349/165298655-f26d5afe-7179-42a5-8029-56490bf6fecf.gif)
Command: `$ vipsthumbnail --size 200x input.gif[n=-1] -o out.gif`

| quantizr v1.2.0 | imagequant v2.17 | imagequant v4.0 |
| - | - | - |
| quant |  |  |

The issue is gone with the latest `libimagequant` version (v4.0), but I just wanted to mention this.

jcupitt:
Hi again, could you explain the motivation behind the `use_lct` feature? Does this help with a significant percentage of images? It looks like a rather niche thing at first glance.

dloebl:
While testing this change with a dataset of GIFs, I noticed that about 25% of the animated GIFs we process have at least one local colour table (might be because some of the GIFs come from libvips originally).
Therefore I would like to extend this improvement to GIFs with local colour tables. But I guess that's something for a follow-up PR - I'll remove the `use_lct` feature for now :-)

libvips/foreign/cgifsave.c
```diff
@@ -72,6 +73,8 @@ typedef struct _VipsForeignSaveCgif {
 	int *delay;
 	int delay_length;
 	int loop;
+	int *gct;
+	int gct_length;
```
jcupitt, line 77: Could we call this `global_colour_table`? Abbreviations in local variables are OK, but I think members should be written out in full, since they start spreading way beyond a small local context.

libvips/foreign/cgifsave.c
```diff
@@ -585,6 +590,35 @@ vips_foreign_save_cgif_build( VipsObject *object )
 	vips__quantise_set_quality( cgif->attr, 0, 100 );
 	vips__quantise_set_speed( cgif->attr, 11 - cgif->effort );
 
+	/* Quantize attached global palette (if present)
```
jcupitt, line 593: How about `Initialise quantization result using global palette`?

jcupitt (commented):
Just two tiny comments.

jcupitt:
25% is higher than I expected! Sure, let's do that in another PR.

dloebl:
Sounds good 👍
All done. I also updated the documentation/changelog.

jcupitt:
Great! It seems to work well, this is a nice improvement. Good job!

jcupitt:
With `reoptimise` (more or less the old behaviour) I see:

```
$ time vips copy 3198.gif[n=-1] x.gif[reoptimise]
real	0m1.076s
user	0m1.386s
sys	0m0.013s
$ ls -l x.gif
-rw-r--r-- 1 john john 4153849 May 23 13:05 x.gif
```

And with this patch:

```
$ time vips copy 3198.gif[n=-1] x.gif
real	0m0.886s
user	0m1.206s
sys	0m0.039s
$ ls -l x.gif
-rw-r--r-- 1 john john 4041446 May 23 13:05 x.gif
```

So in many cases: smaller GIFs, lower CPU load, same memory use.

Issue #2801, opened 2022-05-13
vips_target_finish can fail silently resulting in corrupted or null output


flga:
## Bug report

**Describe the bug**
Currently `vips_target_finish` returns `void`, this can be a problem because it needs to call `vips_target_flush` which can fail.
This silences any errors that might occur on the last write to the target, and can be particularly devious if all the writes up to that point fit into the target buffer, causing the only write to actually occur to fail silently.

In summary, when using the streaming api (at least _custom) any output that is less than `VIPS_TARGET_BUFFER_SIZE` can fail unbeknownst to the caller.

**To Reproduce**
Load and save a small image with a write implementation that fails.
```c
#include <stdio.h>
#include <unistd.h>
#include <vips/vips.h>

gint64 write_cb(VipsTargetCustom *target_custom, gpointer data, gint64 length, gpointer user_data) {
  printf("quick, duck! it's a cosmic ray!\n");
  return -1;
};

int main() {
  VipsSource *source = vips_source_new_from_file("box.jpg");
  VipsTargetCustom *target_custom = vips_target_custom_new();
  g_signal_connect(target_custom, "write", G_CALLBACK(write_cb), NULL);

  VipsImage *img;
  if (vips_jpegload_source(source, &img, "access", VIPS_ACCESS_SEQUENTIAL, NULL)) {
    printf("%s\n", vips_error_buffer());
    return 1;
  }

  if (vips_jpegsave_target(img, VIPS_TARGET(target_custom), NULL)) {
    printf("%s\n", vips_error_buffer());
    return 1;
  }

  printf("phew!\n");
  return 0;
}
```
box.jpg
![box](https://user-images.githubusercontent.com/3987005/168396878-07aaf23a-fbaa-4206-9bfb-dc7d0178df38.jpg)

**Environment**
 - Vips: 7561d4e52c9961e9f644fe23247248704f4498b6

jcupitt:
You're right, I remember worrying about this.

I've made a PR for a fix.

PR #2802, opened 2022-05-14
add vips_target_end

jcupitt:
since finish did not return an error code

deprecate vips_target_finish

also make sure we don't call target_end from inside _dispose, since that
can't signal error either

see https://github.com/libvips/libvips/issues/2801

### DIFF
--- a/libvips/iofuncs/targetcustom.c
+++ b/libvips/iofuncs/targetcustom.c
@@ -64,6 +64,7 @@ G_DEFINE_TYPE( VipsTargetCustom, vips_target_custom, VIPS_TYPE_TARGET );
  */
 enum {
 	SIG_WRITE,		
+	SIG_END,		
 	SIG_FINISH,		
 	SIG_LAST
 };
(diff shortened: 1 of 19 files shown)


libvips/iofuncs/targetcustom.c
```diff
@@ -148,11 +176,27 @@ vips_target_custom_class_init( VipsTargetCustomClass *class )
...
+	 *
+	 * Returns: 0 on success, -1 on error.
+	 */
+	vips_target_custom_signals[SIG_END] = g_signal_new( "end",
+		G_TYPE_FROM_CLASS( class ),
+		G_SIGNAL_ACTION,
+		G_STRUCT_OFFSET( VipsTargetCustomClass, end ), 
+		NULL, NULL,
+		vips_INT__VOID,
+		G_TYPE_NONE, 0 );
```
kleisauke, line 193:
```suggestion 
		G_TYPE_INT, 0 );
```

I think that would fix this error:
```
(process:20208): GLib-GObject-CRITICAL **: 16:25:55.597: g_value_set_int: assertion 'G_VALUE_HOLDS_INT (value)' failed
```
(untested)

jcupitt: Ooop crap. Thanks!!

PR #2803, opened 2022-05-14
Don't fail if PNG contains too many text chunks

DarthSim:
Hey!

One of imgproxy users gave me a sample of proper PNG with 61 text chunks that expectedly is rejected by vips. First I was going to increase the text chunks limit. But then I thought that we already loaded text chunks at the moment of their number check. So if we have too many text chunks, we can just throw a warning, get the maximum allowed text chunks, and set them to the vips image. It doesn't seem to me that we SHOULD return an error in this case.

WDYT?

### DIFF
--- a/libvips/foreign/spngload.c
+++ b/libvips/foreign/spngload.c
@@ -258,9 +258,9 @@ vips_foreign_load_png_set_header( VipsForeignLoadPng *png, VipsImage *image )
 		 */
 
 		if( !png->unlimited && n_text > MAX_PNG_TEXT_CHUNKS ) {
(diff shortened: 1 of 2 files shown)


jcupitt:
Wow, 61!!!??!!? Some people are crazy!

Sure, I guess. Could you make the matching change to the spng loader as well? Somewhere around here https://github.com/libvips/libvips/blob/master/libvips/foreign/spngload.c#L253

DarthSim:
Yeah, this does look crazy :)
Added the same change to the spng loader.

jcupitt:
OK, thanks!

Issue #995, opened 2018-06-07
Support transparent backgrounds when using PDFium


tropperstyle:
I was able to run `vips pdfload input.pdf output.png` using PDFium after following #689 - however the output PNG renders a white background. When using pdftocairo, the -transp flag enabled this. Is this possible to achieve with the vips command line?

jcupitt:
Hello, I'll have a look. Do you have an sample PDF with transparent elements?

tropperstyle:
[test.pdf](https://github.com/jcupitt/libvips/files/2081426/test.pdf)

![test-convert](https://user-images.githubusercontent.com/28314/41115214-afe70ab0-6a3b-11e8-8e16-68a364d8d559.png)
![test-vips](https://user-images.githubusercontent.com/28314/41115215-b002084c-6a3b-11e8-92b6-b2c0fac2ba3b.png)

jcupitt:
I had a go with a `background` option, for example:

```
vips copy transparent.pdf[background=0] x.png
```

To set a black (transparent) background. Would you be able to test it?

jcupitt:
It'll be in 8.7, closing.

PR #2804, opened 2022-05-15
Don't paint the background if the PDF has transparency

DarthSim:
Hey again!

Some PDF pages require the background to be transparent to be rendered properly. Luckily, `FPDFPage_HasTransparency` returns `TRUE` for such pages so we can easily determine if we should fill the canvas before rendering. PDFium sample tool does the same check BTW: https://pdfium.googlesource.com/pdfium/+/refs/heads/main/samples/pdfium_test.cc#864.

Also, I used `FPDFBitmap_FillRect` to fill the background of a single page instead of a whole canvas. And I changed `pdf->ink` type so it can be used with `FPDFBitmap_FillRect`, hope it's ok.

Unfortunately, I can't share a sample I have as it was provided by a customer of our customer so I definitely have no right to share it anywhere.

### DIFF
--- a/libvips/foreign/pdfiumload.c
+++ b/libvips/foreign/pdfiumload.c
@@ -104,6 +104,7 @@ EOF
 
 #include <fpdfview.h>
 #include <fpdf_doc.h>
+#include <fpdf_edit.h>
 
 typedef struct _VipsForeignLoadPdf {
 	VipsForeignLoad parent_object;
@@ -151,9 +152,9 @@ typedef struct _VipsForeignLoadPdf {
 	VipsRect image;
 	VipsRect *pages;
 
-	/* The [double] background converted to the image format.
+	/* The [double] background converted to the PDFium format.
 	 */
-	VipsPel *ink;
+	FPDF_DWORD ink;
 
 } VipsForeignLoadPdf;
 
@@ -437,6 +438,7 @@ vips_foreign_load_pdf_header( VipsForeignLoad *load )
 
 	int top;
 	int i;
+	VipsPel *ink;
 
 #ifdef DEBUG
 	printf( "vips_foreign_load_pdf_header: %p\n", pdf );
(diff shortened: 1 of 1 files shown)


jcupitt:
Hey @DarthSim, this looks good.

I remember this coming up before: https://github.com/libvips/libvips/issues/995

Here's the test image from that issue: [test.pdf](https://github.com/libvips/libvips/files/8699675/test.pdf)

I tried with poppler:

```
vips copy "test.pdf[background=0 128 0 128]" x.png
```

To make:

x

ie. a transparent green background.

I'll have a go with your branch.

jcupitt:
(poppler `background` is premultiplied BGRA, we should fix this too)

jcupitt:
I'm not sure this change is a good idea. Try this with the master branch and pdfium enabled:

```
vips copy "test.pdf[background=128 0 0 128]" x.png
```

To make:

x

(it should be a green background, we should fix this)

Or like this:

```
vips copy test.pdf[background=0] x.png
```

To make:

x

(try opening that in an image viewer to see the transparent background)

With this patch, the first example breaks since the transparent blue background is never applied.

Could you solve your user's problem by converting with `background=0`?

DarthSim:
The color in `FPDFBitmap_FillRect` should be ARGB. Fixed.

```
vips copy test.pdf[background=0] x.png
```
pdf3_t

```
vips copy "test.pdf[background=0 128 0 128]" x.png
```
pdf3_g

```
vips copy "test.pdf[background=128 0 0 128]" x.png
```
pdf3_r

DarthSim:
> Even though that image obviously has a transparent background

It is, but `FPDFPage_HasTransparency` reports if the page background SHOULD be transparent. I mean, if it returns 0, we should fill the background, otherwise, we should. As I mentioned in the PR description, PDFium's sample tool does the same. Also, if we'll take a look at `FPDFPage_HasTransparency` [implementation](https://pdfium.googlesource.com/pdfium/+/refs/heads/main/fpdfsdk/fpdf_editpage.cpp#285), we'll see that it just returns `pPage->BackgroundAlphaNeeded()` which name actually has more sense in the context of this discussion. And if we take a look at [what changes this value](https://pdfium.googlesource.com/pdfium/+/refs/heads/main/core/fpdfapi/page/cpdf_allstates.cpp#106), we'll see that it depends only on blend mode.

Most raw-text PDFs don't defy their background yet should not be rendered transparent. For example https://code.visualstudio.com/shortcuts/keyboard-shortcuts-windows.pdf.

jcupitt:
Ah I understand, thanks for the explanation. Yes, `BackgroundAlphaNeeded()` seems like a better name.

OK, let's merge, and thank you for pushing for this improvement!

jcupitt:
I changed it to use `vips__bgra2rgba()` -- it should work with any byte order.

I suppose this means that the poppler PDF loader is wrong. I'll look for something like `FPDFPage_HasTransparency()` in their API.

DarthSim:
Great, thanks!

> I changed it to use vips__bgra2rgba() -- it should work with any byte order.

Ink in `FPDFBitmap_FillRect()` should be ARGB: https://pdfium.googlesource.com/pdfium/+/refs/heads/main/public/fpdfview.h#1093

jcupitt:
Yes, `vips__bgra2rgba` swaps B and A, so it also converts RGBA to BGRA (it's a self inverse).

jcupitt:
... and ARGB is BGRA but written LSB first.

PR #2810, opened 2022-05-17
webpsave: add parameter for mixed encoding

dloebl:
We noticed that the `allow_mixed` flag in combination with the `min_size` flag can substantially degrade the encoding time of animated WebP without much improvement on the file size. Allow passing the `allow_mixed` value with a new parameter in `webpsave`.

**Example animation:**
[earth.gif](https://upload.wikimedia.org/wikipedia/commons/2/2c/Rotating_earth_%28large%29.gif)
libwebp version: `v1.2.2`

**Old default `min-size=TRUE` behaviour:**
```
$ time vips copy earth.gif[n=-1] min_mixed.webp[min-size=true,mixed=true]
6.743s
```

**New default `min-size=TRUE` behaviour:**
```
$ time vips copy earth.gif[n=-1] min.webp[min-size=true,mixed=false]
0.758s
```

```
$ du -h min_mixed.webp min.webp
256K	min_mixed.webp
256K	min.webp
```

/cc @gurpreetg

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -21,6 +21,7 @@
 - add "gap" option to vips_reduce[hv]() and vips_resize() [kleisauke]
 - add "ceil" option to vips_shrink() [kleisauke]
 - quality improvements for image resizing [kleisauke]
+- add "mixed" to webpsave [dloebl]
 
 26/11/21 started 8.12.3
 - better arg checking for vips_hist_find_ndim() [travisbell]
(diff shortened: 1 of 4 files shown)


jcupitt:
Wow that's a big difference! Sure, this looks great. 

Could you add a sentence to the docs explaining what `allow_mixed` does, and a line to the global ChangeLog (since this is an API addition)?

jcupitt:
Thank you very much!

PR #2813, opened 2022-05-20
jxl: remove support for never-implemented JXL_TYPE_UINT32

lovell:
This was removed via https://github.com/libjxl/libjxl/pull/1337

Technically it has been deprecated, but can be considered as good as gone due to not being included in the official spec.

### DIFF
--- a/libvips/foreign/jxlload.c
+++ b/libvips/foreign/jxlload.c
@@ -350,10 +350,6 @@ vips_foreign_load_jxl_print_format( JxlPixelFormat *format )
 		printf( "JXL_TYPE_UINT16" );
 		break;
 
-	case JXL_TYPE_UINT32: 
-		printf( "JXL_TYPE_UINT32" );
-		break;
-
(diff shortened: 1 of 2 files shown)


lovell:
Whilst clang reports a warning, gcc cannot compile libvips against the latest main branch of libjxl without the change in this PR, so we might want to backport this to 8.12 if there are going to be further 8.12.x releases.

clang:
```
../libvips/foreign/jxlsave.c:284:27: warning: 'JXL_TYPE_UINT32' is deprecated [-Wdeprecated-declarations]
                jxl->format.data_type = JXL_TYPE_UINT32;
```
gcc:
```
../libvips/foreign/jxlsave.c:185:2: error: case label does not reduce to an integer constant
  185 |  case JXL_TYPE_UINT32:
      |  ^~~~
```

jcupitt:
argh, why did they remove the enum and break all downstream projects? Sure, I'll patch 8.12 too.

PR #2814, opened 2022-05-20
Fix a couple of small typos in help messages

lovell:


jcupitt:
ooop, I should have spotted the `minimum` one :(

Issue #2568, opened 2021-12-03
Add linear color encoding for JXL save operation


f1ac:
Currently JxlColorEncodingSetToSRGB() is always used in jxlsave.c, but it seems calling JxlColorEncodingSetToLinearSRGB() would be more appropriate when input image has VIPS_FORMAT_FLOAT and/or VIPS_INTERPRETATION_scRGB. Currently exported linear images appear as if they have additional gamma of 2.2 applied to them because of this issue. 

Libjxl tests seem to follow this logic (float input means linear output when input file has no ICC profile) at https://github.com/libjxl/libjxl/blob/main/lib/jxl/roundtrip_test.cc

f1ac:
Also it seems libjxl now has JxlEncoderSetICCProfile() method, so it can be used instead of the logic above when input files have a profile. As I can see currently ICC profiles are not handled for JXL.

jcupitt:
Hi @f1ac,

Sure, this all sounds good. Let's tag this as an enhancement for 8.13.

jcupitt:
OK, done! I've made a PR here:

https://github.com/libvips/libvips/pull/2815

Any testing very welcome.

It looks like libjxl now has EXIF, XMP etc. as well, so we should also add support for them, and perhaps bump the minimum supprted libjxl version.

f1ac:
@jcupitt Are there any nightly builds for windows? I gave up trying to compile the library, so can't test this w/o binaries unfortunately.

jcupitt:
With WSL2 it should be a one-liner to build from source now:

https://github.com/libvips/build-win64-mxe

Just:

```
./build.sh all --nightly
```

To build git master libvips for windows. It does take a while :(

jcupitt:
Of course, having looked at this, there's no simple way to build a specific commit. I'll try to add this feature.

f1ac:
I've managed to build vips master, then changed the branch in plugins/nightly/overrides.mk to "vips-all_GH_CONF  := libvips/libvips/branches/add-more-libjxl-metadata". The second time it was much faster, but I guess it did not pick up the new branch as resulting JXL files are identical to the ones converted with v8.12.2. I either need to clean up build folder or the overrides do no work.

kleisauke:
If you want to test specific commits or branches, I think you also need to change `nightly_version` here:
https://github.com/libvips/build-win64-mxe/blob/4ddf96c4c152f129d320d27fcf11ade9f9f73e65/build/build.sh#L133
[...]

PR #2815, opened 2022-05-21
add libjxl ICC and scRGB support

jcupitt:
see https://github.com/libvips/libvips/issues/2568

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -27,6 +27,7 @@
 - add vips_tiffsave_target()
 - add vips_target_end(), deprecate vips_target_finish()
 - add "mixed" to webpsave [dloebl]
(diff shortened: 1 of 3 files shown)


jcupitt:
There's a branch here with the start of exif and xmp support, but libjxl have not released this yet, so we should wait.

jcupitt:
I'm still not sure if this is correct, but it's certainly better than what we had before.

I'm inclined to merge this so we can release libvips 8.13, and do another round of JXL improvements when the next version of libjxl is released. What do you think?

f1ac:
Totally agree, I see no issues with this PR.

jcupitt:
OK, let's do it! Thank you for pushing for this and for doing the testing.

PR #2817, opened 2022-05-22
svgload: recognize dimensions for SVGs without width/height

kleisauke:
Rather than defaulting to 1928x1080. Also, prefer to use the
`LIBRSVG_CHECK_VERSION` macro instead of our configure checks.


jcupitt:
That's much better. What a complicated thing though :(

PR #2823, opened 2022-05-26
cgifsave: fill transparent pixels with zeroes

DarthSim:
Hey!

When we make pixels fully transparent in `cgifsave` it's better to fill them with zeroes. This removes nasty artifacts when saving resized GIF with transparency. Also this reduces the number of colors in the image and helps a quantizer to build a better palette.


jcupitt:
Hi @DarthSim, thanks for this!

Issue #2790, opened 2022-05-07
The long and winding road to 8.13


jcupitt:
Aim for a release in late May.

## TODO

- [x] dzsave to target
- [x] tiffsave to target see PR #2798 
- [x] GIF save improvements see PR #2797
- [x] support linear light in JXL see PR #2815
- [x] resolve JXL lossless encode issues
- [x] update downstream projects where necessary
- [x] fix JXL save of float RGB (need new SAVEABLE enum)
- [x] another round of GIF fixes
- [x] website docs update
- [x] release notes and a news post to write
- [x] put out an rc for evaluation
- [x] .... and get the rc at least mostly approved by downstream users
- [ ] publicity

Milestone: https://github.com/libvips/libvips/milestone/11

jcupitt:
@dloebl we should get some GIF fixes in too. Please add some notes to the TODO above if you like.

vansante:
Would love to see [this one](https://github.com/libvips/libvips/issues/2668) :wink:

jcupitt:
pyvips now has support in master for the new targetcustom stuff

kleisauke:
I just emailed Remi Collet with a patch that migrates the vips spec file to build with Meson for the upcoming 8.13 release. While doing so, I noticed 2 things:
- The `VIPS_VERSION_STRING` definition doesn't include the ChangeLog date when building with Meson, but perhaps that's OK.
https://github.com/libvips/libvips/blob/111eef0049068bd12e0ab5ffa1e26246f90a0a5f/libvips/include/vips/meson.build#L132
https://github.com/libvips/libvips/blob/111eef0049068bd12e0ab5ffa1e26246f90a0a5f/configure.ac#L29
- nifticlib usage needs to be disabled on Fedora due to missing pkg-config files. This will hopefully be fixed once PR https://github.com/NIFTI-Imaging/nifti_clib/pull/140 is merged.

jcupitt:
OK, I think we're almost ready. Shall we feature-freeze and do a release?

kleisauke:
I have nothing to add for 8.13, so I'm fine with a feature freeze and a possible release candidate. Though, I noticed some pending cgifsave PRs for possible inclusion.

jcupitt:
I think I found another cgifsave issue --- this  GIF:

![sample](https://user-images.githubusercontent.com/580843/171033026-2c46be03-117e-4aad-b8be-33faf75ce806.gif)

When copied like this:

```
[...]

PR #2824, opened 2022-05-27
cgifsave: palette change POC

DarthSim:
Hey!

This is the POC of my approach to generating palettes for `cgifsave`.  How it works:

1. We compute a palette for each frame. The first frame's palette we use as a global one.
2. For each frame except the first we compare the palette computed in step 1 with the global palette and the local palette of the previous frame (if any).
3. If there's a local palette in the previous frame, and the diff between it and the current frame palette is small enough, we use the palette from the previous frame as local for the current frame.
4. Otherwise, if the diff between the global palette and the current frame palette is small enough, we don't use a local palette.
5. Otherwise, we use the palette computed in step 1 as a local palette for the current frame.

With this approach, we mitigate the difference between neighbor frame colors which helps transparency and frame optimizers do their jobs.

#### Master speeed:

```
$ /usr/bin/time -f %M:%e vips copy /images/animated_gif12.gif[n=-1] /images/animated_gif12_vips_out_master.gif[reoptimise=1]
40344:5.21

$ /usr/bin/time -f %M:%e vips copy /images/animated_gif13.gif[n=-1] /images/animated_gif13_vips_out_master.gif[reoptimise=1]
32552:5.33

$ /usr/bin/time -f %M:%e vips copy /images/animated_gif15.gif[n=-1] /images/animated_gif15_vips_out_master.gif[reoptimise=1]
27096:3.36

$ /usr/bin/time -f %M:%e vips copy /images/animated_gif16.gif[n=-1] /images/animated_gif16_vips_out_master.gif[reoptimise=1]
41744:2.18
```

#### New speeed:

```
$ /usr/bin/time -f %M:%e vips copy /images/animated_gif12.gif[n=-1] /images/animated_gif12_vips_out_new.gif[reoptimise=1]
40340:4.98

$ /usr/bin/time -f %M:%e vips copy /images/animated_gif13.gif[n=-1] /images/animated_gif13_vips_out_new.gif[reoptimise=1]
32972:5.11

$ /usr/bin/time -f %M:%e vips copy /images/animated_gif15.gif[n=-1] /images/animated_gif15_vips_out_new.gif[reoptimise=1]
26884:3.28

$ /usr/bin/time -f %M:%e vips copy /images/animated_gif16.gif[n=-1] /images/animated_gif16_vips_out_new.gif[reoptimise=1]
41592:2.06
```

#### Result sizes

Here I resized the images a bit to increase the color number.

```
$ ls -l /images/animated_gif*_vips_out_*
-rw-r--r-- 1 *** ***   687545 May 27 06:34 /images/animated_gif12_vips_out_master.gif
-rw-r--r-- 1 *** ***   631936 May 27 06:50 /images/animated_gif12_vips_out_new.gif
-rw-r--r-- 1 *** ***  1051173 May 27 06:34 /images/animated_gif13_vips_out_master.gif
-rw-r--r-- 1 *** ***   997631 May 27 06:50 /images/animated_gif13_vips_out_new.gif
-rw-r--r-- 1 *** ***  3355198 May 27 06:51 /images/animated_gif15_vips_out_master.gif
-rw-r--r-- 1 *** ***  3320212 May 27 06:50 /images/animated_gif15_vips_out_new.gif
-rw-r--r-- 1 *** *** 14747250 May 27 06:35 /images/animated_gif16_vips_out_master.gif
-rw-r--r-- 1 *** *** 14557252 May 27 06:51 /images/animated_gif16_vips_out_new.gif
```

### Findings

As we see, the memory footprint is more or less the same for both approaches, yet my approach is slightly faster and produces slightly smaller files. Also, this approach doesn't have a frame size limitation.

### DIFF
--- a/libvips/foreign/cgifsave.c
+++ b/libvips/foreign/cgifsave.c
@@ -307,8 +307,8 @@ vips_foreign_save_cgif_write_frame( VipsForeignSaveCgif *cgif )
 		quantisation_result = cgif->quantisation_result;
 		lp = vips__quantise_get_palette( quantisation_result );
 	} else {
-		if( vips__quantise_image_quantize( cgif->input_image, cgif->attr,
-			&quantisation_result ) ) {
+		if( vips__quantise_image_quantize_fixed( cgif->input_image,
+			cgif->attr, &quantisation_result ) ) {
 			vips_error( class->nickname,
 				"%s", _( "quantisation failed" ) );
 			return( -1 );
@@ -709,7 +709,7 @@ vips_foreign_save_cgif_build( VipsObject *object )
 		tmp_image = vips__quantise_image_create_rgba( cgif->attr,
 			tmp_gct, gct_length + 1, 1, 0 );
 
-		if( vips__quantise_image_quantize( tmp_image,
+		if( vips__quantise_image_quantize_fixed( tmp_image,
 		       cgif->attr, &cgif->quantisation_result ) ) {
 		       vips_error( class->nickname,
 		       	"%s", _( "quantisation failed" ) );
--- a/libvips/foreign/cgifsave.c
+++ b/libvips/foreign/cgifsave.c
@@ -61,8 +61,9 @@ typedef struct _VipsForeignSaveCgif {
 	double dither;
 	int effort;
 	int bitdepth;
-	double maxerror;
+	double interframe_maxerror;
 	gboolean reoptimise;
+	double interpalette_maxerror;
 	VipsTarget *target;
 
 	/* Derived write params.
@@ -85,8 +86,7 @@ typedef struct _VipsForeignSaveCgif {
 	 */
 	VipsQuantiseAttr *attr;
 	VipsQuantiseImage *input_image;
-	VipsQuantiseResult *quantisation_result;
-	const VipsQuantisePalette *lp;
+	VipsQuantiseResult *quantisation_result, *local_quantisation_result;
 
 	/* The current colourmap, updated on a significant frame change.
 	 */
@@ -133,6 +133,8 @@ vips_foreign_save_cgif_dispose( GObject *gobject )
 	VIPS_FREEF( cgif_close, cgif->cgif_context );
 
 	VIPS_FREEF( vips__quantise_result_destroy, cgif->quantisation_result );
+	VIPS_FREEF( vips__quantise_result_destroy,
+		cgif->local_quantisation_result );
 	VIPS_FREEF( vips__quantise_image_destroy, cgif->input_image );
 	VIPS_FREEF( vips__quantise_attr_destroy, cgif->attr );
 
@@ -182,6 +184,58 @@ vips_foreign_save_cgif_pixels_are_equal( const VipsPel *cur, const VipsPel *bef,
 	return( dR * dR + dG * dG + dB * dB <= sq_maxerror );
 }
 
+static double
+vips__cgif_compare_palettes(const VipsQuantisePalette *new,
+	const VipsQuantisePalette *old)
+{
+	g_assert( new->count <= 256 );
+	g_assert( old->count <= 256 );
+
+	int i, j;
+	double best_dist, dist, rd, gd, bd;
+	double total_dist = 0.0;
+
+	for( i = 0; i < new->count; i++ ) {
+		best_dist = 255.0 * 255.0 * 3;
+
+		for( j = 0; j < old->count; j++ ) {
+			if( new->entries[i].a >= 128 ) {
+				/* The new entry is solid.
+				 * If the old entry is transparent, ignore it.
+				 * Otherwise, compare RGB.
+				*/
+				if( old->entries[j].a < 128 )
+					continue;
+
+				rd = new->entries[i].r - old->entries[j].r;
+				gd = new->entries[i].g - old->entries[j].g;
+				bd = new->entries[i].b - old->entries[j].b;
+				dist = rd*rd + gd*gd + bd*bd;
+
+				best_dist = VIPS_MIN(best_dist, dist);
+
+				/* We found the closest entry
+				 */
+				if( best_dist == 0 )
+					break;
+			} else {
+				/* The new entry is transparent.
+				 * If the old entry is transparent too, it's
+				 * the closest color. Otherwise, ignore it.
+				 */
+				if( old->entries[j].a < 128 ) {
+					best_dist = 0;
+					break;
+				}
+			}
+		}
+
+		total_dist += best_dist;
+	}
+
+	return total_dist / new->count;
+}
+
 /* We have a complete frame --- write!
  */
 static int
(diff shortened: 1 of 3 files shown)


libvips/foreign/cgifsave.c
```diff
@@ -203,6 +204,39 @@ vips_foreign_save_cgif_check_alpha_constraint( const VipsPel *cur,
...
+	double total_dist = 0.0;
+
+	for( i = 0; i < new->count; i++ ) {
+		double best_dist = 255.0 * 255.0 * 3;
+
+		for( j = 0; j < old->count; j++ ) {
+			if(
+				(new->entries[i].a >= 128) !=
+				(old->entries[j].a >= 128)
+			) continue;
```
jcupitt, line 224: I think this will ignore palette entries which stay transparent or which stay solid. You'll only consider entries which have changed from solid to transparent, or transparent to solid.

Don't you want to only consider pixels which are solid now and were also solid in the previous frame?

jcupitt, line 224: I think this will ignore palette entries which stay transparent or which stay solid. You'll only consider entries which have changed from solid to transparent, or transparent to solid.

Don't you want to only consider pixels which are solid now and were also solid in the previous frame?

jcupitt:
Oh, nice! It seems to work well on the test images I tried. @dloebl, any thoughts?

libvips/foreign/cgifsave.c
DarthSim: The point is to find the closest entry from the previous palette. If transparency doesn't match, we just ignore the entry completely.

jcupitt: Oop, yes, I had that flipped. But you don't need to consider RGB for transparent pixels, do you? How about:

```C
if( (new->entries[i].a >= 128) && (old->entries[j].a >= 128) )
    continue;
```

DarthSim: Good point!

DarthSim: Updated the function.

libvips/foreign/cgifsave.c
```diff
@@ -572,10 +665,6 @@ vips_foreign_save_cgif_build( VipsObject *object )
 	frame_rect.top = 0;
 	frame_rect.width = cgif->in->Xsize;
 	frame_rect.height = page_height;
-	if( (guint64) frame_rect.width * frame_rect.height > 5000 * 5000 ) {
```
kleisauke, line 575: Perhaps we should still check the theoretical dimension limit?:
```c
	/* GIF has a limit of 64k per axis -- double-check this.
	 */
	if( frame_rect.width > 65536 || frame_rect.height > 65536 ) {
```

DarthSim: Good point! Added the check

libvips/foreign/cgifsave.c
```diff
@@ -251,90 +308,125 @@ vips_foreign_save_cgif_write_frame( VipsForeignSaveCgif *cgif )
...
+			vips_error( class->nickname, 
+				"%s", _( "quantisation failed" ) );
+			return( -1 );
 		}
+		lp = vips__quantise_get_palette( quantisation_result );
 
-		if( cgif->frame_checksum == 0 ||
-			checksum != cgif->frame_checksum ) { 
-			cgif->frame_checksum = checksum;
+		static double pal_change_threshold = 64.0;
```
dloebl, line 325: This method seems to be similar to the way the `maxerror` parameter is handled right now: [cgifsave.c#L177](https://github.com/libvips/libvips/blob/6260a37136973f3d231d0e2c28af6190126c664e/libvips/foreign/cgifsave.c#L177)
We could make the threshold dynamic (e.g. use `maxerror` here) and set `maxerror` to a value > 0 by default.

dloebl:
Looks like a nice improvement to me 👍
I've added some comments inline.

libvips/foreign/cgifsave.c
```diff
@@ -203,6 +204,58 @@ vips_foreign_save_cgif_check_alpha_constraint( const VipsPel *cur,
...
+					continue;	
+
+				rd = new->entries[i].r - old->entries[j].r;
+				gd = new->entries[i].g - old->entries[j].g;
+				bd = new->entries[i].b - old->entries[j].b;
+				dist = rd*rd + gd*gd + bd*bd;
+
+				best_dist = VIPS_MIN(best_dist, dist);
+
+				/* We fond the closest entry
```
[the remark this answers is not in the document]
DarthSim: Thanks, fixed

libvips/foreign/cgifsave.c
DarthSim: I don't think `maxerror` fits well here. For example, 64 is a good value for the inter-palette error while it is too large for the inter-frame error. Yet I agree that this number should be dynamic. Maybe we should rename `maxerror` to, for example, `iterframe_maxerror` and add a new param for inter-palette max error.

DarthSim:
Could you give a sample, please? I have like 16 test GIFs, and all of them work fine.

jcupitt:
Sorry, it's not a GIF I can share :( I'm trying to find another example, but no luck so far.

Let's fix this up and merge it, then try to resolve any fails in another PR.

jcupitt:
Hi, shall I fix the merge conflict?

jcupitt:
I made a GIF I can share that fails.

This terrible GIF:

http://www.rollthepotato.net/~john/skate.gif

If you run this:

```
$ vips copy skate.gif[n=-1] x.gif
```

Becomes:

http://www.rollthepotato.net/~john/x.gif

I think every pixel which is equal to the previous pixel becomes transparent in the output, which causes a lot of artefacts at the start of the clip.

DarthSim:
Rebased and squashed the branch.

> I made a GIF I can share that fails.

Oh, thank you! I'll dig it.

DarthSim:
I've tested `vips copy` built from this branch, and I didn't get any artifacts: http://img.darthsim.me/ufHVruZ3o2.gif. I've tested it with both Quantizr and LIQ from the Debian package.

@jcupitt could you test it again on your end? Maybe it was fixed after the rebase?

jcupitt:
I wiped everything and rebuilt, and I see still bad artefacts.

Could it be the cgif version? I'm using git master.

dloebl:
Maybe it's similar to the issue mentioned [here](https://github.com/libvips/libvips/issues/2790#issuecomment-1141338058)?
I can reproduce the artefact with libimagequant 2.17 - but only if dithering is enabled. With Quantizr/libimagequant 4.0 the issue is gone on my side.

DarthSim:
@dloebl That's it, thank you! I'll try to fix it.

jcupitt:
Ah good spot.

But these artefacts only appear with this branch -- master seems to work fine. I think something in this PR has triggered this behaviour.

DarthSim:
BTW it seems like the same problem may occur when we reuse the global palette from the source image, too.

jcupitt:
Perhaps we should merge, and then try to find a workaround for this in another PR. What do you think?

jcupitt:
That's great! All my tests pass too, well done.

I see some small leaks in valigrind, but maybe that's just libimagequant. Let's merge and maybe try to patch them in another PR.

jcupitt:
I did a quick benchmark against imagemagick:

```
$ /usr/bin/time -f %M:%e vipsthumbnail 3198.gif[n=-1] -o vips-13.gif --size 224
57344:0.62
$ /usr/bin/time -f %M:%e convert 3198.gif -resize 75% im.gif
200796:6.44
$ ls -l im.gif vips-13.gif 
-rw-rw-r-- 1 john john 3176859 Jun  8 16:26 im.gif
-rw-r--r-- 1 john john 2487189 Jun  8 16:25 vips-13.gif
```

So for this GIF at least (a short video clip) libvips is now 10x faster, needs 4x less memory, and makes GIFs which are 20% smaller!

Nice work everyone.

jcupitt:
.... and the dithering of the libvips one is better too, so they look nicer.

PR #2828, opened 2022-05-30
nsgifload: add "palette-bit-depth" header

DarthSim:
Hey!

It'd be nice if `nsgifload` set the `palette-bit-depth` header as `pngload` does. It's pretty handy when saving the result.

If there is only a global palette, `palette-bit-depth` is set to its bit size. Otherwise, it's set to the biggest palette bit size.


jcupitt:
Sure, looks useful. Thanks!

PR #2847, opened 2022-06-08
text: ensure fontconfig cache of Pango is invalidated

kleisauke:
See: https://gitlab.gnome.org/GNOME/pango/-/issues/551.

Resolves: https://github.com/kleisauke/net-vips/issues/170.


jcupitt:
LGTM!

PR #2848, opened 2022-06-08
text: use MT-safe `pango_cairo_font_map_get_default`

kleisauke:
> Note that since Pango 1.32.6, the default fontmap is per-thread. Each thread gets its own default fontmap. In this way, PangoCairo can be used safely from multiple threads.
> https://docs.gtk.org/PangoCairo/type_func.FontMap.get_default.html

Depends on: #2847.


jcupitt:
Oh, sorry, maybe I merged in the wrong order.

jcupitt:
Oh, nice to get rid of that awful lock!

PR #2849, opened 2022-06-10
cgifsave: fix heap-use-after-free

kleisauke:
`liq_set_dithering_level` could somehow free the palette, so
do the transparency check prior to that.

Note that GCC still prints this warning:
```
In function ‘vips_foreign_save_cgif_write_frame’,
    inlined from ‘vips_foreign_save_cgif_sink_disc’ at ../libvips/foreign/cgifsave.c:587:8:
../libvips/foreign/cgifsave.c:401:35: warning: ‘pal_local’ may be used uninitialized [-Wmaybe-uninitialized]
  401 |                 for( i = 0; i < lp->count; i++ ) {
      |                                 ~~^~~~~~~
../libvips/foreign/cgifsave.c: In function ‘vips_foreign_save_cgif_sink_disc’:
../libvips/foreign/cgifsave.c:261:54: note: ‘pal_local’ was declared here
  261 |         const VipsQuantisePalette *lp, *pal_global, *pal_local;
      |                                                      ^~~~~~~~~

```
(probably a false positive)

### DIFF
--- a/libvips/foreign/cgifsave.c
+++ b/libvips/foreign/cgifsave.c
@@ -290,7 +290,7 @@ vips_foreign_save_cgif_write_frame( VipsForeignSaveCgif *cgif )
 		if( cur[3] >= 128 )
 			cur[3] = 255;
 		else {
-			/* Helps the quanizer generate a better palette.
+			/* Helps the quantizer generate a better palette.
 			 */
 			cur[0] = 0;
 			cur[1] = 0;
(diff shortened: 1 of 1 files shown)


kleisauke:
Marking as draft, since the palette is still fetched after `vips__quantise_set_dithering_level`.
https://github.com/libvips/libvips/blob/8afdbbef0bf57251d1a58ea7c2a371b89bf6826f/libvips/foreign/cgifsave.c#L487
This seems to cause a new ASan failure:
https://github.com/libvips/libvips/runs/6828919349?check_suite_focus=true#step:18:327

jcupitt:
I'm half-way through reworking cgifsave, I should have a PR with a rewritten (and easier to maintain) version later today.

jcupitt:
Should be fixed with https://github.com/libvips/libvips/pull/2853

PR #2850, opened 2022-06-10
CI: upgrade runner to Ubuntu 22.04

kleisauke:
- Compile with GCC 11 and Clang 14.
- Remove redundant PPAs.
- Use system provided Meson.
- Update LeakSanitizer suppressions.

Note that this GitHub Actions runner is still in beta phase, so we might have to wait.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,14 +10,14 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - name: "Linux x64 (Ubuntu 20.04) - GCC 9"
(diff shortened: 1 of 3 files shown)


jcupitt:
Hello Kleis, this looks good, do you think we should merge? Or wait a little longer?

kleisauke:
I think it's probably safe to land, but I'm not sure if the ASan / UBsan build would succeed, so I think I'll re-base this after PR #2853 lands.

jcupitt:
OK, I merged #2853

jcupitt:
LGTM!

PR #2854, opened 2022-06-11
Prefer to use `flags |=` everywhere

kleisauke:
Use the bitwise OR assignment operator to ensure that derived
classes does not overwrite the flags from the base class. Also,
move the flags from `openslideload_source` to its base class.


jcupitt:
Great!

PR #2855, opened 2022-06-11
Mark Poppler as untrusted

kleisauke:
`pdfload` using PDFium is continuously fuzzed by OSS-Fuzz, but Poppler not.

### DIFF
--- a/libvips/foreign/popplerload.c
+++ b/libvips/foreign/popplerload.c
@@ -519,6 +519,7 @@ vips_foreign_load_pdf_class_init( VipsForeignLoadPdfClass *class )
 {
 	GObjectClass *gobject_class = G_OBJECT_CLASS( class );
 	VipsObjectClass *object_class = (VipsObjectClass *) class;
+	VipsOperationClass *operation_class = VIPS_OPERATION_CLASS( class );
 	VipsForeignLoadClass *load_class = (VipsForeignLoadClass *) class;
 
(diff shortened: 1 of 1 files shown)


jcupitt:
I remember looking into poppler and it is being fuzzed by oss-fuzz, just not by us. Do you think that's not enough? I suppose it won't be checking our poppler code.

kleisauke:
The same policy is also applied to `matload`, `jp2kload` and `openexrload`, so I thought we should do also the same for `pdfload` using Poppler.
https://github.com/libvips/libvips/blob/1214f942f9397db25cfec19d5712b88480965a0f/libvips/foreign/matload.c#L124-L126
https://github.com/libvips/libvips/blob/1214f942f9397db25cfec19d5712b88480965a0f/libvips/foreign/jp2kload.c#L1052-L1054
https://github.com/libvips/libvips/blob/1214f942f9397db25cfec19d5712b88480965a0f/libvips/foreign/openexrload.c#L131-L133

But perhaps the other loaders are far less utilized (for example, [all modern browsers ship with a built-in PDF viewer](https://caniuse.com/pdf-viewer), but [JPEG 2000 is supported solely by Safari](https://caniuse.com/jpeg2000). MATLAB and OpenEXR images are not supported at all in web browsers).

jcupitt:
Ah, true. Sure, let's untrust poppler.

PR #2857, opened 2022-06-12
Minor cleanups

kleisauke:
- Remove `HAVE_LCMS` definition in favor of `HAVE_LCMS2`.
- Remove `HAVE_WINDOWS_H` definition in favor of `G_OS_WIN32`.
- Remove stray `vips_text_get_type` in `conversion.c`.
- Remove duplicated `unistd.h` include.
- Remove redundant `strcasecmp` definition, we use `g_ascii_strcasecmp` everywhere.
- Remove unnecessary header checks in `configure.ac` and `meson.build`.
- Ensure `unistd.h` include is guarded with `HAVE_UNISTD_H`.
- Fail early when `-Dfontconfig=enabled` and `pangoft2` is not found.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -304,7 +304,7 @@ EXTRA_LIBS_USED=""
 # Checks for header files.
 AC_HEADER_DIRENT
 AC_HEADER_STDC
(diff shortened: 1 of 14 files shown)


jcupitt:
Ahhh much tidier. It's surprising how quickly dead wood builds up.

PR #2862, opened 2022-06-13
cgifsave: fix stack-use-after-scope

kleisauke:
By reusing `cgif->palette_rgb`, this ensures that `pLocalPalette`
doesn't become indeterminate when it reaches `cgif_addframe`.


jcupitt:
Oooop yes that's bad. I've done too much C# recently :( 

I think I wanted to keep the RGB version of the global palette around, but you're right, it's never used, so we might as well put the local palette in that array.

Issue #2852, opened 2022-06-10
cgifsave: number of colours in palette is one less than it could/should be


lovell:
The number of colours in the palette of GIF output via cgifsave is currently set to one less than it could be due to this line:

https://github.com/libvips/libvips/blob/bae0342d9506a4e6f2e5e8ba0bb612c6a601044e/libvips/foreign/cgifsave.c#L685

This can be highlighted by setting `bitdepth` to 1, resulting in a call to `liq_set_max_colors` with the value 1, which is out of the 2-256 range it supports. We currently ignore the error return code so the output GIF has the (default) bitdepth of 8, which is incorrect.

I'm probably to blame here as I originally wrote this code, with the logic of decrementing by 1 to allow space for an extra "colour" for the transparency, if any. However I think with all the recent updates to this code that this logic is no longer required, so the following change should fix this.

```diff
- vips__quantise_set_max_colors( cgif->attr, (1 << cgif->bitdepth) - 1 );
+ vips__quantise_set_max_colors( cgif->attr, 1 << cgif->bitdepth );
```

I'm aware there's more refactoring going on in and around this code so am creating an issue to track it rather than open a PR that may conflict with any work-in-progress.

jcupitt:
Ooop, good catch, I'll fix it in my branch.

jcupitt:
Should be fixed with https://github.com/libvips/libvips/pull/2853

kleisauke:
I noticed that this change produces artefacts with `skate.gif` from https://github.com/libvips/libvips/pull/2824#issuecomment-1147855022, see:
```bash
$ vips copy skate.gif[n=-1] x.gif
```
https://t0.nl/x.gif

Reverting it back to `(1 << cgif->bitdepth) - 1` seems to fix it.
```diff
@@ -731,7 +731,7 @@ vips_foreign_save_cgif_build( VipsObject *object )
 	/* Set up libimagequant.
 	 */
 	cgif->attr = vips__quantise_attr_create();
-	vips__quantise_set_max_colors( cgif->attr, 1 << cgif->bitdepth );
+	vips__quantise_set_max_colors( cgif->attr, (1 << cgif->bitdepth) - 1 );
 	vips__quantise_set_quality( cgif->attr, 0, 100 );
 	vips__quantise_set_speed( cgif->attr, 11 - cgif->effort );
```

lovell:
[...]

PR #2863, opened 2022-06-13
cgifsave: limit max number of colours to 255

dloebl:
Limit the maximum number of colours to 255, so we always leave one index free for the transparency.
Avoids the artefact mentioned here: https://github.com/libvips/libvips/issues/2852#issuecomment-1153728379

### DIFF
--- a/libvips/foreign/cgifsave.c
+++ b/libvips/foreign/cgifsave.c
@@ -733,7 +733,7 @@ vips_foreign_save_cgif_build( VipsObject *object )
 	 * free for the transparency optimization.
 	 */
 	vips__quantise_set_max_colors( cgif->attr,
-		MIN( 255, 1 << cgif->bitdepth) );
+		VIPS_MIN( 255, 1 << cgif->bitdepth ) );
 	vips__quantise_set_quality( cgif->attr, 0, 100 );
 	vips__quantise_set_speed( cgif->attr, 11 - cgif->effort );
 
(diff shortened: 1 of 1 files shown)


dloebl:
Hi @lovell. Good question!
If a `bitdepth` of 7 is specified with this branch we will create a colour palette with a maximum of 128 entries and set the `transparency index` to the number of colours:
https://github.com/libvips/libvips/blob/a7697b462fd1d54292fae98a1180bb89767fc9ef/libvips/foreign/cgifsave.c#L556

In this case, the valid indices are `0-127` for the image data and `128` is the `transparency index`.
cgif supports transparency indices that are outside of the active colour palette, but the highest possible index is `255` of course.

As far as I can tell, `gifsicle` and `ImageMagick` handle this the same way and allow the `transparency index` to be exactly one position outside of the active colour palette.

libvips chooses the `transparency index` and provides the transparent areas (so which areas are identical to the frame before) to cgif as `CGIF_FRAME_ATTR_HAS_SET_TRANS` is set (`CGIF_FRAME_GEN_USE_TRANSPARENCY` is ignored by cgif as libvips provides the transparency information). It is necessary that libvips provides the transparency information to support the `interframe-maxerror` parameter.

**With this branch I see:**
```
vips copy earth.gif[n=-1] 7.gif[bitdepth=7,reoptimise=1]
```
[earth.gif](https://user-images.githubusercontent.com/34104349/173533936-22b48c75-9a44-47b9-9068-93bff6bde868.gif)
| bitdepth 8 | bitdepth 7 | bitdepth 1 |
| - | - | - |
|  |  |  |

bitdepth 1 might look unexpected at first because sometimes one can see 3 colours. The colours per frame are limited to 2, but the transparency might cause pixels to shine through.

jcupitt:
Thanks @dloebl !

PR #2866, opened 2022-06-15
doc: make pandoc optional

kleisauke:
Since it's currently unavailable on RHEL 9, see:
https://bugzilla.redhat.com/show_bug.cgi?id=2089013


jcupitt:
Looks good!

PR #2870, opened 2022-06-17
cgifsave: remove regressions from 2853

### LINKED ISSUE #2864: cgifsave: inter-frame maxerror might be unlimited
With some animated GIFs, the inter-frame `maxerror` can be unlimited. This can happen because the error is within the threshold from `Frame A` to `Frame B` but adds up with every frame.
Previously, we always [compared against a HEAD frame](https://github.com/libvips/libvips/blob/b74504b9858a88c7b0e2524645d82afc839af0d4/libvips/foreign/cgifsave.c#L502) in `cgifsave`. Unfortunately, this HEAD frame logic has been recently removed with #2853.

**Example:**
[issue.gif](https://user-images.githubusercontent.com/34104349/173402325-1a5b4000-53f2-4e28-985d-1acaa9190921.gif) 
```
vips copy issue.gif[n=-1] out.gif[interframe-maxerror=4]
```
| original | output |
| - | - |
| issue | out |

As you can see, the output animation is just black - which is not as expected of course.

DarthSim:
Hey!

This PR fixes a few regressions introduced in #2853.

e4271e16de1f5ee75c27d3115edbe951597db3da: My original approach intent picking the best suitable palette. But after #2835 it picks the global palette if it is good enough even if the previous local palette is better.
c33d2a2766fe542856e2bcaf766ca5cd018c4505: #2835 changed palette comparison and `interpalette_maxerror` of 40 became too high. Also, `int` is not prcisive enough for this new method and `double` should be used.
4a9c96a6a49443522f7fca2a9edc74e2493cd81d: only changed pels should be copied to `cgif->previous_frame`.

The first two commits fix #2864.

### DIFF
--- a/libvips/foreign/cgifsave.c
+++ b/libvips/foreign/cgifsave.c
@@ -219,13 +219,13 @@ vips_foreign_save_cgif_set_transparent( VipsForeignSaveCgif *cgif,
 	}
 }
 
-static int
+static double
 vips__cgif_compare_palettes( const VipsQuantisePalette *new,
 	const VipsQuantisePalette *old )
 {
 	int i, j;
-	int best_dist, dist, rd, gd, bd;
-	int total_dist;
+	double best_dist, dist, rd, gd, bd;
+	double total_dist;
 
 	g_assert( new->count <= 256 );
 	g_assert( old->count <= 256 );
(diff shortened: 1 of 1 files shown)


jcupitt:
Hi @DarthSim, sorry for breaking this :( I'll look at your PR tomorrow.

jcupitt:
Hi again,
 
I noticed a failure with `skate.gif`. With:

```
$ vips copy skate.gif[n=10] x.gif
```

I see:

x

I think it's just because with this GIF there are no transparent palette entries. If I change the transparancy section to this:

```C
        /* Pixels which are equal to pixels in the previous frame can be made
         * transparent, provided no alpha channel constraint is present.
         */
        if( page_index > 0 &&
                has_transparency &&
                !has_alpha_constraint ) {
                vips_foreign_save_cgif_set_transparent( cgif,
                        cgif->previous_frame, frame_bytes, cgif->index, 
                        n_pels, 0 );

                frame_config.attrFlags &= ~CGIF_FRAME_ATTR_HAS_ALPHA;
                frame_config.attrFlags |= CGIF_FRAME_ATTR_HAS_SET_TRANS;
                frame_config.transIndex = 0;
        }
        else {
                /* Take a copy of the RGBA frame.
                 */
                memcpy( cgif->previous_frame, frame_bytes, 4 * n_pels );
        }
```

It seem to work.

Of course this means that GIFs which are video clips will hardly ever use transparency optimisation, since they will usually fill the palette and not have a transparent entry.

Perhaps `vips__quantise_image_quantize_fixed()` should always make palettes with index 0 as transparent?

We have a little code in `vips_foreign_save_cgif_build()` that tries to do this for the first frame, but of course that's often not enough.

jcupitt:
Ah, Kleis suggests another fix: https://github.com/libvips/libvips/issues/2852#issuecomment-1153728379

```C
        vips__quantise_set_max_colors( cgif->attr, (1 << cgif->bitdepth) - 1 );
```

To leave a palette entry clear for transparency. We'd still need to add the transparent element though.

jcupitt:
Ah of course that's also in @dloebl's PR: https://github.com/libvips/libvips/pull/2863 ahem I'm slowly catching up.

jcupitt:
... so combined with that other PR, all my test GIFs pass. Thanks you for fixing my mess, @DarthSim.

PR #2874, opened 2022-06-19
thumbnail: avoid upsizing via libwebp

kleisauke:
Test case:
```bash
$ vips black x.webp 100 100
$ vipsthumbnail x.webp -s 16384x
```

Context: https://github.com/lovell/sharp/issues/3262.


jcupitt:
Heh I hadn't thought of libwebp upsize.

PR #2889, opened 2022-06-23
Cleanup .gitignore

kleisauke:
Sorted difference can be viewed here:
https://gist.github.com/kleisauke/d28225be08b202e0daea070948dd48c0


libvips/foreign/libnsgif/Makefile.am
```diff
@@ -12,7 +12,6 @@ endif
 
 EXTRA_DIST = \
 	meson.build \
-	patches \
```
jcupitt, line 15: `patches` is needed for `Makefile.am`. I think it's harmless to have it here and empty.

jcupitt: Sure, sounds like a good idea.

kleisauke: OK, done: https://github.com/libvips/libvips/pull/2889/commits/e4c5039f61ed7f3c98462f5d2a86d5376c4a1ed6.

PR #2891, opened 2022-06-23
doc: ensure `pandoc-docbook-template.docbook` is distributed

kleisauke:
Since that is also used by Meson (when configuring with `-Dgtk_doc=true`
and when pandoc is available).


jcupitt:
Good find!

PR #2895, opened 2022-06-26
Meson: avoid using deprecated feature

kleisauke:
Fixes the build with `-Ddoxygen=true --fatal-meson-warnings`:
```
cplusplus/meson.build:48: WARNING: Project targeting '>=0.56' but tried to use feature deprecated since '0.56.0': meson.source_root. use meson.project_source_root() or meson.global_source_root() instead.

cplusplus/meson.build:47:17: ERROR: Fatal warnings enabled, aborting
```


jcupitt:
I half-remember a version of doxygen not working unless it had an absolute path, but it seems fine now, so let's do it!

PR #2896, opened 2022-06-26
Doxygen: use SVG for the call graph images

kleisauke:
Now that IE is finally gone, all the main browsers support this, which
provides higher quality and smaller sizes.

Inspired by commit https://github.com/libexif/libexif/commit/4a7ab4cc1332afbc63c44a8a96c93421d6aa07a0.


jcupitt:
:thumbs-up:

PR #2919, opened 2022-07-12
svgload: use `rsvg_handle_get_geometry_for_element` instead

kleisauke:
See: https://gitlab.gnome.org/GNOME/librsvg/-/issues/853.


jcupitt:
Nice!

Issue #2878, opened 2022-06-20
[8.13.0-rc1] broken build on RHEL-7 - undefined reference to `iszip'


remicollet:
```
/bin/sh ../libtool  --tag=CC   --mode=link gcc -std=gnu11 -std=gnu11  -O2 -g -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong --param=ssp-buffer-size=4 -grecord-gcc-switches   -m64 -mtune=generic -ftree-vectorize -no-undefined -shared -module -avoid-version -Wl,-z,relro  -o vips-magick.la -rpath /usr/lib64/vips-modules-8.13 module/vips_magick_la-magick.lo foreign/vips_magick_la-magick.lo foreign/vips_magick_la-magick2vips.lo foreign/vips_magick_la-magick6load.lo foreign/vips_magick_la-magick7load.lo foreign/vips_magick_la-vips2magick.lo libvips.la -pthread -lgmodule-2.0 -lgio-2.0 -lgobject-2.0 -lglib-2.0   -lMagickCore-6.Q16   
libtool: link: ( cd ".libs" && rm -f "vips-openslide.la" && ln -s "../vips-openslide.la" "vips-openslide.la" )
libtool: link: ( cd ".libs" && rm -f "vips-poppler.la" && ln -s "../vips-poppler.la" "vips-poppler.la" )
libtool: link: gcc -std=gnu11 -shared  -fPIC -DPIC  module/.libs/vips_magick_la-magick.o foreign/.libs/vips_magick_la-magick.o foreign/.libs/vips_magick_la-magick2vips.o foreign/.libs/vips_magick_la-magick6load.o foreign/.libs/vips_magick_la-magick7load.o foreign/.libs/vips_magick_la-vips2magick.o   -Wl,-rpath -Wl,/builddir/build/BUILD/vips-8.13.0/libvips/.libs ./.libs/libvips.so -lgmodule-2.0 -lgio-2.0 -lgobject-2.0 -lglib-2.0 -lMagickCore-6.Q16  -O2 -g -fstack-protector-strong -grecord-gcc-switches -m64 -mtune=generic -Wl,-z -Wl,relro -pthread   -pthread -Wl,-soname -Wl,vips-magick.so -o .libs/vips-magick.so
[...]

PR #2927, opened 2022-07-14
Fix `dzsave_target` for ZIP output

kleisauke:
See: https://github.com/libvips/libvips/issues/2878#issuecomment-1184199505.

### DIFF
--- a/libvips/foreign/dzsave.c
+++ b/libvips/foreign/dzsave.c
@@ -248,9 +248,30 @@ gsf_output_target_write( GsfOutput *output,
 static gboolean
 gsf_output_target_seek( GsfOutput *output, gsf_off_t offset, GSeekType whence )
 {
+#ifdef HAVE_GSF_ZIP64
+	/* No seek needed.
+	 */
+	return( FALSE );
+#else
+	GsfOutputTarget *output_target = (GsfOutputTarget *) output;
+	int stdio_whence;
+
+	switch( whence ) {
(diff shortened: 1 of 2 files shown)


libvips/iofuncs/target.c
```diff
@@ -569,7 +569,7 @@ vips_target_read( VipsTarget *target, void *buffer, size_t length )
  * Seeking a target sounds weird, but libtiff needs this. This method will 
  * fail for targets like pipes.
  *
- * Returns: 0 on success, -1 on error.
+ * Returns: the new seek position, -1 on error.
```
jcupitt, line 572: Ooop!

libvips/foreign/dzsave.c
```diff
@@ -248,9 +248,30 @@ gsf_output_target_write( GsfOutput *output,
...
+	 */
+	return( FALSE );
+#else
+	GsfOutputTarget *output_target = (GsfOutputTarget *) output;
+	int stdio_whence;
+
+	switch( whence ) {
+	case G_SEEK_CUR: stdio_whence = SEEK_CUR; break;
+	case G_SEEK_END: stdio_whence = SEEK_END; break;
+	case G_SEEK_SET: stdio_whence = SEEK_SET; break;
```
jcupitt, line 262: I looked up `G_SEEK_SET` and they don't use the standard values! I agree this switch is crazy, but sadly necessary :(

PR #2929, opened 2022-07-15
pngsave: Change default row filter from 'all' to 'none'

lovell:
As discussed in libvips/devchat.

Preventing the use of adaptive row filtering is usually faster and produces a smaller file size, especially as PNG is typically used for diagrams and line art.

This has been the default behaviour of sharp since early 2018 - see https://github.com/lovell/sharp/issues/872

This will help alleviate https://github.com/libvips/libvips/discussions/2900

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -30,6 +30,7 @@
 - add support for ICC profiles and linear encoding to JXL load and save [f1ac]
 - add "reoptimise" to gifsave [dloebl]
 - add "bitdepth" to magicksave [dloebl]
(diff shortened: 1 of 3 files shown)


jcupitt:
I tried a few tests and `none` is indeed better for line art and synthetic images. It's quite a bit worse for photos, but as you say that's not the usual use of PNG. Let's do it!

jcupitt:
Yes, I saw about a 2x speedup as well. Thanks for testing!

PR #2934, opened 2022-07-20
Link the main library with `-Wl,-z,nodelete` to prevent unloading

kleisauke:
See: https://github.com/libvips/php-vips-ext/pull/44

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -457,6 +457,20 @@ else
   fi
 fi
 
+SAVE_LDFLAGS="$LDFLAGS"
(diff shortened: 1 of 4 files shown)


lovell:
The musl libc already (correctly) considers `dlclose` a no-op, so this change will bring glibc-based flavours of Linux up to par. This should also remove a problem in sharp for people using it with Node.js "worker threads", which uses reference counting before calling `dlclose` but can end up doing so too early on a child thread rather than the parent thread.

PR #2940, opened 2022-07-23
Simplify/modernize `vips_crop_get_type()`

kleisauke:


jcupitt:
Oh nice. That new once system is neater.

PR #2948, opened 2022-07-27
Meson: bump version to 8.13.1

kleisauke:
Also, increment revision in the Autotools build.


jcupitt:
Ooof, sorry, I got distracted half-way through and forgot meson.build heh.

PR #2955, opened 2022-07-28
Fix libpng fall-back when build with `-Dspng=disabled`

kleisauke:
> **Note**: this PR targets the [`master`](https://github.com/libvips/libvips/tree/master) branch, since it also alters the summary table (which is not present in the  [`8.13`](https://github.com/libvips/libvips/tree/8.13) branch).


kleisauke:
Should I open a separate PR that brings that summary table and `vips --vips-config` enhancements to 8.13?

jcupitt:
Good idea, I think not having the summary table in 8.13 is going to become annoying quite quickly.

PR #2966, opened 2022-07-30
Allow to modify/create EXIF GPS* tags

kleisauke:
Make `tag_is_ascii` aware of the EXIF GPS* ASCII tags that are
available since libexif 0.6.23.

See: https://github.com/lovell/sharp/issues/2767

> **Note**: this PR targets the [`master`](https://github.com/libvips/libvips/tree/master) branch. I'm not sure if this is also necessary in 8.13.x.


libvips/foreign/exif.c
```diff
@@ -788,6 +788,20 @@ tag_is_ascii( ExifTag tag )
 		|| tag == EXIF_TAG_OFFSET_TIME
 		|| tag == EXIF_TAG_OFFSET_TIME_ORIGINAL
 		|| tag == EXIF_TAG_OFFSET_TIME_DIGITIZED
+		|| tag == EXIF_TAG_GPS_LATITUDE_REF
```
kleisauke, line 791: I noticed that this tag overlaps with the IFD Interoperability index tag (`EXIF_TAG_INTEROPERABILITY_INDEX`), which is also an ASCII tag, but with limited choices (i.e. an enumeration) - see:
https://github.com/libexif/libexif/blob/4a7ab4cc1332afbc63c44a8a96c93421d6aa07a0/libexif/exif-tag.h#L175

I'm unsure if this could cause issues.

jcupitt: I'm also unsure. Let's ignore this issue until someone complains.

jcupitt:
Useful! I think this can stay on master.

PR #2974, opened 2022-08-02
convf: bake the scale into the mask

kleisauke:
Similar to the vector path of `vips_convi`.


kleisauke:
Diff is best viewed with the "Hide whitespace" option, as I also fixed a few tab/space inconsistencies.

kleisauke:
PR https://github.com/google/oss-fuzz/pull/8122 should fix the CIFuzz and OSS-Fuzz integration.

Issue #2965, opened 2022-07-29
Vips jp2ksave creates grayscale JPEG2000 image from SRGB sources


felixbuenemann:
There's a bug in the JPEG2000 saver in Vips 8.13.0, that causes converted images to be grayscale, even though they have 3 bands.

Example:

```sh
vips jp2ksave sample.jpg sample.jp2
```

I tested this with both Vips 8.13.0 and OpenJPEG 2.5.0 on macOS/arm64 and Linux/aarch64, with the same result.

kleisauke:
I could not reproduce this using OpenJPEG 2.4.0 and libvips 8.13.0 on my Fedora PC. Are you using the macOS preview app? If so, this could be a duplicate of https://github.com/libvips/ruby-vips/issues/345. A handy way to check is to convert the JPEG2000 image back to PNG, e.g:
```bash
$ vips copy sample.jp2 x.png
```

felixbuenemann:
The JPEG2000 image shows as grayscale in any app on my Mac: Preview, Safari, Pixelmator.

You are right, converting the file back to jpg gives the correct colors.

But if I convert the file back using ImageMagick, I get completely wrong colors. Looking at the channels the red channel looks ok, but green and blue have wrong brightness and look inverted.

However if I compress from the same source image using `opj_compress` from OpenJPEG the problem does not occur.
Also converting that JPEG2000 back to JPEG using ImageMagick works fine.

So something appears to be wrong with the file.

jcupitt:
Could you try an image with an odd number of columns?

felixbuenemann:
@jcupitt Yes, the bug doesn't happen with odd widths. Eg. 1023x1023px buggy, 1024x1024px OK.

felixbuenemann:
Here's an example what it looks like after converting back from JPEG200 to JPEG using ImageMagick:

Source 1023x1023px:
![Earth1023im](https://user-images.githubusercontent.com/909587/181789168-0eaf2c3d-427c-401e-b541-3ee78160bc69.jpg)

Source 1024x1024px:
![Earth1024im](https://user-images.githubusercontent.com/909587/181789225-03189570-55a3-4415-ae7a-ef45bfd5e2c5.jpg)

felixbuenemann:
I noticed that converting the even width JPEG200 to PNG using `opj_decompress` from the OpenJPEG CLI tools also works fine.

felixbuenemann:
[...]

PR #2975, opened 2022-08-02
Remove even pixel check for J2K subsample_mode auto

felixbuenemann:
Having a different colorspace and chroma subsampling just because the source image happens to be an even or odd number of pixels is surprising and hard to debug, so let's remove it.

See https://github.com/libvips/libvips/issues/2965#issuecomment-1201354156 for discussion.

### DIFF
--- a/libvips/foreign/jp2ksave.c
+++ b/libvips/foreign/jp2ksave.c
@@ -795,8 +795,6 @@ vips_foreign_save_jp2k_build( VipsObject *object )
 		jp2k->subsample =
 			!jp2k->lossless &&
 			jp2k->Q < 90 &&
-			save->ready->Xsize % 2 == 0 &&
-			save->ready->Ysize % 2 == 0 &&
 			(save->ready->Type == VIPS_INTERPRETATION_sRGB ||
 			 save->ready->Type == VIPS_INTERPRETATION_RGB16) &&
 			save->ready->Bands == 3;


jcupitt:
I think we should also get rid of the libvips jp2k YCC conversion and just use mct (if I understand your discoveries correctly). And mct should be enabled iff subsample is enabled, ie. for RGB-style images with three or more bands.

Should we make a further PR for these changes, or do it as part of this PR?

felixbuenemann:
@jcupitt I think you misunderstood my findings. MCT only does RGB->YCC, but no chroma subsampling, so it doesn't make sense to use the subsample_mode to trigger MCT RGB to YCC. Also if we do subsampling nad RGB->YCC ourselves, we cannot use MCT since it requires RGB source data.

jcupitt:
But we can use mct instead of implementing our own RGB<->YCC conversion, can't we? Aren't they functionally equivalent? Or that was my understanding.

Yes, we still need to implement subsample/upsample, but we should enable `mct` for all the same cases we enable subsampling, ie. RGB-like images with three or more bands. 

Although thinking about it again, we will want `mct` without subsample for Q > 90, so I was wrong there.

jcupitt:
Oh heh I tried removing the ycc code and it's worse:

```
$ vips copy k2.jpg x.jp2[subsample-mode=on]
$ ls -l x.jp2
-rw-r--r-- 1 john john 570288 Aug  2 18:23 x.jp2
$ vips copy k2.jpg x.jp2[subsample-mode=off]
$ ls -l x.jp2
-rw-r--r-- 1 john john 435075 Aug  2 18:23 x.jp2
```

Looking with `opj_dump` it seems that openjpeg will disable `mct` if the image planes are not all the same size (ie. subsampling is enabled). You probably said this and I missed it.

So I suppose we need to keep our own RGB->YCC code for use with subsampling, and in this case save as `OPJ_CLRSPC_YCC`. Though it seems very few decoders support YCC JP2, so perhaps it's not worth it.

felixbuenemann:
Yes, because we do YCbCr 4:2:0 with subsample_mode on, but without subsample_mode off and mct 1 OpenJPEG does YCbCr 4:4:4, so of course file size is larger.

I can't judge how widely supported "native" (mct 0) JPEG2000 with YCbCr colourspace is, so I'm not sure, if we should keep it or not.

jcupitt:
Sorry, I was pulled on to something else.

OK, let's merge this and then adjust the CLRSPC in a separate commit.

PR #2991, opened 2022-08-15
Updated README.md to add Elixir language bindings (Vix)

akash-akya:
Hi there, 

I'm working on [Elixir bindings](https://github.com/akash-akya/vix) for libvips for some time now.
I think now it is good enough for people to try out. If possible, please take a look. Any type of feedback is most welcome. 

Thanks to the libvips design, especially immutability aspect of it. It is a perfect match for Elixir/Erlang.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ Full bindings are available for :
 | Go | [govips](https://github.com/davidbyttow/govips) |
 | Lua | [lua-vips](https://github.com/libvips/lua-vips) |
 | Crystal | [crystal-vips](https://github.com/naqvis/crystal-vips) |
+| Elixir | [vix](https://github.com/akash-akya/vix) |
 
 libvips is used as an image processing engine by:
 


jcupitt:
Oh, that looks great.

A tiny point -- `thumbnail_image` is supposed to be for emergencies only, it's MUCH better (eg. 10x faster for jpegs) to call `thumbnail` if you possibly can. Perhaps the example in README could use `thumbnail` instead?

I did my PhD in Miranda, the thing that turned into Haskell, so libvips is supposed to look like a functional programming language. I'm glad the design worked for you.

akash-akya:
I see, is it because of [shrink on-load optimization](https://github.com/libvips/libvips/blob/794d8d03a10f0d64ff30ace978af302875ed6e8c/libvips/resample/thumbnail.c#L1096-L1176) which is possible for the loaders like JPEG?

Thanks for pointing it out. I'll update the README to mention this 👍🏼

jcupitt:
Yes, I wrote a comment on Image a few months ago about this: https://github.com/kipcole9/image/issues/4#issue-1251596053

It has a benchmarks in python:

```
$ ./bench.py ~/pics/nina.jpg x.jpg
random access, using thumbnail_image:
took 398.28ms
sequential access, using thumbnail:
took 39.30ms
```

PR #2995, opened 2022-08-19
Mention mozjpeg as the best libjpeg option

atjn:
Just a quick suggestion to make it more clear how the `libjpeg` dependency works.

The most important change is recommending `mozjpeg`, because it is the package that will give you access to the most features and the best compression. Before, it seemed like `libjpeg-turbo` was the best option.

Another little detail I put in there, is just to make it clear that you can use any lib which supports the standard IJG JPEG library interface. As a new user, I was confused by the original wording, and thought I had to use the IJG library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,8 @@ so make sure that is working.
 
 ### libjpeg
 
-The IJG JPEG library. Use the `-turbo` version if you can. 
+Anything that is compatible with the IJG JPEG library. Use `mozjpeg` if you
+can. Another option is `libjpeg-turbo`. 
 
 ### libexif
 


jcupitt:
Hi, thanks for this. 

mozjpeg is *much* slower, so for some users libjpeg-turbo is a better choice. I think it depends on the use case.

atjn:
I see, I wasn't aware that it was that bad. But `mozjpeg` is only slower if you enable any of the settings that are specific to it, and those are all disabled by default right?

If I am right about that, I see no reason not to use `mozjpeg` and just not use the extra features if you want speed.

If I am wrong, I can rewrite the message to reflect that :)

jcupitt:
Ah maybe it's better now, I've not benchmarked it for a while, but mozjpeg used to be based on plain libjpeg, not libjpeg-turbo, so it was missing all the vector speedups. mozjpeg used to be maybe 4x slower, even when not using the extensions.

jcupitt:
I benchmarked again and it seems mozjpeg with no options is the same speed as libjpeg-turbo now -- they must have merged the vector enhancements.

https://github.com/jcupitt/docker-builds/tree/master/libvips-mozjpeg-ubuntu20.04

kleisauke:
Great! The last time I compared mozjpeg with libjpeg-turbo it produced byte-identical images, thanks to the `JCP_FASTEST` option (see e.g. https://github.com/lovell/sharp-libvips/pull/89). 

Ubuntu 20.04 ships with libjpeg-turbo 2.0.3, so a test with mozjpeg v4.0.0 would probably also produce identical images with the same performance.

PR #2996, opened 2022-08-19
nsgifload: choose tile-height based on page height

DarthSim:
The context
When we process animated images in imgproxy, we split them into frames with `vips_extract_area`. `vips_webpsave` basically does the same: https://github.com/libvips/libvips/blob/v8.13.0/libvips/foreign/vips2webp.c#L359-L361

The problem
In `vips_nsgifsave`, libvips sets `tile-height` of `vips_sequential` to 16. When the GIFs page height is not dividable by 16, this leads to decoding multiple pages when we try to cut out a single page. This happens because the first tile of the page often contains pixels of the previous page and the list tile contains pixels of the nest page. And this leads to the situation when nsgif has to start decoding from the start. If we turn on debug messages we'll see the following:

```
  nsgif_frame_decode(0) = 0
  nsgif_frame_decode(1) = 0
  nsgif_frame_decode(0) = 0
  nsgif_frame_decode(1) = 0
  nsgif_frame_decode(2) = 0
  nsgif_frame_decode(1) = 0
  nsgif_frame_decode(2) = 0
  nsgif_frame_decode(3) = 0
  nsgif_frame_decode(2) = 0
  nsgif_frame_decode(3) = 0
  nsgif_frame_decode(4) = 0
...
```

The solution
The solution is to choose `tile-height` that would be a divisor of GIFs page height so every tile would contain pixels of a single page.

The result
Here's a GIF provided by one of imgproxy users: https://bucketeer-e05bbc84-baa3-437e-9518-adb32be77984.s3.amazonaws.com/public/images/8863f61e-9fa7-452f-bd6e-153b2e2b9236_616x451.gif. This is a hell of a GIF that has 200+ frames, rare "keyframes" and the height that is not dividable by 16. I'll convert it to WebP to reveal the problem.

Before the fix:
```
$ /usr/bin/time -f %M:%e vips copy /images/animated_gif20.gif[n=-1] /images/animated_gif20_out.webp
71924:50.25
```

After the fix:
```
$ /usr/bin/time -f %M:%e vips copy /images/animated_gif20.gif[n=-1] /images/animated_gif20_out.webp
70984:7.63
```

Loading is 6+ times faster and memory consumption is a bit lower! I also see an average 2+ speed up on the other GIFs that height is not divisible by 16.

### DIFF
--- a/libvips/foreign/nsgifload.c
+++ b/libvips/foreign/nsgifload.c
@@ -487,6 +487,26 @@ vips_foreign_load_nsgif_generate( VipsRegion *or,
 	return( 0 );
 }
 
+int
+vips_foreign_load_nsgif_tile_height(VipsForeignLoadNsgif *gif) {
+	int i;
+	int height = gif->info->height;
+
+	// First, check the perfect size
+	if (height % 16 == 0)
+		return 16;
+
+	// Next, check larger and smaller sizes
+	for (i = 1; i < 16; i++) {
+		if (height % (16 + i) == 0)
+			return (16 + i);
+		if (height % (16 - i) == 0)
+			return (16 - i);
+	}
+
+	return 1;
+}
+
 static int
 vips_foreign_load_nsgif_load( VipsForeignLoad *load )
 {
@@ -507,7 +527,7 @@ vips_foreign_load_nsgif_load( VipsForeignLoad *load )
 	if( vips_image_generate( t[0],
 		NULL, vips_foreign_load_nsgif_generate, NULL, gif, NULL ) ||
 		vips_sequential( t[0], &t[1],
-			"tile_height", VIPS__FATSTRIP_HEIGHT,
+			"tile_height", vips_foreign_load_nsgif_tile_height(gif),
 			NULL ) ||
 		vips_image_write( t[1], load->real ) )
 		return( -1 );


jcupitt:
Hi @DarthSim,

I think the problem here is the way webpsave is coded. It shouldn't loop with a crop, that's very slow, instead it should use `vips_sink_disc()`, just like `cgifsave`, `tiffsave`, etc.

For example:

```
$ time vips copy animated_gif20.gif[n=-1] x.webp

real	0m43.090s
user	0m43.973s
sys	0m0.793s

$ time vips copy animated_gif20.gif[n=-1] x.gif

real	0m4.166s
user	0m8.516s
sys	0m0.167s
```

That time difference is just caused by the bad webp save loop, I think. Because of the loop structure, progress reporting for webpsave is broken too :(

jcupitt:
... so I think a better fix would be to rework webpsave.

DarthSim:
Well, yes, that’s what I was going to do next, but the problem with splitting GIF into frames will remain a problem, current webpsave implementation just illustrates it. I’ll apply this changes to libvips in the imgproxy distribution anyway, but I was thinking that this would be useful to other libvips users. At least I don’t see any drawbacks.

DarthSim:
That was my first idea :) But then I thought like "Well, John set it to 16 for a reason" and wrote that little function that picks `tile-weight`. Setting `tile-height` equal to the page height gives us pretty much the same speed (which is expectable) but leads to larger memory consumption (which I don't fully understand):

```
$ /usr/bin/time -f %M:%e vips copy /images/animated_gif20.gif[n=-1] /images/animated_gif20_out.webp
77048:7.66
```

`tile-height` of 1 gives us larger memory consumption AND lower speed:

```
$ /usr/bin/time -f %M:%e vips copy /images/animated_gif20.gif[n=-1] /images/animated_gif20_out.webp
74244:8.01
```

So it looks to me like 16 is a kind of golden mean, and `vips_foreign_load_nsgif_tile_height` tries to pick something close to it.

jcupitt:
Hi again, let's merge Daniel's fixed up webpsave, then merge this. See https://github.com/libvips/libvips/pull/3018

jcupitt:
Phew so we have the new webpsave merged now and I tried your example again.

Git master libvips:

```
$ /usr/bin/time -f %M:%e vips copy animated_gif20.gif[n=-1] animated_gif20.webp
memory: high-water mark 16.25 MB
94044:5.77
```

Your patch:

```
$ /usr/bin/time -f %M:%e vips copy animated_gif20.gif[n=-1] animated_gif20.webp
memory: high-water mark 16.23 MB
93848:5.78
```

I tried setting `tile_height` to `gif->info->height` as well:

```
 $ /usr/bin/time -f %M:%e vips copy animated_gif20.gif[n=-1] animated_gif20.webp
memory: high-water mark 15.53 MB
91316:5.78
```

So I think there are some wins here, but probably not too significant. What do you think?

DarthSim:
Since `webpsave` is fixed and doesn't demonstrate the problem anymore, I wrote a little program that does: https://gist.github.com/DarthSim/671d1ee8a2f03b5b9077bed494842123. It's pretty simple: load a GIF, split it into frames, join frames back together, and save the new GIF. I intensionally threw WebP from the stage so it won't confuse us anymore.

(The program is not memory-optimized at all, it doesn't clean its stuff properly, but it's just a test and it does its best to demonstrate the problem. So let's be indulgent to it :) )

Here are my results:

Git master libvips (the worst speed, medium memory footprint):

```
$ /usr/bin/time -f %M:%e .tmp/vipsgiftest /images/animated_gif20.gif /images/animated_gif20_out.gif
295652:48.64
```

My patch (almost the best speed, the best memory footprint):

```
$ /usr/bin/time -f %M:%e .tmp/vipsgiftest /images/animated_gif20.gif /images/animated_gif20_out.gif
294020:5.97
```

`gif->info->height` (speed is a bit better than with my patch, the worst memory footprint):

```
$ /usr/bin/time -f %M:%e .tmp/vipsgiftest /images/animated_gif20.gif /images/animated_gif20_out.gif
299224:5.87
```

jcupitt:
Ah you're right. I messed up my tests :( I see a good improvement here too.

I agree, let's merge. Thanks for pushing on this.

Issue #2722, opened 2022-03-16
improve animated webp write


jcupitt:
Animated webp images are written with a series of crops and saves. Instead, they should use `sink_disc` and stream the image to the output in a single iteration.

See https://github.com/lovell/sharp/issues/3068

adityapatadia:
Hey,
Can you point us to files which has these specific functions and how can we solve this? Not too good with CPP but will try to see if we can raise a PR. This issue of not having timeout is causing production problems.

jcupitt:
It's a little complex, unfortunately. You'd need to rework the webp writer using something like the gif writer as a model. It should use the sink API and accumulate pages in a temporary buffer.

https://github.com/libvips/libvips/blob/master/libvips/foreign/cgifsave.c

jcupitt:
Let's switch discussion to the PR.

PR #3018, opened 2022-08-31
webpsave: switch to sink_disc

dloebl:
Switch to `sink_disc` instead of cropping frame by frame in `webpsave` (based on `cgifsave`).
Closes https://github.com/libvips/libvips/issues/2722.

The first results look promising: Copying animated GIFs to WebP is quite a bit faster.

**Benchmark (76 GIFs; overall time):**
| **tool** | **version** | **time** |
| - | - | - | 
| libvips | v8.13.0 | 306 seconds |
| libvips | [webpsave-sink-disc](https://github.com/dloebl/libvips/tree/webpsave-sink-disc) | 92 seconds |
| gif2webp | v1.2.4 | 90 seconds |

**Settings:** _Q=85, effort=4, kmin=0, kmax=0_

`webpsave` is about on a par with [gif2webp](https://developers.google.com/speed/webp/docs/gif2webp) (standalone tool for converting GIFs to WebP) now.


jcupitt:
Thanks for working on this Daniel! Give me a ping when you're done and I'll read it.

dloebl:
Ready for review @jcupitt.

**Some notes:**
- References to `webpsave_mime` were missing in some places (I added them with 0dba7ab).
- I didn't change the way static WebP write works, because memory usage was quite a bit higher for static images if the same `sink_disc` path as for animated write was in use.
- Not sure if this PR improves timeout handling for animated WebP write.

libvips/foreign/webpsave.c
```diff
@@ -109,15 +155,639 @@ typedef struct _VipsForeignSaveWebp {
...
+	/* We need the frame as a contiguous RGB(A) buffer for libwebp.
+	 */
+	for( y = 0; y < frame_rect->height; y++ )
+		memcpy( webp->frame_bytes + y * webp->image->Bands * frame_rect->width,
+			VIPS_REGION_ADDR( webp->frame, 0, frame_rect->top + y ),
+			webp->image->Bands * frame_rect->width );
+
+	if( write_webp_image( webp, webp->frame_bytes, &pic ) ) {
+		return( -1 );
+	}
```
jcupitt, line 285: No `{}` for single-line blocks.

libvips/foreign/webpsave.c
```diff
@@ -109,15 +155,639 @@ typedef struct _VipsForeignSaveWebp {
...
+			"%s", _( "anim add error" ) );
+		return( -1 );
+	}
+
+	/* Adjust current timestamp
+	 */
+	if( webp->delay &&
+		page_index < webp->delay_length )
+		webp->timestamp_ms += webp->delay[page_index] <= 10 ?
+			100 : webp->delay[page_index];
```
jcupitt, line 300: Ooof that's a bit ugly. How about moving this to be next to the handling of small `gif_delay`? So have a small loop at line 635 squashing out low values.

libvips/foreign/webpsave.c
```diff
@@ -109,15 +155,639 @@ typedef struct _VipsForeignSaveWebp {
...
+
+	/* New images have an array of ints instead.
+	*/
+	webp->delay = NULL;
+	if( vips_image_get_typeof( webp->image, "delay" ) &&
+		vips_image_get_array_int( webp->image, "delay",
+			&webp->delay, &webp->delay_length ) )
+		return( -1 );
+
+	webp->timestamp_ms = 0;
```
jcupitt, line 635: Classes are inited to zero, so you can remove this.

libvips/foreign/webpsave.c
```diff
@@ -109,15 +155,639 @@ typedef struct _VipsForeignSaveWebp {
...
+		/* Init config for animated write (if necessary)
+		 */
+		if( vips_foreign_save_webp_init_anim_enc( webp ) )
+			return( -1 );
+
+		/* Write animated (via sink_disc)
+		 */
+		if( vips_foreign_save_webp_write_anim( webp ) ) {
+			return( -1 );
+		}
```
jcupitt, line 759: No `{}` here.

libvips/foreign/webpsave.c
```diff
@@ -109,15 +155,639 @@ typedef struct _VipsForeignSaveWebp {
...
+vips_foreign_save_webp_write_anim( VipsForeignSaveWebP *webp )
+{
+	if( vips_sink_disc( webp->image,
+		vips_foreign_save_webp_sink_disc, webp ) )
+		return( -1 );
+	return( 0 );
+}
+
+static int
+vips_foreign_save_webp_write_single( VipsForeignSaveWebP *webp )
```
jcupitt, line 683: Let's not have two paths for write -- this is going to add complexity and make testing harder. 

If there's a significant difference in memory use, we should find out why and fix that.

libvips/foreign/webpsave.c
```diff
@@ -109,15 +155,639 @@ typedef struct _VipsForeignSaveWebp {
...
 
-typedef VipsForeignSaveClass VipsForeignSaveWebpClass;
+typedef VipsForeignSaveClass VipsForeignSaveWebPClass;
 
-G_DEFINE_ABSTRACT_TYPE( VipsForeignSaveWebp, vips_foreign_save_webp, 
+G_DEFINE_ABSTRACT_TYPE( VipsForeignSaveWebP, vips_foreign_save_webp,
 	VIPS_TYPE_FOREIGN_SAVE );
 
+static void
+vips_webp_write_unset( VipsForeignSaveWebP *write )
```
jcupitt, line 193: This is a gobject method, so it ought to be `vips_foreign_save_webp_unset` (yuk).

libvips/foreign/webpsave.c
```diff
@@ -109,15 +155,639 @@ typedef struct _VipsForeignSaveWebp {
...
+	VIPS_UNREF( webp->target );
+
+	VIPS_FREE( webp->frame_bytes );
+
+	G_OBJECT_CLASS( vips_foreign_save_webp_parent_class )->
+		dispose( gobject );
+}
+
+static gboolean
+vips_webp_pic_init( VipsForeignSaveWebP *write, WebPPicture *pic )
```
jcupitt, line 217: Change to `vips_foreign_save_webp_pic_init()`.

libvips/foreign/webpsave.c
```diff
@@ -109,15 +155,639 @@ typedef struct _VipsForeignSaveWebp {
...
+		write->near_lossless ||
+		write->smart_subsample;
+
+	return( TRUE );
+}
+
+/* Write a VipsImage into an unintialised pic.
+ */
+static int
+write_webp_image( VipsForeignSaveWebP *write, const VipsPel *imagedata, WebPPicture *pic )
```
jcupitt, line 239: `vips_foreign_save_webp_write_webp_image()` maybe?

The line needs to fit in 80 columns, so maybe wrap before `const VipsPel ..`.

libvips/foreign/webpsave.c
```diff
@@ -109,15 +155,639 @@ typedef struct _VipsForeignSaveWebp {
...
+	WebPPicture pic;
+	int y;
+	VipsObjectClass *class = VIPS_OBJECT_GET_CLASS( webp );
+	VipsRect *frame_rect = &webp->frame->valid;
+	int page_index = frame_rect->top / frame_rect->height;
+
+	/* We need the frame as a contiguous RGB(A) buffer for libwebp.
+	 */
+	for( y = 0; y < frame_rect->height; y++ )
+		memcpy( webp->frame_bytes + y * webp->image->Bands * frame_rect->width,
```
jcupitt, line 279: This needs to fit in 80 columns. Perhaps add a pointer that steps down the output to get rid of this huge expression?

libvips/foreign/webpsave.c
```diff
@@ -109,15 +155,639 @@ typedef struct _VipsForeignSaveWebp {
...
+			"%s", _( "config version error" ) );
+		return( -1 );
+	}
+
+	/* These presets are only for lossy compression. There seems to be
+	 * separate API for lossless or near-lossless, see
+	 * WebPConfigLosslessPreset().
+	 */
+	if( !(webp->lossless || webp->near_lossless) &&
+		!WebPConfigPreset( &webp->config, get_preset( webp->preset ), webp->Q ) ) {
```
jcupitt, line 538: This needs to fit in 80 cols. Maybe break after `config,`?

libvips/foreign/webpsave.c
```diff
@@ -109,15 +155,639 @@ typedef struct _VipsForeignSaveWebp {
...
+	        return( -1 );
+	}
+
+	/* Get delay array
+	 *
+	 * There might just be the old gif-delay field. This is centiseconds.
+	 */
+	webp->gif_delay = 10;
+	if( vips_image_get_typeof( webp->image, "gif-delay" ) &&
+		vips_image_get_int( webp->image, "gif-delay", &webp->gif_delay ) )
```
jcupitt, line 617: This is more than 80 columns. Maybe break after `image,`.

libvips/foreign/webpsave.c
```diff
@@ -109,15 +155,639 @@ typedef struct _VipsForeignSaveWebp {
...
+		return( -1 );
+	}
+	page_height = vips_image_get_page_height( webp->image );
+
+	/* Init generic WebP config
+	 *  (needed for both animated and single image write)
+	 */
+	if( vips_foreign_save_webp_init_config( webp ) ) {
+		return( -1 );
+	}
```
jcupitt, line 739: No `{}` here.

libvips/foreign/webpsave.c
```diff
@@ -109,15 +155,639 @@ typedef struct _VipsForeignSaveWebp {
...
+		 */
+		if( vips_foreign_save_webp_write_anim( webp ) ) {
+			return( -1 );
+		}
+
+		/* Finish animated write
+		 */
+		if( vips_foreign_save_webp_finish_anim( webp ) )
+			return( -1 );
+	} else {
```
jcupitt, line 765: Break before `else`.

jcupitt:
Nice! It seems to work well.

I realize a lot of the stuff I commented on was things you inherited from the old `vips2webp.c`, but let's fix that too while doing this.

The history is that back in the vips7 days, all the loaders and savers were structured rather like `vips2webp.c`. During the vips8 rewrite, we adopted the more modern gobject style used for eg. `cgifsave.c`, but we had to keep some of the old `vips2xxx` savers for the vips7 compat layer. This is why webp save was split into two parts, with `vips2webp` in a slightly different style.

You've moved the vips2webp code into the save class, which is great, but it means various old naming conventions need to be updated.

libvips/foreign/webpsave.c
dloebl: Good point. Let's have a look what's wrong. I reverted the sink_disc path for animated and single image write with b2d263f.

dloebl:
Reverted to the same sink_disc path for animated and single image write. I see the following times/max memory usage:

**current master:**
```
gtime -f%M:%e vips copy rocket.jpg old.webp
186640:1.35

gtime -f%M:%e vips copy demo.gif"[n=-1]" old_gif.webp
41552:4.49
```

**webpsave-sink-disc:**
```
gtime -f%M:%e vips copy rocket.jpg new.webp
277136:1.36

gtime -f%M:%e vips copy demo.gif"[n=-1]" new_gif.webp
39744:0.75
```

The resulting WebPs are byte-identical compared to current master.

[rocket.jpg](https://user-images.githubusercontent.com/34104349/188421806-e5803cfd-6397-481f-a5e9-9dfcb94cae09.jpg)
[demo.gif](https://user-images.githubusercontent.com/34104349/188422159-bf512e63-285c-4d84-b302-5a2ce3d1f3cf.gif)

dloebl:
It seems that one part of the memory usage regression is that we have aggregated RGB(A) image data in two buffers:
In `frame_bytes` and the `frame` VipsRegion.

@jcupitt I added a commit which removes `frame_bytes` and instead gets the RGB(A) data for libwebp via 
 `VIPS_REGION_ADDR_TOPLEFT()` directly from the VipsRegion:
```
gtime -f%M:%e vips copy rocket.jpg new.webp          
198976:1.36

gtime -f%M:%e vips copy demo.gif"[n=-1]" new_gif.webp
38896:0.75
```

From my understanding on how we call [vips_region_buffer()](https://www.libvips.org/API/current/VipsRegion.html#vips-region-buffer), the underlying buffer should be guaranteed to be contiguous.

jcupitt:
Unfortunately, the region isn't guaranteed to be contiguous :( We really do need to copy to a buffer before calling into libwebp :( If libwebp had a `stride` option on its pixel buffers we'd be OK, but sadly it doesn't.

jcupitt:
Background: operators are allowed to modify the region they are asked to write to. This can save a LOT of copies, for example, `crop` doesn't need to copy pixels into the region it is passed, instead it can change the region to point to the input image of the crop. This way, `crop` turns into just a pointer copy.

Lots of operators do this. For example, `insert` will check the region it has to fill against the geometry of the two images it is combining, and if the region will fit entirely inside either of the inputs, it'll just forward the pixel request. `insert` will only copy and compute pixels if the area that is being calculated falls on an overlap.

This saves copies all the way down the pipeline, but the problem comes at the end, where the pixel sink can't be sure it has contiguous memory. We have to put a copy in at this stage, just in case.

dloebl:
> Unfortunately, the region isn't guaranteed to be contiguous

Ok I see, let's revert 84fc10e again.
Thanks for your detailed explanation on this! 👍

dloebl:
I had another thought: Maybe we could change the way `vips_foreign_save_webp_sink_disc()` works and accumulate the the contiguous image data directly in `frame_bytes`?

With 8e9cdb6, I see:

```
gtime -f%M:%e vips copy rocket.jpg out.webp
196752:1.36

gtime -f%M:%e vips copy demo.gif"[n=-1]" out.webp
35552:0.75
```

jcupitt:
Oh dear, I'm so sorry, I misread your post up there ^^ I hadn't read your patch carefully and thought you were talking about the region coming in from `sink_disc`.

Yes, the memory in `frame` was guaranteed to be contiguous. I think you should flip it back to `frame` (so no `frame_bytes` member), and use `vips_region_copy()` rather than your own `memcpy()` loop. The code in the cgif saver is very similar:

https://github.com/libvips/libvips/blob/master/libvips/foreign/cgifsave.c#L623-L685

Though I think that would be better as a `while() {}` now I look at it again.

jcupitt:
Oooof, sorry, now I read some more I think you're right, copying directly to `frame_bytes` is a good idea. Argh I need to drink more coffee before posting in technical discussions.

libvips/foreign/webpsave.c
```diff
@@ -49,16 +71,41 @@
...
-#ifdef HAVE_LIBWEBP
+#include <webp/encode.h>
+#include <webp/types.h>
+#include <webp/mux.h>
+
+typedef int (*webp_import)( WebPPicture *picture,
+        const uint8_t *rgb, int stride );
 
-typedef struct _VipsForeignSaveWebp {
+typedef enum _VipsForeignSaveWebPMode {
```
jcupitt, line 92: Sorry, this is such a tiny thing, but libvips is using the name `Webp`, not `WebP`, and we can't change it without breaking source compatibility, or introducing a set of compat macros.

Let's have `Webp` for all libvips names, and leave `WebP` for the WebP API calls.

libvips/foreign/webpsave.c
```diff
@@ -109,15 +161,554 @@ typedef struct _VipsForeignSaveWebp {
...
+	WebPAnimEncoder *enc;
+
+	/* Add metadata with this.
+	 */
+	WebPMux *mux;
+
+	/* The current y position in the frame and the current page index.
+	 */
+	int write_y;
+	int page_index;
```
jcupitt, line 184: I think `page_number` would probably be better.

libvips/foreign/webpsave.c
```diff
@@ -109,15 +161,554 @@ typedef struct _VipsForeignSaveWebp {
...
+
+	/* VipsRegion is not always contiguious, but we need contiguous RGB(A)
+	 * for libwebp. We need to copy each frame to a local buffer.
```
jcupitt, line 187: This is my fault -- that should be `contiguous`.

libvips/foreign/webpsave.c
```diff
@@ -109,15 +161,554 @@ typedef struct _VipsForeignSaveWebp {
...
+		/* If we've filled the frame, write and move it down.
+		 */
+		if( webp->write_y == page_height ) {
+			if( vips_foreign_save_webp_write_frame( webp ) )
+				return( -1 );
+
+			webp->write_y = 0;
+			webp->page_index += 1;
+		}
+	}
```
jcupitt, line 344: This is so much better than the way I had the loop arranged, nice job!

jcupitt (commented):
Just a few renames and some tiny layout things. I ran the tests and valgrind and it all passed. Nice work Daniel.

jcupitt:
I copied your nice loop idea for `cgifsave.c`. The memory saving is not very significant, but it saves 60 lines of code.

dloebl:
> I copied your nice loop idea for cgifsave.c

Sounds good. Thanks! 👍
Ok, I revised `webpsave.c` based on your feedback.

adityapatadia:
Would it be possible to backport this to 8.13? It is an important change to use GIF well.

jcupitt:
8.14 is due in a month or two, so I think it'd be better to just wait for that. 8.13 is supposed to be stable, except for major bugfixes.

PR #3022, opened 2022-09-03
CI: remove redundant glib-utils dependency

kleisauke:
The GLib utilities, which was split into its own `glib-utils`
formula has been merged back into `glib`, see:
https://github.com/Homebrew/homebrew-core/pull/108307

This reverts commit a846de017d9f369ea3cdd1b0f5510bebbb71b5c7.


jcupitt:
+1!

PR #3031, opened 2022-09-08
pngsave: ensure 8-bit palette images can be created

lovell:
This PR ensures the logic matches the comment and, I think, intended behaviour. (It also fixes comment indentation.)

It targets the 8.13 branch as this appears to be a bug introduced in v8.13.1 via commit https://github.com/libvips/libvips/commit/198920398509465ab00e7c6f5584fd345bda20e7

Reported downstream at https://github.com/lovell/sharp/issues/3357


jcupitt (commented):
Ooop that's a bit of a brainfart, well done for finding this.

PR #3034, opened 2022-09-08
readme: meson build flags - fix small typo

lucor:


jcupitt:
Oooop! Thank you for fixing this!

PR #3047, opened 2022-09-12
Inverted order of TIFF tag writing when saving multi-page subifd pyramids

ruven:
This ensures that subifd pyramid layers are correctly tagged as reduced-image subfile type and not as pages when creating a multi-page image stack. With the previous tag writing order, the reduced-image subfile type was being over-written.

### DIFF
--- a/libvips/foreign/vips2tiff.c
+++ b/libvips/foreign/vips2tiff.c
@@ -822,7 +822,12 @@ wtiff_write_header( Wtiff *wtiff, Layer *layer )
 	else
 		TIFFSetField( tif, TIFFTAG_ROWSPERSTRIP, wtiff->tileh );
 
-	if( wtiff->toilet_roll ) {
+
+	if( layer->above ) 
+		/* Pyramid layer.
+		 */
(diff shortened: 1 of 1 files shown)


jcupitt:
Oh good point, that's dumb, thanks!

How about instead of swapping the blocks, changing it to be:

```C
        if( layer->above )
                /* Pyramid layer.
                 */
                TIFFSetField( tif, TIFFTAG_SUBFILETYPE, FILETYPE_REDUCEDIMAGE );
        else if( wtiff->toilet_roll ) {
                /* One page of many.
                 */
                TIFFSetField( tif, TIFFTAG_SUBFILETYPE, FILETYPE_PAGE );

                TIFFSetField( tif, TIFFTAG_PAGENUMBER,
                        wtiff->page_number, wtiff->n_pages );
        }
```

Now there's no overwriting of tags.

ruven:
Sure, an if-else indeed avoids tag overwriting

jcupitt:
\o/ thank you!

PR #3050, opened 2022-09-14
jxlsave: sync quality to distance calculation with libjxl

kleisauke:
Ensures the quality to distance conversion is continuous at 30.

See:
https://github.com/libjxl/libjxl/commit/ea5fa8074d0146ec0f4388f52990d428ea3b67c3

> **Note**: this PR targets the [`8.13`](https://github.com/libvips/libvips/tree/8.13) branch.


mo271:
This change does not all saving JPEG XL with a quality below 5. It only makes sure the quality->distance conversion is continuous at 30, which it wasn't before...

kleisauke:
@mo271 Thanks for the clarification! You're right, I assumed that commit was related to the (previously) linked issue, but it wasn't.

I've updated the commit and PR description accordingly.

jcupitt:
Thanks!

PR #3052, opened 2022-09-15
convi/reducev: initialize the sum with the addition

kleisauke:
Instead of adding to the sum during the clip, we could just intialize the sum with the addition. This avoids an Orc opcode, so I expect it to be slightly faster.

I also added a couple of comments to magic numbers in `vector.c`.


jcupitt:
Nice!

PR #3056, opened 2022-09-20
WIP: experiment with gi-docgen

jcupitt:
Minimal hacky attempt, it does sort-of work.

Sample output here:

http://www.rollthepotato.net/~john/Vips/index.html

Issues:

- [x] the markdown chapters need a bit of extra formatting to appear in the table of contents correctly
- [x] some docbook chapters need rewriting in markdown
- [x] gi-docgen will only document functions with GIR binding, is this OK?
- [ ] check that the various discussion sections have made it
- [ ] inline code samples in doc comments won't work (look for `|[ ]|` pairs)
- [ ] how will this work with the API section of the website? the thing that drops the pages into the template will need revising
- [x] the python that generates the function list needs to output markdown, not xml 
- [x] maybe we need to need to remove `images/` from the start of image references?
- [x] could rename the `gtk_doc` meson option as `docs`?
- [ ] we have a few docbook-isms in the markdown (search for `</` maybe?)
- [ ] gi-docgen doesn't automatically turn eg. "vips_add()" or "#VipsAdd" into a link ... it looks like this will be a lot of work to fix, see https://gnome.pages.gitlab.gnome.org/gi-docgen/linking.html ... the new syntax is `[func@add]` to generate a link to `vips_add()`.
- [ ] probably more stuff!


jcupitt:
.... this will be a huge amount of work, I'm not sure the improvement is worth it for us. 

Sample output here: http://www.rollthepotato.net/~john/Vips/

kleisauke:
I had a look at how many work this is in this changeset:
https://github.com/libvips/libvips/compare/master...kleisauke:libvips:gi-docgen-rebase

Sample output: https://kleisauke.nl/vips-gi-docgen/.

That branch moves all `SECTION:` to the relevant struct docs. The remaining ones needs to be moved to a separate file, similar to https://kleisauke.nl/vips-gi-docgen/libvips-header.html (see e.g. commit https://github.com/kleisauke/libvips/commit/749aeb0317b602d60cc36077a087954c2f7fac82). Also, many gtk-doc links and syntax still need to be converted to gi-docgen syntax.

jcupitt:
That's very cool! 

Was what your impression after doing this work -- do you think it's worth doing more on this branch?

kleisauke:
There's still a lot of work to do, maybe another weekend or two.

I'd like the HTML output it generates, so it may be worthwhile. FWIW, the reason for this is that searching online for libvips often leads to outdated documentation, like https://www.libvips.org/API/8.6/, which strangely lacks all its styling. :(

> * [ ]  how will this work with the API section of the website? the thing that drops the pages into the template will need revising

I'm unsure whether we should add our libvips header to these pages, it might be better to leave it as-is. We may also consider revising Doxygen to produce a standalone output.

jcupitt:
I agree, the result is quite a bit better than gtk-doc, and the search feature is nice.

Yes, pasting a header onto each page is a bit fragile :( 

I tried to merge all your work into this PR, I hope I didn't drop anything. It seems to build at least haha.

kleisauke:
> I tried to merge all your work into this PR, I hope I didn't drop anything. It seems to build at least haha.

I just forced pushed to prevent duplicated commits and to ensure commit 599ee6c22b65084f0fb2bc95c14936598069731a is also included.

jcupitt:
If we've decided to switch it over, how about merging this to master and fixing it in a series of follow-on PRs? Having broken documentation in master will motivate us to do the boring work.

kleisauke:
> If we've decided to switch it over, how about merging this to master and fixing it in a series of follow-on PRs? Having broken documentation in master will motivate us to do the boring work.

I agree; it's better to do this in batches rather than all at once. These changes are mostly mechanical changes, the only potential issue might be when `gi-docgen` is unavailable or when users depend on older options, as I renamed the  `doxygen` and `gtk_doc` options to `cpp-docs` and `docs`, respectively.

jcupitt:
OK, let's push the button!

To void treading on each other's feet, shall we divide the work up a bit? I'll make a PR for `iofuncs/`.

kleisauke:
:tada:

> To avoid treading on each other's feet, shall we divide the work up a bit? I'll start with a PR for `iofuncs/`.

Sure. I'll work on converting the remaining sections to separate pages.

PR #3057, opened 2022-09-21
text: simplify check for 32k pixel limit

kleisauke:
Cairo also does this check internally in `cairo_image_surface_create_for_data()`, see:
https://github.com/freedesktop/cairo/blob/2cfd08664d080ff22e357a5feeca9bc29ec1823c/src/cairo-image-surface.c#L532-L533

Test case:
```bash
$ vips text x.png $(printf '=%.0s' {1..9999})
text: invalid value (typically too big) for the size of the input (surface, pattern, etc.)
```

Context: https://github.com/kleisauke/net-vips/issues/178.

> **Note**: this PR targets the [`8.13`](https://github.com/libvips/libvips/tree/8.13) branch.

### DIFF
--- a/libvips/create/text.c
+++ b/libvips/create/text.c
@@ -348,6 +348,7 @@ vips_text_build( VipsObject *object )
 	VipsImage *image;
 	cairo_surface_t *surface;
 	cairo_t *cr;
+	cairo_status_t status;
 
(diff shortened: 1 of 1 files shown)


jcupitt:
Ah, that's neater. I merged to master as well.

PR #3058, opened 2022-09-22
SubIFD offsets should only be added if smaller pyramid layers actually exist

ruven:
For images that are smaller than the tile size, empty SubIFD offsets get added resulting in corrupted TIFF files


ruven:
When you try to create a tiled TIFF image pyramid using SubIFDs, vips creates a corrupted TIFF file with an empty SubIFD TIFFTAG if the image is too small and does not in fact have any pyramid layers to create.

jcupitt:
Oh, nice! I'll patch 8.13 too, thanks!

PR #3064, opened 2022-09-24
tiff2vips: avoid `g_assert_not_reached` in the default clause

kleisauke:
Since that could terminate the application in debug builds. Also, add the missing case clause for `PHOTOMETRIC_MINISWHITE`, see:
https://gitlab.com/libtiff/libtiff/-/blob/4c3ddd99b6efbdca9d0f2e91a53671edb0656fd7/libtiff/tif_jpeg.c#L1706-1731

Resolves: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=51013.


jcupitt:
Nice find.

PR #3068, opened 2022-09-28
spngsave: ensure ICC profiles have a name

lovell:
The PNG spec requires that the iCCP chunk has a profile name with a minumum length of 1 (we use the same "icc" name as pngsave).

http://www.libpng.org/pub/png/spec/1.2/PNG-Chunks.html#C.iCCP

https://github.com/libvips/libvips/blob/4176ab21061f425b4ee274f7f9c6634dacf5ca7c/libvips/foreign/vipspng.c#L1113-L1115

Targets the 8.13 branch as this is a bug fix.

Originally reported at https://github.com/lovell/sharp/issues/3390

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,7 @@
 - reduce latency on dzsave kill [kleisauke]
 - improve text too large check [kleisauke]
 - fix subifd writing for small images [ruven]
+- name ICC profiles in spngsave [lovell]
 
 24/7/22 started 8.13.1
 - fix im7 feature detection in meson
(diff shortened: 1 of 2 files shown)


jcupitt:
Huh what a strange spec. Good fix!

jcupitt:
... I'll merge to master as well.

Should we make 8.13.2? It's picked up quite a few useful fixes now.

lovell:
Thanks, and yes please to 8.13.2 when you're ready.

jcupitt:
OK, done!

PR #3074, opened 2022-10-02
spngsave: fix transparency

DarthSim:
Fixes https://github.com/lovell/sharp/issues/3395

LIQ may not sort the palette by alpha properly so the current way of filling `trns` may not work in some cases. This PR makes libvips fill `trns.type3_alpha` for every palette entry but update `trns.n_type3_entries` only when the palette entry has transparency. Thus we don't care if the palette is sorted by alpha or not.


kleisauke:
Since it's a bug fix, should this target the [`8.13`](https://github.com/libvips/libvips/tree/8.13) branch instead?

DarthSim:
Yup, my bad

jcupitt:
Thanks! I've added a quick note to changelog and also merged to master.

PR #3079, opened 2022-10-06
Introspection improvements

tintou:
Enhance the introspection coverage and increase Vala API coverage


jcupitt:
Looks useful, nice.

PR #3085, opened 2022-10-07
GObject Introspection: Image unskip constructors (cont. #3079)

vixalien:
Unskips constructors for the Image classes. This, for example, exposes the `from_source` constructor. Continues #3079.

However, generating a vapi with `vapigen Vips-8.0.gir --metadatadir ~/Projects/libvips/libvips/ --library vips` reports:

```
Vips-8.0.gir:5168.7-5170.21: warning: `Image' already contains a definition for `memory'
 5168 |       <constructor name="new_memory"
      |       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 5169 |                    c:identifier="vips_image_new_memory"
      | ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 5170 |                    introspectable="0">
      | ~~~~~~~~~~~~~~~~~~~~~                 
Vips-8.0.gir:4562.7-4562.54: note: previous definition of `memory' was here
 4562 |       <constructor name="memory" c:identifier="vips_image_memory">
      |       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~            
Vips-8.0.gir:23300.7-23300.54: warning: Creation method of abstract class cannot be public.
23300 |       <constructor name="new" c:identifier="vips_interpolate_new">
      |       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~            
Vips-8.0.gir:24309.7-24310.73: warning: Creation method of abstract class cannot be public.
24309 |       <constructor name="new_from_string"
      |       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
24310 |                    c:identifier="vips_object_new_from_string">
      | ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Vips-8.0.gir:25669.7-25669.50: warning: Creation method of abstract class cannot be public.
25669 |       <constructor name="new" c:identifier="vips_operation_new">
      |       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~              
```

So check this PR, maybe I made a mistake

### DIFF
--- a/libvips/Vips-8.0.metadata
+++ b/libvips/Vips-8.0.metadata
@@ -10,6 +10,7 @@ ArrayDouble.newv skip=false
 ArrayImage.newv skip=false
 ArrayInt.newv skip=false
 
+Image.*#constructor skip=fase
 Image.*#method skip=false
 Image.eval#virtual_method name="eval_impl"
 Image.invalidate#virtual_method name="invalidate_impl"


jcupitt:
Hi @vixalien,

I don't think goi should expose these varargs functions -- all the varargs stuff is just for human C programmers, not for language bindings. I think it's much better to reimplement them using the libvips gobject API.

Here's what the python binding does, for example:

https://github.com/libvips/pyvips/blob/master/pyvips/vimage.py#L636-L668

Or ruby:

https://github.com/libvips/ruby-vips/blob/master/lib/vips/image.rb#L425-L462

Or C++:

https://github.com/libvips/libvips/blob/master/cplusplus/VImage.cpp#L617-L637

vixalien:
Oh yeah. I understand. This means we have to create a whole set of vala files though, if I understand correctly.

jcupitt:
It shouldn't be too much code, but yes there would have to be some. There needs to be some anyway to support features like optional arguments and optional return values.

libvips is designed for run-time binding rather than compile time --- you're supposed to use GOI for just the core API, then look up and call the operators dynamically. There's a chapter in the docs:

https://www.libvips.org/API/current/binding.html

vixalien:
yea. this is understood.

tintou (approved):
In the very specific case of Vala (which transpiles code into C) this is correct. You might want to prefix with `vala:` instead of `GObject Introspection:` because you're only touching the vala binding here.

tintou:
To remove the warning, you can do something like `Image.new_memory name="new_from_memory"`

vixalien:
@jcupitt ideas?

tintou:
@vixalien the suggestions were for your change actually :sweat_smile:

vixalien:
I'm sorry but I don't fully understand what "varargs" are. I think they are just variable arguments but still dont fully get it. I think if they are easy to work with in C, they should be easy to work with Vala too as Vala just produces C code based on the ".vapi" file generated.

I could use a different "vapi" file (modified by a .metadata) and still get similar results similar to a vapi file generated by using the right™ way.

If Vala compiles directly to C, I don't think there's a reason one would create other bindings, as `Vips.add` would be as good as `image.add`. However, this view of mine I think is not correct because I don't understand what varargs are.

jcupitt:
Yes, just the `...` variable arguments in C.

They are not typechecked at all, even at runtime, so it's very easy to write code that crashes, or equally, very difficult to write code that you are sure can't crash. It'd be great if there was some way to avoid using them.

The C# binding (for example) uses parameters with default values:

https://github.com/kleisauke/net-vips/blob/master/src/NetVips/Image.Generated.cs#L157

The C++ binding does it as runtime with an option object:

https://github.com/libvips/libvips/blob/master/cplusplus/include/vips/VImage8.h#L2109

Would either of those work for vala? (sorry, I don't know much about it, I should read up on the syntax).

there's any way to avoid using them it'd be  (parameters with default values, perhaps?)

vixalien:
oh i think those would be unsafe for vala. vala is very javascript-like. you can read up it's syntax on the learnxinyminutes website in a short time.

I don't know what those `...` arguments are, but they sound unsafe.

vixalien:
I think someone with skills should create a vala binding.

jcupitt:
I'll have a very quick go at a proof of concept binding.

jcupitt:
I tried:

```
vapigen --library vips --metadatadir metadata ~/vips/share/gir-1.0/Vips-8.0.gir
```

With this metadata file:

```
new_from_string skip
print_summary_class skip
ArgumentClass skip
ArgumentClassMapFn skip
ClassMapFn skip
class_find skip
```

But I get:

```
Vips-8.0.metadata:1.1-1.20: warning: metadata never used
    1 | new_from_string skip
      | ^~~~~~~~~~~~~~~~~~~~
Vips-8.0.gir:24594.69-24594.69: error: The type name `ObjectClass' could not be found
24594 |             <type name="ObjectClass" c:type="VipsObjectClass*"/>
      |                                                                 
```

Line 24594 in `Vips-8.0.gir` is:

```
      <constructor name="new_from_string"
                   c:identifier="vips_object_new_from_string">
        <source-position filename="../libvips/include/vips/object.h"
                         line="648"/>
        <return-value transfer-ownership="full">
          <type name="Object" c:type="VipsObject*"/>
        </return-value>
        <parameters>
          <parameter name="object_class" transfer-ownership="none">
            <type name="ObjectClass" c:type="VipsObjectClass*"/>
^^^^^ this line
          </parameter>
          <parameter name="p" transfer-ownership="none">
            <type name="utf8" c:type="const char*"/>
          </parameter>
        </parameters>
      </constructor>
```

So it looks like it's not skipping `new_from_string` for some reason. I don't suppose anyone knows what I'm doing wrong? This is vapigen 0.56 on ubuntu 22.04.

vixalien:
Hey, I don't know exactly what's wrong, but I think the strings in metadata should be attached to an object like `Image` like so:

```
Image.new_from_string skip
```

jcupitt:
> but I think the strings in metadata should be attached to an object like Image 

That did it, thanks! I have a working vapi now, I'll see if I can make a micro vala binding.

vixalien:
wow. didn't know that would work. bon courage.

vixalien:
Do you have any progress @jcupitt. I would be interesting to see your .metadata file.

jcupitt:
I got distracted onto other projects (of course). I pushed what I did here:

https://github.com/jcupitt/vala-vips

jcupitt:
Oh heh I just realized that the new libvips meson build system will make a vapi for you. Try:

```
cd libvips
meson build --prefix=~/vips --libdir lib --buildtype release -Dvapi=true
cd build
ninja
ninja install
```

Then:

```
$ ls ~/vips/share/vala/vapi/
vips.deps  vips.vapi
```

I've not tested it yet. You'll need git master libvips.

vixalien:
oh let me try it.

vixalien:
i tested and it works. here is a pro tip by the way, to generate documentation for that vips.vapi (like the docs at https://valadoc.org), you can use valadoc like so:

```
valadoc vips.vapi -o docs --pkg=vips
```

and browse the plain html docs using your fav tool. python for ref:

```
python -m http.server --directory docs
```

vixalien:
other tip: you can use `-C` or `--ccode` to compile to C instead to binary. since i think you are familiar with C, I think this may be handy for you @jcupitt .

jcupitt:
Since setting `-Dvapi=true` seems to work, let's close this PR. I'll update here if I get anywhere with vala.

PR #3096, opened 2022-10-16
CI: various improvements

kleisauke:


lovell (approved):
LGTM

kmartinez:
I just realised its Ubuntu specific - so perhaps apt-get is now just apt but it may help people with older versions to install so... probably not worth a change

jcupitt:
:heart:

PR #3101, opened 2022-10-19
spngsave: Don't shift indexed pixels

DarthSim:
Hey!

The commit https://github.com/libvips/libvips/commit/ca2796aa7ff75c5bb5a74a1f3a4de9538bb9c45e have broken saving paletted PNGs with low bit-depth (< 8). This happens because palette indexes can't be larger than `1 << bitdepth`, so shifting makes all of them zeroes.

Fixes https://github.com/imgproxy/imgproxy/issues/1013


jcupitt:
D'oh! Thanks for fixing this dumb thing.

Issue #3099, opened 2022-10-19
Upgrading Libvips Meson build problem.


web-apply:
Hello, 
I am using debian 11.5 and  indtalled > apt install libvips42   last version ( vips-8.10.5  Thu Jan 21 18:55:16 UTC 2021).
I tried update to 8.13.2 with meson build but i am getting errors.   
How can I update packages that have remained in the old version more easily and automatically?
I installed meson ninja python 3 and all required packages.
When i run **ninja test**

```
Summary of Failures:

1/8 cli         FAIL             0.32s   exit status 1
2/8 formats     FAIL             0.35s   exit status 1
3/8 connections FAIL             0.19s   exit status 1

Ok:                 5   
Expected Fail:      0   
Fail:               3   
Unexpected Pass:    0   
Skipped:            0   
Timeout:            0  

```

**And here: meson setup build-dir --prefix=/aaa/bbb/ccc
I don't understand what is this: --prefix=/aaa/bbb/ccc**

```
==================================== 1/8 =====================================
test:         cli
start time:   07:13:37
duration:     0.32s
result:       exit status 1
command:      MALLOC_PERTURB_=123 /bin/sh /root/vips-813/test/test_cli.sh
----------------------------------- stdout -----------------------------------
testing with sample.jpg
testing nearest ... difference is 0.000000
----------------------------------- stderr -----------------------------------
memory: high-water mark 0 bytes
memory: high-water mark 1.46 MB
memory: high-water mark 692.03 KB
memory: high-water mark 1012.82 KB
memory: high-water mark 646.96 KB
memory: high-water mark 1.03 MB
memory: high-water mark 445.32 KB
memory: high-water mark 13.72 KB
/root/vips-813/test/test_cli.sh: 26: bc: not found
==============================================================================

==================================== 2/8 =====================================
test:         formats
start time:   07:13:37
duration:     0.35s
result:       exit status 1
command:      MALLOC_PERTURB_=104 /bin/sh /root/vips-813/test/test_formats.sh
----------------------------------- stdout -----------------------------------
testing sample.jpg v ... save / load difference is 0.000000
----------------------------------- stderr -----------------------------------
memory: high-water mark 707.54 KB
memory: high-water mark 1.10 MB
memory: high-water mark 1.06 MB
memory: high-water mark 293.48 KB
memory: high-water mark 912.03 KB
memory: high-water mark 220.13 KB
memory: high-water mark 1.03 MB
memory: high-water mark 448.75 KB
[...]

PR #3102, opened 2022-10-19
skip some skippable tests

eli-schwartz:
There are two types of tests that are implemented here:
- Some tests have an if block, and only run the entire test if some single prerequisite condition is available. This is a simple fix, because the tests are already skipped, but they don't say so. Make these `exit 77` so they are marked as having been skipped.
- Some tests use `bc`, and fail if that is not installed. Make this a bit nicer, and abort the test, marking it as skipped. Oftentimes, `bc` is not installed.

### DIFF
--- a/test/test_cli.sh
+++ b/test/test_cli.sh
@@ -7,32 +7,6 @@
 
 . ./variables.sh
 
-# is a difference beyond a threshold? return 0 (meaning all ok) or 1 (meaning
-# error, or outside threshold)
-# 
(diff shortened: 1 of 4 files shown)


eli-schwartz:
Ref. #3099

jcupitt:
Hi Eli, nice! Thanks for improving this.

These tests are a bit of a mess. They predate the python test suite and should perhaps be removed altogether.  I think we kept them around mostly to exercise the libvips CLI.

PR #3105, opened 2022-10-20
Revised threading system

jcupitt:
done:

- add rgb mode to openslideload
- new thread recycler
- threadpools size dynamically with load
- operations can hint threadpool size
- add VIPS_MAX_THREADS

todo:

- [x] doc comments
- [x] test wasm compatibility


libvips/iofuncs/threadset.c
```diff
@@ -0,0 +1,274 @@
...
+/*
+
+    These files are distributed with VIPS - http://www.vips.ecs.soton.ac.uk
+
+ */
+
+#ifdef HAVE_CONFIG_H
+#include <config.h>
+#endif /*HAVE_CONFIG_H*/
+#include <vips/intl.h>
```
kleisauke, line 36:
```suggestion
#include <glib/gi18n-lib.h>
```

libvips/iofuncs/threadset.c
```diff
@@ -0,0 +1,274 @@
...
+
+        member->kill = TRUE;
+        vips_semaphore_up( &member->idle );
+        g_thread_join( member->thread );
+
+	vips_semaphore_destroy( &member->idle );
+
+        g_mutex_lock( set->lock );
+        set->free = g_slist_remove( set->free, member );
+        set->n_threads += 1;
```
kleisauke, line 236: Since this kills a thread, perhaps it should subtract `->n_threads` instead?
```suggestion
        set->n_threads -= 1;
```

kleisauke (commented):
Great! I left some comments inline and was able to resolve the LSan CI failures by adding libMagickCore to the suppression file (commit https://github.com/kleisauke/libvips/commit/53386765f8ff7252b45ec67d20baf5cadb1edb55). The `sink->sslock` also no longer seems necessary after this (commit https://github.com/kleisauke/libvips/commit/b584140dae84b74b94067d895911509f84785b6c).

I had a quick go to test this implementation in wasm-vips, see commit https://github.com/kleisauke/wasm-vips/commit/da8956e6fd0ed6bc12d9abcb72b4a21083bd9425. The trick there is to initialize a fixed thread pool during `vips_threadset_new()` and not let `vips_threadset_run()` spawn more threads, see [the corresponding patch](https://github.com/kleisauke/wasm-vips/blob/da8956e6fd0ed6bc12d9abcb72b4a21083bd9425/build/patches/vips-3105-emscripten.patch#L711-L779) and [branch](https://github.com/kleisauke/libvips/tree/reimplement-threadpool-wasm). Though, this didn't seem to work, it somehow exhausts the fixed thread pool. I'll investigate.

libvips/iofuncs/threadset.c
jcupitt: Oooops! You're right, nice spot. Though I don't think it matters (fortunately).

jcupitt:
Ah I finally understand what you mean about the sslock in sink.c. I agree that can go, good discovery. I've merged your lsan fix too, nice!

I'll add something to threadset to enable creation of a fixed size pool at startup.

kleisauke:
I noticed that some tests on this branch became flaky or might crash suspiciously, which I cannot reproduce on the master branch.

<details>
    <summary>Details</summary>

```
test/test-suite/test_arithmetic.py::TestArithmetic::test_hough_circle FAILED
...
>           assert pytest.approx(x) == 50
E           assert 48 ± 4.8e-05 == 50
E             +48 ± 4.8e-05
E             -50

test/test-suite/test_arithmetic.py:397: AssertionError
```
```
test/test-suite/test_resample.py::TestResample::test_reduce FAILED
...
>                   assert d < 2
E                   assert 20.600368175034006 < 2

test/test-suite/test_resample.py:91: AssertionError
```
```
test/test-suite/test_conversion.py::TestConversion::test_scaleimage corrupted size vs. prev_size
Fatal Python error: Aborted
```
```
test/test-suite/test_histogram.py::TestHistogram::test_hist_ismonotonic free(): invalid next size (fast)
Fatal Python error: Aborted
```
</details>

Test script:
```bash
#!/usr/bin/env bash
set -ex

n=$1
exit_code=0

  echo "run $i"
  python3 -m pytest -sv --log-cli-level=WARNING test/test-suite || exit_code=$?
  if [ $exit_code -ne 0 ]; then
    echo "Python exited with non-zero exit code ($exit_code) after $i runs"
    exit $exit_code
  fi
done
```
```bash
./check-regression.sh 1024
```

I tried running this with ThreadSanitizer, but didn't spot anything wrong (see commit https://github.com/kleisauke/libvips/commit/8cfaf68971662f9ff4af922c533dd8b87698e7ed).

> I'll add something to threadset to enable creation of a fixed size pool at startup.

Thanks!

jcupitt:
I had a go ... now there's `VIPS_MAX_THREADS` to preallocate the threadset, and set a hard limit. What do you think? Allocation happens on first use, is that OK for wasm?

jcupitt:
Ouch, I agree, I see what looks like memory corruption too.

jcupitt:
I reverted the sslock revert and it seems fine now -- your test prog did 80 iterations without any problems. 

I'm not sure I understand why that broke things, I'll have a look.

kleisauke:
I confirm that fixes it.

> I had a go ... now there's `VIPS_MAX_THREADS` to preallocate the threadset, and set a hard limit. What do you think? Allocation happens on first use, is that OK for wasm?

Thanks! For Wasm, it needs to setup the fixed threadpool during `vips_init()`, so I moved it to `vips__threadpool_init()` instead with commit https://github.com/kleisauke/wasm-vips/commit/ee410bf5d6b84f737260d0ec979f4cf430359f5c.

wasm-vips had previously reverted commit https://github.com/libvips/libvips/commit/6946c3b9d9a047fbef5014341ced1ef4ad4d4be3 for this purpose, but that probably wouldn't work on macOS, where we need to delay setting up the threadpool to ensure `fork()` safety.

I wonder if we should expose `VIPS_MAX_THREADS`? As far as I know, WebAssembly is the only platform that requires this, so I'm not sure if we should include it by default, we could perhaps `#ifdef __EMSCRIPTEN__` it instead.

Note that running the wasm-vips test suite still seems to exhaust the fixed threadpool, see for example this run:
https://github.com/kleisauke/wasm-vips/actions/runs/3312785036/jobs/5470249864

I'm not sure what's causing this, the previous GLib threadpool implementation didn't had this issue. It can be reproduced by running the test suite on the [`reimplement-threadpool-wasm`](https://github.com/kleisauke/libvips/tree/reimplement-threadpool-wasm) branch or with:
```bash
$ export VIPS_MAX_THREADS="$(($(nproc) + 3))"
$ echo $VIPS_MAX_THREADS
9
```
(+3 is the amount of background threads required by `vips_sink_disc`, I'm aware that `dzsave` may require more threads than this, but I haven't compiled libgsf for Wasm yet)

kleisauke:
Ah, according to the [`g_thread_pool_push()`](https://docs.gtk.org/glib/method.ThreadPool.push.html) documentation, new threads are only started when the number of currently running threads is less than the maximal allowed number of threads. Otherwise, it stays in the queue until a thread in the pool finishes its previous task.

jcupitt:
> I confirm that fixes it.

Hooray!

> For Wasm, it needs to setup the fixed threadpool during vips_init(),

Ah I wasn't sure, OK, let's move it. I hoped to delay making the threads until first use to help systems like macOS, but it sounds like that's not possible (or useful).

> I wonder if we should expose VIPS_MAX_THREADS? 

I was thinking it would be handy for testing, but you're right, not useful for most people outside wasm. We can leave it undocumented and unsupported.

> +3 is the amount of background threads required by vips_sink_disc

I think it's two, isn't it? There's one thread for each of the two write behind buffers, plus the main thread (which we don't count). I see:

```
$ VIPS_MAX_THREADS=5 VIPS_CONCURRENCY=3 vips gaussblur k2.jpg x.jpg 100
$
```

So concurrency 3 runs with six threads: 3 workers, 2 write-behind, one main thread.

jcupitt:
> Otherwise, it stays in the queue until a thread in the pool finishes its previous task.

Perhaps we should add a queue like this to threadset? At the moment it just hard fails. Though I think there's maybe a danger of deadlock, ouch.

kleisauke:
> I think it's two, isn't it?

Oh huh, perhaps I miscounted. Though, this one seems to fail.
```shell
$ VIPS_MAX_THREADS=5 VIPS_CONCURRENCY=3 vips autorot zebra.jpg x.jpg
VipsThreadset: threadset is exhausted
```
Commit https://github.com/kleisauke/libvips/commit/a11afaa746cc3734c407c03f73a7e748a6f842f9 on the [`wasm-master`](https://github.com/kleisauke/libvips/tree/wasm-master) branch also seems to fail on the test suite. That branch contains the rebased wasm-vips patches applied to the master branch for reference.

> Perhaps we should add a queue like this to threadset? At the moment it just hard fails. Though I think there's maybe a danger of deadlock, ouch.

Deadlocks are indeed the symptoms when things don't work out. For example, I also had to apply commit https://github.com/kleisauke/libvips/commit/6100d34f995eed0bbea601fb72398204b3dd410d and needed to configure libwebp with `--disable-threading` for Wasm, since [this fixed thread limit](https://github.com/kleisauke/wasm-vips/blob/ec44ecd6fbc7b3da7c6a9f71a8989cd6b06043a4/src/CMakeLists.txt#L130) applies to the entire binary.

jcupitt:
I think that example is running two pipelines:

```
$ VIPS_MAX_THREADS=7 VIPS_CONCURRENCY=3 vips autorot k2.jpg x.jpg --vips-progress
vips temp-4: 1450 x 2048 pixels, 3 threads, 1450 x 16 tiles, 128 lines in buffer
vips k2.jpg: 1450 x 2048 pixels, 3 threads, 1450 x 16 tiles, 128 lines in buffer
vips k2.jpg: done in 0.0154s          
vips temp-4: done in 0.0304s          
```

Some libvips threads are optional (eg. when it tries to expand a threadpool), but some must happen or we'll deadlock (eg. it must have a thread watching a write-behind buffer).

We could add API for the optional case. It'd make it a bit more robust against threadset exhaustion.

kleisauke:
Ah, you're right. `vips_autorot()` is a non-sequential operation, so two pipelines are needed. For example, it can also be reproduced with the gaussblur sample when this line is commented out:
https://github.com/libvips/libvips/blob/ea0912f23bf61d3ddf1e7f59d6cce2ef7dbe07cb/libvips/convolution/gaussblur.c#L129

I've clarified with commit https://github.com/kleisauke/wasm-vips/commit/acbfff65e9b671d31906158810bd0535b25b107e why the +3 is needed.

> We could add API for the optional case. It'd make it a bit more robust against threadset exhaustion.

GLib threadpool's uses the [`g_async_queue_*`](https://docs.gtk.org/glib/struct.AsyncQueue.html) functions for this. Though, I'm not sure if we should do the same, we previously passed `max_threads = -1` to [`g_thread_pool_new()`](https://docs.gtk.org/glib/type_func.ThreadPool.new.html), so a new thread was spawned when no idle threads were available. IIUC, a queue implementation is only handy for a fixed threadpool (`VIPS_MAX_THREADS`), it probably adds some overhead when there's no hard limit.

It's a bit unfortunate that WebAssembly has this limitation. :(

libvips/iofuncs/threadset.c
```diff
@@ -0,0 +1,330 @@
...
+vips_threadset_new( int max_threads )
+{
+        VipsThreadset *set;
+
+        set = g_new0( VipsThreadset, 1 );
+	set->lock = vips_g_mutex_new();
+	set->max_threads = max_threads;
+
+        if( set->max_threads > 0 )
+                for( int i = 0; i < set->max_threads; i++ ) {
```
kleisauke, line 206: Should we declare `i` at the beginning of the scope block to ensure C89 compat?

libvips/iofuncs/threadpool.c
```diff
@@ -538,251 +221,308 @@ vips_thread_state_new( VipsImage *im, void *a )
...
+	/* Ask threads to exit, either set by allocate, or on free.
 	 */
 	gboolean stop;
-} VipsTask;
 
-/* Allocate some work (single-threaded), then do it (many-threaded).
+        /* Set this and the next worker to see it will clear the flag and exit
+         * (used to downsize the threadpool).
+         */
+        int exit;
```
kleisauke, line 279: nit: the newly added files and this one contains mixed indentations (4-space versus tabs). Perhaps we should re-indent?

test/test-suite/test_resample.py
```diff
@@ -88,7 +88,7 @@ def test_reduce(self):
                     x = im.cast(fmt)
                     r = x.reduce(fac, fac, kernel=kernel)
                     d = abs(r.avg() - im.avg())
-                    assert d < 2
+                    assert d < 5
```
kleisauke, line 91: Was this threshold increase necessary?

jcupitt: It failed for me, but I think that was perhaps just because of the crash bug we had. I'll revert it.

libvips/iofuncs/threadset.c
jcupitt: I thought we decided to switch to C99? I'll see if I can find the discussion.

jcupitt: We build with meson in C gnu99 mode, in line with glib, so I think this means we are effectively a c99 project now.

libvips/iofuncs/threadpool.c
jcupitt: Good point, I've been fiddling with my vim settings and I've probably messed it up.

Perhaps we should do tabs everywhere (I don't really mind). Do you have a preference?

kleisauke: I don't have a preference, as long as it's consistent. Tabs are the most commonly used indentation throughout the codebase, so that's fine.

jcupitt: I had a quick go and it touches lots of files outside this commit. Let's resolve whitespaces issues in a separate PR (and fix trailing spaces and tabs too).

jcupitt:
I think we're done, let's merge and try living with it for a bit.

Thank you very much for reviewing this thing!

kleisauke:
> GLib threadpool's uses the [`g_async_queue_*`](https://docs.gtk.org/glib/struct.AsyncQueue.html) functions for this. Though, I'm not sure if we should do the same, we previously passed `max_threads = -1` to [`g_thread_pool_new()`](https://docs.gtk.org/glib/type_func.ThreadPool.new.html), so a new thread was spawned when no idle threads were available. IIUC, a queue implementation is only handy for a fixed threadpool (`VIPS_MAX_THREADS`), it probably adds some overhead when there's no hard limit.

FWIW, I implemented this with commit https://github.com/kleisauke/libvips/commit/fe02527015b6d4740730162cdc3f4d3b6e3086e6 for wasm-vips (only the changes in `threadpool.c` and `threadset.c` are interesting). It pushes tasks to the global `GAsyncQueue` when the threadset is limited with `VIPS_MAX_THREADS` and exhausted. `vips_threadset_work()` is slightly modified to pop these tasks from the global queue and process them before the thread becomes idle. This seems to pass the wasm-vips testsuite using:
```shell
VIPS_MAX_THREADS="$(($(nproc) + 3))"
```

I'm not sure if it's worth upstreaming that, since it could easily deadlock when being used in combinatin with `dzsave` or when `VIPS_MAX_THREADS` is lower than 3.

jcupitt:
Oh, interesting.

It'd be great if we could upstream everything and have wasm as a fully supported target. Maybe we could hide the extra code behind ifdef WASM? 

I suppose the downside is that libvips would then have code that was difficult to test, and added maintenance effort. What do you think?

kleisauke:
I'd like to officially support Wasm as a supported target in libvips. Indeed, you can just `#ifdef` Wasm-specific code behind `__EMSCRIPTEN__` or `__wasm__` (if necessary).

Unfortunately, the main blocker for this is GLib, see for example:
https://github.com/GNOME/glib/compare/2.75.0...kleisauke:wasm-vips-2.75.0

I suppose this should be discussed in GLib first, especially commit https://github.com/kleisauke/glib/commit/4ed9ec074c3a015516849b023043a904699befcb (since that would cause an ABI break). GLib currently deliberately ignores the `-Wcast-function-type` and `-Wbad-function-cast` warnings due to this (and I'm aware that running with [`-fsanitize=cfi-icall`](https://clang.llvm.org/docs/ControlFlowIntegrity.html) is not supported either) - see:
https://gitlab.gnome.org/GNOME/glib/-/blob/2.75.0/meson.build#L483-485
https://gitlab.gnome.org/GNOME/glib/-/blob/2.75.0/meson.build#L502-504

For this reason, I marked the GLib patches as stalled in the "Upstream patches" item at https://github.com/users/kleisauke/projects/1, but hopefully I can revisit that in the future.

kleisauke:
I mentioned the patches and my intention to upstream them in [this comment](https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2882#note_1554298), but I plan to discuss this also on [GNOME's Discourse instance](https://discourse.gnome.org) (using the [`glib`](https://discourse.gnome.org/tag/glib) tag and possibly the [`strawman`](https://discourse.gnome.org/tag/strawman) tag for the function pointer cast issues), this to avoid that either the issue or MR (GitLab's meaning of PR) is immediately closed (e.g. [like this one](https://gitlab.gnome.org/GNOME/glib/-/issues/2483) :sweat_smile:).

Though, perhaps I should check the `#gtk` IRC channel on irc.gnome.org first to see whether the maintainers are interested in this. Hopefully they won't reject this in favor of https://gitlab.gnome.org/GNOME/gjs.

/cc @RReverser you might also be interested in this.

jcupitt:
Ah sensible, yes I remember that PR being closed without comment. How can a PR not be actionable? Ah well.

Sure, discourse is maybe a better place to start discussion. Please post a link here, I'd be interested to read.

PR #3100, opened 2022-10-19
exif: ensure prefix is present before parsing

lovell:
It looks like some newer Apple devices are attaching EXIF to images without the `Exif\0\0` prefix, however libexif still requires it due to JPEG/JFIF/APP1 marker heritage.

There's an upstream issue about this at https://github.com/libexif/libexif/issues/58

Originally reported at https://github.com/lovell/sharp/issues/3412

Before:
```
exif-data: 2402 bytes of binary data
exif-ifd0-XResolution: 72009/1000 (72.009, Rational, 1 components, 8 bytes)
exif-ifd0-YResolution: 72009/1000 (72.009, Rational, 1 components, 8 bytes)
exif-ifd0-ResolutionUnit: 2 (Inch, Short, 1 components, 2 bytes)
exif-ifd2-ExifVersion: Exif Version 2.1 (Exif Version 2.1, Undefined, 4 components, 4 bytes)
exif-ifd2-FlashpixVersion: FlashPix Version 1.0 (FlashPix Version 1.0, Undefined, 4 components, 4 bytes)
exif-ifd2-ColorSpace: 65535 (Uncalibrated, Short, 1 components, 2 bytes)
```

After:
```
exif-data: 2402 bytes of binary data
exif-ifd0-Make: Apple (Apple, ASCII, 6 components, 6 bytes)
exif-ifd0-Model: iPhone 11 Pro Max (iPhone 11 Pro Max, ASCII, 18 components, 18 bytes)
exif-ifd0-Orientation: 1 (Top-left, Short, 1 components, 2 bytes)
exif-ifd0-XResolution: 72/1 (72, Rational, 1 components, 8 bytes)
exif-ifd0-YResolution: 72/1 (72, Rational, 1 components, 8 bytes)
exif-ifd0-ResolutionUnit: 2 (Inch, Short, 1 components, 2 bytes)
exif-ifd0-Software: 15.5 (15.5, ASCII, 5 components, 5 bytes)
exif-ifd0-DateTime: 2022:07:08 13:00:46 (2022:07:08 13:00:46, ASCII, 20 components, 20 bytes)
exif-ifd2-ExposureTime: 1/122 (1/122 sec., Rational, 1 components, 8 bytes)
exif-ifd2-FNumber: 12/5 (f/2.4, Rational, 1 components, 8 bytes)
exif-ifd2-ExposureProgram: 2 (Normal programme, Short, 1 components, 2 bytes)
[...]

PR #3109, opened 2022-10-22
Ensure EXIF is at least 4 bytes before inspection

lovell:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=52606

This was exposed by me in https://github.com/libvips/libvips/pull/3100, sorry.


jcupitt:
No problem, I should have spotted it :( isprefix avoids buffer overflow, but only for null-terminated strings, which of course this isn't :(

PR #3119, opened 2022-10-26
make whitespace consistent

jcupitt:
- 8 space hard tabs for indent
- no trailing spaces or tabs

The result of running this script:

```bash

function forall()
{
        find . -name "*.[hc]" | xargs sed -i -e "$1"
        find . -name "*.cpp" | xargs sed -i -e "$1"
}

forall 's/^        /\t/'

forall 's/\t        /\t\t/'
forall 's/\t        /\t\t/'
forall 's/\t        /\t\t/'
forall 's/\t        /\t\t/'

forall 's/[ \t]*$//'
```


jcupitt:
This is a huge change and is likely to cause some problems with git blame etc. Is it worth it? Dunno!

kleisauke:
A lot more changes than I expected. Maybe we need to do the trailing spaces/tabs change in a separate PR? But perhaps that doesn't matter much.

> This is a huge change and is likely to cause some problems with git blame etc.

`git blame` has an option to ignore whitespace only modifications (`-w`), but I'm not sure if GitHub's UI has a similar option. `git rebase` features the `--ignore-whitespace` option.

jcupitt:
Let's use clang-format on everything instead and get away from manual code layout.

PR #3125, opened 2022-10-30
Enforce code consistency with clang-format

kleisauke:
This PR adds a `.clang-format` configuration file (commit 149ca4348b8bdc596762e6bcc6a12aa74c31fb3e) which I have been using to ensure consistent formatting.

With commit 6149b40b40df39e8f8e706be8e0ca86fe3b543b4, I applied this configuration to the entire project with (using clang-format 16.0.5):
```bash
$ find . \
  \( -name "*.[hc]" -o -name "*.cc" -o -name "*.cpp" \) \
  -not \( -path "./libvips/foreign/libnsgif/*" -o \
    -name vips-operators.cpp -o \
    -name StandaloneFuzzTargetMain.c -o \
    -name profiles.c \) | \
  xargs clang-format -i
```

Though, it still required some manual edits after that:
- line wrap at 80 columns and some minor formatting (commit e45de7fd4fc83ff91cc95c1778ccf5b832589df9);
- remove parentheses from return statements (commit bcf8e3f490ac5c437589d9516a15b7fe87df8cd2);
- rename all `or` variables (commit 5fedbc1f904337b12df5518b3e311c353cd6bb0b);
- adapt documentation to  `.clang-format` style;
  - `doc/` directory (commit 03ca6532d7345c58e4ee340fb091b4f8c3afebe2);
  - code comments (commit cdcfe2dca25e16e17d054849bf618dcfb34d9205);
- adapt auto-generated files to `.clang-format` style;
  - C++ bindings (commit dc1c707959a6b366ab16bf1c3ea744bd66dc66d2);
  - `profiles.c` (commit 5316e69007e5279a69d009fb95ebc41e43955af4);
  - `*.in` files (commit 33033d0bfa39c70594c270dade48e67fb1f99ef5).

While we're here, commit 7450863131a12ff1391eff6b68b372e3601d64f6 updates `.editorconfig` to reflect the clang-format changes and commit c322194087710db680655c26c696bccf57e38f81 add some usage instructions for this.

Last but not least, commit c31cdbe7d9b466fd93a7c3e8a03c04a0f8f5984c adds CI integration for this.


.clang-format
```diff
@@ -0,0 +1,25 @@
...
+TabWidth: 4
+UseTab: Always
+# Prevent clang-format from adding a space after the return statement, i.e:
+# `return( TRUE );` -> `return ( TRUE );`
+WhitespaceSensitiveMacros: [ 'return' ]
+# Our column limit is actually 80, but setting that results in clang-format
+# making a lot of dubious hanging-indent choices; disable it and assume the
+# developer will line wrap appropriately. clang-format will still check
+# existing hanging indents.
+ColumnLimit: 0
```
kleisauke, line 25: This idea is borrowed from GLib:
https://github.com/GNOME/glib/blob/9f108815740c7bb64bd8fae80d89161ffe49add6/.clang-format#L7-L11

jcupitt:
This is a good idea, and I like clang-format.

Having manual fixups after an automatic process is not good though :( Perhaps we should just adopt one of the more standard K&R-derived layouts? Would just getting rid of the spaces inside parenthesis be enough?

kleisauke:
Indeed, I don't like manual fixes either. Getting rid of the spaces inside parenthesis might help, see the subsequent commits. Line wrapping and adding comments in code regions we want to exclude is still necessary, but that's all.

Diff is best viewed with GitHub's "Hide whitespace" option:
https://github.com/libvips/libvips/pull/3125/files?w=1
(which could be handy if we do this on more files)

libvips/foreign/webpsave.c
```diff
@@ -724,15 +726,17 @@ static const char *vips__save_webp_suffs[] = { ".webp", NULL };
 
 /* Type promotion for save ... just always go to uchar.
  */
+// clang-format off
 static int bandfmt_webp[10] = {
-/* UC  C   US  S   UI  I   F   X   D   DX */
-   UC, UC, UC, UC, UC, UC, UC, UC, UC, UC
+/*  UC  C   US  S   UI  I   F   X   D   DX */
+	UC, UC, UC, UC, UC, UC, UC, UC, UC, UC
```
kleisauke, line 732: Looks like GitHub's UI is using `TabWidth: 8` by default. Appending `?ts=4` to the URL makes this comment aligned
https://github.com/libvips/libvips/pull/3125/files?ts=4

This can also be controlled in https://github.com/settings/appearance.

jcupitt: I suppose that's just a coincidence though --- I think that was aligned with tabwidth 8, and the alignment is thrown off by github adding the leading + char. Anyway, a minor thing!

kleisauke: Ah, you're right. Unfortunately, `TabWidth: 4` is still required in `.clang-format` because otherwise it uses spaces for continuation, despite using `UseTab: Always` and/or `ContinuationIndentWidth: 8`. WDYT about commit c2a037b07c20c96560c219ff2005047a66103af2 instead?

We could also consider preferring 4 spaces over tabs, but I don't want to interfere with anyone's workflow.

kleisauke:
I was thinking, perhaps we should run clang-format only on new files (or files we touch)? I'm not sure if we should run clang-format over all files at once (since line wrapping still has to be done manually after that).

jcupitt:
Let's clang-format everything in one huge commit. Moving to automatic formatting in a series of steps will cause even more disruption.

We could automatically format on push, or maybe just add a format check to the test suite. Any preference?

kleisauke:
I agree. Let's clang-format everything in one go.

> We could automatically format on push, or maybe just add a format check to the test suite. Any preference?

I added CI integration with commit 6efac0e6ac9fa62343f0dee8799598e2eed75774 and deliberately triggered a formatting error with the (temporary) commit [`477d8cc`](https://github.com/libvips/libvips/pull/3125/commits/477d8cc14fa14d94a014440ebbe79aff54f19200). On that commit (and on the "Files changed"-tab), I see:
Screenshot from 2022-11-23 13-36-35

However, it does not seem to detect all the files in CI (probably because this patch is huge). For example:
https://api.github.com/repos/libvips/libvips/pulls/3125/files

Returns only the modified `cplusplus/*` files, so it doesn't perform clang-format on the other files.

libvips/foreign/nsgifload.c
```diff
@@ -340,130 +340,131 @@ vips_foreign_load_nsgif_set_header( VipsForeignLoadNsgif *gif,
...
 }
 
 static int
-vips_foreign_load_nsgif_generate( VipsRegion *or,
-	void *seq, void *a, void *b, gboolean *stop )
+vips_foreign_load_nsgif_generate(VipsRegion * or,
+	void *seq, void *a, void *b, gboolean *stop)
 {
-	VipsRect *r = &or->valid;
+	VipsRect *r = & or->valid;
```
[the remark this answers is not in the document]
jcupitt: Oh dear, that's probably going to need quite a few changes :(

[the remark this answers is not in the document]
kleisauke: I'll investigate, it doesn't do this when it's named `class`, so it's a bit odd.

libvips/conversion/smartcrop.c
```diff
@@ -217,68 +217,70 @@ vips_smartcrop_attention( VipsSmartcrop *smartcrop,
...
 	vscale = 32.0 / in->Ysize;
-	sigma = VIPS_MAX( sqrt( pow( smartcrop->width * hscale, 2 ) +
-		pow( smartcrop->height * vscale, 2 ) ) / 10, 1.0 );
-	if ( vips_resize( in, &t[17], hscale,
-		"vscale", vscale,
-		NULL ) )
-		return( -1 );
+	sigma = sqrt(pow(smartcrop->width * hscale, 2) +
+		pow(smartcrop->height * vscale, 2));
+	sigma = VIPS_MAX(sigma / 10, 1.0);
```
kleisauke, line 222: I used this contstuction, otherwise clang-format does strange things their.

libvips/deprecated/im_gradcor.c
```diff
@@ -47,80 +46,89 @@
...
+			!grads ||
+			im_grad_x(small, xgrad) ||
+			im_grad_y(small, ygrad))
+			return (-1);
+
+		if (im_generate(out,
+				gradcor_start, gradcor_gen, gradcor_stop, (void *) large, (void *) grads))
+			return (-1);
+
+		return (0);
```
kleisauke, line 130: Split into two if-statements, otherwise clang-format placed it in one long line.

I also manually added parentheses to the return statements in this file, but maybe I should do that in all C files.

kleisauke:
Clarified the PR description and title. I confirm that re-running clang-format on all files does not change anything.

TODO:
- [ ] Remove all parentheses in return statements? Commit fed944ee414df2850ee13ed64342e92ece2ffc4f does this only for C++ files.
- [ ] Rename variables named `or` to `out_region`, just like in C++? Or perhaps there's a clang-format setting to not detect `or` as a language keyword (and thus avoid the addition of spaces)?
- [ ] Split the minor changes into another PR?
- [ ] Enforce code consistency in CI?:
  https://github.com/libvips/libvips/blob/0fff6c6384fd0c2bb5108e65119d1643e6f53ffc/.github/workflows/lint.yml#L21
- [ ] Should we disable the use of file annotations as PR feedback? To avoid dangling annotations, see for example:
  https://github.com/cpp-linter/cpp-linter-action/issues/119
- [ ] Sanity check changed files and the use of `clang-format off` comments.

kleisauke:
> * Rename variables named `or` to `out_region`, just like in C++? Or perhaps there's a clang-format setting to not detect `or` as a language keyword (and thus avoid the addition of spaces)?

I reported this issue at https://github.com/llvm/llvm-project/issues/59188.

jcupitt:
Shall we delay 8.14 for this PR? It looks like it could take a while to finish up.

kleisauke:
Alternatively, I'm fine with deferring this to the 8.15 cycle, hopefully the above clang-format issue will be addressed in LLVM 16 (which would avoid the need to rename all those `or` variables to `out_region` 😅).

jcupitt:
OK, let's do 8.14 alpha then.

jcupitt:
Ooop, once #3164 is done, I mean.

libvips/foreign/nsgifload.c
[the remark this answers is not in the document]
kleisauke: I worked around this with commit 60162ec4f7d43004d49036270fd906a25f884b26.

kleisauke:
This is ready for review now.

After this PR has landed, I'll create another PR to add a `.git-blame-ignore-revs` file to ensure these changes are ignored in the blame view, see:
https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

> hopefully the above clang-format issue will be addressed in LLVM 16

LLVM 16 was branched yesterday, so it probably won't be fixed in that release. I worked around this issue with commit 60162ec4f7d43004d49036270fd906a25f884b26 instead. FWIW, the touched files in that commit are annotated within the "Files changed"-tab, for example:
Screenshot from 2023-01-26 13-45-06

jcupitt:
I tried this PR out locally.

```
 563 files changed, 97803 insertions(+), 96343 deletions(-)
```

/me falls off his chair

jcupitt:
This is great! I had some general thoughts:

1. Perhaps we should exclude the libnsgif sources. If we reformat that, it'll make comparing to the upstream repo difficult.
2. That clang-format bug is tricky. Right now we need to run a sed after every clang-format :( This feels very bad. And even if clang fix this bug, it'll still break with older clang-formats :( So everyone would have to install the latest version of clang-format before they could do any work. Maybe we should just rename all those `or` variables? `out_region` would be clearer anyway.
3. A tiny "cheat sheet" in the README with some notes on running clang-format would help users who hadn't seen the formatter before.

kleisauke:
> ```
>  563 files changed, 97803 insertions(+), 96343 deletions(-)
> ```

:sweat_smile: 

Leaving out the whitespace changes results in:
```
 516 files changed, 16244 insertions(+), 14784 deletions(-)
```

> 1\. Perhaps we should exclude the libnsgif sources. If we reformat that, it'll make comparing to the upstream repo difficult.

Ah, I had excluded those sources within commit 219afdbb8cfe9da8c66baeed84d7edec4dd61526, but I accidentally touched `libvips/foreign/libnsgif/test/cli.c` in commit fa933a4ae654944e77fca089056b94909b1b4076. Let me fix that.

FWIW, the bundled files are also ignored in the CI lint configuration:
https://github.com/libvips/libvips/blob/d93ee2c52a9e7a83e1a413f886e143bde609103c/.github/workflows/lint.yml#L17-L18

> 2\. That clang-format bug is tricky. Right now we need to run a sed after every clang-format :( This feels very bad. And even if clang fix this bug, it'll still break with older clang-formats :( So everyone would have to install the latest version of clang-format before they could do any work. Maybe we should just rename all those `or` variables? `out_region` would be clearer anyway.

I wasn't sure how to approach this correctly, but this is indeed not the right solution. I'll change all `or` variables to `out_region` instead.

> 3\. A tiny "cheat sheet" in the README with some notes on running clang-format would help users who hadn't seen the formatter before.

Good idea! I'll add some instructions for this.

jcupitt:
Oh duh, sorry, you talk about this right at the top, forget that comment about nsgif.

I ran it like this, for reference:

```
find . \
  \( -name "*.[hc]" -o -name "*.cc" -o -name "*.cpp" \) \
  -not \( -path "./libvips/foreign/libnsgif/*" -o \
    -name vips-operators.cpp -o \
    -name StandaloneFuzzTargetMain.c -o \
    -name profiles.c \) | \
  xargs clang-format -i
find . -name "*.[hc]" | xargs sed -i -e "/VipsRegion \*/s/ or/or/" -e "/REGION \*/s/ or/or/" -e "s/& or/\&or/"
```

And indeed it all seems to work well, though it's a bit slow. And having to use `-i` followed by a fixup pass is not great :(

kleisauke:
Feedback incorporated, PTAL.

Somehow the CI says the `libvips/iofuncs/window.c` doesn't conform to code style, see e.g.:
Screenshot from 2023-01-27 14-55-32

But I cannot reproduce that using:
```shell
$ docker run -v $(pwd):/src -w /src -it ubuntu:22.04 sh -c "\
    apt-get update -y && \
    apt-get install -y clang-format && \
    clang-format -i libvips/iofuncs/window.c"
$ git status
On branch clang-format
nothing to commit, working tree clean
```

:man_shrugging:

jcupitt:
Oh, `git clang-format` is really nice! I didn't know about that.

I tested pytest, made a few formatting errors, tried a release build, tried a gtk-doc build, it all seems good and worked as expected.

I suppose one thing is the 4-space tabs we set now. All line breaks are currently set for 8-space tabs, so there's quite a bit of "prettyfying" we could do. But I think this is probably better left to happen piecewise as we work on the various files.

Summary: LGTM! Nice work on this, Kleis.

kleisauke:
Indeed, we may have to (re-)adjust some code to the 80-column limit.

I'll do another careful review tomorrow. :)

kleisauke:
> Somehow the CI says that the `libvips/iofuncs/window.c` file doesn't conform to the code style

I reported this issue at https://github.com/llvm/llvm-project/issues/60357, see e.g. commit 7a09a5a3d36f887b6c8ee048063cae3a64bff3e3 for a possible workaround.

jcupitt:
I made a branch here that fixes the merge conflicts https://github.com/libvips/libvips/tree/kleisauke-clang-format

Perhaps we should merge this PR now? It seems good to me.

kleisauke:
I held back this PR for a while because it could make backporting commits to the 8.14 branch difficult (as a different code style is used there).

I suppose there won't be any backports to 8.14 now, so let me rebase and merge this.

kleisauke:
\o/, PR #3566 will hide this landed commit from git blame.

jcupitt:
Hooray! This is a very good thing.

Issue #3134, opened 2022-11-02
Single-frame truncated GIF returns `gifload: no frames in GIF`


evazion:
Attempting to read the GIF below returns a `gifload: no frames in GIF` error:

```bash
$ wget https://user-images.githubusercontent.com/8430473/199390128-317eaf2b-8da3-495a-b8ba-a191c443ebf8.gif -O test.gif
$ vipsheader test.gif
vipsheader: gifload: no frames in GIF
$ vips copy test.gif test.jpg
gifload: no frames in GIF
$ vips copy test.gif[fail-on=none] test.jpg
gifload: no frames in GIF
$ vips --version
vips-8.13.3-Tue Nov  1 09:09:54 UTC 2022
```

I guess the GIF is truncated, but Firefox and Chrome are still able to render most of it, and other tools like ImageMagick and ExifTool can at least read the dimensions:

```
$ identify test.gif
test.gif GIF 575x800 575x800+0+0 8-bit sRGB 8c 40524B 0.000u 0:00.000
identify-im6.q16: corrupt image `test.gif' @ error/gif.c/PingGIFImage/964.
$ exiftool -v test.gif                            
  ExifToolVersion = 12.40
  FileName = test.gif
  Directory = .
  FileSize = 40524
  FileModifyDate = 1667360491
  FileAccessDate = 1667360628
  FileInodeChangeDate = 1667360628
  FilePermissions = 33152
  FileType = GIF
  FileTypeExtension = GIF
  MIMEType = image/gif
  GIFVersion = 89a
  ScreenDescriptor (SubDirectory) -->
  + [BinaryData directory, 7 bytes]
  | ImageWidth = 575
  | ImageHeight = 800
  | HasColorMap = 1
  | ColorResolutionDepth = 2
  | BitsPerPixel = 2
  | BackgroundColor = 0
  | PixelAspectRatio = 0
Graphic Control: delay=0.00
Image: left=0 top=0 width=575 height=800
  Error = File format error
```

I think this only applies to single-frame GIFs. I'm able to read multi-frame truncated GIFs.

I'd like to be able to read the as much of the GIF as possible and have the missing pixels replaced with transparency, or some other placeholder color.

Tested on vips 8.13.3 on Ubuntu 22.10 and 8.12.1 on Ubuntu 22.04.

[...]

PR #3141, opened 2022-11-04
Update to latest upstream libnsgif and call nsgif_data_complete

tlsa:
This allows libnsgif to distinguish between awaiting more data, and a broken truncated GIF. In the latter case we can display what we have.

Using libnsgif's test utility with the sample truncated GIF from https://github.com/libvips/libvips/issues/3134 was failing:

```
test_nsgif -i ~/Downloads/bad-truncated.gif 
gif:
  width: 575
  height: 800
  max-loops: 1
  frame-count: 0
  global palette: yes
  background:
    red: 0xb2
    green: 0xb2
    blue: 0xb2
  frames:
nsgif_frame_prepare failed: Frame can't be displayed
```

Now we get:

```
test_nsgif -i ~/Downloads/bad-truncated.gif
gif:
  width: 575
  height: 800
  max-loops: 1
  frame-count: 1
  global palette: yes
  background:
    red: 0xb2
    green: 0xb2
    blue: 0xb2
  frames:
  - frame: 0
    local palette: no
    disposal-method: Unspecified
    transparency: yes
    display: yes
    delay: 0
    rect:
      x: 0
      y: 0
      w: 575
      h: 800
```

And using vips:

```
tools/vips copy ~/Downloads/bad-truncated.gif converted.png
memory: high-water mark 3.72 MB
```

Produces the following PNG:

converted

### DIFF
--- a/libvips/foreign/libnsgif/README-ns.md
+++ b/libvips/foreign/libnsgif/README-ns.md
@@ -67,6 +67,10 @@ it has reached the end of the source data.
 > buffer. Just be sure to call `nsgif_data_scan()` again with the new pointer
 > before making any other calls against that nsgif object.
 
+When all the source data has been provided to `nsgif_data_scan()` it is
+advisable to call `nsgif_data_complete()` (see below), although this is not
+necessary to start decoding frames.
+
 To decode the frames, you can call `nsgif_get_info()` to get the frame_count,
 and then call `nsgif_frame_decode()` for each frame, and manage the animation,
 and non-displayable frames yourself, or you can use the helper function,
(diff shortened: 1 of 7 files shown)


jcupitt:
That's much neater than my PR, nice!

jcupitt:
... is there some way libvips could detect truncated GIFs? Maybe scan could return something other than OK, but still allow frames to be read?

tlsa:
> ... is there some way libvips could detect truncated GIFs? Maybe scan could return something other than OK, but still allow frames to be read?

Yes, good point! https://github.com/libvips/libvips/pull/3142

PR #3168, opened 2022-11-17
add exif support to png load/save

jcupitt:
the "Exif\0\0" header isn't being added and removed correctly

needs tests


jcupitt:
Tested with libpng and libspng, it all seems to work.

PR #3188, opened 2022-11-27
dzsave: add error handling for `vips_gsf_path()`

kleisauke:
It can return `NULL` when it exceeds the path limits.

Resolves: https://github.com/kleisauke/net-vips/issues/184.


jcupitt:
Ah, nice!

PR #3189, opened 2022-11-27
nsgifload: avoid minimise after mapping 

kleisauke:
Since it's not reliable on Windows, this corresponds to what is done in the PPM and WebP loaders. Previously on Windows, it somehow caused an "access denied"-error when trying to delete the GIF file after processing it.

While we're at it, improve `test_descriptors.c` by printing all the leaked paths, if there's a leak (previously it only printed the file passed as the first argument, which is not necessarily the case).

Context: https://github.com/libvips/libvips/discussions/3185.


jcupitt:
Useful!

Issue #3077, opened 2022-10-04
Support incremental webp decode


jcupitt:
Looks like libwebp supports block at a time decoding:

https://docs.rs/libwebp-sys2/0.1.1/libwebp_sys/fn.WebPIDecGetRGB.html

We could use this to avoid buffering the entire image in memory during load. Is there also incremental write now? Someone should check.

See https://github.com/libvips/libvips/discussions/3076 for referring discussion.

kleisauke:
> Is there also incremental write now?

I had [investigated that 2 years ago](https://github.com/libvips/libvips/discussions/2044#discussioncomment-377029), and it seems so far that this is only supported for non-animated WebP images where no metadata needs to be added - see:
https://github.com/webmproject/libwebp/blob/abf73d628e62cb8747125b4b9e993ccb401c174f/examples/cwebp.c#L989-L996

jcupitt:
Ah I'd forgotten about that, well done.

PR #3190, opened 2022-11-27
nsgifload: avoid using sequential flags

kleisauke:
It cannot be considered sequential since the entire source is mapped into memory. This corresponds to `webpload`, whose semantics are the same.

This also means that we can revert commit ce31c04cd28ab4d2c2872dd94d69b8e337ad0738. I verified that this does not cause any regressions using the test image and program mentioned in #2996.

Setup:
```console
$ gcc -O3 vipsgiftest.c `pkg-config vips --cflags --libs`
$ sha256sum animated_gif20.gif
58bb2293ff14273e37277a4f5b5f39f9f430d85b96b2728a133b5a99fb7d27a2  animated_gif20.gif
```

Before ([`master`](https://github.com/libvips/libvips/tree/master) branch - commit 4611651d90d5afe3f8628be2735a1d35be32f3b2):
```console
$ /usr/bin/time -f %M:%e ./a.out animated_gif20.gif x.gif
306912:5.54
```

Commit 0f1de55ccbb818f985039b912bb2ac6c31f64bff:
```console
$ /usr/bin/time -f %M:%e ./a.out animated_gif20.gif x.gif
304400:5.62
```

Commit 0abe70eee2eb6176fa68fbefdf9aee493546c6c8:
```console
$ /usr/bin/time -f %M:%e ./a.out animated_gif20.gif x.gif
304388:5.59
```

So, slightly lower memory use, similar performance. This was tested with cgif v0.3.0 and libimagequant v2.17.0, as provided by Fedora 37.

/cc @DarthSim

### DIFF
--- a/libvips/foreign/nsgifload.c
+++ b/libvips/foreign/nsgifload.c
@@ -259,7 +259,8 @@ vips_foreign_load_nsgif_set_header( VipsForeignLoadNsgif *gif,
 		gif->has_transparency ? 4 : 3,
 		VIPS_FORMAT_UCHAR, VIPS_CODING_NONE,
 		VIPS_INTERPRETATION_sRGB, 1.0, 1.0 );
-	vips_image_pipelinev( image, VIPS_DEMAND_STYLE_FATSTRIP, NULL );
+	if( vips_image_pipelinev( image, VIPS_DEMAND_STYLE_FATSTRIP, NULL ) )
+		return( -1 );
 
 	/* Only set page-height if we have more than one page, or this could
 	 * accidentally turn into an animated image later.
@@ -440,7 +441,8 @@ vips_foreign_load_nsgif_header( VipsForeignLoad *load )
 
 	gif->gif_delay = gif->delay[0] / 10;
 
-	vips_foreign_load_nsgif_set_header( gif, load->out );
+	if( vips_foreign_load_nsgif_set_header( gif, load->out ) )
+		return( -1 );
 
 	return( 0 );
 }
--- a/libvips/foreign/nsgifload.c
+++ b/libvips/foreign/nsgifload.c
@@ -173,13 +173,13 @@ vips_foreign_load_nsgif_dispose( GObject *gobject )
 static VipsForeignFlags
 vips_foreign_load_nsgif_get_flags_filename( const char *filename )
 {
-	return( VIPS_FOREIGN_SEQUENTIAL );
+	return( 0 );
 }
 
 static VipsForeignFlags
 vips_foreign_load_nsgif_get_flags( VipsForeignLoad *load )
 {
-	return( VIPS_FOREIGN_SEQUENTIAL );
+	return( 0 );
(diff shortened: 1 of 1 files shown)


kleisauke:
I cancelled the CIFuzz build on this PR, as it seems to be stuck (2+ hours running time). I'll investigate this further.

kleisauke:
... reported this CIFuzz issue at https://github.com/google/oss-fuzz/issues/9079.

kleisauke:
Just restarted the CIFuzz job, which should be safe after commit https://github.com/google/oss-fuzz/commit/675c27588499f3213e10d556ca92a9cbd1af9516.

jcupitt:
I think this PR will break seq mode for gifs, won't it?

With master (with the SEQ tag in gif load) you can run:

```
vips extract_area 3198.gif[n=-1] x.png 0 0 10 10 
```

On a large GIF and it'll only decode the first 10 scanlines of the first frame. This is because the loader and the crop are tagged as seq, so vips_foreign_load_temp() will link the crop operation directly to the output of gif load.

With this PR, the same command will load the entire GIF into memory (or disc) before starting the crop.

kleisauke:
Ah, good point. I assumed this was fine, since `nsgif_frame_decode()` in the `*_load`/`*_generate` phase operates on the `gif->anim` pointer, which already contains the entire decoded/`mmap()`'d image as processed by the `*_header` phase.

Note that these seq flags can probably be applied to `webpload` if we address issue #3077. 

Temporarily marking this PR as draft, I'll investigate if we could do incremental decode for GIF images instead.

kleisauke:
I confirm that this causes a performance issue (and uses slightly more memory) with partial decoding.

Using the provided animated GIF with 207 frames.
```console
$ vipsheader -f n-pages animated_gif20.gif
207
$ vipsheader animated_gif20.gif[n=-1]
animated_gif20.gif: 616x93357 uchar, 4 bands, srgb, gifload
```

I see this on the [`master`](https://github.com/libvips/libvips/tree/master) branch:
```console
$ /usr/bin/time -f %M:%e vips extract_area animated_gif20.gif[n=-1] x.png 0 0 10 10
53948:0.04
```

But on this PR, I see:
```console
$ /usr/bin/time -f %M:%e vips extract_area animated_gif20.gif[n=-1] x.png 0 0 10 10
55964:0.60
```

Performance will probably still be the same if you crop 10x10 pixels from the bottom of the GIF image. I'm aware that partially decoding a JPEG image also has this issue, although libjpeg-turbo (since v1.5.0) provides the `jpeg_skip_scanlines()` and `jpeg_crop_scanline()` for this purpose (IIRC, I mentioned using these APIs somewhere, but I couldn't find the corresponding issue for that, it was probably mentioned somewhere outside GitHub).

I'll abandon this PR, we should indeed experiment with applying these flags to `webpload` instead (and address the above issues regarding incremental and partial decoding).

PR #3198, opened 2022-11-29
heifsave: make it possible to use a specific encoder

dloebl:
[libheif](https://github.com/strukturag/libheif) supports various backends: For instance `aom`, `svt` and `rav1e` for AV1 encoding.
As of now, we use the first encoder we find (default encoder). However, in some cases it's useful to force [libheif](https://github.com/strukturag/libheif)  to use a different encoder (e.g. `svt` instead of `aom`).
Make it possible to specify the encoder to use (e.g. `aom`, `svt` etc).

**Question to the reviewer:**
~~Do we want to add a list of encoders to libvips or is it fine to just pass a string to libheif directly (as in this PR with the `encoder` parameter)?~~

### DIFF
--- a/libvips/foreign/foreign.c
+++ b/libvips/foreign/foreign.c
@@ -2250,6 +2250,7 @@ vips_heifload_source( VipsSource *source, VipsImage **out, ... )
  * * @compression: #VipsForeignHeifCompression, write with this compression
  * * @effort: %gint, encoding effort
  * * @subsample_mode: #VipsForeignSubsample, chroma subsampling mode
+ * * @encoder: %gchararray, select encoder to use
  *
  * Write a VIPS image to a file in HEIF format. 
  *
@@ -2270,6 +2271,8 @@ vips_heifload_source( VipsSource *source, VipsImage **out, ... )
  *
  * Use @bitdepth to set the bitdepth of the output file. HEIC supports at
  * least 8, 10 and 12 bits; other codecs may support more or fewer options.
+ *
+ * Use @encoder if you want to use a specific encoder (e.g. aom, svt etc).
  *
  * See also: vips_image_write_to_file(), vips_heifload().
  *
@@ -2303,6 +2306,7 @@ vips_heifsave( VipsImage *in, const char *filename, ... )
  * * @compression: #VipsForeignHeifCompression, write with this compression
  * * @effort: %gint, encoding effort
  * * @subsample_mode: #VipsForeignSubsample, chroma subsampling mode
+ * * @encoder: %gchararray, select encoder to use
  *
  * As vips_heifsave(), but save to a memory buffer. 
  *
@@ -2356,6 +2360,7 @@ vips_heifsave_buffer( VipsImage *in, void **buf, size_t *len, ... )
  * * @compression: #VipsForeignHeifCompression, write with this compression
  * * @effort: %gint, encoding effort
  * * @subsample_mode: #VipsForeignSubsample, chroma subsampling mode
+ * * @encoder: %gchararray, select encoder to use
  *
  * As vips_heifsave(), but save to a target.
  *
--- a/libvips/foreign/foreign.c
+++ b/libvips/foreign/foreign.c
@@ -2250,7 +2250,7 @@ vips_heifload_source( VipsSource *source, VipsImage **out, ... )
  * * @compression: #VipsForeignHeifCompression, write with this compression
  * * @effort: %gint, encoding effort
  * * @subsample_mode: #VipsForeignSubsample, chroma subsampling mode
- * * @encoder: %gchararray, select encoder to use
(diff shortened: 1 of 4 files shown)


libvips/foreign/heifsave.c
```diff
@@ -651,6 +668,12 @@ vips_foreign_save_heif_class_init( VipsForeignSaveHeifClass *class )
 		G_STRUCT_OFFSET( VipsForeignSaveHeif, speed ),
 		0, 9, 5 );
 
+	VIPS_ARG_STRING( class, "encoder", 18,
+		_( "encoder" ),
+		_( "select encoder to use" ),
```
kleisauke, line 673: nit: argument labels and descriptions should start with a leading capital letter. 
```suggestion
		_( "Encoder" ),
		_( "Select encoder to use" ),
```

jcupitt:
I think passing a string is fine.

dloebl:
I thought about it again and I think it's better to have an Enum in this case, so it's clearer which options (aom, svt, rav1e, x265) are available.
Ready for review.

jcupitt:
Oooops, I've just noticed:

```
$ vips heifsave
save image in HEIF format
usage:
   heifsave in filename [--option-name option-value ...]
where:
optional arguments:
...
   compression  - Compression format, input VipsForeignHeifCompression
			default: hevc
			allowed: hevc, avc, jpeg, av1
...
   encoder      - Select encoder to use, input VipsForeignHeifEncoder
			default: auto
			allowed: auto, aom, rav1e, svt, x265
```

Do we need both?

dloebl:
Do you mean the `compression` and the `encoder` parameter?
Yes, I think so: The `compression` specifies the compression format to use (e.g. HEVC or AV1). The `encoder` specifies the encoder you _want_ to use. Per compression format multiple encoders can be available. For instance:
`compression` = `av1`: `encoder` = `aom`,`rav1e` or `svt`.
As of now, you can only specify the `compression` parameter - which means that libvips uses the default encoder set in libheif.

Our use case is to choose a different encoder for AVIF encoding at runtime (e.g. SVT-AV1 instead of aom).

libvips/foreign/heifsave.c
```diff
@@ -465,9 +470,29 @@ vips_foreign_save_heif_build( VipsObject *object )
...
+	if( heif->selected_encoder != VIPS_FOREIGN_HEIF_ENCODER_AUTO ) {
+		const int count = heif_context_get_encoder_descriptors( heif->ctx,
+					(enum heif_compression_format) heif->compression,
+					vips_enum_nick( VIPS_TYPE_FOREIGN_HEIF_ENCODER, heif->selected_encoder ),
+					&out_encoder, 1 );
+
+		if( count > 0 ) {
+			error = heif_context_get_encoder( heif->ctx,
+				out_encoder, &heif->encoder );
+		} else {
```
jcupitt, line 484: nit: libvips "style" is newline before else.

libvips/foreign/foreign.c
```diff
@@ -2271,6 +2272,8 @@ vips_heifload_source( VipsSource *source, VipsImage **out, ... )
  * Use @bitdepth to set the bitdepth of the output file. HEIC supports at
  * least 8, 10 and 12 bits; other codecs may support more or fewer options.
  *
+ * Use @encoder if you want to use a specific encoder (e.g. aom, svt etc).
```
jcupitt, line 2275: I think this is a little unclear. Suggestion:

- In the "compression" paragraph above, change to "Use @compression to set the compression format, e g. ..."
- Change this text to be "Use @encoder to set the encode library, eg. aom, rav1e, etc."

jcupitt (commented):
Looks good (now I understand the compression / encoder difference heh). Just some small nits.

jcupitt:
Nice!

suphamster:
Can  this be implemented for Windows also since I dont't have "encoder" option available?
```
C:\Test\vips>vips --vips-progress heifsave
save image in HEIF format
usage:
   heifsave in filename [--option-name option-value ...]
where:
optional arguments:
   Q            - Q factor, input gint
                        default: 50
                        min: 1, max: 100
   bitdepth     - Number of bits per pixel, input gint
                        default: 12
                        min: 1, max: 16
   lossless     - Enable lossless compression, input gboolean
                        default: false
   compression  - Compression format, input VipsForeignHeifCompression
                        default: hevc
                        allowed: hevc, avc, jpeg, av1
   effort       - CPU effort, input gint
                        default: 4
                        min: 0, max: 9
   subsample-mode - Select chroma subsample operation mode, input VipsForeignSubsample
                        default: auto
                        allowed: auto, on, off
   strip        - Strip all metadata from image, input gboolean
                        default: false
   background   - Background value, input VipsArrayDouble
   page-height  - Set page height for multipage save, input gint
                        default: 0
                        min: 0, max: 10000000
operation flags: sequential nocache
```

PR #3199, opened 2022-11-29
Ensure consistent formatting of `->format_table`

kleisauke:
Also, ensure it's declared as a `VipsBandFormat` array instead of an `int` array. Note that it can only be declared const outside the `foreign/` directory, see for example:
https://github.com/libvips/libvips/blob/25444cd9275a154f4f24f071bdfdcd996529d117/libvips/include/vips/foreign.h#L378
https://github.com/libvips/libvips/blob/25444cd9275a154f4f24f071bdfdcd996529d117/libvips/arithmetic/parithmetic.h#L92
According to https://stackoverflow.com/a/5083976, it would not be an ABI/API break to change it to const-qualified type, but reverting back to non-const would be. I can open a new PR for that, if desired.

_Split out from #3125._

### DIFF
--- a/libvips/arithmetic/abs.c
+++ b/libvips/arithmetic/abs.c
@@ -244,8 +244,8 @@ vips_abs_buffer( VipsArithmetic *arithmetic,
 /* Format doesn't change with abs, other than complex -> real.
  */
 static const VipsBandFormat vips_abs_format_table[10] = {
-/* UC  C   US  S   UI  I   F   X   D   DX */
-   UC, C,  US, S,  UI, I,  F,  F,  D,  D 
+	/* Band format:  UC  C  US  S  UI  I  F  X  D  DX */
+	/* Promotion: */ UC, C, US, S, UI, I, F, F, D, D
 };
 
 static void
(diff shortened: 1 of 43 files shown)


kleisauke:
CI failure in the ASan/UBSan build is flaky. Somehow LSan aborts within the atexit handler, which is considered to be undefined behavior.
https://github.com/llvm/llvm-project/blob/1f925d70ce8b8f3fc7b83cf2495966fa44b597c9/compiler-rt/lib/lsan/lsan_common_linux.cpp#L116-L121
https://github.com/llvm/llvm-project/blob/1f925d70ce8b8f3fc7b83cf2495966fa44b597c9/compiler-rt/lib/sanitizer_common/sanitizer_termination.cpp#L50

Given that LSan doesn't provide a useful stack trace, maybe it's racing with the atexit handler that pyvips registers?
https://github.com/libvips/pyvips/blob/98cbdb3b0ad2b66c5d42d226bc2525ccecf282b5/pyvips/__init__.py#L172

Anyway, let me restart that failed CI job.

kleisauke:
... I just opened PR #3205 as an attempt to fix this flakiness.

jcupitt:
OK, LGTM. It's good to fix these small consistency things.

PR #3202, opened 2022-11-29
Minor doc improvements

kleisauke:
Note that I recently became aware that `bash` is not the correct keyword for shell commands. You have to use `console` for that to ensure that syntax highlighting doesn't match on output. For example:

```bash
$ for i in {1..1000}; do cp x.tif test/$i.tif; done
for i in {1..1000}; do cp x.tif test/$i.tif; done
```
(i.e., triple backtick `bash`)

```console
$ for i in {1..1000}; do cp x.tif test/$i.tif; done
for i in {1..1000}; do cp x.tif test/$i.tif; done
```
(i.e., triple backtick `console`, much nicer)

Unfortunately DocBook (or Pandoc?) does not support that, so commit bed04b85d24ee43005bbe37d89489771b76e883c removes the bash highlighting on output where this is a issue.

_Split out from #3125._

### DIFF
--- a/doc/Examples.md
+++ b/doc/Examples.md
@@ -17,7 +17,7 @@ in the API.  It's also a useful source of examples.
 
 # Average a region of interest box on an image
 
-``` python
+```python
 #!/usr/bin/python3
 
 import sys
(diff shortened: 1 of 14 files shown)


jcupitt:
Ah, nice.

PR #3205, opened 2022-11-29
CI: disable `__tls_get_addr` interception in ASan

kleisauke:
It might be the cause of the intermittent ASan failures on CI.

This workaround should be safe, since it has been adopted by many projects, e.g. [Chromium](https://github.com/chromium/chromium/commit/536f57c0fd998ebb788a668544e721fa87d0fd74), [Node.js](https://github.com/nodejs/node/commit/347000ccf8c4705721ff60717b234c75e72183cd) and others.

See: https://github.com/google/sanitizers/issues/1322


jcupitt:
Good discovery!

PR #3209, opened 2022-12-01
Adding layout: IIIF and IIIF3 to documentation

aliflux-omo:
Doc patch to add IIIF creation info


jcupitt:
Hi @aliflux-omo, thanks for making this PR!

Issue #3240, opened 2022-12-24
The road to 8.14


jcupitt:
We have 8.14.0-rc1 now, including win32 binaries:

https://github.com/libvips/libvips/releases/tag/v8.14.0-rc1

Website updated in this branch:

https://github.com/libvips/libvips/tree/gh-pages-8.14

Draft release notes here:

https://github.com/libvips/libvips/blob/gh-pages-8.14/_posts/2022-12-22-What's-new-in-8.14.md

## TODO

- [x] there have been some minor doc fixes since rc1, we need to regen the C++ binding 
- [x] and the website docs
- [x] test downstream bindings ... pyvips and ruby-vips seem fine
- [x] test sharp
- [x] benchmark
- [ ] test packaging for deb and rpm now that auto* has gone

lovell:
The sharp tests are passing on all platforms/archs using 8.14.0-rc1 :tada: 

https://github.com/lovell/sharp/actions/runs/3801639451
https://github.com/lovell/sharp/actions/runs/3801639453
https://app.circleci.com/pipelines/github/lovell/sharp/329/workflows/1a6aecda-f7c4-4de4-8820-8ef609f11f62

kleisauke:
The NetVips tests are also passing on all platforms/archs using 8.14.0-rc1. :heavy_check_mark: 
https://www.nuget.org/packages/NetVips.Native/8.14.0-rc1
https://github.com/kleisauke/net-vips/actions/runs/3806284622
https://github.com/kleisauke/net-vips/runs/10362832354
https://ci.appveyor.com/project/kleisauke/net-vips/builds/45802136

The wasm-vips test suite passes locally using 8.14.0-rc1 (on the [`vips-8.14`](https://github.com/kleisauke/wasm-vips/tree/vips-8.14) branch). :heavy_check_mark: 

I deployed an A/B test on the [`weserv/images`](https://github.com/weserv/images) project using 8.14.0-rc1, which I'll be monitoring for a few days. :mag:

jcupitt:
That's great! Let's make 14.0!

I shouldn't, but I'd like to squeeze that `arrayjoin` improvement in -- it really has a huge effect on performance (runtime goes from 2m to 1s on that benchmark). I'll try to finish testing that today.

kleisauke:
[...]

PR #3253, opened 2023-01-01
threadset: stop idle threads after a regular timeout

kleisauke:
In line with the previous behaviour of GLib's threadpool, which would help in long-running processes. This PR is best reviewed per commit.

Unfortunately, the added internal function (`vips__semaphore_down_timed()`) was based on `vips_semaphore_downn()`, so there are currently two copies of almost the same function. I don't think we can safely change `vips_semaphore_downn()` as that could be called with `n > 1`, which could conflict with the error code (`-1`) upon return.

Also, deprecate `vips_g_thread_join()` in favor of `g_thread_join()`  while we are here (commit 46aee63a4f43e6486867208f5e049f42541eccd7).

Context: https://github.com/libvips/libvips/issues/3240#issuecomment-1367931803.

---
This branch has successfully deployed on the service associated with the [`weserv/images`](https://github.com/weserv/images) project.
<details>
  <summary>Details</summary>

[graph](https://kleisauke.nl/threads-pinpoint-1671872837-1672564037.png)
</details>


jcupitt:
> I don't think we can safely change vips_semaphore_downn() as that could be called with n > 1

We always have n>=0 in libvips, and it'd be strange to call it with n<0 I think. How about adding "n must be non-negative" to the docs and adding an assert?

I think avoiding two copies of (almost) the same code is a useful win for maintenance and consistency.

libvips/iofuncs/semaphore.c
```diff
@@ -144,3 +144,43 @@ vips_semaphore_down( VipsSemaphore *s )
 {
 	return( vips_semaphore_downn( s, 1 ) );
 }
+
+/* Waits until sem > 0, then decrement.
+ * Returns -1 when the relative time in @timeout_ms (in microseconds)
+ * was passed.
+ */
+int
+vips__semaphore_down_timed( VipsSemaphore *s, double timeout_ms )
```
jcupitt, line 153: How about just using seconds? With a double, the unit doesn't really affect the accuracy, so I think I'd expect the most human-friendly unit. If it's really microseconds, a gint64 would probably be better.

kleisauke:
> > I don't think we can safely change vips_semaphore_downn() as that could be called with n > 1
> 
> We always have n>=0 in libvips, and it'd be strange to call it with n<0 I think. How about adding "n must be non-negative" to the docs and adding an assert?

Ah, indeed. I misunderstood that logic, it would only fail when `n < 0`. Making `n` a `guint32` would fix that, but that would also break API/ABI compatibility. :(

> I think avoiding two copies of (almost) the same code is a useful win for maintenance and consistency.

Resolved with commit 78e251dd7504c860ace6e8e3195127107701c5c6.

libvips/iofuncs/semaphore.c
kleisauke: Commit 78e251dd7504c860ace6e8e3195127107701c5c6 makes this a `gint64` instead.

jcupitt:
Great!

jcupitt:
I added a note to the docs about n>=0 for `_timeout()`.

PR #3256, opened 2023-01-02
Ensure pthread_create suppression stack also matches newer glibc

lovell:
It looks like the internal implementation of `pthread_create` in newer versions of glibc has altered the call stack slightly.

With glibc 2.35 I see:
```
  fun:calloc
  fun:allocate_dtv
  fun:_dl_allocate_tls
  fun:allocate_stack
  fun:pthread_create@@*
```
The frame-level wildcard additions in this PR should allow the suppression to match both ("it works on my machine" :tm: ).

### DIFF
--- a/suppressions/valgrind.supp
+++ b/suppressions/valgrind.supp
@@ -97,7 +97,9 @@
    pthread
    Memcheck:Leak
    fun:calloc
+   ...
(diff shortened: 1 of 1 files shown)


jcupitt:
:+1:

Issue #2879, opened 2022-06-20
Please keep autotools until next major version


remicollet:
As it works well ;)

I'm used to consider such a major change to the build system is suitable for a major version, not in a minor one.
(and it is broken for now)

So please keep autotools for now.

jcupitt:
Hi @remicollet, thanks for testing the rc.

I don't think we want to maintain two build systems, so I'm keen to remove autotools as soon as possible.

In my mind, the major version number represents the API the package supports, and that's supposed to be unchanged with this release. The build system is obviously very important for packaging, but isn't generally a big deal for users.

kleisauke:
For reference, GLib also dropped autotools without bumping the major version. See:

> Overview of changes in GLib 2.59.1
==================================
> * Autotools support is gone. (!580)

...

> Overview of changes in GLib 2.59.0
==================================
> * This will be the last development release with autotools support. As our
>  Meson support has been around since 2.56.x, and was used to release tarballs
>  in 2.58.x, the next development release (2.59.1) will drop autotools as used
>  to build GLib. The macros installed for other packages to use will remain.

https://gitlab.gnome.org/GNOME/glib/-/blob/main/NEWS
https://gitlab.gnome.org/GNOME/glib/-/merge_requests/580

remicollet:
I understand...

BTW, among hundreds of packages I maintain
- lot use autotools
- some use cmake
- none use meson ;)

The reason why I'm always reluctant to switch in a new unknown world :p

eli-schwartz:
As a general rule of thumb, I think it's reasonable enough when changing from one build system to another, to have a *transition* state, where one version supports both build systems. You get the opportunity to shake out bugs in the new build system, without leaving people in the lurch if they do run into problems. You can always mark the old build system as deprecated (and possibly print warnings to that effect, even), and a common tactic is to delete the old build system immediately after the next (non-bugfix) release.

But I don't think that you need to wait for a semver major library API/ABI change in order to cycle the build system!

With that in mind, the release notes for the current feature release seem reasonable:
> We’ve settled on [meson](https://mesonbuild.com/) and tintou has very generously done all the work. The old autotools system is still there for this release, but it will be removed for 8.14 and meson will the only one we support.

autotools is available for a transition period of this release, but will be removed in the next release. People are urged to switch over if at all possible, but if they cannot, they can still fall back to autotools for now, while the meson.build gets fixed.

[...]

PR #3257, opened 2023-01-02
Lower Meson version requirement to 0.55.x

kleisauke:
Allowing libvips to be compiled on RHEL 7 (and its derivatives). Tested with:
```console
$ pip3 install meson==0.55.0
$ meson setup full_build -Ddoxygen=true -Dgtk_doc=true -Dvapi=true
$ meson compile -Cfull_build
$ meson install -Cfull_build
```

See: https://github.com/libvips/libvips/issues/2879#issuecomment-1368742798.

### DIFF
--- a/doc/meson.build
+++ b/doc/meson.build
@@ -51,7 +51,7 @@ private_libvips_headers = [
 
 private_headers = []
 foreach private_libvips_header : private_libvips_headers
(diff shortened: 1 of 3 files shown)


kleisauke:
Perhaps I could lower it further (e.g. to support Ubuntu 20.04), but I'm not sure if that would be worth it.

jcupitt:
I agree, let's try this.

PR #3270, opened 2023-01-10
sharpen: reduce max sigma to 10

lovell:
See https://github.com/lovell/sharp/issues/3521

Plus starts a new changelog for 8.15.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,6 +4,7 @@
 	* use with (un)premultiply for ~10% perf gain on AVX CPUs
 - dedupe FITS header write [ewelot]
 - add fast path to extract_band and bandjoin for uchar images [lovell]
+- reduce `vips_sharpen` max `sigma` to 10 [lovell]
 
 TBD 8.14.2
 
--- a/libvips/convolution/sharpen.c
+++ b/libvips/convolution/sharpen.c
@@ -330,7 +330,7 @@ vips_sharpen_class_init( VipsSharpenClass *class )
 		_( "Sigma of Gaussian" ),
 		VIPS_ARGUMENT_OPTIONAL_INPUT,
 		G_STRUCT_OFFSET( VipsSharpen, sigma ),
-		0.000001, 10000.0, 0.5 );
+		0.000001, 10.0, 0.5 );
 
 	VIPS_ARG_DOUBLE( class, "x1", 5, 
 		_( "x1" ), 


kleisauke:
Given that this value is passed to `vips_gaussmat()`, perhaps the max should be reduced there as well?
https://github.com/libvips/libvips/blob/b00a4a80d6e9bcca8f2a3e02e3fcdff68e646538/libvips/convolution/sharpen.c#L206-L213
https://github.com/libvips/libvips/blob/f11b58ff6c602549d553c57e506142de5b6d28e9/libvips/create/gaussmat.c#L189

FWIW, I'm aware that passing a large `sigma` value to `vips_gaussblur()` [is also computationally intensive](https://github.com/lovell/sharp/blob/081debd055257426ad355049e212e524cc63ddfb/test/unit/timeout.js#L9-L15), so we could also reduce that one as a possible follow-up (or possibility all `sigma` arguments while we are at it(?)).

kleisauke:
I see. Somehow I was able to halve the RSS usage by short-circuiting the `vips_rot()` operation in `vips_convsep()`.
```diff
--- a/libvips/convolution/convsep.c
+++ b/libvips/convolution/convsep.c
@@ -88,7 +88,7 @@ vips_convsep_build( VipsObject *object )
 	else { 
 		/* Take a copy, since we must set the offset.
 		 */
-		if( vips_rot( convolution->M, &t[0], VIPS_ANGLE_D90, NULL ) ||
+		if( vips_rot( convolution->M, &t[0], VIPS_ANGLE_D0, NULL ) ||
 			vips_copy( t[0], &t[3], NULL ) )
 			return( -1 ); 
 		vips_image_set_double( t[3], "offset", 0 );
```
(tested with `--sigma 100 --vips-concurrency 36` using the provided image in the linked issue)

Of course that would produce incorrect images, but I didn't expect this `vips_rot()` call on the mask to consume that amount of memory.

jcupitt:
No that seems strange to me too. It probably needs investigating.

kleisauke:
I'll investigate. The exact same short-circuit change would also resolve a suspicious "out of order read" error on my 6-core workstation, see: https://gist.github.com/kleisauke/5129e16749f1b3de35bfb29d2853fc93.

Note that I haven't encountered both issues between sigma 1 and 10, inclusive.

kleisauke:
I think I found a possible fix, adding seq to `vips_convsep()` seems to fix this. See e.g. commit https://github.com/kleisauke/libvips/commit/f9123c6630e2872e78ca80d3e22a70b3c676019f.

With that, I see:

Before:
```console
$ /usr/bin/time -f "%M" vips sharpen 211210991-13a20ac3-e099-4813-bf2b-894d6be414cf.png out.png --sigma 100 --vips-concurrency 36
1508268
```

After:
```console
$ /usr/bin/time -f "%M" vips sharpen 211210991-13a20ac3-e099-4813-bf2b-894d6be414cf.png out.png --sigma 100 --vips-concurrency 36
180848
```

This will probably(?) also help `vips_gaussblur()`, since that also uses `vips_convsep()`.

lovell:
Rebased to remove the changelog conflict. In terms of the `vips_sequential` suggestion, are we thinking this is a :why_not_both: situation?

kleisauke (approved):
LGTM! Sorry for having sat on this.

Issue #3273, opened 2023-01-13
Duplicated header keywords when combining FITS images


ewelot:
Combining FITS images (e.g. by using add() or bandjoin()) results in a FITS image with duplicated BSCALE and BZERO header keywords if the individual input images do contain those keywords. This violates the FITS standard.

I believe a useful approach would be to not copy those keywords from input files and write them once if necessary according to the image data.

Example FITS image:
[rose.fits.zip](https://github.com/libvips/libvips/files/10410983/rose.fits.zip)

jcupitt:
Hi again, 

Yes, I guess this will need sorting out :( 

The FITS format is rather difficult to write: some header fields can be duplicated, some must not, and there's no way to find out if a field can or cannot be duplicated using cfitsio. I think libvips will need to maintain it's own list of "fields that should not be duplicated", track written field names in its own hash table (since cfitsio wont tell you if a field has been set already), and knock out dupes there. It'll also need its own thing to parse FITS header lines, since cfitsio doesn't seem to provide this either.

`vips_fits_write_meta()` has some code a bit like this already, it probably just needs expanding.

ewelot:
Thanks for the quick reply!

According to the [FITS standard](url=https://fits.gsfc.nasa.gov/standard40/fits_standard40aa-le.pdf) the header keywords should be unique (see end of section 4.1) with the exception of long-string keywords (ref. 4.2.1.2) and commentary keywords (ref. 4.4.2.4).

As far as I can tell, cfitsio has quite a number of functions to read and write header records. It is a bit hdden, because the function names are somewhat cryptic in fitsio.h but there are more user friendly aliases in longnam.h. Unfortunately I have almost no experience in using them. ;(

Hope that helps a bit,
Thomas

jcupitt:
Ah you're right, it looks like `COMMENT, `HISTORY`, `CONTINUE` and blank line can appear more than once, but all others should be unique. I suppose cfitsio might handle `CONTINUE` for us.

ewelot:
To me it looks like the following scheme could be applied for the header part of writing FITS images:
- write required header keyword records
- add BSCALE and BZERO (e.g. if image data are 16bit, ref. to section 5.2.5)
- copy all records from first FITS image for keywords not already written

[...]

PR #3282, opened 2023-01-15
Dedupe fits header

jcupitt:
We mustn't write most FITS header keywords more than once. Something like `vips add a.fits b.fits x.fits` was duplicating header keywords.

See https://github.com/libvips/libvips/issues/3273

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,7 @@
 
 - add support for target_clones attribute [lovell]
 	* use with (un)premultiply for ~10% perf gain on AVX CPUs
+- dedupe FITS header write [ewelot]
 
 9/1/23 8.14.1
 
--- a/libvips/foreign/fits.c
+++ b/libvips/foreign/fits.c
@@ -346,12 +346,6 @@ vips_fits_get_header( VipsFits *fits, VipsImage *out )
 	}
 
 	for( i = 0; i < keysexist; i++ ) {
-		/* We try to make key names unique, so when you eg. "add
-		 * a.fits b.fits c.fits" the writer sees all keys from all
-		 * inputs.
-		 */
-		static int key_id = 0;
-
 		char record[81];
 		char vipsname[100];
 
@@ -368,7 +362,7 @@ vips_fits_get_header( VipsFits *fits, VipsImage *out )
 		 * have to include the key index in the vips name we assign.
 		 */
 
-		vips_snprintf( vipsname, 100, "fits-%d", key_id++ );
+		vips_snprintf( vipsname, 100, "fits-%d", i );
 		vips_image_set_string( out, vipsname, record );
 	}
 
--- a/libvips/foreign/fits.c
+++ b/libvips/foreign/fits.c
@@ -68,8 +68,8 @@
  */
 
 /*
-#define VIPS_DEBUG
  */
+#define VIPS_DEBUG
 
 #ifdef HAVE_CONFIG_H
 #include <config.h>
@@ -127,6 +127,12 @@ typedef struct {
 	/* One line of pels ready for scatter/gather.
 	 */
 	VipsPel *line;
+
+	/* All the field names we've written, eg. "NAXIS", so we can dedupe 
+	 * metadata.
+	 */
+	GSList *names;
+
 } VipsFits;
 
 const char *vips__fits_suffs[] = { ".fits", ".fit", ".fts", NULL };
@@ -147,6 +153,7 @@ vips_fits_close( VipsFits *fits )
 {
 	VIPS_FREE( fits->filename );
 	VIPS_FREEF( vips_g_mutex_free, fits->lock );
+	VIPS_FREEF( vips_slist_free_all, fits->names );
 
 	if( fits->fptr ) {
 		int status;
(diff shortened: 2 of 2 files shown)


jcupitt:
We should wait for a FITS expert to have a look at this (paging @ewelot !).

ewelot:
This is my findings about Multi-Extension FITS files and libvips.

A few general comments at first: A FITS file which has a single header-data-unit (HDU) is sometimes referred to as a Single Image FITS (SIF) and a FITS file containing multiple HDUs is referred to as a Multi-Extension FITS (MEF). The latter may consist of several image HDUs (even of different sizes and dimensions) and table data HDUs and often it may have a data-less primary HDU (just storing metadata). MEF are often used to store multiple images from observations at multiple wavelengths (different filters). A conscise description of the FITS standard can be found [here](https://fits.gsfc.nasa.gov/standard40/fits_standard40aa-le.pdf). Example image crops from Hubble Space Telescope science images containing elaborated header structures are provided for both [Single Image FITS](https://fits.gsfc.nasa.gov/samples/WFPC2ASSNu5780205bx.fits) and [Multi-Extension FITS](https://fits.gsfc.nasa.gov/samples/NICMOSn4hk12010_mos.fits) files.

Although this thread is about fixing the FITS header part I should place it within the context of the general support of FITS files by libvips. With this respect I believe we should aim for the following basics:
1. read all 1D, 2D, 3D image HDUs by joining bands (possibly enlarging axes), ignore image HDUs of higher dimensions
2. always write a single image HDU where multiple bands form a FITS cube instead of multiple image extensions
3. create the output FITS header from required keywords and allowed records originating from the first image HDU (non data-less) of the first input image (I would call it header reference HDU)

Currently libvips reads the first image HDU only. It already writes image data in the required way.
The header part is almost there. It already picks the correct HDU to copy header fields from, even with input from MEF. I am currently spotting two issues, though: 
1. Propagate comment dupes and extended comments to the output header.
2. Do not write XTENSION, PCOUNT, GCOUNT records which are not allowed in a single image HDU. This has to be taken into account if the first input image is a MEF and the header reference HDU is not the primary HDU but a FITS extension (see e.g. the MEF example above).

Sorry for his long post ...
Thomas

jcupitt:
OK, I think I've fixed everything. It correctly (I hope!) dedupes the two forms of comment and blank line, and knows about the multi HDU header fields. 

Could you test again?

ewelot:
Initial tests look promising (YEAH!), will have more time for testing tomorrow ...

ewelot:
OK, it looks like the issue regarding XTENSION, PCOUNT, GCOUNT is fixed now.

Most of the duplicated empty line records are correctly preserved now, but not those of comment lines as indicated by the following example. Using an input MEF file `sample.fits` containing this  handcrafted header
```
SIMPLE  =                    T / file does conform to FITS standard
BITPIX  =                    8 / number of bits per data pixel
NAXIS   =                    3 / number of data axes
NAXIS1  =                   70 / length of data axis 1
NAXIS2  =                   46 / length of data axis 2
NAXIS3  =                    3 / length of data axis 3
EXTEND  =                    T / FITS dataset may contain extensions
COMMENT   FITS (Flexible Image Transport System) format is defined in 'Astronomy
COMMENT   and Astrophysics', volume 376, page 359; bibcode: 2001A&A...376..359H
BSCALE  =                    1
BZERO   =                    0


COMMENT   Header mods by TL
COMMENT   abc
COMMENT

        # TITLE 1 / ' @ :=+~*
HISTORY   mod2
HISTORY
HISTORY
COMMENT
COMMENT   abc

COMMENT
END
```
as input to the command `vips add sample.fits sample.fits sample2.fits` the resulting SIF file header is
```
SIMPLE  =                    T / file does conform to FITS standard
BITPIX  =                   16 / number of bits per data pixel
NAXIS   =                    3 / number of data axes
NAXIS1  =                   70 / length of data axis 1
NAXIS2  =                   46 / length of data axis 2
NAXIS3  =                    3 / length of data axis 3
EXTEND  =                    T / FITS dataset may contain extensions
COMMENT   FITS (Flexible Image Transport System) format is defined in 'Astronomy
COMMENT   and Astrophysics', volume 376, page 359; bibcode: 2001A&A...376..359H
BZERO   =                32768 / offset data range to that of unsigned short
BSCALE  =                    1 / default scaling factor


COMMENT   Header mods by TL
COMMENT   abc
COMMENT

        # TITLE 1 / ' @ :=+~*
HISTORY   mod2
HISTORY
END
```
The missing records are
- the last empty spaces record
- any duplicates of COMMENT and HISTORY records

jcupitt:
Yes, it dedupes identical comments. So:

```
COMMENT  hello
COMMENT  hello
```

It'll remove the second one. Or:

```
        # TITLE 1 / ' @ :=+~*
        # TITLE 1 / ' @ :=+~*
```

Again, it'll remove one of these. If you don't do this, stuff like:

```
COMMENT   FITS (Flexible Image Transport System) format is defined in 'Astronomy
```

And section separators will get doubled up. Do you think it's essential to preserve *all* comment fields? I thought doubled section divider comments could be an issue for some readers.

It preserves all blank lines (zero length lines) as a special case.

jcupitt:
... I added something to make fits metadata keys unique, it should improve the `vips add a.fits b.fits c.fits` case (I think).

ewelot:
> Do you think it's essential to preserve all comment fields? I thought doubled section divider comments could be an issue for some readers.

It is not essential with respect to compliance with the FITS standard. But it contradicts the purpose of using the comment records. Especially the HISTORY records are meant to describe processing steps applied to an image to allow the reader to reproduce/verify the image reduction pipeline. This involves - sometimes - repeating processing steps, hence the presence of possibly identical records. This is of cause not obvious with my rather stupid `sample.fits` file .... ;-)

jcupitt:
OK, I removed the dedupe-identical-lines logic.

ewelot:
> ... I added something to make fits metadata keys unique, it should improve the `vips add a.fits b.fits c.fits` case (I think).

Sorry, it looks like I missed to emphasize that we should propagate allowed header records from the _first_ input FITS file only.
The idea is, that an image reduction pipeline is working on a FITS image which gets modified by successive operations. The header of the (first) FITS image is kept and gets additional header records like parameters used, measurements or comments added along the way of processing. Libvips should just propagate the header records of the first image. Metadata from second and every other FITS image loaded must be ignored. If necessary, dealing with these other images FITS headers must be done by the specific user program which (hopefully) has some knowledge about their context.

ewelot:
Testing the new version using the same `sample.fits` file and the same command `vips add sample.fits sample.fits sample2.fits` results in
```
SIMPLE  =                    T / file does conform to FITS standard
BITPIX  =                   16 / number of bits per data pixel
NAXIS   =                    3 / number of data axes
NAXIS1  =                   70 / length of data axis 1
NAXIS2  =                   46 / length of data axis 2
NAXIS3  =                    3 / length of data axis 3
EXTEND  =                    T / FITS dataset may contain extensions
COMMENT   FITS (Flexible Image Transport System) format is defined in 'Astronomy
COMMENT   and Astrophysics', volume 376, page 359; bibcode: 2001A&A...376..359H
BZERO   =                32768 / offset data range to that of unsigned short
BSCALE  =                    1 / default scaling factor


COMMENT   Header mods by TL
COMMENT   abc
COMMENT

        # TITLE 1 / ' @ :=+~*
HISTORY   mod2
HISTORY
COMMENT   abc
END
```
This shows that some of the comment lines (those we no real text content) are still being rejected. Not quite as expected but no big issue either.

jcupitt:
OK, I reverted the unique header id change. If I run `vips add hubble.fits hubble.fits x.fits` I now see an identical header in `x.fits` after processing.

It won't quite propagate fields from the first image only -- if you have 10 fields in the first and 20 in the second, the output image with have the fields 1 - 10 from the first image, and fields 11 - 20 from the second. I don't think we can implement your first-image-only rule within the libvips metadata system without some larger changes.

Could you share your `sample.fits`?

ewelot:
> It won't quite propagate fields from the first image only -- if you have 10 fields in the first and 20 in the second, the output image with have the fields 1 - 10 from the first image, and fields 11 - 20 from the second. I don't think we can implement your first-image-only rule within the libvips metadata system without some larger changes.

I see. That's probably not worth the effort. But at least I would like to see the header records of the first input image first, if that is possible.

> Could you share your `sample.fits`?

Sure. I am attaching a zipped version (FITS is not allowed directly):
[sample.fits.zip](https://github.com/libvips/libvips/files/10467621/sample.fits.zip)

jcupitt:
>  But at least I would like to see the header records of the first input image first, if that is possible.

Yes, you should see the first image fields first.

ewelot:
I have done some more testing. And it looks like I do get the header records from the first input image only written to output, which is better than expected :+1: 
Also I have created a set of FITS files with some annotated header records which could be helpful for reference and testing:
- test_sif.fits - a basic single image plane
- test_cube.fits - a basic single image cube (RGB cube)
- test_mef2.fits - a multi-extension FITS containing an image plane in the second HDU
- test_mef3.fits - a multi-extension FITS containing RGB image planes in distict HDUs

FITS test files together with a descriptive README file are packed in
[test_fits.zip](https://github.com/libvips/libvips/files/10474111/test_fits.zip)
 
At the moment I believe there is nothing left to be done.
Thanks a lot for taking care of this sideshow topic,
Thomas

jcupitt:
Oh, great! Let's add your test images to the libvips test suite and put in some more FITS tests. Would you be able to write some tests yourself? The existing FITS tests are here:

https://github.com/libvips/libvips/blob/master/test/test-suite/test_foreign.py#L920-L931

... pretty basic. Some metadata testing would be useful.

ewelot:
Well, I can try, but have no experience writing tests. It will probably take some time to get me into this business so i'll PM to you if I get something done next couple of days.

jcupitt:
Ah don't worry. Let's make another PR after this is merged and do the tests on that. Thanks for your advice on this one!

@kleisauke I think this is ready to merge now.

kleisauke (approved):

PR #3287, opened 2023-01-18
tiffload: increase sanity checks on tile dimensions

lovell:
TIFF tile dimensions must be a multiple of 16 ~~and no larger than the image to which they belong (AFAIK, TIFF files don't have a "viewport" concept)~~.

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=55036 and https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=55148

We might want to backport this to 8.14 also.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,7 @@
 
 - add support for target_clones attribute [lovell]
 	* use with (un)premultiply for ~10% perf gain on AVX CPUs
+- increase sanity checks on TIFF tile dimensions [lovell]
 
 9/1/23 8.14.1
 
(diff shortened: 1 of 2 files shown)


jcupitt:
lgtm!

lovell:
It looks like one of the tests is writing a TIFF that cannot be read back with these checks - I'm taking a look now.

lovell:
The TIFF spec allows tiles to be larger than the image dimensions but recommends against doing so, however this clearly occurs in the real world.

I've modified the existing `> 10000` sanity check to also round up the longest image edge to the nearest 256 (a very common tile size) and use that as a possibly lower limit.

kleisauke (approved):
LGTM!

kleisauke:
Backport to [`8.14`](https://github.com/libvips/libvips/tree/8.14) available as #3290.

PR #3288, opened 2023-01-18
remove duplicated vips_dbuf_destroy declaration in headers,

UrielCh:
Hi,

I'm trying to write a libvips binding for deno, so have to map all functions.

the following declaration in `dbuf.h` is present 2 times:

```C
VIPS_API
void vips_dbuf_destroy( VipsDbuf *dbuf );
```

that the only duplicate exported C API.


jcupitt:
Thank you for reporting this dumbness, @UrielCh !

PR #3302, opened 2023-01-22
Add fast path to extract_band and bandjoin for single uchar, up to ~60% faster for 1 and 3 channels

lovell:
Such images are common and using these two functions for single channel remove/add is also common.

Removing the inner `z` loop allows the compiler to auto-vectorise the outer `x` loop.

Here are some `callgrind` stats for extracting a single channel from a 3 channel 1.4 megapixel image:

Before:
```
28,466,438 (19.51%)  ../libvips/conversion/extract.c:vips_extract_band_buffer [/usr/local/lib/x86_64-linux-gnu/libvips.so.42.17.0]
```
After:
```
 8,568,560 ( 6.80%)  ../libvips/conversion/extract.c:vips_extract_band_buffer [/usr/local/lib/x86_64-linux-gnu/libvips.so.42.17.0]
```

...and some stats for joining a channel to the same 3 channel image:

Before:
```
71,131,060 (36.56%)  ../libvips/conversion/bandjoin.c:vips_bandjoin_buffer [/usr/local/lib/x86_64-linux-gnu/libvips.so.42.17.0]
```

After:
```
51,234,183 (29.32%)  ../libvips/conversion/bandjoin.c:vips_bandjoin_buffer [/usr/local/lib/x86_64-linux-gnu/libvips.so.42.17.0]
```

I haven't done it here, but it might be worth adding a fast path for 3 channel joining (`ips == 3`) as well as 1 channel, as RGB+A is a common scenario. Happy to add to this PR or as a follow-up.

There is already a fast path for `vips_bandjoin_const` and this approach keeps things consistent.

https://github.com/libvips/libvips/blob/07d3a53f0d29fa5e611ed372b879117bb53af3de/libvips/conversion/bandjoin.c#L328

### DIFF
--- a/libvips/conversion/bandjoin.c
+++ b/libvips/conversion/bandjoin.c
@@ -135,7 +135,7 @@ vips_bandjoin_buffer( VipsBandarySequence *seq,
 				q1 += ops;
 			}
 
-			q += 3;
+			q += ips;
 		}
 		else {
 			for( x = 0; x < width; x++ ) {
--- a/libvips/conversion/bandjoin.c
+++ b/libvips/conversion/bandjoin.c
@@ -123,7 +123,7 @@ vips_bandjoin_buffer( VipsBandarySequence *seq,
 				q1 += ops;
 			}
 
-			q += 1;
+			q += ips;
 		}
 		else if( ips == 3 ) {
 			for( x = 0; x < width; x++ ) {
--- a/libvips/conversion/bandjoin.c
+++ b/libvips/conversion/bandjoin.c
@@ -131,7 +131,7 @@ vips_bandjoin_buffer( VipsBandarySequence *seq,
 				q1[1] = p1[1];
 				q1[2] = p1[2];
 
-				p1 += 3;
+				p1 += ips;
 				q1 += ops;
 			}
 
--- a/libvips/conversion/bandjoin.c
+++ b/libvips/conversion/bandjoin.c
@@ -116,15 +116,38 @@ vips_bandjoin_buffer( VipsBandarySequence *seq,
 		q1 = q;
 		p1 = p[i];
 
-		for( x = 0; x < width; x++ ) {
-			for( z = 0; z < ips; z++ )
-				q1[z] = p1[z];
+		if( ips == 1 ) {
+			for( x = 0; x < width; x++ ) {
+				q1[0] = p1[x];
 
-			p1 += ips;
-			q1 += ops;
+				q1 += ops;
+			}
+
+			q += 1;
 		}
+		else if( ips == 3 ) {
+			for( x = 0; x < width; x++ ) {
+				q1[0] = p1[0];
+				q1[1] = p1[1];
+				q1[2] = p1[2];
 
-		q += ips;
+				p1 += 3;
+				q1 += ops;
+			}
+
+			q += 3;
+		}
+		else {
+			for( x = 0; x < width; x++ ) {
+				for( z = 0; z < ips; z++ )
+					q1[z] = p1[z];
+
+				p1 += ips;
+				q1 += ops;
+			}
+
+			q += ips;
+		}
 	}
 }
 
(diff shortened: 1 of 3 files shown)


libvips/conversion/bandjoin.c
```diff
@@ -116,15 +116,27 @@ vips_bandjoin_buffer( VipsBandarySequence *seq,
 		q1 = q;
 		p1 = p[i];
 
-		for( x = 0; x < width; x++ ) {
-			for( z = 0; z < ips; z++ )
-				q1[z] = p1[z];
+		if( ips == 1 ) {
+			for( x = 0; x < width; x++ ) {
+				q1[0] = p1[0];
```
jcupitt, line 121: You could swap this for `p1[x]` and get rid of the `p1 += 1` below, it should save a little more.

libvips/conversion/extract.c
```diff
@@ -355,17 +355,27 @@ vips_extract_band_buffer( VipsBandarySequence *seq,
...
 
 	p = in[0] + extract->band * es;
 	q = out;
-	for( x = 0; x < width; x++ ) {
-		for( z = 0; z < ops; z++ )
-			q[z] = p[z];
+	if( ops == 1 ) {
+		for( x = 0; x < width; x++ ) {
+			q[0] = p[0];
+			p += ips;
```
jcupitt, line 367: And here it could be `q[x]` and no `+= 1`.

jcupitt:
Nice! I added two possible tiny speedups (I've not actually timed them).

lovell:
I've added John's suggestions. They don't appear to offer any additional performance improvement so I guess the compiler had already realised the two variables were in step.

I also added the fast path for joining 3 channels, which makes the common RGB+A scenario much faster:

Before:
```
71,131,060 (36.56%)  ../libvips/conversion/bandjoin.c:vips_bandjoin_buffer [/usr/local/lib/x86_64-linux-gnu/libvips.so.42.17.0]
```
After:
```
22,813,791 (15.59%)  ../libvips/conversion/bandjoin.c:vips_bandjoin_buffer [/usr/local/lib/x86_64-linux-gnu/libvips.so.42.17.0]
```

libvips/conversion/bandjoin.c
```diff
@@ -116,15 +116,38 @@ vips_bandjoin_buffer( VipsBandarySequence *seq,
...
+		if( ips == 1 ) {
+			for( x = 0; x < width; x++ ) {
+				q1[0] = p1[x];
 
-			p1 += ips;
-			q1 += ops;
+				q1 += ops;
+			}
+
+			q += 1;
```
kleisauke, line 126:
```suggestion
			q += ips;
```
To make it consistent with the `else` statement, although I'm not sure if that would cause a different codegen.

libvips/conversion/bandjoin.c
```diff
@@ -116,15 +116,38 @@ vips_bandjoin_buffer( VipsBandarySequence *seq,
...
 		}
+		else if( ips == 3 ) {
+			for( x = 0; x < width; x++ ) {
+				q1[0] = p1[0];
+				q1[1] = p1[1];
+				q1[2] = p1[2];
+
+				p1 += 3;
```
kleisauke, line 134:
```suggestion
				p1 += ips;
```
ditto.

libvips/conversion/bandjoin.c
```diff
@@ -116,15 +116,38 @@ vips_bandjoin_buffer( VipsBandarySequence *seq,
...
+				q1 += ops;
+			}
 
-		q += ips;
+			q += 3;
```
kleisauke, line 138:
```suggestion
			q += ips;
```
ditto.

kleisauke (approved):
Great! I left a few minor comments inline, but I'm not sure if that would affect performance.

kleisauke:
FWIW, I see no difference in assembly output when I apply my suggestions locally on both GCC 12.2.1 and Clang 15.0.7. Tested with `CFLAGS="-O3 -save-temps"`.

kleisauke:
Thank you! :+1:

PR #3309, opened 2023-01-23
colour: add AVX-specific XYZ2Lab implementation, ~10% faster

lovell:
The XYZ to LAB conversion is most commonly used for colour difference calculations and is often their bottleneck. It is also used for the smartcrop and sharpen operations. ICC profile transforms can use LAB as the connection space, although XYZ is usually fast/good enough.

Before:
```
127,870,765 (55.01%)  ../libvips/colour/XYZ2Lab.c:vips_XYZ2Lab_line [/usr/local/lib/x86_64-linux-gnu/libvips.so.42.17.0]
```
After:
```
115,085,981 (52.39%)  ../libvips/colour/XYZ2Lab.c:vips_XYZ2Lab_line.avx.0 [/usr/local/lib/x86_64-linux-gnu/libvips.so.42.17.0]
```

This adds up to ~4KB to the binary size (debug build, worst case).

This PR also moves the `GOnce` status to live in same scope as the lookup table it protects, which helps provide a small speed-up as its status no longer has to be checked for every pixel when converting a line.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,7 @@
 
 - add support for target_clones attribute [lovell]
 	* use with (un)premultiply for ~10% perf gain on AVX CPUs
+	* use with XYZ to LAB colourspace conversion for ~10% perf gain on AVX CPUs
 - dedupe FITS header write [ewelot]
 - add fast path to extract_band and bandjoin for uchar images [lovell]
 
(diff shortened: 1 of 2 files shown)


kleisauke (approved):
Nice! LGTM.

PR #3311, opened 2023-01-23
tiffload: sanitise JPEG dimensions before decompression

lovell:
This PR ensures we call `jpeg_calc_output_dimensions()` to populate the `cinfo` struct with dimensions before calling `jpeg_start_decompress()`, like jpegload already does, to prevent libjpeg from allocating memory based on untrusted input.

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=55111

I haven't added an entry to the changelog yet as we might want to backport this to 8.14.

### DIFF
--- a/libvips/foreign/tiff2vips.c
+++ b/libvips/foreign/tiff2vips.c
@@ -1949,7 +1949,7 @@ rtiff_decompress_jpeg_run( Rtiff *rtiff, j_decompress_ptr cinfo,
                 break;
         }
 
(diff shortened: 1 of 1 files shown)


kleisauke (approved):
LGTM!

kleisauke:
Backport (+ changelog entry) to [`8.14`](https://github.com/libvips/libvips/tree/8.14) available as #3314.

PR #3318, opened 2023-01-28
radiance: clean-up and update

kleisauke:
This PR started with resolving these Clang warnings (commit 5d925edf9257ae298e2eea324769d21348924082):
```
../libvips/foreign/radiance.c:429:1: warning: a function definition without a prototype is deprecated in all versions of C and is not supported in C2x [-Wdeprecated-non-prototype]
resolu2str(buf, rp)             /* convert resolution struct to line */
^
../libvips/foreign/radiance.c:446:1: warning: a function definition without a prototype is deprecated in all versions of C and is not supported in C2x [-Wdeprecated-non-prototype]
str2resolu(rp, buf)             /* convert resolution line to struct */
^
2 warnings generated.
```

But then I realised that these sources were actually bundled. So, I also cleaned up the unused macros/externs (commit 7dbd710970162370994906937e57ffb094665bdb) and re-synced this with Radiance 5.3 (commit 0764c5840319d58a301efa8b222f747e9416a84c).

### DIFF
--- a/libvips/foreign/radiance.c
+++ b/libvips/foreign/radiance.c
@@ -31,6 +31,7 @@
  * 	- revise for VipsConnection
  * 28/1/23 kleisauke
  * 	- clean-up unused macros/externs
+ * 	- sync with radiance 5.3
  */
 
 /*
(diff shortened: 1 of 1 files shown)


jcupitt:
LGTM!

Yes, Greg gave us permission to copy-paste this code rather than having to implement our own reader. Though I don't know if many people use it :/

PR #3324, opened 2023-02-03
Bits per sample

MathemanFlo:
Implements https://github.com/libvips/libvips/discussions/3321

- I added a `VIPS_META_BITS_PER_SAMPLE` flag which contains the bits per pixel of each channel. Hope you are okay with the name. Alternative would be `VIPS_META_BITS_PER_CHANNEL` or `VIPS_META_BITS_PER_BITDEPTH`. Avoided the latter because it is unclear whether it is per channel or the total color depth.
- Support for tiffs via `TIFFTAG_BITSPERSAMPLE`. According to [libtiff reference](https://www.awaresystems.be/imaging/tiff/tifftags/bitspersample.html) a tiff file could contain different bits per sample for each channel but libtiff does not support that, so it is okay to only read the tag once. 
- Support for libpng and libspng. [libpng reference](https://refspecs.linuxfoundation.org/LSB_5.0.0/LSB-TrialUse/LSB-TrialUse/libpng15-png-get-ihdr.html)

I noticed that the code for png and gif contains `"palette-bit-depth"` which is very similar to `VIPS_META_BITS_PER_SAMPLE`.
Haven't looked at gif and webp yet, but will asap.

### DIFF
--- a/libvips/foreign/nsgifload.c
+++ b/libvips/foreign/nsgifload.c
@@ -10,6 +10,8 @@
  * 	- avoid minimise after mapping -- not reliable on Win32
  * 25/1/23 kleisauke
  * 	- set interlaced=1 for interlaced images
+ * 13/3/23 MathemanFlo
+ *  - add bits per sample metadata
  */
 
 /*
--- a/libvips/foreign/nsgifload.c
+++ b/libvips/foreign/nsgifload.c
@@ -334,6 +334,8 @@ vips_foreign_load_nsgif_set_header( VipsForeignLoadNsgif *gif,
 
 	vips_image_set_int( image, "palette-bit-depth", 
 		ceil( log2( colours ) ) ); 
+	vips_image_set_int( image, VIPS_META_BITS_PER_SAMPLE, 
+		ceil( log2( colours ) ) ); 
 
 	/* Let our caller know if the GIF is interlaced.
 	 */
--- a/libvips/foreign/nsgifload.c
+++ b/libvips/foreign/nsgifload.c
@@ -11,7 +11,7 @@
  * 25/1/23 kleisauke
  * 	- set interlaced=1 for interlaced images
  * 13/3/23 MathemanFlo
- *  - add bits per sample metadata
+ * 	- add bits per sample metadata
  */
 
 /*
--- a/libvips/foreign/nsgifload.c
+++ b/libvips/foreign/nsgifload.c
@@ -334,10 +334,13 @@ vips_foreign_load_nsgif_set_header( VipsForeignLoadNsgif *gif,
 		}
 	}
 
-	vips_image_set_int( image, "palette-bit-depth", 
-		ceil( log2( colours ) ) ); 
-	vips_image_set_int( image, VIPS_META_BITS_PER_SAMPLE, 
-		ceil( log2( colours ) ) ); 
+	vips_image_set_int( image, VIPS_META_BITS_PER_SAMPLE,
+		ceil( log2( colours ) ) );
+
+	/* Deprecated "palette-bit-depth" use "bits-per-sample" instead.
+	 */
+	vips_image_set_int( image, "palette-bit-depth",
+		ceil( log2( colours ) ) );
 
 	/* Let our caller know if the GIF is interlaced.
 	 */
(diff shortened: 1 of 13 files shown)


kleisauke:
> I noticed that the code for png and gif contains `"palette-bit-depth"` which is very similar to `VIPS_META_BITS_PER_SAMPLE`.

Good point, I think the best way to deal with this is just to deprecate `palette-bit-depth` in favor of `bits-per-sample` (`VIPS_META_BITS_PER_SAMPLE`). Deprecated header fields are tracked here: https://github.com/libvips/libvips/blob/76f2178a31d62ebc84e30a0d0a04216271bd7e34/libvips/iofuncs/header.c#L1454-L1460

libvips/foreign/nsgifload.c
```diff
@@ -332,7 +332,7 @@ vips_foreign_load_nsgif_set_header( VipsForeignLoadNsgif *gif,
 		}
 	}
 
-	vips_image_set_int( image, "palette-bit-depth", 
+	vips_image_set_int( image, VIPS_META_BITS_PER_SAMPLE,
```
kleisauke, line 335: I think we still need to set the `palette-bit-depth` field for compatibility.

libvips/foreign/spngload.c
```diff
@@ -269,11 +271,7 @@ vips_foreign_load_png_set_header( VipsForeignLoadPng *png, VipsImage *image )
 		vips_image_set_blob_copy( image, VIPS_META_EXIF_NAME, 
 			exif.data, exif.length );
 
-	/* Attach original palette bit depth, if any, as metadata.
-	 */
-	if( png->ihdr.color_type == SPNG_COLOR_TYPE_INDEXED )
-		vips_image_set_int( image, 
-			"palette-bit-depth", png->ihdr.bit_depth );
```
kleisauke, line 276: ditto

libvips/foreign/vipspng.c
```diff
@@ -592,11 +596,6 @@ png2vips_header( Read *read, VipsImage *out )
 				return( -1 ); 
 	}
 
-	/* Attach original palette bit depth, if any, as metadata.
-	 */
-	if( color_type == PNG_COLOR_TYPE_PALETTE )
-		vips_image_set_int( out, "palette-bit-depth", bitdepth );
-
```
kleisauke, line 599: ditto

kleisauke (commented):
Thanks for working on this! I left a comment about compatibility regarding the `palette-bit-depth` field (for the same reason we still set need to set the `gif-loop` and `gif-delay` fields, unfortunately).

> * As far as I understand WEBP can not have different values for bits per sample as assumed in [Accessing tiff's bits per sample #3321 (comment)](https://github.com/libvips/libvips/discussions/3321#discussioncomment-4849971) . According to my research it is always 8 bit per channel.

Maybe we should always set it to 8 for WebP images?

> * We potentially could support HEIC/AVIF. Look for `"heif-bitdepth"` and `bits_per_pixel` in heifload.c
> * We could also support JPEG2000

I think this would be also useful to have :+1:. JXL images could support this too (via `jxl->info.bits_per_sample`).

libvips/foreign/nsgifload.c
MathemanFlo: I fear I miunderstood how deprecation works. I added the deleted code lines again.

kleisauke:
> Added support for jpeg xl, heif/avif, jpeg 2000

Great!

> This is indeed possible but then the question arises why do it for WebP and not for all other file formats like jpeg (which also only supports 8 bits per pixel per channel)

I see. It probably makes more sense to avoid setting this for images that do not support it.

kleisauke (approved):
LGTM, I've updated the PR to support this within the *magick loader and deprecated the `heif-bitdepth` meta field while we're here. I also covered this with a few unit tests.

I'll leave this here for a few days for any further feedback from other maintainers before merging. Thank you!

lovell (approved):
LGTM, thanks both! There will be (optional) support for 12-bit JPEG in the forthcoming libjpeg-turbo v3, but that's very much a future possible enhancement.

PR #3326, opened 2023-02-06
Improve cpp wrapper

MathemanFlo:
I added two missing functions to the cplusplus wrapper.

Question: Besides `inplace` which I use for pixel modifications afterwards the c++ wrapper is missing methods for functions like `vips_draw_circle1` (with focus on the 1 versions). Would it be okay if I implemented overloads for circle, ... similar to `ifthenelse`. For example 
```
void draw_circle( ... )
{
  draw_circle( to_vector( ink), ... ); 
}
```

### DIFF
--- a/cplusplus/VImage.cpp
+++ b/cplusplus/VImage.cpp
@@ -781,22 +781,20 @@ VImage
 VImage::thumbnail_buffer( void *buf, size_t len, int width, VOption *options )
 {
 	VipsBlob *blob;
+	VImage out;
 
 	/* We don't take a copy of the data or free it.
 	 */
 	blob = vips_blob_new( NULL, buf, len );
+	options = (options ? options : VImage::option())->
+		set( "buffer", blob )->
+		set( "width", width )->
+		set( "out", &out );
+	vips_area_unref( VIPS_AREA( blob ) );
 
-	try
-	{
-		VImage thumbnail = thumbnail_buffer( blob, width, options );
-		vips_area_unref( VIPS_AREA( blob ) );
-		return( thumbnail );
-	}
-	catch ( VError &ex )
-	{
-		vips_area_unref( VIPS_AREA( blob ) );
-		throw; // rethrow exception
-	}
+	call( "thumbnail_buffer", options );
+
+	return( out );
 }
 
 VRegion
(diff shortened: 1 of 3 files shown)


jcupitt:
Hi @MathemanFlo,

Thanks, this looks useful.

Sure, some extra overloads for `draw_circle` etc. sound good.

kleisauke:
FWIW, CIFuzz failure was supposed to be fixed with PR https://github.com/google/oss-fuzz/pull/9575, I'm not sure why it doesn't test against the latest OSS-Fuzz master branch.

macOS CI failure will be fixed with PR https://github.com/libvips/libvips/pull/3327.

MathemanFlo:
- Now the C++ wrapper for the drawing functions is on par with the underlying C API
- I also implemented a void* buffer version  (instead of a `VipsBlob`) for `thumbnail_buffer`. It is now identical to the C API and easier to use.

cplusplus/VImage.cpp
```diff
@@ -777,6 +777,28 @@ VImage::write_to_target( const char *suffix, VTarget target,
...
+	{
+		VImage thumbnail = thumbnail_buffer( blob, width, options );
+		vips_area_unref( VIPS_AREA( blob ) );
+		return( thumbnail );
+	}
+	catch ( VError &ex )
+	{
+		vips_area_unref( VIPS_AREA( blob ) );
+		throw; // rethrow exception
+	}
```
kleisauke, line 799: I think this try-catch can be avoided like this:
```suggestion
	VImage out;

	/* We don't take a copy of the data or free it.
	 */
	blob = vips_blob_new( NULL, buf, len );
	options = (options ? options : VImage::option())->
		set( "buffer", blob )->
		set( "width", width )->
		set( "out", &out );
	vips_area_unref( VIPS_AREA( blob ) );

	call( "thumbnail_buffer", options );

	return( out );
```

See for example the implementation of `VImage::new_from_buffer`:
https://github.com/libvips/libvips/blob/80f1d89365ecd69b268e7ffd22fed8b0d4f5fa62/cplusplus/VImage.cpp#L596-L606

MathemanFlo:
@kleisauke Thank you for your suggestion. I thought the blob must not be destroyed before `call`.

kleisauke (approved):
Thank you!

MathemanFlo:
Thank you for letting me participate

PR #3327, opened 2023-02-06
CI: fix test failure on macOS

kleisauke:
Remove the ICC profile after the image is converted to 16-bit.

Ensures that pixels are correctly interpreted and that the maximum absolute difference remains consistent across different colour management software (i.e. lcms and skcms).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,8 @@ jobs:
       - name: Install macOS dependencies
         if: runner.os == 'macOS'
         run: |
+          # FIXME: remove when the GitHub Actions runner image ships with Homebrew/homebrew-core#122689
+          brew update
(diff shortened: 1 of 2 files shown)


jcupitt:
Oh, does libjxl always apply any embedded profile on load? That's bad (or I've misunderstood).

Perhaps jxlload should not attach the icc profile? If it's been applied, we must not apply it again.

kleisauke:
I think the issue was that `jxlsave` would save the ICC profile even if it's incompatible with the input image format (i.e. `rgb16` in this case). Then, when `jxlload` tried to load the image, the pixels were misinterpreted.

kleisauke:
I'll land this as-is to make CI green again. Will investigate the ICC profile behavior within `jxlload` / `jxlsave` another time.

PR #3329, opened 2023-02-08
add "vector" switch to find_trim

jcupitt:
disables the median filter to make handling vector art more accurate

see https://github.com/libvips/libvips/discussions/2076

### DIFF
--- a/libvips/arithmetic/find_trim.c
+++ b/libvips/arithmetic/find_trim.c
@@ -5,6 +5,8 @@
  * 18/9/17 kleisauke 
  * 	- missing bandor
  * 	- only flatten if there is an alpha
+ * 8/2/23
+ *	- add @vector
  */
 
 /*
@@ -55,6 +57,7 @@ typedef struct _VipsFindTrim {
 	VipsImage *in;
 	double threshold;
 	VipsArrayDouble *background;
+	gboolean vector;
 
 	int left;
 	int top;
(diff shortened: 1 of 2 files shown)


lovell (approved):
:+1:

libvips/arithmetic/find_trim.c
```diff
@@ -268,6 +287,11 @@ vips_find_trim_init( VipsFindTrim *find_trim )
  *
  * @threshold defaults to 10. 
  *
+ * The detector is designed for photographic or compressed images where there
+ * is a degree of noise that needs filtering. If your images are synthetic
+ * (eg. rendered from vector art, perhjaps), set @vector to disable the
```
kleisauke, line 292:
```suggestion
 * (eg. rendered from vector art, perhaps), set @vector to disable the
```

kleisauke (approved):

jcupitt:
I just had a thought -- perhaps `vector` is too much like the things that turn vector instructions on and off? Maybe `synthetic` (as opposed to `natural`) might be better?

lovell:
I understand "synthetic" to be computer-generated, which is (increasingly) more than line art and/or vector drawings. `skip_median` is consistent with other options e.g. `skip_blanks` so that feels like a suitable does-what-it-says-on-the-tin option. The "no" prefix could be confused with "number".

jcupitt:
Thanks for the feedback! Sorry for the delay. 

In the end I agonised and changed it to `line_art`. I thought about `skip_median`, but that feels operational (ie. it says what it does in terms of the internal operation of the function) and I think options like are usually clearer named after their purpose (ie. why you'd use it).

Line art sounds a bit like why you'd chose PNG over JPEG for some image types, which is a similar choice.

PR #3346, opened 2023-02-16
Avoid retrieving PDFium via pkg-config

kleisauke:
PDFium doesn't ship with a pkg-config file, so it's recommend to use Meson's `find_library()` function to look for it. This would avoid the need to create `pdfium.pc`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,21 +159,11 @@ Untar to the libvips install prefix, for example:
     cd ~/vips
     tar xf ~/pdfium-linux.tgz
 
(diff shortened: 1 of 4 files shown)


kleisauke:
I'll abandon this PR, will revisit this when Meson fixes issue https://github.com/mesonbuild/meson/issues/10172.

FWIW, commit 35008b3b683ecad9a8ad7f63e9c0ed3a99033edd doesn't seem to work with Clang.

Issue #3159, opened 2022-11-14
About cmyk color image


jishi92:
## Questions, enhancements, tips, etc.

I resize/thumbnail this jpeg image and find the result color is changed.
However, it‘s not changed when imagick is used.
So is it a bug or should I change some parameter?
## Bug report

origin:  
![cmyk-test](https://user-images.githubusercontent.com/22196414/201678606-8e23e44e-6051-4ab2-8759-1add4156ee17.jpeg)

after thumbnail=400x400:  

![cmyk400x400](https://user-images.githubusercontent.com/22196414/201678969-d86d9024-5e19-467f-b248-45551ac1e4a8.jpeg)

**Environment**
OS: centos-8.4, MacOS 12.4
Vips: vips-8.12.1

jcupitt:
Hello @jishi92,

There was a bug in vipsthumbnail when used with non-sRGB images and an export profile. If you update to the current stable version (8.13.3) it should be fixed.

I see:

```
$ vipsthumbnail cmyk6.jpeg --size 400
```

To make:

![tn_cmyk6](https://user-images.githubusercontent.com/580843/201690383-3a47fe74-1c06-4dc0-bd5b-f4746cdf912d.jpg)

romanzks:
I have a bug reproduced in the version 8.13.3 even without resizing
```
vipsthumbnail input.jpeg --size 4488 -o output.jpeg
```
But interestingly, when using Linear light, the bug is not reproduced
```
vipsthumbnail input.jpeg --size 4488 --linear -o output.jpeg
```
I think this issues is related to https://github.com/libvips/libvips/issues/3150

kleisauke:
Perhaps issue https://github.com/libvips/php-vips/issues/138 is also related here? The linked commit in that issue was reverted with PR #3027. Processing in linear space (i.e. `icc_import` on the begin, `icc_export` on the end) would indeed fix that.

jishi92:
Hey @jcupitt @romanzks @kleisauke ,
thanks for your answers. I probably understand.
So I should set `icc_export = 'cmyk' ` will solve my question ?
By the way,I don't know what's the meaning of `--linear` in @romanzks 's case.

romanzks:
@jishi92 regarding linear light https://github.com/libvips/libvips/wiki/HOWTO----Image-shrinking#linear-light

I don't think that icc_export = 'cmyk' is solution. If it works in one case, in not means that will work with other images.

jcupitt:
[...]

PR #3369, opened 2023-03-01
thumbnail: skip colourspace conversion when needed

kleisauke:
i.e. if the image needs to transformed with a pair of ICC profiles.

Resolves: #3159.

### DIFF
--- a/libvips/resample/thumbnail.c
+++ b/libvips/resample/thumbnail.c
@@ -37,6 +37,8 @@
  * 	- add a seq to thumbnail_image to stop cache thrashing
  * 28/4/22
  * 	- add fail_on
+ * 1/3/23 kleisauke
+ *	- skip colourspace conversion when needed
  */
 
 /*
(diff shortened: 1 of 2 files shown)


libvips/resample/thumbnail.c
```diff
@@ -697,17 +703,9 @@ vips_thumbnail_build( VipsObject *object )
...
 	preshrunk_page_height = vips_image_get_page_height( in );
 
-	/* RAD needs special unpacking.
-	 */
-	if( in->Coding == VIPS_CODING_RAD ) {
-		g_info( "unpacking Rad to float" );
-
-		/* rad is scrgb.
-		 */
-		if( vips_rad2float( in, &t[12], NULL ) )
```
lovell, line 707: Is this rad2float conversion no longer required?

kleisauke: This was removed since the same thing was done in `colourspace`.
https://github.com/libvips/libvips/blob/65c99219af5b3f4a8e9d4a9daf7886a6c97cf619/libvips/colour/colourspace.c#L556-L563

Though, `colourspace` is not always called after this change, so let me revert this change.

lovell (approved):

PR #3370, opened 2023-03-03
add an experimental "nocache" option

jcupitt:
This experimental PR adds a "nocache" option to all operations. 

The idea is that if this flag is set, the operation is always executed and never fetched from the operation cache. If the operation succeeds, any old cache entry is replaced.

This is supposed to be useful for things like opening file where you know that the file might have changed since you last opened it. Setting `nocache` ensures that you always get a fresh file open.

See https://github.com/libvips/ruby-vips/issues/360 for the reporting issue.

Questions:

- I think it's threadsafe, but I'm often wrong heh
- This is a bit obtrusive, since it appears *everywhere*. We could tag this as DEPRECATED, or perhaps internal API, and only expose it in `new_from_file` and friends.
- I've not done tests or docs
- Any thoughts?

### DIFF
--- a/libvips/iofuncs/cache.c
+++ b/libvips/iofuncs/cache.c
@@ -896,7 +896,9 @@ vips_cache_operation_buildp( VipsOperation **operation )
 	vips_object_print_dump( VIPS_OBJECT( *operation ) );
 #endif /*VIPS_DEBUG*/
 
-	if( (hit = vips_cache_operation_lookup( *operation )) ) {
+	hit = vips_cache_operation_lookup( *operation );
+	if( hit && 
+		!(*operation)->nocache ) {
 #ifdef VIPS_DEBUG
 		printf( "vips_cache_operation_buildp: cache hit %p\n", hit );
 #endif /*VIPS_DEBUG*/
@@ -911,6 +913,16 @@ vips_cache_operation_buildp( VipsOperation **operation )
 
 		if( vips_object_build( VIPS_OBJECT( *operation ) ) ) 
 			return( -1 );
+
+		/* If this is nocache, there might be an old cache entry we
+		 * must replace.
+		 */
+		if( hit ) {
+			g_mutex_lock( vips_cache_lock );
+			vips_cache_remove( hit );
+			g_mutex_unlock( vips_cache_lock );
+			hit = NULL;
+		}
 
 		vips_cache_operation_add( *operation ); 
 	}
--- a/ChangeLog
+++ b/ChangeLog
@@ -14,6 +14,7 @@
 - allow negative line spacing in text [donghuikugou]
 - add VIPS_META_BITS_PER_SAMPLE metadata, deprecate the
   "palette-bit-depth" and "heif-bitdepth" meta fields [MathemanFlo]
+- add "nocache" to operation class [jcupitt]
 
 TBD 8.14.2
 
--- a/libvips/include/vips/operation.h
+++ b/libvips/include/vips/operation.h
@@ -80,6 +80,10 @@ typedef struct _VipsOperation {
 	 */
 	int pixels;
 
+	/* Set to force execution of this operation.
+	 */
+	gboolean nocache;
+
 } VipsOperation;
 
 typedef struct _VipsOperationClass {
--- a/libvips/iofuncs/operation.c
+++ b/libvips/iofuncs/operation.c
@@ -596,6 +596,8 @@ vips_operation_class_init( VipsOperationClass *class )
 
 	gobject_class->finalize = vips_operation_finalize;
 	gobject_class->dispose = vips_operation_dispose;
+	gobject_class->set_property = vips_object_set_property;
+	gobject_class->get_property = vips_object_get_property;
 
 	vobject_class->build = vips_operation_build;
 	vobject_class->summary = vips_operation_summary;
@@ -605,6 +607,13 @@ vips_operation_class_init( VipsOperationClass *class )
 
 	class->usage = vips_operation_usage;
 	class->get_flags = vips_operation_real_get_flags;
+
+	VIPS_ARG_BOOL( class, "nocache", 200, 
+		_( "No cache" ), 
+		_( "Force execution of this operation" ),
+		VIPS_ARGUMENT_OPTIONAL_INPUT,
+		G_STRUCT_OFFSET( VipsOperation, nocache ),
+		FALSE );
 
 	vips_operation_signals[SIG_INVALIDATE] = g_signal_new( "invalidate",
 		G_TYPE_FROM_CLASS( class ),


jcupitt:
... the prog I used for testing:

```ruby
#!/usr/bin/ruby

require "fileutils"
require "vips"

if ARGV.length != 2
  puts "usage: ./nocache.rb IMAGE-FILE-1 IMAGE-FILE-2"
  exit 1
end

if File.exists?("fred")
  puts "file 'fred' exists, delete it first"
  exit 1
end

puts "copying #{ARGV[0]} to fred ..."
FileUtils.cp(ARGV[0], "fred")

image1 = Vips::Image.new_from_file("fred")
puts "fred.width = #{image1.width}"

puts "copying #{ARGV[1]} to fred ..."
FileUtils.cp(ARGV[1], "fred")

puts "plain image open ..."
image2 = Vips::Image.new_from_file("fred")
puts "fred.width = #{image2.width}"

puts "opening with nocache ..."
image2 = Vips::Image.new_from_file("fred", nocache: true)
puts "fred.width = #{image2.width}"

File.delete("fred")
```

I see:

```
$ ./nocache.rb ~/pics/k2.jpg ~/pics/shark.jpg 
copying /home/john/pics/k2.jpg to fred ...
fred.width = 1450
copying /home/john/pics/shark.jpg to fred ...
plain image open ...
fred.width = 1450
opening with nocache ...
fred.width = 537
```

ie. setting `nocache` forces a cache update.

jcupitt:
I renamed it as `force`, since the operation is still cached, it just forces a cache update. The other questions remain, of course!

elohanlon:
Hi @jcupitt.  Do you think that `refresh` might be a clearer name for the option?

`Vips::Image.new_from_file("fred", refresh: true)`

Here's my thinking:

- `force` sounds to me like it's forcing the operation to go through when it would have otherwise failed to run (even though the operation would have still run even if `force` was not present)
- `refresh` sounds like it's running an operation that would have run with or without the `refresh` operation, but `refresh` triggers a refresh of the cache

lovell:
Borrowing from HTTP semantics, the term `revalidate` is used for an operation that should both ignore and update a cache.

jcupitt:
I like `revalidate`, let's try that.

kleisauke:
> * This is a bit obtrusive, since it appears _everywhere_. We could tag this as DEPRECATED, or perhaps internal API, and only expose it in `new_from_file` and friends.

Perhaps another alternative is to do this only for `VipsForeignLoad`? See e.g. commit https://github.com/kleisauke/libvips/commit/0a6b667b4899cce02b6232702be9ad8ea881f954, though I'm not sure how expensive that `VIPS_IS_FOREIGN_LOAD()` call is (in terms of performance).

jcupitt:
> Perhaps another alternative is to do this only for VipsForeignLoad? See e.g. commit https://github.com/kleisauke/libvips/commit/0a6b667b4899cce02b6232702be9ad8ea881f954

Yes, that's nicer!

It's a bit odd having a special case for a subclass in a base class. Perhaps `revalidate` could be a member of `VipsOperationClass`, false by default, and which `VipsForeignLoad` sets true in `_class_init`? You could skip the `IS_FOREIGN_LOAD` in cache build then, and it wouldn't feel like a strange corner case.

kleisauke:
I just noticed that `VIPS_OPERATION_NOCACHE` / `load->nocache` does the same thing.
https://github.com/libvips/libvips/blob/db02d9f949d36766d39c20b9b8c507d9523cc033/libvips/include/vips/foreign.h#L182-L184

So only exposing this setting is necessary, see e.g. commit https://github.com/kleisauke/libvips/commit/d2376b81ce4f50011f4d466d6783bd3ed94f0a84.

(I could push the commits in https://github.com/libvips/libvips/compare/add-nocache...kleisauke:add-nocache to the branch associated with this PR, if that looks good to you)

jcupitt:
I'm not sure that's the same. `nocache` stops things entering cache, but I don't think it will evict a cache entry that's already there.

kleisauke:
Ah, you're right. Though, the previous logic didn't really invalidate the old cache entry. For example, by changing that test program like this:
<details>
  <summary>Details</summary>

```diff
--- a/nocache.rb
+++ b/revalidate.rb
@@ -4,7 +4,7 @@ require "fileutils"
 require "vips"
 
 if ARGV.length != 2
-  puts "usage: ./nocache.rb IMAGE-FILE-1 IMAGE-FILE-2"
+  puts "usage: ./revalidate.rb IMAGE-FILE-1 IMAGE-FILE-2"
   exit 1
 end
 
@@ -26,8 +26,12 @@ puts "plain image open ..."
 image2 = Vips::Image.new_from_file("fred")
 puts "fred.width = #{image2.width}"
 
-puts "opening with nocache ..."
-image2 = Vips::Image.new_from_file("fred", nocache: true)
+puts "opening with revalidate ..."
+image2 = Vips::Image.new_from_file("fred", revalidate: true)
+puts "fred.width = #{image2.width}"
+
+puts "plain image open ..."
+image2 = Vips::Image.new_from_file("fred")
 puts "fred.width = #{image2.width}"
 
 File.delete("fred")
```
</details>

And invoking it with this:
```console
$ vips black x.jpg 100 100
$ vips black x2.jpg 200 200
$ ./revalidate.rb x.jpg x2.jpg
copying x.jpg to fred ...
fred.width = 100
copying x2.jpg to fred ...
plain image open ...
fred.width = 100
opening with revalidate ...
fred.width = 200
plain image open ...
fred.width = 100
```

i.e. that last one should print `fred.width = 200` if the old cache entry was successfully invalidated.

kleisauke:
... that can be fixed by ensuring we don't take the `revalidate` param into account in `vips_operation_hash()`, see e.g. commit https://github.com/kleisauke/libvips/commit/6cf21f042aedcab2c1c0e5b147dc9d0df05a1279.

jcupitt:
> ... that can be fixed by ensuring we don't take the revalidate param into account in `vips_operation_hash()`, s

Argh! You're right, I didn't think of that. D'oh.

It's a bit odd to have a special case like that inside the hasher. How about adding a new argument flag for "don't include in hash"?

kleisauke:
Indeed, that special case is a bit odd. I added `VIPS_ARGUMENT_NON_HASHABLE` with commit https://github.com/kleisauke/libvips/commit/e788c09e9bfca3cc3d779225db111b5e6978189d, which seems to work.

jcupitt:
I merged your commit and added a test. 

I made argument equal also check for nocache (not just the hash func) and tweaked lookup a bit.

jcupitt:
Oh dear, I think this is still not quite right. With this test prog:

```ruby
#!/usr/bin/ruby

# demo the "revalidate" feature in libvips 8.15

require "fileutils"
require "vips"

if ARGV.length != 2
  puts "usage: ./revalidate.rb IMAGE-FILE-1 IMAGE-FILE-2"
  exit 1
end

if File.exists?("fred")
  puts "file 'fred' exists, delete it first"
  exit 1
end

puts "copying #{ARGV[0]} to fred ..."
FileUtils.cp(ARGV[0], "fred")

image1 = Vips::Image.new_from_file("fred")
puts "fred.width = #{image1.width}"

puts "copying #{ARGV[1]} to fred ..."
FileUtils.cp(ARGV[1], "fred")

puts "plain image open ..."
image2 = Vips::Image.new_from_file("fred")
puts "fred.width = #{image2.width}"

puts "opening with revalidate ..."
image2 = Vips::Image.new_from_file("fred", revalidate: true)
puts "fred.width = #{image2.width}"

puts "opening again, should get cached entry ..."
image2 = Vips::Image.new_from_file("fred")
puts "fred.width = #{image2.width}"

File.delete("fred")
```

I see:

```
$ VIPS_TRACE=1 ./revalidate.rb ~/pics/k2.jpg ~/pics/shark.jpg 
copying /home/john/pics/k2.jpg to fred ...
vips cache+: jpegload filename="fred" out=((VipsImage*) 0x55f20dbae000) flags=((VipsForeignFlags) VIPS_FOREIGN_SEQUENTIAL) -
fred.width = 1450
copying /home/john/pics/shark.jpg to fred ...
plain image open ...
vips cache*: jpegload filename="fred" -
fred.width = 1450
opening with revalidate ...
vips_cache_operation_lookup: removing old entry
vips cache : jpegload filename="fred" out=((VipsImage*) 0x55f20dbae190) flags=((VipsForeignFlags) VIPS_FOREIGN_SEQUENTIAL) revalidate=TRUE -
fred.width = 537
opening again ...
vips cache+: jpegload filename="fred" out=((VipsImage*) 0x55f20dbae320) flags=((VipsForeignFlags) VIPS_FOREIGN_SEQUENTIAL) -
fred.width = 537
```

So the final open is not getting a cached result.

The problem is that setting `nocache` with `revalidate` is correctly forcing an old value out of cache, but it's also preventing that new result from being entered into cache.

I think we'll need a separate `revalidate` flag again,

jcupitt:
I made it a flag on the operation class rather than a member. It seems to work!

```
$ VIPS_TRACE=1 ./revalidate.rb ~/pics/k2.jpg ~/pics/shark.jpg 
copying /home/john/pics/k2.jpg to fred ...
vips cache+: jpegload filename="fred" out=((VipsImage*) 0x55f14f040000) flags=((VipsForeignFlags) VIPS_FOREIGN_SEQUENTIAL) -
fred.width = 1450
copying /home/john/pics/shark.jpg to fred ...
plain image open ...
vips cache*: jpegload filename="fred" -
fred.width = 1450
opening with revalidate ...
vips cache+: jpegload filename="fred" out=((VipsImage*) 0x55f14f040190) flags=((VipsForeignFlags) VIPS_FOREIGN_SEQUENTIAL) revalidate=TRUE -
fred.width = 537
opening again, should get cached entry ...
vips cache*: jpegload filename="fred" -
fred.width = 537
```

So the final open gets an updated result from cache.

jcupitt:
... I was looking again at the caching code, and I don't like having the complicated nocache / revalidate logic split between `vips_cache_operation_lookup()` and `vips_cache_operation_add()`. It would be cleaner if all that tricky code was moved into `vips_cache_operation_buildp()`.

We need to change the behaviour of lookup and add to make them both do just lookup and add and no extra stuff, but they are in the public API. 

How about making them both `static` and chopping them out of `operation.h`? They are not used anywhere else, so they don't need to be public. I guess we'd need some stubs for compatibility. Or maybe we should just change the behaviour? If they are unused, perhaps it doesn't matter.

jcupitt:
But maybe that's something for another PR.

kleisauke:
Great! Sorry for the confusion about `revalidate` and `nocache`, we should indeed treat them separately.

I think we can safely mark `vips_cache_operation_lookup()` and `vips_cache_operation_add()` as deprecated now (see e.g. commit https://github.com/kleisauke/libvips/commit/3d3f4d4440306e2c1613d6582e5a2d056bfb1067).

Should we update the PR description and title before landing this? I think it's already in good shape (i.e. not experimental).

jcupitt:
Good idea, I merged your deprecated patch and updated the top comment.

I'll do some quick docs and we're done, I think.

jcupitt:
OK, I added some docs, I think we're done on this one.

libvips/iofuncs/cache.c
```diff
@@ -780,6 +780,16 @@ vips_cache_trim( void )
...
  *
+ * Operators that have been tagged as invalid by the `invalidate` signal are
+ * removed from cache.
+ *
+ * Operators with the #VIPS_OPERATION_BLOCKED flag are never executed.
+ *
+ * Operators with the #VIPS_OPERATION_REVALIDATE flag are always executed and
+ * any old cache value is replaced.
+ *
+ * Operatoes with the #VIPS_OPERATION_NOCACHE flag are never cached.
```
kleisauke, line 791:
```suggestion
 * Operators with the #VIPS_OPERATION_NOCACHE flag are never cached.
```

libvips/iofuncs/cache.c
```diff
@@ -788,7 +788,7 @@ vips_cache_trim( void )
  * Operators with the #VIPS_OPERATION_REVALIDATE flag are always executed and
  * any old cache value is replaced.
  *
- * Operatoes with the #VIPS_OPERATION_NOCACHE flag are never cached.
+ * Operators with the #VIPS_OPERATION_NOCACHE flag are never cached.
```
jcupitt, line 791: Ahem oops, thanks!

Issue #3150, opened 2022-11-08
ICC transforms with XYZ PCS don't roundtrip correctly


jcupitt:
This works:

```
vips icc_import input.jpg t1.v
vips icc_export t1.v output.jpg
```

But this has a visible colour difference:

```
vips icc_import input.jpg t1.v --pcs xyz
vips icc_export t1.v output.jpg
```

See https://github.com/libvips/php-vips/issues/169

romanzks:
I did a little research and noticed that the bug has been reproducible since version 8.11.0.
In earlier versions, error "icc_import: no input profile" is issued.

jcupitt:
I did a little digging and I think the issue is that these two paths are not exactly equal:

```
device rgb -> icc_import(pcs=xyz) -> xyz 
device rgb -> icc_import(pcs=lab) -> lab2xyz -> xyz
```

ie. the lcms option to output xyz does not give exactly the same results as requesting lab, and then using the libvips lab2xyz. Perhaps lcms is using D50 lab, and libvips is using D65 lab?

I'll try reading the lcms sources and see if I can find the formula they are using.

kleisauke:
I just opened PR #3368 for this.

PR #3368, opened 2023-03-01
colour: use `cmsFLAGS_NOOPTIMIZE` to keep all precision

kleisauke:
Resolves: #3150.

> **Note**: this PR targets the [`8.14`](https://github.com/libvips/libvips/tree/8.14) branch.

PR #3373, opened 2023-03-04
colour: use floating-point APIs of lcms

kleisauke:
Avoids the need to pack the buffer of floats into lcms's fixed-point formats and ensures full precision internally in lcms.

Resolves: #3150.
Supersedes: #3368.

### DIFF
--- a/libvips/colour/icc_transform.c
+++ b/libvips/colour/icc_transform.c
@@ -802,21 +802,6 @@ vips_icc_import_build( VipsObject *object )
 }
 
 static void 
-decode_lab( float *in, float *out, int n )
-{
-	int i;
-
-	for( i = 0; i < n; i++ ) {
-		out[0] = in[0];
-		out[1] = in[1];
-		out[2] = in[2];
-
-		out += 3;
-		in += 3;
-	}
-}
-
-static void 
 decode_xyz( float *in, float *out, int n )
 {
 	int i;
@@ -831,10 +816,8 @@ decode_xyz( float *in, float *out, int n )
 	}
 }
 
-/* Process a buffer of data.
- */
 static void
-vips_icc_import_line( VipsColour *colour, 
+vips_icc_import_line_xyz( VipsColour *colour, 
 	VipsPel *out, VipsPel **in, int width )
 {
 	VipsIcc *icc = (VipsIcc *) colour;
@@ -853,15 +836,25 @@ vips_icc_import_line( VipsColour *colour,
 		const int chunk = VIPS_MIN( width - i, PIXEL_BUFFER_SIZE );
 
 		cmsDoTransform( icc->trans, p, encoded, chunk );
-
-		if( icc->pcs == VIPS_PCS_LAB ) 
-			decode_lab( encoded, q, chunk );
-		else
-			decode_xyz( encoded, q, chunk );
+		decode_xyz( encoded, q, chunk );
 
 		p += PIXEL_BUFFER_SIZE * VIPS_IMAGE_SIZEOF_PEL( colour->in[0] );
 		q += PIXEL_BUFFER_SIZE * 3;
 	}
+}
+
+/* Process a buffer of data.
+ */
+static void
+vips_icc_import_line( VipsColour *colour,
+	VipsPel *out, VipsPel **in, int width )
+{
+	VipsIcc *icc = (VipsIcc *) colour;
+
+	if( icc->pcs == VIPS_PCS_LAB )
+		cmsDoTransform( icc->trans, in[0], out, width );
+	else
+		vips_icc_import_line_xyz( colour, out, in, width );
 }
 
 static void
@@ -964,21 +957,6 @@ vips_icc_export_build( VipsObject *object )
 }
 
 static void
-encode_lab( float *in, float *out, int n )
-{
-	int i;
-
-	for( i = 0; i < n; i++ ) {
-		out[0] = in[0];
-		out[1] = in[1];
-		out[2] = in[2];
-
-		in += 3;
-		out += 3;
-	}
-}
-
-static void
 encode_xyz( float *in, float *out, int n )
 {
 	int i;
@@ -993,10 +971,8 @@ encode_xyz( float *in, float *out, int n )
 	}
 }
 
-/* Process a buffer of data.
- */
 static void
-vips_icc_export_line( VipsColour *colour, 
+vips_icc_export_line_xyz( VipsColour *colour,
 	VipsPel *out, VipsPel **in, int width )
 {
 	VipsIcc *icc = (VipsIcc *) colour;
@@ -1014,16 +990,26 @@ vips_icc_export_line( VipsColour *colour,
 	for( x = 0; x < width; x += PIXEL_BUFFER_SIZE ) {
 		const int chunk = VIPS_MIN( width - x, PIXEL_BUFFER_SIZE );
 
-		if( icc->pcs == VIPS_PCS_LAB )
-			encode_lab( p, encoded, chunk );
-		else
-			encode_xyz( p, encoded, chunk );
-
+		encode_xyz( p, encoded, chunk );
 		cmsDoTransform( icc->trans, encoded, q, chunk );
 
 		p += PIXEL_BUFFER_SIZE * 3;
 		q += PIXEL_BUFFER_SIZE * VIPS_IMAGE_SIZEOF_PEL( colour->out );
 	}
+}
+
+/* Process a buffer of data.
+ */
+static void
+vips_icc_export_line( VipsColour *colour, 
+	VipsPel *out, VipsPel **in, int width )
+{
+	VipsIcc *icc = (VipsIcc *) colour;
+
+	if( icc->pcs == VIPS_PCS_LAB )
+		cmsDoTransform( icc->trans, in[0], out, width );
+	else
+		vips_icc_export_line_xyz( colour, out, in, width );
 }
 
 static void
--- a/libvips/colour/icc_transform.c
+++ b/libvips/colour/icc_transform.c
@@ -845,12 +845,11 @@ decode_lab( guint16 *fixed, float *lab, int n )
 	int i;
 
 	for( i = 0; i < n; i++ ) {
-		cmsCIELab Lab;
-		cmsLabEncoded2Float( &Lab, fixed );
-
-		lab[0] = (float) Lab.L;
-		lab[1] = (float) Lab.a;
-		lab[2] = (float) Lab.b;
+		/* cmsLabEncoded2Float inlined.
+		 */
+		lab[0] = (double) fixed[0] / 655.35;
+		lab[1] = ((double) fixed[1] / 257.0) - 128.0;
+		lab[2] = ((double) fixed[2] / 257.0) - 128.0;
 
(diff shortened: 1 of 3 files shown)


jcupitt (approved):
I'm amazed lcms has such different behaviour on this code path. 

Well done for finding this Kleis.

jcupitt:
I tested this PR with:

```
$ vips icc_import input.jpg t1.v --pcs xyz
$ vips icc_export t1.v output.jpg
```

And it now round-trips correctly, though there's some gamut clipping in the extreme reds, as you'd expect.

Open these in two browser tabs and flip. Before:

input

After round-trip:

output

jcupitt:
I benchmarked like this: 

```
$ time vips icc_import input.jpg t1.v --pcs xyz
real	0m0.096s
user	0m0.112s
sys	0m0.076s
```

After:

```
$ time vips icc_import input.jpg t1.v --pcs xyz
real	0m0.097s
user	0m0.345s
sys	0m0.090s
```

So we're maybe 4x more CPU with this change, which is unfortunate :( However, the accuracy improvement is so dramatic, I don't think we have a choice.

The high CPU cost seems alarming, but it's interesting that overall run time is unaffected, since libvips hides the extra processing with extra concurrency (phew!).

lovell:
I've done some quick performance testing/profiling and can confirm the CPU cost (ignoring concurrency) of both approaches is ~4x, with ~30% of the time spent calculating `pow` (gamma calc?). Perhaps the original `cmsFLAGS_NOOPTIMIZE` approach is a better/simpler change after all, especially if we can make it conditional somehow e.g. XYZ-only, or maybe via a new argument?

lovell (approved):
Having slept on it, I agree this PR is the better solution, especially as it doesn't prevent lcms from internally optimising the `icc_transform` operation.

kleisauke:
Ah, I was unaware of the additional CPU cost (I only tested overall run time). Let me check if there's another solution.

I remembered that one of the libjxl maintainers said that it did a `pow()` per pixel, which is probably the reason why libjxl seems to prefer skcms over lcms. This was tracked in issue https://github.com/mm2/Little-CMS/issues/2 and the fast float plugin resolves that (unfortunately GPL-licensed).

jcupitt:
I had a quick go with fast float using this diff to this patch:

```diff
diff --git a/libvips/colour/icc_transform.c b/libvips/colour/icc_transform.c
index 31f0c954c..e7ed52d32 100644
--- a/libvips/colour/icc_transform.c
+++ b/libvips/colour/icc_transform.c
@@ -91,6 +91,7 @@
 /* Has to be before VIPS to avoid nameclashes.
  */
 #include <lcms2.h>
+#include <lcms2_fast_float.h>
 
 #include <vips/vips.h>
 
diff --git a/libvips/iofuncs/init.c b/libvips/iofuncs/init.c
index 93f0b17ba..25089d0bc 100644
--- a/libvips/iofuncs/init.c
+++ b/libvips/iofuncs/init.c
@@ -114,6 +114,11 @@
 #include <vips/vips7compat.h>
 #endif
 
+#ifdef HAVE_LCMS2
+#include <lcms2.h>
+#include <lcms2_fast_float.h>
+#endif /*HAVE_LCMS2*/
+
 /* abort() on the first warning or error.
  */
 int vips__fatal = 0;
@@ -469,6 +474,10 @@ vips_init( const char *argv0 )
                return( 0 );
        started = TRUE;
 
+#ifdef HAVE_LCMS2
+       cmsPlugin(cmsFastFloatExtensions());
+#endif /*HAVE_LCMS2*/
+
        /* Try to set a minimum stacksize, default 2mb. We need to do this
         * before any threads start.
         */
```

But I don't see a speedup :( 

```
$ time vips icc_import input.jpg t1.v --pcs xyz
real	0m0.113s
user	0m0.360s
sys	0m0.085s
$ ldd $(which vips) | grep fast
	liblcms2_fast_float.so => /home/john/vips/lib/liblcms2_fast_float.so (0x00007fd96cbe2000)
```

I've probably messed up somewhere.

jcupitt:
Heh callgrind shows 50% of CPU spent evaluating tone curves:

image

I must have missed something.

kleisauke:
You're right, I don't see any speedup here either with the fast float plugin :(

I did some digging and it looks like this is being caused by the `OptimizeByResampling` builtin optimization in lcms. With this patch:
```diff
--- a/src/cmsopt.c
+++ b/src/cmsopt.c
@@ -1803,8 +1803,7 @@ static _cmsOptimizationCollection DefaultOptimization[] = {
 
     { OptimizeByJoiningCurves,            &DefaultOptimization[1] },
     { OptimizeMatrixShaper,               &DefaultOptimization[2] },
-    { OptimizeByComputingLinearization,   &DefaultOptimization[3] },
-    { OptimizeByResampling,               NULL }
+    { OptimizeByComputingLinearization,   NULL },
 };
 
 // The linked list head
```

I see (click to enlarge):

```console
$ vipsthumbnail php-vips-192.jpg -s 1536x1024 -e srgb -o x.jpg
```

| Original | `8.14` | `8.14` + patch | This PR | Commit https://github.com/kleisauke/libvips/commit/3e93bb07ea9f4327bc16fa2ecc1a56839c8f8ccb + patch
| - | - | - | - | - |
| php-vips-192 |  |lcms-patch|pr-3373 | lcms-simplify-2-patch |

And:

```console
$ vipsthumbnail php-vips-169.jpg --linear -s 1920x1080 -o x.jpg
```

| Original | `8.14` | `8.14` + patch | This PR | Commit https://github.com/kleisauke/libvips/commit/3e93bb07ea9f4327bc16fa2ecc1a56839c8f8ccb + patch
| - | - | - | - | - |
| php-vips-169 | |lcms-patch| lcms-simplify| lcms-simplify-2 |

So, it looks like commit https://github.com/kleisauke/libvips/commit/3e93bb07ea9f4327bc16fa2ecc1a56839c8f8ccb might also help here, but only when that builtin lcms optimization is disabled.

jcupitt:
Oh very cool! And does changing the lcms optimisation table improve performance significantly?

lovell:
Great detective work Kleis. Did you see that the accuracy of the LUTs in `OptimizeByResampling` is based on `nGridPoints` which is optionally increased by the `cmsFLAGS_HIGHRESPRECALC` flag?

kleisauke:
Yes, I also noticed that. I initially tried these two patches:

```diff
--- a/libvips/colour/icc_transform.c
+++ b/libvips/colour/icc_transform.c
@@ -444,6 +444,9 @@ vips_icc_build( VipsObject *object )
 	if( icc->black_point_compensation ) 
 		flags |= cmsFLAGS_BLACKPOINTCOMPENSATION;
 
+	if( icc->pcs == VIPS_PCS_XYZ )
+		flags |= cmsFLAGS_HIGHRESPRECALC;
+
 	if( !(icc->trans = cmsCreateTransform( 
 		icc->in_profile, icc->in_icc_format,
 		icc->out_profile, icc->out_icc_format, 
```

```diff
--- a/libvips/colour/icc_transform.c
+++ b/libvips/colour/icc_transform.c
@@ -444,6 +444,9 @@ vips_icc_build( VipsObject *object )
 	if( icc->black_point_compensation ) 
 		flags |= cmsFLAGS_BLACKPOINTCOMPENSATION;
 
+	if( icc->pcs == VIPS_PCS_XYZ )
+		flags |= cmsFLAGS_GRIDPOINTS( 49 );
+
 	if( !(icc->trans = cmsCreateTransform( 
 		icc->in_profile, icc->in_icc_format,
 		icc->out_profile, icc->out_icc_format, 
```

But both didn't work, unfortunately.

jcupitt:
I tried removing `OptimizeByResampling()` but it didn't reduce the number of `pow()` calls (according to callgrind). It's still doing 3 `pow()` calls per pixel processed :(

jcupitt:
libvips uses a LUT for tone curves -- 1500 elements (from memory?) then linear piecewise interpolation between them, which is accurate enough, I think.

Maybe we could make our own lcms plugin that does something similar? Or perhaps we need to swap to skcms :(

jcupitt:
Oh heh no we don't, it's much more complex, I'd forgotten.

kleisauke:
IIRC, skcms doesn't support conversion _to_ LUT-based or CMYK profiles, so switching to that is not an option.

jcupitt:
Ah shame.

kleisauke:
Commit 4e16c13000813bd8266f9ab69e1329e328c3e54c fixes a small color regression with the `php-vips-192.jpg` testcase.
```console
$ vipsthumbnail php-vips-192.jpg -s 1536x1024 -e srgb -o x.jpg
```

| Original | Before (caf5aa476b9c7195b4c6d7ef9a172044abf76f1b) | After (4e16c13000813bd8266f9ab69e1329e328c3e54c) |
| - | - | - |
| php-vips-192 |pr-3373 | pr-3373 |

libvips/colour/icc_transform.c
```diff
@@ -840,44 +839,50 @@ vips_icc_import_build( VipsObject *object )
...
+
+		lab += 3;
+		fixed += 3;
+	}
 }
 
-#define X_FAC (VIPS_D50_X0 * 32768 / (VIPS_D65_X0 * 100))
-#define Y_FAC (VIPS_D50_Y0 * 32768 / (VIPS_D65_Y0 * 100))
-#define Z_FAC (VIPS_D50_Z0 * 32768 / (VIPS_D65_Z0 * 100))
+#define X_FAC (cmsD50X / VIPS_D65_X0)
```
lovell, line 860: It looks like these constants are always used in a float context so prefixing with a `(float)` cast would prevent an implicit conversion to double, although I'm unsure if this would help performance.

libvips/colour/icc_transform.c
```diff
@@ -840,44 +839,50 @@ vips_icc_import_build( VipsObject *object )
...
 {
 	int i;
 
-        for( i = 0; i < n; i++ ) {
-                lab[0] = (double) fixed[0] / 652.800;
-                lab[1] = ((double) fixed[1] / 256.0) - 128.0;
-                lab[2] = ((double) fixed[2] / 256.0) - 128.0;
+	for( i = 0; i < n; i++ ) {
+		cmsCIELab Lab;
+		cmsLabEncoded2Float( &Lab, fixed );
```
lovell, line 849: Can a compiler optimise things here, or will this always result in a shared library call per-pixel?

(The lcms source appears to use slightly different constants for this calculation e.g. https://github.com/mm2/Little-CMS/blob/9ca6b591e0af9bebe6412dba907cb077f6291a8c/src/cmspcs.c#L204-L215 )

kleisauke: I inlined these calculations with commit 91070f0810f453995f6fcb3f91474600ea9c883d, allowing compilers to optimise.

lcms has two APIs to decode a Lab value:
- `cmsLabEncoded2Float()` - decode a Lab value, encoded on ICC v4 convention.
- `cmsLabEncoded2FloatV2()` - decode a Lab value, encoded on ICC v2 convention.

The reason I used the former as the reference is to match the `cmsCreateLab4Profile()` API.  Previously, libvips used the former as the reference, which is probably(?) only appropriate when using `cmsCreateLab2Profile()`.

PR #3381, opened 2023-03-12
Prevent use of void pointer arithmetic

lovell:
Fixes the specific problems mentioned in #3378 plus a few others I found, and adds a compiler flag (for clang/gcc) to help catch this in the future. Targets the 8.14 branch.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -50,9 +50,7 @@ cc = meson.get_compiler('c')
 cpp = meson.get_compiler('cpp')
 
 # Prevent use of void* pointer arithmetic to support MSVC
-if cc.has_argument('-Werror=pointer-arith')
-    add_project_arguments('-Werror=pointer-arith', language : ['cpp', 'c'])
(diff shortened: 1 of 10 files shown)


meson.build
```diff
@@ -49,6 +49,11 @@ host_os = host_machine.system()
 cc = meson.get_compiler('c')
 cpp = meson.get_compiler('cpp')
 
+# Prevent use of void* pointer arithmetic to support MSVC
+if cc.has_argument('-Werror=pointer-arith')
+    add_project_arguments('-Werror=pointer-arith', language : ['cpp', 'c'])
+endif
```
kleisauke, line 55:
```suggestion
add_project_arguments(cc.get_supported_arguments('-Werror=pointer-arith'), language : ['cpp', 'c'])
```

kleisauke (approved):
LGTM! I left a small suggestion inline.

PR #3383, opened 2023-03-13
Fix compatibility with MSVC

kleisauke:
See: #3378.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -12,6 +12,7 @@ TBD 8.14.2
 - fix thumbnail of CMYK images with an embedded ICC profile [kleisauke]
 - fix ICC handling of RGB images with a monochrome profile [kleisauke]
 - ensure ICC transforms keep all precision [kleisauke]
+- fix compatibility with MSVC [SpaceIm]
 
 9/1/23 8.14.1
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -12,6 +12,7 @@ TBD 8.14.2
 - fix thumbnail of CMYK images with an embedded ICC profile [kleisauke]
 - fix ICC handling of RGB images with a monochrome profile [kleisauke]
 - ensure ICC transforms keep all precision [kleisauke]
+- fix openslideload associated=XXX load [jcupitt]
 - fix compatibility with MSVC [SpaceIm]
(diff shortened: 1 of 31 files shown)


lovell (approved):

jcupitt:
Adding `__restrict` everywhere is a bit ugly. How about something like:

```C
#ifdef _MSC_VER
#define restrict __restrict
#endif 
```

in `internal.h`? We do things like this to hide other MSVC-isms.

SpaceIm:
@jcupitt Current injection `-Drestrict=__restrict` in meson.build breaks Visual Studio (see https://github.com/libvips/libvips/issues/3378#issue-1620482223). I'm not sure that defining it inside cpp files would change anything.

kleisauke:
It looks like moving:
https://github.com/libvips/libvips/blob/3848b5fad4570fdf9fd2e5eee2ce6c74655f617d/meson.build#L33

To `internal.h`, i.e.:
https://github.com/libvips/libvips/blob/47af384585417bf655d1cb8bf58922cca5084b1d/libvips/include/vips/internal.h#L75-L79

Also works without any problems. I think defining this through the compiler command line caused it to be propagated to the Windows headers, resulting in these build failures (it probably interfered with `__declspec(restrict)` / `_CRTRESTRICT`).

FWIW, I tested this with:
<details>
  <summary>Details</summary>

```cmd
.\vcpkg\vcpkg install cfitsio expat fftw3 glib graphicsmagick libexif libheif libjpeg-turbo libjxl libpng librsvg libwebp matio nifticlib openexr openjpeg openslide poppler tiff --triplet=x64-windows
pip3 install meson

set PKG_CONFIG=%cd%\vcpkg\installed\x64-windows\tools\pkgconf\pkgconf.exe
set PKG_CONFIG_PATH=%cd%\vcpkg\installed\x64-windows\lib\pkgconfig
set PATH=%PATH%;%cd%\vcpkg\installed\x64-windows\tools\glib\

cd libvips
meson setup build --buildtype=release -Dintrospection=false -Dmagick-package=GraphicsMagick
meson compile -C build
```
</details>

jcupitt:
Nice!

PR #3384, opened 2023-03-13
Fix openslide associated

jcupitt:
Before this PR I see:

```
$ vips copy CMU-1.svs[associated=label] x.png
```

I suppose the recent refactoring broke this case.

This should probably be cherry-picked for 8.14 as well.

openslideload could use a mild rewrite -- it's still split into two parts.


kleisauke (approved):

kleisauke:
Backport to [`8.14`](https://github.com/libvips/libvips/tree/8.14) available as #3386.

PR #3391, opened 2023-03-19
try to improve module load messages

jcupitt:
We were warning about module load failure, but (I think?) this can fail if a module if already resident, so we should probably only have an info for this.

If load fails, don't call `make_resident`, since this will trigger a glib critical wrror.

See https://github.com/libvips/ruby-vips/discussions/364#discussioncomment-5357118

@kleisauke any thoughts? I'm a bit out of the loop on this, perhaps this isn't the correct fix.

### DIFF
--- a/libvips/iofuncs/init.c
+++ b/libvips/iofuncs/init.c
@@ -263,7 +263,6 @@ vips_load_plugins( const char *fmt, ... )
         char dir_name[VIPS_PATH_MAX];
         GDir *dir;
 	const char *name;
-        int result;
 
 	/* Silently succeed if we can't do modules.
 	 */
@@ -281,7 +280,6 @@ vips_load_plugins( const char *fmt, ... )
 		 */
                 return( 0 );
 
-        result = 0;
         while( (name = g_dir_read_name( dir )) )
                 if( vips_ispostfix( name, "." G_MODULE_SUFFIX )
 #if ENABLE_DEPRECATED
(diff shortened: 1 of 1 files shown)


kleisauke:
I'm not sure if we should reduce the severity of module load failures. The linked issue seems to be a packaging issue in MSYS2 where the optional dynamic modules are distributed without the dependencies of those modules, so these failures are real.

Guarding the `g_module_make_resident()` call for `NULL` looks good to me.

jcupitt:
OK, I put the warning back.

kleisauke:
FWIW, I had a look at GLib's sources and noticed that there is a check to ensure that the module has not been loaded earlier:
https://gitlab.gnome.org/GNOME/glib/-/blob/e979fd9a157688a1a65fb58250b2a9fe94d14d50/gmodule/gmodule.c#L584-592

Also, `g_module_make_resident()` only sets an internal property, so that's probably harmless to call multiple times:
https://gitlab.gnome.org/GNOME/glib/-/blob/e979fd9a157688a1a65fb58250b2a9fe94d14d50/gmodule/gmodule.c#L855

kleisauke (approved):
LGTM!

kleisauke:
I just noticed that `vips_load_plugins()` is internal API, so we could change it to a void function instead. See e.g. commit https://github.com/kleisauke/libvips/commit/aefa02b9ca37818f734227350b264a6245b7c290.

kleisauke:
Let me know if I should apply https://github.com/kleisauke/libvips/commit/aefa02b9ca37818f734227350b264a6245b7c290 to this branch or if you prefer a separate PR for that. 

(we should probably also backport this PR to 8.14)

jcupitt:
Good idea, do add the void patch too.

kleisauke:
Backport to [`8.14`](https://github.com/libvips/libvips/tree/8.14) available as #3395.

PR #3400, opened 2023-03-22
colour: swap built-in profiles with ICC v4 variants

kleisauke:
See: https://www.color.org/whyusev4.xalter


lovell:
We might need to wait until https://github.com/saucecontrol/Compact-ICC-Profiles/issues/16 is fixed to ensure compatibility.

lovell (commented):
Should the `libvips/colour/profiles/*.icm` files be replaced rather than removed?

kleisauke:
Updated to incorporate commit https://github.com/saucecontrol/Compact-ICC-Profiles/commit/ca0a34b4047e3ebdefc0196759a4d88b85b104b4, making the profiles spec-compliant.

lovell (approved):

Issue #3411, opened 2023-03-28
segv in tiffsave


jcupitt:
tiffsave seems to segv reliably for me in master. 

Make a 32k x 32k jpeg:

```
$ vips replicate ../st-francis.jpg x.jpg 2 2 
$ vips crop x.jpg y.jpg 0 0 32768 32768
```

Copy 32 times and join horizontally:

```
$ for i in {1..32}; do cp y.jpg x_$i.jpg; done
$ time vips arrayjoin "$(echo x_*)" final.tif[tile,compression=jpeg,pyramid] --across 32
Segmentation fault (core dumped)

real	3m5.574s
user	4m16.189s
sys	0m37.318s
```

gdb shows:

```
(gdb) bt
#0  0x0000000000000000 in  ()
#1  0x00007ffff7c2e157 in vips_target_seek ()
    at /home/john/vips/lib/libvips.so.42
#2  0x00007ffff6f34d12 in TIFFLinkDirectory (tif=0x7fffe82720f0)
    at /build/tiff-2akeH2/tiff-4.4.0/libtiff/tif_dirwrite.c:3125
#3  TIFFWriteDirectorySec
    (tif=0x7fffe82720f0, isimage=1, imagedone=1, pdiroff=0x0)
    at /build/tiff-2akeH2/tiff-4.4.0/libtiff/tif_dirwrite.c:903
#4  0x00007ffff6f3e1fe in TIFFFlush (tif=0x7fffe82720f0)
    at /build/tiff-2akeH2/tiff-4.4.0/libtiff/tif_flush.c:53
#5  0x00007ffff6f1ea45 in TIFFCleanup (tif=0x7fffe82720f0)
    at /build/tiff-2akeH2/tiff-4.4.0/libtiff/tif_close.c:51
#6  0x00007ffff6f1ea8d in TIFFClose (tif=<optimised out>)
    at /build/tiff-2akeH2/tiff-4.4.0/libtiff/tif_close.c:131
#7  0x00007ffff7a99c75 in layer_free_all ()
    at /home/john/vips/lib/libvips.so.42
#8  0x00007ffff7a99baf in layer_free_all ()
    at /home/john/vips/lib/libvips.so.42
#9  0x00007ffff7a99baf in layer_free_all ()
    at /home/john/vips/lib/libvips.so.42
#10 0x00007ffff7a99dd6 in wtiff_free () at /home/john/vips/lib/libvips.so.42
[...]

PR #3413, opened 2023-03-28
tiffsave: fix use-after-free during pyramid save

kleisauke:
`TIFFClose()` could still use the target after it had been unreffed.

Resolves: #3411.

Targets the 8.14 branch.


jcupitt (approved):

jcupitt:
Very nice!

PR #3419, opened 2023-03-29
bump min version for libheif

jcupitt:
We use heif_image_handle_get_luma_bits_per_pixel() in heifload, which was only added in libheif 1.4


jcupitt:
It compiles with libheif 1.4.1, but pytest crashes. I think in practice users should probably use a more recent libheif.

I had a quick look at getting `heifload.c` to compile with older libheifs, but it's not trivial, and doesn't seem worth the effort.

kleisauke (approved):

PR #3432, opened 2023-04-06
Change introspection option to feature type

kleisauke:
Avoids having to explicitly disable introspection support if GOI is not installed.

Also, disable it by default for cross builds unless the option was explicitly enabled by the user.


jcupitt (approved):
Nice! This will make life easier for a lot of people I think.

Issue #3422, opened 2023-03-31
Request: Scharr edge detection


Zireael07:
Title. I see Sobel and Canny are on offer, but Scharr is conspicuously missing

jcupitt:
Hi @Zireael07,

It's just a couple of lines of eg. python, but I agree it'd be useful to have an implementation built in. Let's tag this as an enhancement.

Zireael07:
Thanks a lot, this means libvips/its official GUI should suit my use case well <3

Leaving this open because of the feature request itself

jcupitt:
I made a PR for scharr and prewitt detectors, and made sobel a bit more accurate for non-uchar images.

PR #3436, opened 2023-04-12
add scharr and prewitt edge detectors

jcupitt:
And Sobel is slower but more accurate for non-uchar images.

See https://github.com/libvips/libvips/issues/3422


kleisauke (approved):
LGTM!

Issue #3446, opened 2023-04-17
webpsave_timeout test runs even when libwebp is not available


thesoulless:
## Bug report

**Describe the bug**
Meson does not check libwebp availability before running `webpsave_timeout` test.

**To Reproduce**
1. Use centos:7 as the base image (or any other distros where libwebp is not available)
2. Build with meson (meson setup build && cd build && meson compile)
3. Run `meson test`, it will fail on webpsave_timeout test case

**Expected behavior**
Run the test only if `libwebp_dep.found()`

**Actual behavior**
Running the test regardless of `libwebp_dep.found()` result.

**Environment**
 - OS: CentOs 7
 - Vips: 8.14.2

kleisauke:
Fixed with commit afe6c606c998e193d80246fa5bc6d9acf5ee18d2.

PR #3447, opened 2023-04-17
fix: check libwebp availability

thesoulless:
`webpsave_timeout` test runs even when libwebp is not available

Resolves: https://github.com/libvips/libvips/issues/3446.

Targets the 8.14 branch.

### DIFF
--- a/test/meson.build
+++ b/test/meson.build
@@ -64,13 +64,15 @@ if host_os == 'linux'
     )
 endif
 
-test_timeout_webpsave = executable('test_timeout_webpsave',
-    'test_timeout_webpsave.c',
-    dependencies: libvips_dep,
-)
+if libwebp_dep.found()
+    test_timeout_webpsave = executable('test_timeout_webpsave',
+        'test_timeout_webpsave.c',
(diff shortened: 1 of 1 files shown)


kleisauke:
Nice catch! Perhaps we should check availability at runtime instead? For example:
```diff
--- a/test/test_timeout_webpsave.c
+++ b/test/test_timeout_webpsave.c
@@ -22,6 +22,11 @@ main( int argc, char **argv )
 	if( VIPS_INIT( argv[0] ) )
 		vips_error_exit( NULL );
 
+	if( !vips_type_find( "VipsOperation", "webpsave" ) )
+		/* webpsave not available, skip test with return code 77.
+		 */
+		return( 77 );
+
 	if( vips_black( &im, 16383, 16383, NULL ) )
 		vips_error_exit( NULL );
 
```

This would be in line with the other tests and ensures that the user is aware that the test is skipped.

thesoulless:
That's much better and more informative for the user.

kleisauke (approved):

PR #3460, opened 2023-04-22
Add a magick sniffer for CR2

kleisauke:
This image format uses the standard TIFF header, so a special sniffer is needed to prevent *magick from trying to load the buffer through its TIFF loader.

### DIFF
--- a/libvips/foreign/magick.c
+++ b/libvips/foreign/magick.c
@@ -71,6 +71,7 @@
  * * https://www.dca.fee.unicamp.br/~martino/disciplinas/ea978/tgaffs.pdf
  * * http://www.paulbourke.net/dataformats/tga/
  * * https://en.wikipedia.org/wiki/Truevision_TGA#Technical_details
+ * * http://lclevy.free.fr/cr2/
  */
 static const char *
 magick_sniff(const unsigned char *bytes, size_t length)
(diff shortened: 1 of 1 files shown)


libvips/foreign/magick.c
```diff
@@ -102,6 +105,54 @@ magick_sniff( const unsigned char *bytes, size_t length )
...
+	if( length >= 12 &&
+		ifd_offset >= 8 &&
+		((bytes[8] == 28 &&
+		  bytes[9] == 00 &&
+		  bytes[10] == 254 &&
+		  bytes[11] == 00) ||
+		 (bytes[8] == 31 &&
+		  bytes[9] == 00 &&
+		  bytes[10] == 11 &&
+		  bytes[11] == 00) ||
```
kleisauke, line 149: FWIW, this was based on:
https://github.com/sindresorhus/file-type/blob/1c75cfbaab544bd0882e498bcbfa54882ea7300e/core.js#L1521

But that unfortunately did not recognize the buffer in the linked issue, which contains `0x00 0x1B 0x00 0xFE` starting from offset 8.

Perhaps there's a better way to recognize these NEF images?

416c616e:
Would it be advantageous to include CR3 magic byte handling I've proposed with #3546 ?

kleisauke:
I omitted the NEF sniffer from this PR because there is no reliable way to distinguish it from TIFF images when inspecting the first number of bytes. For the same reason, GIMP probably doesn't have a sniffer for this either:
https://gitlab.gnome.org/GNOME/gimp/-/blob/7219245bfb2ae9590bd483cdb0d980ae92add83b/plug-ins/file-raw/file-raw-formats.h#L62

libvips/foreign/magick.c
kleisauke: https://github.com/libvips/libvips/pull/3460#issuecomment-1615815041

jcupitt:
LGTM!

jcupitt (approved):

PR #3481, opened 2023-05-05
lsan: suppress leak warnings from libheif

lovell:
Since https://github.com/strukturag/libheif/commit/8f9ac3d6db58d91ebb36f68244cdc534253b889b libheif creates static conversion tables on initialisation. We rely on libheif implicitly calling `heif_init` for us, so `heif_uninit` is never called and these tables will appear to leak.

Should silence https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=58618

### DIFF
--- a/suppressions/lsan.supp
+++ b/suppressions/lsan.supp
@@ -5,3 +5,4 @@ leak:libIlmImf-2_5.so
 leak:libIlmThread-2_5.so
 leak:libMagickCore-6.Q16.so
 leak:libx265.so
+leak:libheif.so


kleisauke:
I don't think this would work in OSS-Fuzz, since all dependencies are statically linked.

It looks like libheif resolved this with commit https://github.com/strukturag/libheif/commit/c242822ab5f6919a823bec08ba8dff77252394bb. Though, it's a bit strange if we need to do the same thing within libvips' fuzzers.

Perhaps marking `heif_deinit()` with `__attribute__((destructor))` would work in OSS-Fuzz? For example:
```diff
--- a/fuzz/oss_fuzz_build.sh
+++ b/fuzz/oss_fuzz_build.sh
@@ -67,6 +67,8 @@ popd
 pushd $SRC/libheif
 # Ensure libvips finds heif_image_handle_get_raw_color_profile
 sed -i '/^Libs.private:/s/-lstdc++/-lc++/' libheif.pc.in
+# Suppress leak warnings from libheif
+sed -i '/void heif_deinit()/i __attribute__((destructor))' libheif/heif_init.cc
 cmake \
   -DCMAKE_BUILD_TYPE=RelWithDebInfo \
   -DCMAKE_INSTALL_PREFIX=$WORK \
```

(FWIW, `vips_shutdown()` could probably also marked with this attribute, see e.g. commit https://github.com/kleisauke/libvips/commit/832f9c97644f92d888830f769c3eecc327f8d219, but that would only ensure that the internal leak checker and profiler would run, so it's not necessary in most cases)

lovell:
Thanks Kleis, if `heif_deinit` will always be called then should we also ensure `heif_init` is always called too? It looks like not doing so could trigger the line commented with "This case should never happen" :)

https://github.com/strukturag/libheif/blob/e411241b9afd0ac96e78045cb5fb815e906f09ff/libheif/init.cc#L165

kleisauke:
It seems that the semantics of `heif_deinit()` is the same as `vips_shutdown()`, i.e. it causes the actual cleaning to happen once if `heif_init()` has been called previously. When `heif_init()` has not been called, the function is no-op. So marking `heif_deinit()` as dtor should not be a issue, I think.

I just opened PR #3489, hopefully that will work.

lovell:
Should we keep this change in addition to #3489? Perhaps it would be useful for anyone running lsan locally?

kleisauke:
Good idea, let's do both. I intend to upstream #3489, if that works.

kleisauke (approved):

PR #3485, opened 2023-05-09
Don't use 0 as a default for pngsave bitdepth

jcupitt:
pngsave introspection lists 0 as the default value for bitdepth, but this value does not work:

```
$ vips pngsave
save image to png file
...
   bitdepth     - Write as a 1, 2, 4, 8 or 16 bit image, input gint
			default: 0
			min: 0, max: 16
...

$ vips pngsave k2.jpg x.png --bitdepth 0

(vips:1393285): VIPS-WARNING **: 13:15:09.883: Invalid bit depth in IHDR

(vips:1393285): VIPS-WARNING **: 13:15:09.883: Invalid IHDR data
vips2png: unable to write to target x.png
```

Instead, set the default value to 8. This will be overridden in _build() if bitdepth has not been set.


jcupitt:
Looks like a copy-paste from tiffsave, which used 0 as a default value because it is still a vips7 saver. I can't see any other bad bitdepths.

kleisauke (approved):
Nice catch!

PR #3489, opened 2023-05-10
OSS-Fuzz: mark `heif_deinit()` as dtor

kleisauke:
Ensures the 'static' conversion tables are properly cleaned up.

Might resolve https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=58618

---

I'm not sure if this will work, as I could not reproduce that issue. I tried:
```console
$ python infra/helper.py build_fuzzers libvips > build_log.txt
$ mkdir -p tmp
$ python infra/helper.py run_fuzzer --corpus-dir=$PWD/tmp/ libvips pngsave_buffer_fuzzer -- -max_total_time=1800
```
(i.e. run the fuzzer for 30 minutes without pressing CTRL+C)

### DIFF
--- a/fuzz/oss_fuzz_build.sh
+++ b/fuzz/oss_fuzz_build.sh
@@ -67,6 +67,8 @@ popd
 pushd $SRC/libheif
 # Ensure libvips finds heif_image_handle_get_raw_color_profile
 sed -i '/^Libs.private:/s/-lstdc++/-lc++/' libheif.pc.in
+# Ensure heif_deinit is called at program exit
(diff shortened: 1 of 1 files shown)


kleisauke:
> OSS-Fuzz complained about a leak with `Op_YCbCr444_to_YCbCr420_average<uint16_t>`, which was moved within that commit.

Ah, I looked wrong. All new-expression allocated operations seem to be leaky.
https://github.com/strukturag/libheif/blob/e411241b9afd0ac96e78045cb5fb815e906f09ff/libheif/color-conversion/colorconversion.cc#L183-L208

lovell (approved):

PR #3501, opened 2023-05-20
Move rounding from reduce(h/v)_notab to cast

DarthSim:
This PR fixes the unexpected behavior brought by https://github.com/libvips/libvips/pull/1872 while keeping the rounding issue fixed.

https://github.com/libvips/libvips/pull/1872 added rounding to `reduce(h/v)_notab` functions. This behavior is unexpected. For example, I created an image with the `double` format and a pixel range of 0.0-1.0. If I apply `vips_reduce` to this image, all its pixels will become 0.0 or 1.0.

The rounding issue fixed with https://github.com/libvips/libvips/pull/1872 was caused by a lack of rounding when casting an image from a floating-point format to an integer format. So this PR removes rounding from `reduce(h/v)_notab` functions and adds it to the `cast` operation instead.

/cc @kleisauke

### DIFF
--- a/libvips/conversion/cast.c
+++ b/libvips/conversion/cast.c
@@ -227,7 +227,7 @@ G_DEFINE_TYPE( VipsCast, vips_cast, VIPS_TYPE_CONVERSION );
 	OTYPE * restrict q = (OTYPE *) out; \
 	\
 	for( x = 0; x < sz; x++ ) \
-		q[x] = CAST( (double) p[x] ); \
+		q[x] = CAST( VIPS_ROUND_INT( (double) p[x] ) ); \
 }
 
 /* Cast complex types to an int type. Just take the real part.
@@ -240,7 +240,7 @@ G_DEFINE_TYPE( VipsCast, vips_cast, VIPS_TYPE_CONVERSION );
 	OTYPE * restrict q = (OTYPE *) out; \
 	\
 	for( x = 0; x < sz; x++ ) { \
-		q[x] = CAST( (double) p[0] ); \
+		q[x] = CAST( VIPS_ROUND_INT( (double) p[0] ) ); \
 		p += 2; \
 	} \
 }
--- a/libvips/resample/reduceh.cpp
+++ b/libvips/resample/reduceh.cpp
@@ -284,12 +284,8 @@ reduceh_notab( VipsReduceh *reduceh,
 
 	vips_reduce_make_mask( cx, reduceh->kernel, reduceh->hshrink, x ); 
 
-	for( int z = 0; z < bands; z++ ) {
-		double sum;
-		sum = reduce_sum<T, double>( in + z, bands, cx, n );
-
-		out[z] = VIPS_ROUND_UINT( sum );
-	}
+	for( int z = 0; z < bands; z++ )
+		out[z] = reduce_sum<T, double>( in + z, bands, cx, n );
 }
 
 /* Tried a vector path (see reducev) but it was slower. The vectors for
--- a/libvips/resample/reducev.cpp
+++ b/libvips/resample/reducev.cpp
@@ -501,12 +501,8 @@ reducev_notab( VipsReducev *reducev,
 
 	vips_reduce_make_mask( cy, reducev->kernel, reducev->vshrink, y ); 
 
-	for( int z = 0; z < ne; z++ ) {
-		double sum;
-		sum = reduce_sum<T, double>( in + z, l1, cy, n );
-
-		out[z] = VIPS_ROUND_UINT( sum );
-	}
+	for( int z = 0; z < ne; z++ )
+		out[z] = reduce_sum<T, double>( in + z, l1, cy, n );
 }
 
 static int


jcupitt:
Hi @DarthSim,

I think changing the rounding behaviour in `cast` is likely to break a lot of code. It's supposed to match the behaviour of C for eg. `float a = (int) b;` and users are supposed to use `round` if they want `floor` / `ceil` / `rint`.

Instead, how about splitting `reducev_notab()` into two paths, one for int and one for float, with rounding only on the float path? That would match the behaviour of other parts of libvips.

DarthSim:
`reducev_notab` is used only on the `double` path.

After some thought, the whole test from https://github.com/libvips/libvips/pull/1872 seems incorrect to me. [resamplescope](https://github.com/jsummers/resamplescope) uses `uchar` images for tests, so it's incorrect to use it for testing resampling in the `double` space.

Also, we do not do rounding on the `float` path.

IMHO, even if we don't want to add rounding to `cast`, we still should remove it from `reduce(v/h)`.

kleisauke:
PR #1872 was tested by doing: 
```console
$ vips cast pd.png pd.v double
$ vips reduceh pd.v pd_vips_lanczos3.png 1.003603604 --kernel lanczos3
$ ./rscope -name "lanczos3 reduceh with libvips" -nologo pd_vips_lanczos3.png pd_vips_lanczos3-out.png
```

i.e. the resampling takes place in the `reduceh_notab` function. But indeed, `VIPS_ROUND_UINT` shouldn't probably be used in these `*_notab` functions, it should do `+ 0.5` instead to ensure the fractional part isn't lost (which ResampleScope doesn't cover since it operates on `uchar` images).

Note that this particular issue(?) is also present within the `reduce{h,v}_{un,}signed_int32_tab` functions, e.g. with commit https://github.com/kleisauke/pyvips-issue-148/commit/d4d97c634ceade04fa1ac018ec5f88a2a5f1eeed, I see:
| int | uint |
| :---: |  :---: |
|  |  |

Commit https://github.com/kleisauke/libvips/commit/90195477255040842b14b93803ba67d6632c1ea3 seems to fix all ResampleScope test cases.

DarthSim:
@kleisauke That's what I'm talking about – you used a `double` source but converted it to `uchar` during saving the result to PNG since ResampleScope only operates PNGs. So it's not a `reduce(v/h)_notab` issue per se, it's a problem of rounding during saving the image.

> it should do + 0.5 instead

It will keep the unexpected behavior. If I provide a `double` image with `0.0-1.0` pixels, I expect to get the same range of values in the output.

When using ResampleScope to test `reduce(v/h)_notab`, it seems more correct to add 0.5 during the test but not inside of `reduce(v/h)_notab`:

```
vips cast pd.png pd.vips double
vips reduceh pd.vips pd_vips_lanczos3.vips 1.003603604 --kernel lanczos3
vips linear pd_vips_lanczos3.vips pd_vips_lanczos3.png 1 0.5
./rscope -name "lanczos3 reduceh with libvips" -nologo pd_vips_lanczos3.png pd_vips_lanczos3-out.png
```

pd_vips_lanczos3-out

DarthSim:
BTW `reduce(v/h)_float_tab` has the same "issue", but fixing it with rounding or adding 0.5 will break resizing images in the scRGB colorspace. I believe that `reduce(v/h)_float_tab` and `reduce(v/h)_notab` behavior should be the same.

kleisauke:
You're right, `reduce{h,v}_notab` should be more in-line with `reduce{h,v}_float_tab`, i.e. no rounding.

I think it still makes sense to do this rounding in the `reduce{h,v}_{un,}signed_int32_tab` functions. For example, with this script:
```bash
# Resize with hshrink=1.4 to check against http://entropymine.com/imageworsener/filterfairness/
cat <<EOT > 7x1.mat
7 1 1
0 0 0 250 0 0 0
EOT

# Reference double-precision
vips reduceh 7x1.mat .csv 1.5 --kernel linear

# Reference rounding
vips cast 7x1.mat 7x1-uchar.v uchar
vips reduceh 7x1-uchar.v .csv 1.5 --kernel linear

vips cast 7x1.mat 7x1-int.v int
vips reduceh 7x1-int.v .csv 1.5 --kernel linear

vips cast 7x1.mat 7x1-uint.v uint
vips reduceh 7x1-uint.v .csv 1.5 --kernel linear
```

I see:
```
0	0	149.99999999999997	0	0
0	0	150	0	0
0	0	149	0	0
0	0	149	0	0
```

> `reduce(v/h)_float_tab` has the same "issue"

The weird thing is that this cannot be reproduced in ResampleScope, see:
| reduceh float | reducev float |
| :---: |:---: |
|  |  |

DarthSim:
> I think it still makes sense to do this rounding in the reduce{h,v}_{un,}signed_int32_tab functions

Totally agree here

BTW why don't use `long long` (`int64_t`) intermediate for `int` and `uint`? Theoretically, this should be faster as we won't need to cast from integer types to double and back (casting between integer types should be much faster).

kleisauke:
> I think it still makes sense to do this rounding in the `reduce{h,v}_{un,}signed_int32_tab` functions.

> BTW why don't use `long long` (`int64_t`) intermediate for `int` and `uint`? Theoretically, this should be faster as we won't need to cast from integer types to double and back (casting between integer types should be much faster).

I just opened PR #3514 for this.

kleisauke:
I just opened PR https://github.com/libvips/libvips/pull/3532 as an alternative. Using a `long double` intermediate type resolves the ResampleScope issue.

PR #3502, opened 2023-05-20
Fix signed_fixed_round( 0 )

DarthSim:
The current implementation of `signed_fixed_round` returns `-1` for `0` while it seems like it should return `0`. This PR fixes this behavior.

### DIFF
--- a/libvips/resample/templates.h
+++ b/libvips/resample/templates.h
@@ -200,7 +200,7 @@ bicubic_unsigned_int(
 static int inline
 signed_fixed_round( int v )
 {
-	const int sign_of_v = 2 * (v > 0) - 1;
+	const int sign_of_v = 2 * (v >= 0) - 1;
 	const int round_by = sign_of_v * (VIPS_INTERPOLATE_SCALE >> 1);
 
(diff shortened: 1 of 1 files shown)


lovell:
The proposed change here looks like it does the right thing, thank you. Was there a specific image that allowed you to discover this? Is it possible to add a test that might have caught it?

DarthSim:
I noticed this while developing a SIMD-optimized version of `signed_fixed_round`. I don't know about a test, but it seems logical that rounding of 0 should produce 0 😅

lovell (approved):

kleisauke:
FWIW, the `double` implementations, which were possibly used as a reference, do not seem to have this issue:
https://github.com/libvips/libvips/blob/16aeae9e76999d5bd8ac2fb2c4f356bd84b6f7cb/libvips/resample/templates.h#L46
https://github.com/libvips/libvips/blob/16aeae9e76999d5bd8ac2fb2c4f356bd84b6f7cb/libvips/resample/templates.h#L112

kleisauke (approved):

PR #3514, opened 2023-05-28
reduce{h,v}: ensure rounding for the uint/int paths

kleisauke:
See: #3501.

### DIFF
--- a/libvips/resample/templates.h
+++ b/libvips/resample/templates.h
@@ -146,10 +146,10 @@ bilinear_nosign(
  * Bicubic (Catmull-Rom) interpolation templates:
  */
 
-static int inline
-unsigned_fixed_round( int v )
+template <typename T> static T inline
+unsigned_fixed_round( T v )
 {
-	const int round_by = VIPS_INTERPOLATE_SCALE >> 1;
+	const T round_by = VIPS_INTERPOLATE_SCALE >> 1;
 
 	return( (v + round_by) >> VIPS_INTERPOLATE_SHIFT );
 }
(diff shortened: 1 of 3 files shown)


libvips/resample/templates.h
```diff
@@ -444,7 +444,24 @@ calculate_coefficients_lanczos( double *c,
  */
 template <typename T, typename IT>
 static IT
-reduce_sum( const T * restrict in, int stride, const IT * restrict c, int n )
+reduce_sum( const T * restrict in, int stride, const int * restrict c, int n,
```
lovell, line 447: Do we need to `#include <type_traits>`? (I feel like this is the sort of thing that MSVC would complain about, maybe one day I'll get round to adding it to CI.)

lovell (approved):
LGTM, with one question inline. I wonder if this change might improve the chance of auto-vectorisation?

libvips/resample/templates.h
kleisauke: Good point, I added `#include <type_traits>` on the top of this file to ensure compat with MSVC.

kleisauke:
> I wonder if this change might improve the chance of auto-vectorisation?

I'm not sure. I tried to verify a similar function in the past by passing `-fopt-info-vec-all` to GCC, see e.g.:
https://gist.github.com/kleisauke/dcf24beb95cdf2b3037ff63f71b7f9bd

Though, that commit (https://github.com/kleisauke/libvips/commit/67d1aa5d5f4fc83e18e6a428213b8ec878f263bc) didn't actually produce a useful speed-up.

libvips/resample/templates.h
```diff
@@ -439,12 +441,35 @@ calculate_coefficients_lanczos( double *c,
 		c[i] /= sum;
 }
 
+/* Simplified version of std::enable_if<cond, bool>::type
+ */
+template<bool Cond>
+using Requires = typename std::enable_if<Cond, bool>::type; /* C++11 */
```
lovell, line 447: :sunglasses:

lovell (approved):

PR #3520, opened 2023-06-03
Avoid copying data in vips_jpegload_buffer

DarthSim:
Hey everyone!

I noticed that `vips_jpegload_buffer` copies data to an intermediate buffer even though we already have all the data in memory. This PR checks if the source is mappable to memory, and if so sets its data right to `next_input_byte` eliminating the need to copy. It also replaces the `init_source` and `fill_input_buffer` functions with stubs as we don't need to do anything if we already have all data in `next_input_byte`. And it sets a special implementation of `skip_input_data` that considers that we already read all the data (which we actually did).

### DIFF
--- a/libvips/foreign/jpeg2vips.c
+++ b/libvips/foreign/jpeg2vips.c
@@ -239,6 +239,9 @@ source_init_source( j_decompress_ptr cinfo )
 	src->pub.bytes_in_buffer = 0;
 }
 
+static void
+source_init_source_mappable( j_decompress_ptr cinfo ) {}
+
 /* Fill the input buffer --- called whenever buffer is emptied.
  */
 static boolean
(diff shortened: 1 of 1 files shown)


lovell:
Hi Sergey, thanks for the PR, it looks like it will change the behaviour around corrupted/truncated images.

Using [corrupt-header.jpg](https://github.com/lovell/sharp/blob/main/test/fixtures/corrupt-header.jpg) as a test case, libvips will currently generate a "Premature end of JPEG file" warning or error (depending on the `VipsFailOn` setting). With this PR I always get an "Improper call to JPEG library in state 201" error, with the `VipsFailOn` setting ignored.

DarthSim:
@lovell Good catch! Fixed the issue

lovell (approved):
LGTM, thank you.

kleisauke (approved):

PR #3532, opened 2023-06-13
reduce{h,v}: remove rounding from the double paths

kleisauke:
This PR ensures that `reduce{h,v}_notab` is in-line with `reduce{h,v}_float_tab`, i.e. no rounding. 

`vips_reduce_make_mask()` needed a slight refactor, since the `*_notab` functions requires a `long double` intermediate instead of a `double`. Additionally, as part of the process, I consolidated the `calculate_coefficients_*` functions into a single function.

See commit https://github.com/kleisauke/pyvips-issue-148/commit/c6d0049686b97f97feb5a8a5ade14aa2ad905e73 for the ResampleScope difference.

Supersedes: #3501.

### DIFF
--- a/libvips/resample/templates.h
+++ b/libvips/resample/templates.h
@@ -28,6 +28,9 @@
 
  */
 
+#include <cstdint>
+#include <type_traits>
+
 /*
  * Various casts which assume that the data is already in range. (That
  * is, they are to be used with monotone samplers.)
@@ -149,7 +152,7 @@ static T inline bilinear_nosign(
 template <typename T>
 static T inline unsigned_fixed_round(T v)
 {
-	const T round_by = VIPS_INTERPOLATE_SCALE >> 1;
+	const int round_by = VIPS_INTERPOLATE_SCALE >> 1;
 
 	return (v + round_by) >> VIPS_INTERPOLATE_SHIFT;
 }
@@ -200,8 +203,8 @@ static int inline bicubic_unsigned_int(
 template <typename T>
 static T inline signed_fixed_round(T v)
 {
-	const T sign_of_v = 2 * (v >= 0) - 1;
-	const T round_by = sign_of_v * (VIPS_INTERPOLATE_SCALE >> 1);
+	const int sign_of_v = 2 * (v >= 0) - 1;
+	const int round_by = sign_of_v * (VIPS_INTERPOLATE_SCALE >> 1);
 
 	return (v + round_by) >> VIPS_INTERPOLATE_SHIFT;
 }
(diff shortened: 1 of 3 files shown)


kleisauke:
For reference, by applying this patch (which causes the `*_notab` functions to use `double`'s as intermediates):

<details>
  <summary>Details</summary>

```diff
--- a/libvips/resample/reduceh.cpp
+++ b/libvips/resample/reduceh.cpp
@@ -237,13 +237,13 @@ reduceh_notab( VipsReduceh *reduceh,
 	const T* restrict in = (T *) pin;
 	const int n = reduceh->n_point;
 
-	long double cx[MAX_POINT];
+	double cx[MAX_POINT];
 
 	vips_reduce_make_mask( cx, reduceh->kernel, reduceh->n_point,
 		reduceh->hshrink, x );
 
 	for( int z = 0; z < bands; z++ )
-		out[z] = reduce_sum<T, long double>( in + z, bands, cx, n );
+		out[z] = reduce_sum<T, double>( in + z, bands, cx, n );
 }
 
 /* Tried a vector path (see reducev) but it was slower. The vectors for
--- a/libvips/resample/reducev.cpp
+++ b/libvips/resample/reducev.cpp
@@ -496,13 +496,13 @@ reducev_notab( VipsReducev *reducev,
 	const int n = reducev->n_point;
 	const int l1 = lskip / sizeof( T );
 
-	long double cy[MAX_POINT];
+	double cy[MAX_POINT];
 
 	vips_reduce_make_mask( cy, reducev->kernel, reducev->n_point,
 		reducev->vshrink, y );
 
 	for( int z = 0; z < ne; z++ )
-		out[z] = reduce_sum<T, long double>( in + z, l1, cy, n );
+		out[z] = reduce_sum<T, double>( in + z, l1, cy, n );
 }
 
 static int
```

</details>

It produces this incorrect ResampleScope output:
| `reduceh` | `reducev` |
| :---: |  :---: |
|  |  |

(i.e., the same output that PR #1872 and #3501 tried to resolve)

lovell:
How portable is the use of `long double`? Does it matter if it maps to `double`, as both [MSVC](https://developercommunity.visualstudio.com/t/Add-80-bit-or-more-long-double-in-Visual/1533848?entry=myfeedback&space=8) and [LLVM on ARM64](https://discourse.llvm.org/t/support-for-128bit-long-double-on-apple-silicon/66108) might do this?

Useful compiler `sizeof(long double)` checker at https://godbolt.org/z/46ac7b9fq

kleisauke:
Temporarily rebased against PR #3553, as it depends on that one.

libvips/resample/templates.h
```diff
@@ -473,10 +476,41 @@ vips_reduce_make_mask(T *c, VipsKernel kernel, const int n_points,
...
+ */
+template <typename T>
+struct intermediate<T, false, 4> {
+	typedef double type;
+};
+
+/* 64-bit floating-point types needs a 128-bits intermediate.
+ */
+template <typename T>
+struct intermediate<T, false, 8> {
```
lovell, line 506: If `long double` is unavailable should we use `4` instead of `8` here? Perhaps something more dynamic? I can see the `intermediate` struct uses `sizeof()` for missing values, but by passing `8` I think this is overridden. Or have I misunderstood?

lovell: Thank you, that's much clearer.

PR #3546, opened 2023-06-24
Add support for Canon CR3 magic bytes

416c616e:
Canon CR3 formats currently get incorrectly detected as TGA, so explicitly handling their edge case first.

### DIFF
--- a/libvips/foreign/magick.c
+++ b/libvips/foreign/magick.c
@@ -75,6 +75,17 @@
 static const char *
 magick_sniff( const unsigned char *bytes, size_t length )
 {
+	if( length >= 12 &&
+		bytes[4] == 'f' &&
+		bytes[5] == 't' &&
+		bytes[6] == 'y' &&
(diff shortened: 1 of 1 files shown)


416c616e:
Canon CR3 formats currently get incorrectly detected as TGA, so explicitly handling their edge case first.

lovell:
Thanks for the PR, did you see #3460 ?

416c616e:
> Thanks for the PR, did you see #3460 ?

I did see that, I think I'll amend my PR to only include CR3.

kleisauke:
Given that CR3 images uses ISOBMFF as container format, I think this could be easily fixed in ImageMagick instead. 

I just opened PR https://github.com/ImageMagick/ImageMagick/pull/6450 for this.

kleisauke:
... I also opened https://github.com/libvips/libvips/pull/3555 to ensure CR3 images are not loaded through ImageMagick's TGA loader.

PR #3550, opened 2023-06-29
Fix clang warnings

jcupitt:
Fix a few warnings from clang. With this PR, libvips builds cleanly on macOS and on Ubuntu in debug and release mode, and pytest passes.

### DIFF
--- a/libvips/iofuncs/init.c
+++ b/libvips/iofuncs/init.c
@@ -270,31 +270,26 @@ vips_load_plugins( const char *fmt, ... )
 	if( !(dir = g_dir_open( dir_name, 0, NULL )) )
 		return;
 
-        while( (name = g_dir_read_name( dir )) )
-                if( vips_ispostfix( name, "." G_MODULE_SUFFIX )
-#if ENABLE_DEPRECATED
-				|| vips_ispostfix( name, ".plg" ) 
-#endif
-			) { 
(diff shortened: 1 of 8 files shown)


libvips/iofuncs/init.c
```diff
@@ -270,31 +270,26 @@ vips_load_plugins( const char *fmt, ... )
 	if( !(dir = g_dir_open( dir_name, 0, NULL )) )
 		return;
 
-        while( (name = g_dir_read_name( dir )) )
-                if( vips_ispostfix( name, "." G_MODULE_SUFFIX )
-#if ENABLE_DEPRECATED
-				|| vips_ispostfix( name, ".plg" ) 
-#endif
-			) {
```
kleisauke, line 278: I think we could safely remove this (i.e. build modules as `*.dylib` on macOS):
https://github.com/libvips/libvips/blob/00e4047ff9a223132adfdf874f723139e399c6f2/libvips/meson.build#L75-L80
Now that we no longer use the deprecated `G_MODULE_SUFFIX` macro. Let me open a follow-up PR for this.

kleisauke (approved):

libvips/iofuncs/init.c
kleisauke: #3554

PR #3553, opened 2023-06-30
reduce{h,v}: simplify coefficients handling

kleisauke:
- De-duplicate `n_points` calculations.
- Extract the various filters in inline functions.
- Merge the `calculate_coefficients*` functions into a single one.
- Simplify `reduce_sum()`.

Split from: #3532.

### DIFF
--- a/libvips/resample/presample.h
+++ b/libvips/resample/presample.h
@@ -70,8 +70,6 @@ GType vips_resample_get_type(void);
 #define MAX_POINT (2000)
 
 int vips_reduce_get_points(VipsKernel kernel, double shrink);
-void vips_reduce_make_mask(double *c,
-	VipsKernel kernel, int n_points, double shrink, double x);
 
 void vips_reduceh_uchar_hwy(VipsPel *pout, VipsPel *pin,
 	int n, int width, int bands,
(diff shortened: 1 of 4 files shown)


jcupitt:
Does this mean we will build on C++ before C++11? Maybe the "C++ version required" declration in meson.build should change too.

kleisauke:
This will indeed restore compatibility with C++98. Commit 963deded75666b0899c66c53389219751a716cd5 lowers the minimum required C++ version to C++98.

We could also remove that line, but that would not allow us to detect regressions (since most compilers use C++17 as the default nowadays).

kleisauke:
I just rebased this PR to resolve the merge conflicts.

kleisauke:
Friendly ping, the diff is probably best viewable with the `Hide whitespace changes` option.

kleisauke:
Split into multiple commits to make reviewing easier.

kleisauke:
Godbolt link to verify the inlining behaviour: https://godbolt.org/z/cYbKWTW6T (i.e. there is only a call made to `sin()`).

lovell (approved):
I've done some testing with this, everything appears to work as before. Much simpler, thanks Kleis.

PR #3554, opened 2023-06-30
Use `.dylib` as dynamic module extension on macOS

kleisauke:
Should be safe, we no longer depend on `G_MODULE_SUFFIX`.


jcupitt:
Looks good here!

PR #3557, opened 2023-07-02
Add direct mode to dzsave

jcupitt:
Rather than making a pipeline for each tile, call directly into vips2jpeg. For small (eg. 256x256 tiles), this is a lot quicker.

Before:

```
$ vips copy CMU-3.svs[rgb] x.szi --vips-progress
vips temp-2: 66000 x 45402 pixels, 32 threads, 128 x 128 tiles, 640 lines in buffer
vips temp-2: done in 17.7s         
```

With this PR:

```
$ vips copy CMU-3.svs[rgb] x.szi --vips-progress
vips temp-2: 66000 x 45402 pixels, 32 threads, 128 x 128 tiles, 640 lines in buffer
vips temp-2: done in 7.99s        
```

This is just a test of the idea, there's a lot of missing stuff.

- [x] add a `compression=` option and implement at least JPG and PNG (don't do this -- pngsave is so complex and annoying, and we'd need to do spngsave as well)
- [x] add a `Q=` option, and perhaps some others (tiffsave could be a guide)
- [x] implement google tile padding
- [x] quite a bit of cleaning up and refactoring needs doing
- [x] it should be quicker -- we still don't saturate all cores, though it's much better
- [x] check for memleaks etc.
- [x] deprecate `suffix`? or we could leave it


fredprodibi:
Would this work also with webp? and for tiles that are 512x512?

jcupitt:
It'll do any tilesize.

It's JPEG only as it stands, but other formats will probably be added.

fredprodibi:
The performance gain looks amazing, do you know when it's going to be released?

Thank you!

jcupitt:
I found a `vips_copy()` in jpegsave we can skip so it's a bit quicker again:

```
$ vips copy CMU-3.svs[rgb] x.szi --vips-progress
vips_foreign_save_dz_build: enabling direct mode
vips temp-2: 66000 x 45402 pixels, 32 threads, 128 x 128 tiles, 640 lines in buffer
vips temp-2: done in 7.71s          
```

jcupitt:
I think this is ready to review and merge now. I implemented google maps mode, and all tests now pass.

Before:

```
$ time vips copy CMU-3.svs[rgb] x.szi --vips-progress
vips temp-2: 66000 x 45402 pixels, 32 threads, 128 x 128 tiles, 640 lines in buffer
vips temp-2: done in 18.5s          

real	0m18.658s
user	3m7.196s
sys	1m56.522s
```

After:

```
$ time vips copy CMU-3.svs[rgb] x.szi --vips-progress
vips temp-2: 66000 x 45402 pixels, 32 threads, 128 x 128 tiles, 640 lines in buffer
vips temp-2: done in 8.53s          

real	0m8.723s
user	1m56.327s
sys	0m8.524s
```

So it's a bit more than 2x faster. The saving comes from now setting up and tearing down a pipeline for every tile, as you can see from the drop in systime!

jcupitt:
There are two other big changes here:

1. I moved the libarchive stuff out to `archive.c`. I thought `dzsave.c` was getting too big.
2. We used to do padding of tiles in google mode in the tile save pipeline. The new direct mode doesn't have a tile save pipeline, so instead it now has a single pipeline which pads the input image, and tile save just writes each block of computed pixels.

libvips/foreign/dzsave.c
```diff
@@ -2081,78 +2043,97 @@ vips_foreign_save_dz_build(VipsObject *object)
...
+		// level
+		width *= 1 << level->n;
+		height *= 1 << level->n;
+
+		level_free(level);
+
+#ifdef DEBUG
+#endif /*DEBUG*/
+		printf("vips_foreign_save_dz_build: "
+			"google mode outputs a %d x %d pixel image\n", width, height);
```
kleisauke, line 2088:
```suggestion
		printf("vips_foreign_save_dz_build: "
			"google mode outputs a %d x %d pixel image\n", width, height);
#endif /*DEBUG*/
```

libvips/iofuncs/target.c
```diff
@@ -666,6 +666,9 @@ vips_target_finish(VipsTarget *target)
  * The data is NOT automatically null-terminated. vips_target_putc() a '\0'
  * before calling this to get a null-terminated string.
  *
+ * This won't work after the tyarget has been ended -- read the blob in that
```
kleisauke, line 669:
```suggestion
 * This won't work after the target has been ended -- read the blob in that
```

libvips/foreign/archive.c
```diff
@@ -0,0 +1,364 @@
...
+
+// write to a filesystem directory
+VipsArchive *
+vips__archive_new_to_dir(const char *base_dirname)
+{
+	VipsArchive *archive;
+
+	vips__archive_init();
+
+	archive = VIPS_NEW(NULL, VipsArchive);
```
kleisauke, line 129:
```suggestion
	if (!(archive = VIPS_NEW(NULL, VipsArchive)))
		return NULL;
```

libvips/foreign/dzsave.c
```diff
@@ -580,66 +423,64 @@ vips_foreign_save_dz_dispose(GObject *gobject)
...
+static Level *
+pyramid_build(VipsForeignSaveDz *dz,
+	Level *above, int width, int height, VipsRect *save_area)
 {
 	VipsForeignSave *save = VIPS_FOREIGN_SAVE(dz);
-	Layer *layer = VIPS_NEW(dz, Layer);
+	Level *level = VIPS_NEW(dz, Level);
 
 	VipsRect strip;
 	int limit;
```
kleisauke, line 439:
```suggestion

	Level *level;
	VipsRect strip;
	int limit;

	if (!(level = VIPS_NEW(dz, Level)))
		return NULL;

```

kleisauke (approved):
LGTM! I added a few minor comments.

libvips/foreign/dzsave.c
```diff
@@ -1574,43 +1378,26 @@ strip_work(VipsThreadState *state, void *a)
...
-		}
-		g_object_unref(x);
-
-		x = t;
-	}
-
-	out = tile_name(layer,
-		state->x / dz->tile_step, state->y / dz->tile_step);
-
-	/* g_build_filename() can return NULL when it exceeds the path limits.
```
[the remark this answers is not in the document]
jcupitt: Yes, exactly. I think it was a rename thing.

jcupitt:
Thanks for the review!

kleisauke:
I fixed some clang-format errors and added a ChangeLog note.

This PR is ready to land. Should I use this as final commit message?:
```
Add direct mode to dzsave (#3557)

Rather than making a pipeline for each tile, call directly into
vips2jpeg. For small (eg. 256x256 tiles), this is a lot quicker.
```

jcupitt:
Sure, sounds good!

(sorry, I should have done the clang-format thing)

PR #3558, opened 2023-07-05
make icc_transform autoset depth

jcupitt:
default to 16 output for a 16-bit source image


jcupitt:
Klies said:

> /cc @tufty perhaps we should set depth to 16  automatically for VIPS_INTERPRETATION_{RGB,GREY}16? This would make the following code unnecessary downstream:
> https://github.com/weserv/images/blob/5.x/src/api/processors/thumbnail.cpp#L514-L515
> https://github.com/lovell/sharp/blob/main/src/pipeline.cc#L320

kleisauke (approved):
LGTM!

PR #3559, opened 2023-07-07
Release build by default

jcupitt:
I've had a few users forget to enable release mode and then have performance issues. meson defaults to a debug build, so it's an easy mistake to make.

This PR makes meson default to a release build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,9 +107,6 @@ configuration.
   see `meson_options.txt` and the list below for a summary of all the libvips
   dependencies.
 
-- Meson will do a debug build by default. Add `--buildtype release` for a 
-  release (optimised) build.
-
 - You might need to add `--libdir lib` on Debian if you don't want the arch 
   name in the library path.
 
--- a/meson.build
+++ b/meson.build
@@ -6,6 +6,8 @@ project('vips', 'c', 'cpp',
         'c_std=gnu99',
 	# we use some C++11 features 
 	'cpp_std=c++11',
+        # do a release (optimised) build by default
+        'buildtype=release',
         # turn off asserts etc. in release mode
         'b_ndebug=if-release'
     ]


libvips/iofuncs/init.c
```diff
@@ -648,6 +648,13 @@ vips_init( const char *argv0 )
 	if( g_getenv( "VIPS_BLOCK_UNTRUSTED" ) )
 		vips_block_untrusted_set( TRUE );
 
+	/* Since it's so easy to forget the "--buildtype release" arg to meson.
+	 */
+#ifdef DEBUG_BUILD
```
kleisauke, line 653: Given that we use Meson's `b_ndebug=if-release`.
https://github.com/libvips/libvips/blob/f807d7b1aaa09d10e4b38583fb417fb7f74b5e3c/meson.build#L9-L10

Perhaps it would be simpler to just check for `NDEBUG` instead?:
```suggestion
#ifndef NDEBUG
```

jcupitt: We could do, though of course someone could build with `-DNDEBUG=1`, so I thought maybe a define that is *just* used for signalling debug mode might be safer.

kleisauke:
Thinking about this further, perhaps we should do a release build by default? For example:
```diff
--- a/meson.build
+++ b/meson.build
@@ -6,6 +6,8 @@ project('vips', 'c', 'cpp',
         'c_std=gnu99',
 	# we use some C++11 features 
 	'cpp_std=c++11',
+        # do a release (optimised) build by default
+        'buildtype=release',
         # turn off asserts etc. in release mode
         'b_ndebug=if-release'
     ]
--- a/README.md
+++ b/README.md
@@ -107,9 +107,6 @@ configuration.
   see `meson_options.txt` and the list below for a summary of all the libvips
   dependencies.
 
-- Meson will do a debug build by default. Add `--buildtype release` for a 
-  release (optimised) build.
-
 - You might need to add `--libdir lib` on Debian if you don't want the arch 
   name in the library path.
 
```

This corresponds to the Autotools semantics where one had to pass `--enable-debug` to produce debug builds.

jcupitt:
Good idea, that's probably a better fix.

jcupitt:
OK, I changed it to do a release build by default, as you suggested.

kleisauke (approved):

PR #3563, opened 2023-07-11
Run codespell on pr and push to master

jcupitt:


jcupitt:
I've not fixed any of the spelling errors yet.

jcupitt:
That's a lot of stilly spelling errors!

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,22 @@
+---
+name: Codespell
```
kleisauke, line 2: The three dashes (`---`) used to indicate the start of a YAML document should not be mandatory.
```suggestion
name: Codespell
```

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,22 @@
...
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
```
kleisauke, line 8: Perhaps we should run this workflow on other branches too?:
```suggestion
on: [push, pull_request]
```
(for example, it could be useful when someone targets the 8.14 branch directly)

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,22 @@
...
+
+permissions:
+  contents: read
```
kleisauke, line 11: According to https://github.com/codespell-project/codespell/blob/master/.github/workflows/codespell.yml#L6, the `codespell-project/actions-codespell` action doesn't need any permissions.
```suggestion
permissions: {}
```

kleisauke (approved):
Looks great! I left a few inline comments regarding the GitHub workflow.

kleisauke (approved):

RatStar811 (approved):

Issue #3564, opened 2023-07-12
Linear thumbnail CMYK embedded ICC profile error


jjonesrs:
## Bug report

**Describe the bug**
Possibly related to #3159, vipsthumbnail with `--linear` fails on the attached image, but succeeds without that option. I've also tried setting the input profile and output profile to cmyk and srgb respectively. It's also possible that the image itself has issues, but it seems to have a profile of Coated GRACoL 2006 (ISO 12647-2:2004) and `vipsImage.Contains("icc-profile-data")` returns true.

The real application uses NetVips, but I believe vipsthumbnail provides simpler reproduction steps.

**To Reproduce**
1. Use the image `sample.tif` in [sample.zip](https://github.com/libvips/libvips/files/12033952/sample.zip)
2. Run vipsthumbnail with the following arguments: `sample.tif --linear`
3. Vips prints `profile incompatible with image`
4. Vips terminates with `icc_import: unable to load or find any compatible input profile`

**Environment**
 - OS: Windows 11 and AWS Lambda Amazon Linux x86_64
 - Vips: 8.14.2

kleisauke:
I was able to reproduce this. It seems that the provided test image is a CMYK image with 5 bands and the embedded ICC profile is not an N-colour profile (i.e. the ICC profile ought to have the `5CLR` signature in this case).
```console
$ vipsheader sample.tif 
sample.tif: 2000x2000 uchar, 5 bands, cmyk, tiffload
$ exiftool -icc_profile:ColorSpaceData sample.tif 
Color Space Data                : CMYK
```

With this patch:
```diff
--- a/libvips/colour/icc_transform.c
+++ b/libvips/colour/icc_transform.c
@@ -580,9 +580,12 @@ vips_image_expected_bands( VipsImage *image )
 			expected_bands = 3;
 			break;
 
+		case VIPS_INTERPRETATION_CMYK:
+			expected_bands = 4;
+			break;
+
 		case VIPS_INTERPRETATION_MULTIBAND:
 		case VIPS_INTERPRETATION_HISTOGRAM:
-		case VIPS_INTERPRETATION_CMYK:
 		case VIPS_INTERPRETATION_MATRIX:
 		case VIPS_INTERPRETATION_FOURIER:
 		default:
```

I'm able to `icc_import` the image with the embedded ICC profile, but this would probally break NCLR support, as noted in: https://github.com/libvips/libvips/pull/3372#discussion_r1125432552.

jcupitt:
[...]

PR #3372, opened 2023-03-04
colour: fix ICC handling of RGB images w/ a monochrome profile

kleisauke:
This restores the `vips_image_expected_bands` function and matching logic (which had been removed in commit 1297f2c) to almost its original state.

See: #3361.

> **Note**: this PR targets the [`8.14`](https://github.com/libvips/libvips/tree/8.14) branch.

Issue #3139, opened 2022-11-03
icc_transform: incompatible profile isn't ignored


dloebl:
Since 1297f2c6d790aba928c1f9376674890ad75c6e39 incompatible source profiles do not seem to be ignored anymore:

**Before 1297f2c6d790aba928c1f9376674890ad75c6e39:**
```
$ vips icc_transform --input-profile=sRGB.icm test.jpg bef.jpg sRGB.icm

(vips:9111): VIPS-WARNING **: 10:55:58.075: profile incompatible with image
```
| original | output |
| - | - |
| ![test](https://user-images.githubusercontent.com/34104349/199690097-5a26bbdc-76d5-4774-be65-625d2ba7e7d9.jpg) |  ![bef](https://user-images.githubusercontent.com/34104349/199692114-50a786d9-3fa2-48ee-95b2-6511442726b5.jpg) |

**Current master (17ca29adc9c52ef48cf5a9a59c26a83d0693fbbe):**
```
$ vips icc_transform --input-profile=sRGB.icm test.jpg now.jpg sRGB.icm
```
| original | output |
| - | - |
| ![test](https://user-images.githubusercontent.com/34104349/199690239-f8edae02-78b9-4f05-9e49-08b942ebb679.jpg) | ![now](https://user-images.githubusercontent.com/34104349/199692196-c3d1cfed-14f4-4449-b645-6c64ea19d265.jpg) |

[sRGB.icm](https://github.com/libvips/libvips/blob/master/libvips/colour/profiles/sRGB.icm)
[test.jpg](https://user-images.githubusercontent.com/34104349/199690097-5a26bbdc-76d5-4774-be65-625d2ba7e7d9.jpg)

```
$ vipsheader test.jpg
test.jpg: 640x426 uchar, 4 bands, cmyk, jpegload
```

jcupitt:
Hi Daniel, 

I added a test for this specific case, but I don't think we can do the general test we used to have :( Thanks for the report!

BTW you could write:

```
$ vips icc_transform --input-profile=sRGB.icm test.jpg bef.jpg sRGB.icm
```

As:

```
$ vips icc_transform --input-profile=srgb test.jpg bef.jpg srgb
```

Since you can use the libvips built in profile names anywhere you use a profile filename. master has `p3` and `cmyk` as built in profiles as well.

Issue #730, opened 2017-09-07
vips_thumbnail: (CIE)LAB TIFF to sRGB


kleisauke:
It seems that `vips_thumbnail` has some issues with processing a (CIE)LAB TIFF to the sRGB colourspace.

[Input image](https://github.com/lovell/sharp/raw/master/test/fixtures/cielab-dagams.tiff)
[sRGB profile](https://github.com/jcupitt/nip2/blob/master/share/nip2/data/sRGB.icm)

**Input image to sRGB png**
```bash
vipsthumbnail cielab-dagams.tiff \
    --size 200x \
    -o %s_thumbnail_srgb.png \
    --eprofile sRGB.icm
````

Warnings:
```bash
(vipsthumbnail:17830): VIPS-WARNING **: Incompatible type for "RichTIFFIPTC"; tag ignored (x8)
(vipsthumbnail:17503): VIPS-WARNING **: icc_transform: intent 1 (relative) not supported by input profile; falling back to default intent
```

Output (wrong):
![cielab-dagams_thumbnail_srgb](https://user-images.githubusercontent.com/12746591/30172888-54d44e92-93f6-11e7-8f32-a16a45f3d938.png)

**Input image to sRGB png with the perceptual intent**
```bash
vipsthumbnail cielab-dagams.tiff \
    --size 200x \
    -o %s_thumbnail_srgb_percep.png \
    --intent perceptual \
    --eprofile sRGB.icm
````

Warning (x8):
```bash
(vipsthumbnail:17830): VIPS-WARNING **: Incompatible type for "RichTIFFIPTC"; tag ignored
```

Output (wrong):
![cielab-dagams_thumbnail_srgb_percep](https://user-images.githubusercontent.com/12746591/30172875-4745652c-93f6-11e7-829c-bc14cab3c4a9.png)

**Input image to png (without the `--eprofile` flag)**
```bash
vipsthumbnail cielab-dagams.tiff \
    --size 200x \
    -o %s_thumbnail.png
````

Same warnings as the second command.

Output (correct):
![cielab-dagams_thumbnail](https://user-images.githubusercontent.com/12746591/30172867-4479db0c-93f6-11e7-8c8b-0530db26764c.png)

Tested with libvips 8.6.0-alpha2.

jcupitt:
Hi Kleis, I think your image is wrong. CIELAB images should not have profiles attached to them, they are in profile connection space already. This is like an RGB image with a CYMK profile. 

[...]

PR #3565, opened 2023-07-16
colour: improve ICC compatibility check for CMYK images

kleisauke:
See: #3564.

Targets the 8.14 branch.

### DIFF
--- a/libvips/colour/icc_transform.c
+++ b/libvips/colour/icc_transform.c
@@ -545,23 +545,18 @@ vips_icc_print_profile( const char *name, cmsHPROFILE profile )
 }
 #endif /*DEBUG*/
 
-/* How many bands we expect to see from an image after preprocessing by our
- * parent classes. This is a bit fragile :-( 
- *
- * FIXME ... split the _build() for colour into separate preprocess / process
- * / postprocess phases so we can load profiles after preprocess but before
- * actual processing takes place.
+/* TRUE if the number of bands in the ICC profile is compatible with the
+ * image.
  */
-static int
-vips_image_expected_bands( VipsImage *image )
+static gboolean
+vips_image_is_profile_compatible( VipsImage *image, int profile_bands )
 {
-	int expected_bands;
-
 	switch( image->Type ) {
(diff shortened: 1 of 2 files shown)


libvips/colour/icc_transform.c
```diff
@@ -575,22 +570,24 @@ vips_image_expected_bands( VipsImage *image )
...
 
 		case VIPS_INTERPRETATION_MULTIBAND:
 		case VIPS_INTERPRETATION_HISTOGRAM:
-		case VIPS_INTERPRETATION_CMYK:
 		case VIPS_INTERPRETATION_MATRIX:
 		case VIPS_INTERPRETATION_FOURIER:
 		default:
-			expected_bands = image->Bands;
-			break;
+			return( image->Bands == profile_bands );
```
jcupitt, line 589: Maybe ">="? There can be more image bands than profile bands if there are alpha channels.

kleisauke: Commit bec4942c6af0e629857da158c7e8657dfa0e1fff relaxes this check to `>=`.

jcupitt:
This looks too easy! I can't help feeling there's some corner case we've overlooked, but I don't know what it is.

kleisauke:
As well as checking it against the test case in #3564, I also checked it against issue #3139 and the test cases in PR #3372. But indeed, perhaps we overlooked something, I will verify some old issues.

jcupitt:
I checked it against CMYKs with sRGB profiles, RGBs with CMYK profiles, and mono images with RGB profiles, and it all seems fine. I didn't try alpha channels.

I think maybe I'm only uneasy because I remember changing this logic before and breaking everything :( It's probably fine.

kleisauke:
The test case mentioned in https://github.com/libvips/libvips/issues/730#issuecomment-1345375891 that was fixed with PR #3139 also still produces the expected behavior.

I'll do some more testing with CMYK images, since that's the only behavior change in this PR (the change in commit bec4942c6af0e629857da158c7e8657dfa0e1fff is also a behavior change, but that's much more uncommon, I think).

Issue #2282, opened 2021-06-03
Strip all metadata except the color profile?


grishka:
I'd like to write a webp file while stripping metadata like EXIF but keeping the color profile so the colors always look accurate. I'm using libvips in a server application and the files I'm processing are uploaded by users, so it's important to not leak any sensitive data that might be in EXIF tags. I'm currently passing `strip=true` to `vips_image_write_to_file` but that strips everything, including the color profile.

jcupitt:
Hi @grishka,

I would remove all metadata except the ICC profile. In Python, something like:

```python
# take a private copy of the image before we start modifying it 
image = image.copy()

for field in image.get_fields():
    if field != "icc-profile-data":
        image.remove(field)

image.write_to_file(filename)
```

See https://libvips.github.io/libvips/API/current/libvips-header.html#vips-image-get-fields

PR #3585, opened 2023-07-27
Add "keep" flag to savers

a3mar:
(edited to reflect the current state of this PR)

This PR introduces a "keep" flag option for all savers, allowing more control over which metadata to retain, unlike the now-deprecated "strip" option, which used to remove all metadata.

Old description:

---

Hello!

There are two common save parameters:
* strip - strips all meta from result image
* profile - adds custom ICC profile to result

But its impossible to use both at one time. When i'm trying to save with [strip,profile=./AdobeRGB1998.icc], there is no icc profile in resulting image.

It's very useful to get this ability, for example, i using stream image processing (custom source and custom target) and access=sequential. I need to make some operation, like thumbnail, remove all meta, and add ICC profile (same or different). There are two variants for me:
1. Copy image object and remove all unnecessary fields in loop ( like in #2282 ) - this solution breaks streaming
2. Encode result to stream with [strip], then decode, add color profile (with import_icc operation) and encode to stream - this works good, but performance is lower by ~25% in my case.

In this PR i tried to separate strip and profile options for most popular web-formats (where ICC may be needed). It works, i wrote test for this behavior.


jcupitt:
Hi Dmitry, thanks for working on this!

> Copy image object and remove all unnecessary fields in loop ( like in Strip all metadata except the color profile? #2282 ) - this solution breaks streaming

It shouldn't -- copy is a streaming operator, so copy + delete non-icc metadata ought to work ok. But I agree something to strip and attach a custom profile would be handy at the CLI.

How about adding a bool flag to `VipsForeignSave` called `keep_profile`? The writers can then check this to see if they should also remove the ICC profile (as well as everything else). We could set this flag automatically if we see a `profile` argument.

So you could then have:

```
# no extra metadata written
vips copy k2.jpg x.jpg[strip]
# no metadata, except the profile
vips copy k2.jpg x.jpg[strip,keep-profile]
# no metadata except the profile
vips copy k2.jpg x.jpg[strip,profile=banana.icc]
```

This should keep back compat and be nice and flexible. What do you think?

a3mar:
Hi
> It shouldn't -- copy is a streaming operator

I investigated realization in elixir binding, used in my project, under the hood [it uses "copy_memory"](https://github.com/akash-akya/vix/blob/v0.22.0/lib/vix/vips/mutable_image.ex#L180) instead of "copy". I'll try to patch it for using "copy" and reproduce situation.

> How about adding a bool flag to `VipsForeignSave` called `keep_profile`?

It's interesting idea, i think i can implement it, but first i need more research the code. For my current project this flag is not needed, but it will be very handy for processing something like photo archive.

> CI / macOS (12.6) - Xcode 14.2 (pull_request) Failing after 3m 

Maybe you have some ideas why xcode build fails? I have no mac and xcode, and cant reproduce that. It fails at check that when saving with [strip,profile="file.icc"], there is no icc in resulting image.

jcupitt:
macOS fails with:

> FAILED: meson-internal__test 

argh, I bet it's some homebrew thing, I'll have a look, thanks!

a3mar:
Hello, i added "keep_profile" flag for jpg, webp and png, updated test to check that its working (also added test that not stripped really not).

libvips/foreign/webpsave.c
```diff
@@ -450,6 +453,67 @@ vips_webp_add_chunks(VipsForeignSaveWebp *write)
...
+vips_webp_get_webp_name(const char *vips_name)
+{
+	int position = 0;
+	while (position < vips__n_webp_names) {
+		gboolean needed = strcmp(vips_name, vips__webp_names[position].vips) == 0;
+		const char *webp_name = vips__webp_names[position].webp;
+
+		if (needed) {
+			return webp_name;
+			break;
```
jcupitt, line 466: Not needed.

libvips/foreign/webpsave.c
```diff
@@ -486,22 +550,39 @@ vips_webp_add_metadata(VipsForeignSaveWebp *webp)
...
+				return -1;
+		}
+	}
+
+	/* If save is called with "keep_profile" and
+	 * "strip" options, we need to copy ICC profile
+	 * from original image to new webp
+	 */
+	if (save->keep_profile && save->strip) {
+		if (vips_webp_add_original_icc(webp))
```
jcupitt, line 574: Again, just one call to `vips_webp_add_original_icc()` would be clearer.

libvips/foreign/vips2jpeg.c
```diff
@@ -729,20 +733,24 @@ write_vips(Write *write, int qfac, const char *profile,
...
 				write_profile_meta(write))
 				return -1;
 		}
 	}
 
+	if (keep_profile) {
+		if (vips_image_get_typeof(in, VIPS_META_ICC_NAME) &&
+			write_profile_meta(write))
+			return -1;
+	}
```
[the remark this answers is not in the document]
jcupitt: Another improvement might be to add `profile` as a member of `VipsForeignSave`, along with `keep_profile`, and do away with the `profile` members for webp, jpeg, etc.

In `vips_foreign_save_build()` we could have the logic for what the various combinations of these arguments mean. Perhaps:

```
if !strip or is_set(profile)
    keep_profile = true
```

ie. `keep_profile` is true if we should write a profile. Now you could change this code to be:

```C
    if (!strip) {
        if (vips__exif_update(write->in) ||
            write_exif(write) ||
            write_xmp(write) ||
            write_blob(write,
                VIPS_META_IPTC_NAME, JPEG_APP0 + 13))
            return -1;
    }

    if (keep_profile) {
        if (profile) {
            if (write_profile_file(write, profile))
                return -1;
        }
        else if (vips_image_get_typeof(in, VIPS_META_ICC_NAME)) {
            if (write_profile_meta(write))
                return -1;
        }
    }
```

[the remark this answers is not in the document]
jcupitt: That makes it seem that `keep_profile` should be renamed as `write_profile`. What do you think?

test/test_strip_profile.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
+#
```
jcupitt, line 2: This is really nice!

jcupitt (commented):
This is great! Thank you very much for doing this work, Dmitry.

I saw a few tiny things, and had a suggestion for making the keep-profile logic in the writers a bit simpler.

libvips/foreign/vips2jpeg.c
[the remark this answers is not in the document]
a3mar: I didn't correctly understand the idea before doing, sorry)  But behavior may be changed in this variant. 
Now, if saving with (for example) [Q=85], we have resulting image with the original icc, if we do changes like that, when saving with same options we'll get image without it.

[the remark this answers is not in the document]
jcupitt: Ah maybe I missed something, though I think something like that should work.

jcupitt:
TODO:

- we should add tiff and heif save as well
- spngsave will need this too
- docs need revising
- add a line to changelog

jcupitt:
I had a go at tiff and spng:

https://github.com/libvips/libvips/commit/58a30dbbb2dd4ea5c912241a296d80bcfb370de0

https://github.com/libvips/libvips/commit/f1cabf91f0e09cc77e2cb20bb5f8531060cadbf3

a3mar:
Added HEIF , but it cannot be tested from build folder and current script, because  it tries to found vips-heif.so in installation prefix. I'll rewrite test to python and fix bugs if any.

a3mar:
interesting fact, if i build with custom prefix, like that:
```
meson setup build --prefix /home/azmar/Projects/vips-install/
meson compile
meson install
```
in build folder it all ok, in installation folder i have default prefix:

```
localhost:~/Projects/libvips/build > VIPS_INFO=1 tools/vips --vips-config|grep VIPS_
VIPS-INFO: 14:23:02.428: VIPS_PREFIX = /home/azmar/Projects/vips-install
VIPS-INFO: 14:23:02.428: VIPS_LIBDIR = /home/azmar/Projects/vips-install/lib64

localhost:~/Projects/libvips/build > VIPS_INFO=1 /home/azmar/Projects/vips-install/bin/vips --vips-config|grep VIPS_
VIPS-INFO: 14:23:07.012: VIPS_PREFIX = /usr
VIPS-INFO: 14:23:07.012: VIPS_LIBDIR = /usr/lib64
```

jcupitt:
> in build folder it all ok, in installation folder i have default prefix:

I would guess it's picking up your system libvips at runtime. Try setting `LD_LIBRAY_PATH`.

jcupitt:
ie.

```
meson setup build --prefix /home/azmar/Projects/vips-install
meson compile
meson install
cd
export PATH=/home/azmar/Projects/vips-install/bin:$PATH
# you might need arch in that path too, depending on your distro
export LD_LIBRARY_PATH=/home/azmar/Projects/vips-install/lib
# and this should now work
VIPS_INFO=1 vips --vips-config
```

kleisauke:
Thanks for working on this! I was thinking, perhaps we should just deprecate the `strip` flag in favor of a `preserve` flag as [outlined here](https://github.com/bleroy/core-imaging-playground/pull/26#issuecomment-1101263632)?

I had a quick go:
https://github.com/a3mar/libvips/compare/strip-with-profile...kleisauke:add-preserve-flag

So, you could write this:
```bash
# no extra metadata written
$ vips copy k2.jpg x.jpg[strip]
# no metadata, except the profile
$ vips copy k2.jpg x.jpg[strip,keep-profile]
# no metadata except a custom profile
$ vips copy k2.jpg x.jpg[strip,profile=banana.icc]
# all metadata + custom ICC profile
$ vips copy k2.jpg x.jpg[profile=banana.icc]
```

As:
```bash
# no extra metadata written
$ vips copy k2.jpg x.jpg[strip] # deprecated, so can still be used
$ vips copy k2.jpg x.jpg[preserve=none]
# no metadata, except the profile
$ vips copy k2.jpg x.jpg[preserve=icc]
# no metadata except a custom profile
$ vips copy k2.jpg x.jpg[preserve=icc,profile=banana.icc]
# all metadata + custom ICC profile
$ vips copy k2.jpg x.jpg[profile=banana.icc]
```

jcupitt:
Using space to separate a list of strings is a bit stupid (we implement this now, I think). Comma is taken, so we could maybe add `|`?

```
vips copy a.jpg b.jpg[preserve=icc-profile-data|exif-data]
```

Looks nice.

In python it would look like:

```python
image.write_to_file("x.jpg", preserve=["icc-profile-data", "exif-data"])
```

Which is kind-of OK. Ruby would be:

```python
image.write_to_file  "x.jpg", preserve: %i[icc-profile-data exif-data]
```

kleisauke:
Supporting a list of strings would require introducing a new GValue marshaller (I think `G_TYPE_STRV`?) in the various bindings, unfortunately.

You're right that we could centralize this in `vips_foreign_save_build()`, I implemented this with commit https://github.com/kleisauke/libvips/commit/457f6cba78a7f6aeba064f33ea3f142c603e0c86.

Using a `|` as a separator for list items sounds good to me, perhaps we should do this for flags as well, since that is currently not supported either.
```console
$ vips copy test/test-suite/images/sample.jpg x.png[filter=avg\|paeth]
pngsave_target: flags 'VipsForeignPngFilter' has no member 'avg|paeth', should be one of: none, sub, up, avg, paeth, all
pngsave_target: 'avg|paeth' is not an integer
$ vips copy test/test-suite/images/sample.jpg x.png[filter=192]
$ pngcheck -vv x.png | grep filters -A 1 | tail -5
    row filters (0 none, 1 sub, 2 up, 3 avg, 4 paeth):
      4 4 4 4 4 4 4 4 4 3 4 4 4 4 4 4 4 (440 out of 442)
--
    row filters (0 none, 1 sub, 2 up, 3 avg, 4 paeth):
      3 4 (442 out of 442)
```

jcupitt:
> Supporting a list of strings would require introducing a new GValue marshaller (I think G_TYPE_STRV?) in the various bindings, unfortunately.

Ooop you're right, I thought we had an `ARRAY_STRING` type for some reason. Yes, we'd need to marshal STRV, that's annoying. 

Yup, `|` for flag enums would be very handy. I suppose we could allow `|` for `ARRAY_INT` too, eg.:

```
vips copy x.png y.jpg[background=255|0|0]
```

At the moment we only allow spaces, which means it has to be quoted. Though maybe a CSS-style hex string would be even better heh.

kleisauke:
I've updated the PR to improve the tests and documentation. Commit cfd891cb09ce1ec35c5764d76f2c9e68e069d9aa could possibly be split to its own PR.

It needs another round of review from someone other than me, since I've changed it too much now, and can't review my own work.

jcupitt (commented):
It looks very nice!

I wonder about testing `preserve` in savers though -- if we are stripping metadata centrally, do we also need to test `preserve` in every save operation?

jcupitt:
I had a two more random thoughts:

- The pipe character `|` is a word break in many shells, so you need to start quoting everything, which is irritating. Maybe colon would be better?
- Other operations would benefit from an array-of-strings type, such as `text`. You could have an array of strings for the set of fonts to load. Perhaps we should bite the bullet and add this feature (though not in this PR).

This:

```
vips copy test/test-suite/images/sample.jpg x.png[filter=avg:paeth]
```

looks quite nice.

doc/How-it-opens-files.xml
```diff
@@ -11,11 +11,20 @@
   <refnamediv> <refname>Opening</refname> <refpurpose>How libvips opens files</refpurpose> </refnamediv>
 </para>
 <para>
-  libvips now has at least four different ways of opening image files, each best for different file types, file sizes and image use cases. libvips tries hard to pick the best strategy in each case and mostly you don’t need to know what it is doing behind the scenes, except unfortunately when you do.
```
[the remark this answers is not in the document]
kleisauke: These files are being used when pandoc is not available (if building with `-Dgtk_doc=true`):
https://github.com/libvips/libvips/blob/8ae3e5eb9a8075b4951cb2d928b5a455bfe96fc8/doc/meson.build#L95-L105

We could make pandoc a required dependency, but I'm aware that some distros doesn't distribute pandoc as it depends on [many Haskell dependencies](https://src.fedoraproject.org/rpms/pandoc/blob/fb1cbb02adaa7a2e05fa7159fac6c0e2bde20409/f/pandoc.spec#_60-288).

It's a bit unfortunate that we still need this Markdown to DocBook conversion. It would be much easier if gtk-doc supported Markdown out-of-the-box.

libvips/foreign/heifsave.c
```diff
@@ -293,7 +334,7 @@ vips_foreign_save_heif_write_page(VipsForeignSaveHeif *heif, int page)
 		}
 	}
 
-	if (!save->strip &&
+	if (save->preserve &&
 		vips_foreign_save_heif_write_metadata(heif))
```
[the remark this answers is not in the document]
kleisauke: You're right, I think this check was needed to prevent `vips__exif_update()` from being called when `preserve=none`. In that case, the entire EXIF ​​block does not need to be rebuilt.

Fixed with commit 41bcdefb94dcb2080c065ee8ee088905a90aad16.

doc/How-it-opens-files.xml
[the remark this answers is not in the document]
jcupitt: Ooof, I'd forgotten, sigh.

libvips/foreign/foreign.c
```diff
@@ -1669,34 +1705,80 @@ vips__foreign_convert_saveable(VipsImage *in, VipsImage **ready,
...
+	{ VIPS_FOREIGN_PRESERVE_ICC, VIPS_META_ICC_NAME }
+};
+
+static void *
+vips_foreign_save_remove_other(VipsImage *image,
+	const char *field, GValue *value, void *user_data)
+{
+	if (vips_isprefix("png-comment-", field) ||
+		strcmp(field, VIPS_META_PHOTOSHOP_NAME) == 0 ||
+		strcmp(field, VIPS_META_IMAGEDESCRIPTION) == 0)
```
jcupitt, line 1733: This looks hard to maintain :( 

How about doing this in a single pass over the metadata? 

At the moment, you have one loop over the array of preserve names, then a second loop over all the metadata items for "other" handling.

Instead, only have the `vips_image_map()` loop and in the handler test each field name against preserve names. You could remove these special cases.

libvips/include/vips/foreign.h
```diff
@@ -330,13 +330,48 @@ typedef enum {
...
+ */
+typedef enum /*< flags >*/ {
+	VIPS_FOREIGN_PRESERVE_NONE = 0,
+	VIPS_FOREIGN_PRESERVE_EXIF = 1 << 0,
+	VIPS_FOREIGN_PRESERVE_XMP = 1 << 1,
+	VIPS_FOREIGN_PRESERVE_IPTC = 1 << 2,
+	VIPS_FOREIGN_PRESERVE_ICC = 1 << 3,
+	VIPS_FOREIGN_PRESERVE_OTHER = 1 << 4,
+
+	VIPS_FOREIGN_PRESERVE_ALL = (VIPS_FOREIGN_PRESERVE_EXIF |
```
jcupitt, line 353: Could we just use `-1`?

jcupitt (commented):
Just two small thoughts.

jcupitt:
I had a go at improving CLI flags handling: https://github.com/libvips/libvips/commit/7513fa3bf5ffaf388da57fc42becf94f7c4e6b64

So this now works:

```

$ vips pngsave k2.jpg x.png --filter up:avg
$ vips pngsave k2.jpg x.png --preserve icc:exif
```

And help text looks nicer too:

```
$ vips pngsave
save image to png file
usage:
   pngsave in filename [--option-name option-value ...]
where:
optional arguments:
...
   filter       - libpng row filter flag(s), input VipsForeignPngFilter
			default flags: none
			allowed flags: none, sub, up, avg, paeth, all
...
   preserve     - Which metadata should be preserved, input VipsForeignPreserve
			default flags: exif:xmp:iptc:icc:other:all
			allowed flags: none, exif, xmp, iptc, icc, other, all
...

Any thoughts?

File: libvips/include/vips/foreign.h (hunk 8 above)
Comment (kleisauke) [reply libvips/include/vips/foreign.h:353]: I think flags values should always be a signed integer, as `vips_flags_from_nick` uses `-1` to indicate an error.
https://github.com/libvips/libvips/blob/7513fa3bf5ffaf388da57fc42becf94f7c4e6b64/libvips/iofuncs/object.c#L2120-L2121
https://github.com/libvips/pyvips/blob/a2b08d4c5aa203399d03b379ca05145f32ce5b94/pyvips/gvalue.py#L136-L138

File: libvips/foreign/foreign.c (hunk 7 above)
Comment (kleisauke) [reply libvips/foreign/foreign.c]: You're right, fixed with commit e83194b26a86365f0bfa4cbf38baad89d953fcad.

**Comment by kleisauke** [comment]
Great! I merged the improved CLI flags handling commits.

> I think this would help the matching pyvips commit as well, since that calls `vips_flags_from_nick()` for str -> flags conversion.

Indeed, I think it makes this possible:
```python
im.pngsave('x.png', filter='up|avg')
im.pngsave('x.png', preserve='icc|exif')
```

In NetVips, flags are not handled as string values, allowing you to simply write:
```csharp
im.Pngsave("x.png", filter: ForeignPngFilter.Up | ForeignPngFilter.Avg);
im.Pngsave("x.png", preserve: ForeignPreserve.Icc | ForeignPreserve.Exif);
```

See for example:
https://github.com/kleisauke/net-vips/blob/vips-8.15/src/NetVips/Enums.cs#L486-L537

libvips/include/vips/foreign.h
kleisauke: `g_value_set_flags` also expects flags to be unsigned.
https://github.com/GNOME/glib/blob/2.78.0/gobject/genums.c#L741

jcupitt:
pyvips allows numbers for flags, so that should work there as well.

Now I look at `ForeignPngFilter` I see it has:

```python
    NONE = 'none'
    SUB = 'sub'
    UP = 'up'
    AVG = 'avg'
    PAETH = 'paeth'
    ALL = 'all'
```

Those should probably be ints, I'll change them.

jcupitt (approved):

jcupitt:
LGTM!

jcupitt:
... though I suppose it needs a changelog note.

kleisauke:
I added a changelog note and two minor fixes. PR description was updated to reflect the current state.

I think this PR is ready to land using this a final commit message:
```
Add "preserve" flag to savers (#3585)

This commit introduces a "preserve" flag option for all savers,
allowing more control over which metadata should be preserved,
unlike the now-deprecated "strip" option, which used to remove all
metadata.

Co-authored-by: John Cupitt <jcupitt@gmail.com>
Co-authored-by: Kleis Auke Wolthuizen <github@kleisauke.nl>
```

WDYT?

kleisauke:
... thinking about this further, perhaps "preserve" is too long? Renaming the option to "keep" might be clearer (naming things is hard).

jcupitt:
Oh, I like `keep`, good idea! Yes, "preserve" is annoying to type.

libvips/foreign/foreign.c
```diff
@@ -1669,34 +1705,73 @@ vips__foreign_convert_saveable(VipsImage *in, VipsImage **ready,
...
+
+int
+vips__foreign_update_metadata(VipsImage *in,
+	VipsForeignKeep keep)
+{
+	/* Rebuild exif from tags, if we'll be saving it.
+	 */
+	if ((keep & VIPS_FOREIGN_KEEP_EXIF) &&
+		vips__exif_update(in))
+		return -1;
```
kleisauke, line 1751: It looks like this wasn't called by every saver previously. See for example:

Git master:
```console
$ curl -LO https://github.com/kleisauke/net-vips/raw/master/tests/NetVips.Tests/images/йцук.jpg
$ vips copy йцук.jpg x.v
$ diff -u <(vipsheader йцук.jpg -a) <(vipsheader x.v -a)
```
<details>
  <summary>Diff</summary>

```diff
@@ -1,4 +1,4 @@
-йцук.jpg: 1024x768 uchar, 3 bands, srgb, jpegload
+x.v: 1024x768 uchar, 3 bands, srgb, jpegload
 width: 1024
 height: 768
 bands: 3
@@ -9,7 +9,7 @@
 yoffset: 0
 xres: 7.08661
 yres: 7.08661
-filename: йцук.jpg
+filename: x.v
 vips-loader: jpegload
 jpeg-multiscan: 0
 jpeg-chroma-subsample: 4:2:0
@@ -64,3 +64,4 @@
 jpeg-thumbnail-data: 3160 bytes of binary data
 orientation: 1
 icc-profile-data: 1352 bytes of binary data
+vips-sequential: 1
```
</details>



This PR:
```console
$ curl -LO https://github.com/kleisauke/net-vips/raw/master/tests/NetVips.Tests/images/йцук.jpg
$ vips copy йцук.jpg x.v
$ diff -u <(vipsheader йцук.jpg -a) <(vipsheader x.v -a)
```

<details>
  <summary>Diff</summary>

```diff
@@ -1,4 +1,4 @@
-йцук.jpg: 1024x768 uchar, 3 bands, srgb, jpegload
+x.v: 1024x768 uchar, 3 bands, srgb, jpegload
 width: 1024
 height: 768
 bands: 3
@@ -9,11 +9,13 @@
 yoffset: 0
 xres: 7.08661
 yres: 7.08661
-filename: йцук.jpg
+filename: x.v
 vips-loader: jpegload
 jpeg-multiscan: 0
 jpeg-chroma-subsample: 4:2:0
-exif-data: 6626 bytes of binary data
+icc-profile-data: 1352 bytes of binary data
+vips-sequential: 1
+exif-data: 4302 bytes of binary data
 resolution-unit: in
 exif-ifd0-ImageDescription:                                 (                               , ASCII, 32 components, 32 bytes)
 exif-ifd0-Make: Canon (Canon, ASCII, 6 components, 6 bytes)
@@ -45,12 +47,12 @@
 exif-ifd2-MeteringMode: 5 (Pattern, Short, 1 components, 2 bytes)
 exif-ifd2-Flash: 16 (Flash did not fire, compulsory flash mode, Short, 1 components, 2 bytes)
 exif-ifd2-FocalLength: 22500/1000 (22.5 mm, Rational, 1 components, 8 bytes)
-exif-ifd2-MakerNote: 2290 bytes undefined data (2290 bytes undefined data, Undefined, 2290 components, 2290 bytes)
+exif-ifd2-MakerNote: 6 bytes undefined data (6 bytes undefined data, Undefined, 6 components, 6 bytes)
 exif-ifd2-UserComment:  (, Undefined, 264 components, 264 bytes)
 exif-ifd2-FlashpixVersion: FlashPix Version 1.0 (FlashPix Version 1.0, Undefined, 4 components, 4 bytes)
 exif-ifd2-ColorSpace: 1 (sRGB, Short, 1 components, 2 bytes)
-exif-ifd2-PixelXDimension: 3648 (3648, Short, 1 components, 2 bytes)
-exif-ifd2-PixelYDimension: 2736 (2736, Short, 1 components, 2 bytes)
+exif-ifd2-PixelXDimension: 1024 (1024, Short, 1 components, 2 bytes)
+exif-ifd2-PixelYDimension: 768 (768, Short, 1 components, 2 bytes)
 exif-ifd2-FocalPlaneXResolution: 3648000/292 (12493.151, Rational, 1 components, 8 bytes)
 exif-ifd2-FocalPlaneYResolution: 2736000/219 (12493.151, Rational, 1 components, 8 bytes)
 exif-ifd2-FocalPlaneResolutionUnit: 2 (Inch, Short, 1 components, 2 bytes)
@@ -63,4 +65,3 @@
 exif-ifd2-SceneCaptureType: 0 (Standard, Short, 1 components, 2 bytes)
 jpeg-thumbnail-data: 3160 bytes of binary data
 orientation: 1
-icc-profile-data: 1352 bytes of binary data
```

</details>

So, when saving the image in vips format, this PR corrects the `exif-ifd2-PixelXDimension` / `exif-ifd2-PixelYDimension` tags and changes the `exif-ifd2-MakerNote` tag (probably because we don't use the `EXIF_DATA_OPTION_DONT_CHANGE_MAKER_NOTE` option).

I'm not sure if this new behaviour is desired. It caused a test failure in NetVips (which I was able to fix by changing the test fixture, see e.g. commit https://github.com/kleisauke/net-vips/commit/1c4acefc8288ec318bd58ce2a144fa731785ce9f).

jcupitt: Huh that's true. I think this only happens for save via the `foreign/` interface, if you use the vips writer in `iofuncs/` directly, this probably won't happen, which is even more confusing.

The alternative is to move the exif update out of the base saver and only call it from format savers that have direct EXIF support (what we had before). On the other hand, I think this only happens with vips format, which is maybe not very important.

What do you think?

kleisauke: It probably only happens with the vips format, so indeed, let's leave it as is.

Could invoking `vips__exif_update()` potentially impact performance? For instance, formats that don't support Exif at all (e.g. `gifsave`) would not need this.

jcupitt: Yes, there's a performance hit. My instinct says it's negligible, but I've not done any experiments.

kleisauke: I also think it's negligible. If anyone complains, we could set the "keep" flag automatically based on which metadata is supported by the saver.

kleisauke:
\o/, this will be in libvips 8.15. Thanks for working on this @a3mar!

a3mar:
It's great, thank you @kleisauke @jcupitt , all work done by you) Actually, it was my first time with C lang, i'm not professional programmer

PR #3456, opened 2023-04-21
pdfiumload: add support for forms

kleisauke:
Resolves: #3455.

Used https://pdfium.googlesource.com/pdfium/+/main/samples/simple_no_v8.c as reference, without implementing the additional-action calls:
```
FORM_OnAfterLoadPage
FORM_DoPageAAction
FORM_OnBeforeClosePage
FORM_DoDocumentAAction
```
(which is apparently optional?)

PR #3594, opened 2023-08-04
pdfiumload: fix rendering multiple pages with different sizes

DarthSim:
Hey there 👋

Long story short 🙂
Here's a PDF file containing multiple pages of different sizes: https://img.darthsim.me/R2rprQ86Q6.pdf.
Here's how vips renders it: https://img.darthsim.me/mJ8SkNJb_m.png

As you can see, the pages after the first one are rendered incorrectly. That's because vips sets a tile size equal to the first page size and renders a full page in each tile. Also, there are leftovers from the previous pages because vips doesn't clear the region before rendering pages in it.

Luckily, despite what comments in `pdfiumload.c` say, PDFium allows rendering a part of the page.

This PR fixes the `pdfiumload` behavior to work in sequential mode and render part of a page in each tile. Also, I added resetting of the output region to remove previous pages' leftovers.

Here's how vips renders the test PDF now: https://img.darthsim.me/9p4CSGSVC_.png

And here're benchmarks:

```bash
# Before the fix
/usr/bin/time -f "%e %M" vips copy /images/pdf8.pdf[n=-1] /images/pdf8.png
0.35 119368

# After the fix
/usr/bin/time -f "%e %M" vips copy /images/pdf8.pdf[n=-1] /images/pdf8.png
0.21 44324
```

After the fix, `pdfiumload` becomes faster and eats less memory.

PS. https://github.com/libvips/libvips/pull/3456 have broken the rendering of PDFs with multiple pages of different sizes completely. That's because `rect.width` may be zero here – https://github.com/libvips/libvips/blob/master/libvips/foreign/pdfiumload.c#L613-L615, and `FPDF_FFLDraw` segfaults because of that. Changing its arguments to ones of `FPDF_RenderPageBitmap` from this PR fixes the issue.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,7 @@ TBD 8.14.4
 
 - fix null-pointer dereference during svgload [kleisauke]
 - heif{load,save}: guard against NULL strings [kleisauke]
+- pdfiumload: fix rendering of pages with different sizes [DarthSim]
 
 20/7/23 8.14.3
 
--- a/libvips/foreign/pdfiumload.c
+++ b/libvips/foreign/pdfiumload.c
@@ -108,6 +108,8 @@ EOF
 #include <fpdf_doc.h>
 #include <fpdf_edit.h>
 
+#define TILE_SIZE (4000)
+
 typedef struct _VipsForeignLoadPdf {
 	VipsForeignLoad parent_object;
 
@@ -314,8 +316,7 @@ vips_foreign_load_pdf_build( VipsObject *object )
 static VipsForeignFlags
 vips_foreign_load_pdf_get_flags_filename( const char *filename )
 {
-	/* We can't render any part of the page on demand, but we can render
-	 * separate pages. Might as well call ourselves partial.
+	/* We can render any part of the page on demand.
 	 */
 	return( VIPS_FOREIGN_PARTIAL );
 }
@@ -384,9 +385,9 @@ vips_foreign_load_pdf_set_image( VipsForeignLoadPdf *pdf, VipsImage *out )
 	printf( "vips_foreign_load_pdf_set_image: %p\n", pdf );
 #endif /*DEBUG*/
 
-	/* We render to a linecache, so fat strips work well.
-	 */
-        if( vips_image_pipelinev( out, VIPS_DEMAND_STYLE_FATSTRIP, NULL ) )
+        /* We render to a tilecache, so it has to be SMALLTILE.
+         */
+        if( vips_image_pipelinev( out, VIPS_DEMAND_STYLE_SMALLTILE, NULL ) )
 		return( -1 );
 
 	/* Extract and attach metadata. Set the old name too for compat.
@@ -562,7 +563,12 @@ vips_foreign_load_pdf_generate( VipsRegion *or,
 		if( VIPS_RECT_BOTTOM( &pdf->pages[i] ) > r->top )
 			break;
 
-	top = r->top; 
+	/* Reset out region. Otherwise there might be parts of previous pages
+	 * left.
+	 */
+	vips_region_black( or );
+
+	top = r->top;
 	while( top < VIPS_RECT_BOTTOM( r ) ) {
 		VipsRect rect;
 		FPDF_BITMAP bitmap;
@@ -589,11 +595,13 @@ vips_foreign_load_pdf_generate( VipsRegion *or,
 				0, 0, rect.width, rect.height, ink );
 		}
 
-		FPDF_RenderPageBitmap( bitmap, pdf->page, 
-			0, 0, rect.width, rect.height,
-			0, 0 ); 
+		FPDF_RenderPageBitmap( bitmap, pdf->page,
+			pdf->pages[i].left - rect.left,
+			pdf->pages[i].top - rect.top,
+			pdf->pages[i].width, pdf->pages[i].height,
+			0, 0 );
 
-		FPDFBitmap_Destroy( bitmap ); 
+		FPDFBitmap_Destroy( bitmap );
 
 		g_mutex_unlock( vips_pdfium_mutex );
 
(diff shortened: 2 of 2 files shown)


jcupitt:
Hey @DarthSim, this is very cool! Yes, the pdfium loader is rather unloved :( 

I'll have a look.

kleisauke:
Which PDFium version was this tested with? I could not reproduce this issue on the master branch with the latest pre-built binaries from https://github.com/bblanchon/pdfium-binaries.

<details>
  <summary>Details</summary>

```bash
export PDFIUM_VERSION=5921
mkdir -p $HOME/pdfium-linux-x64/lib/pkgconfig
curl -Ls https://github.com/bblanchon/pdfium-binaries/releases/download/chromium%2F$PDFIUM_VERSION/pdfium-linux-x64.tgz | tar xzC $HOME/pdfium-linux-x64
cat > $HOME/pdfium-linux-x64/lib/pkgconfig/pdfium.pc << EOF
prefix=$HOME/pdfium-linux-x64
exec_prefix=\${prefix}
libdir=\${exec_prefix}/lib
includedir=\${prefix}/include

Name: pdfium
Description: PDFium
Version: $PDFIUM_VERSION
Libs: -L\${libdir} -lpdfium
Cflags: -I\${includedir}
EOF
export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:$HOME/pdfium-linux-x64/lib/pkgconfig
export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$HOME/pdfium-linux-x64/lib
pkg-config --exists --print-errors "pdfium >= 4200" && echo OK

CC="gcc" CXX="g++" CFLAGS="-O3" CXXFLAGS="-O3" meson setup build --prefix=/usr --buildtype=release -Ddeprecated=false -Dexamples=false -Dintrospection=disabled
meson compile -Cbuild && sudo meson install -Cbuild
vips copy R2rprQ86Q6.pdf[n=-1] x.png
```
x
</details>

kleisauke:
Ah, sorry for the noise I was able to reproduce this (I had a dangling `vips-poppler.so`).

kleisauke:
I just opened PR #3602 given that this PR targets the 8.14 branch.

> PS. #3456 have broken the rendering of PDFs with multiple pages of different sizes completely. That's because `rect.width` may be zero here – https://github.com/libvips/libvips/blob/master/libvips/foreign/pdfiumload.c#L613-L615, and `FPDF_FFLDraw` segfaults because of that. Changing its arguments to ones of `FPDF_RenderPageBitmap` from this PR fixes the issue.

Confirmed. I wonder why this segv hasn't been caught by OSS-Fuzz. :thinking:

kleisauke:
It appears that PDFium still cannot render any part of the page on demand. I could not reproduce this on this PR, but removing the `vips_sequential()` and switching to `VIPS_DEMAND_STYLE_SMALLTILE` would cause non-deterministic output. See for example commit https://github.com/kleisauke/libvips/commit/aca7051698e761e24cac06cdf7c213070a48ab9e and this test script:
```bash
vips copy R2rprQ86Q6.pdf[n=-1] x.png

expected_sha256=$(sha256sum "x.png" | awk '{ print $1 }')
for run in {1..20}; do
  vips copy R2rprQ86Q6.pdf[n=-1] x.png
  echo "$expected_sha256 x.png" | sha256sum --check --quiet
done
```
```console
$ ./check-deterministic.sh
x.png: FAILED
sha256sum: WARNING: 1 computed checksum did NOT match
...
```

Given this, I'm a bit worried to land this as-is on 8.14. How about splitting the changes in `vips_foreign_load_pdf_load()` to a separate PR?

DarthSim:
This is not a PDFium issue but the issue with the way vips works with it. More specifically, multithreading is the problem here.

`VipsForeignLoadPdf` stores only a single page at a time. Let's assume that the first thread starts generating a tile from page 1 and calls `vips_foreign_load_pdf_get_page`. Then, the second thread starts generating a tile from page 2 and calls `vips_foreign_load_pdf_get_page` before the first thread started rendering its tile. In this case, both threads will render page 2.

The current implementation may also face the same issue, but since it renders huge tiles, it's quite unlikely but still possible.

I'll think what can be done here.

jcupitt:
>  Let's assume that the first thread starts generating a tile from page 1 and calls vips_foreign_load_pdf_get_page.

I think this should be OK -- `pdfiumload` runs behind a `tilecache`, and this will only let one thread at once through to `vips_foreign_load_pdf_generate()`. As long as that function sets the page each call, it should work.

DarthSim:
I guess, we could just use `vips_sequential` as we do in `nsgifload` and friends,

DarthSim:
I can confirm that using using `vips_sequential` works without issues even with 1px `tile_height`.

I changed `tile_height` to `16`. It slows `pdfiumload` down a little bit but the memory footprint is lower:

```
/usr/bin/time -f "%e %M" vips copy /images/pdf8.pdf[n=-1] /images/pdf8.png
0.23 39348
```

jcupitt:
What if you swap the `vips_sequential` for `vips_tilecache`? I'd think it should work.

`popperload` uses:

```C
/* Render PDFs with tiles this size. They need to be pretty big to limit
 * overcomputation.
 *
 * An A4 page at 300dpi is 3508 pixels, so this should be enough to prevent
 * most rerendering.
 */ 
#define TILE_SIZE (4000)

...
        vips_tilecache(t[0], &t[1],
            "tile_width", TILE_SIZE,
            "tile_height", TILE_SIZE,
            "max_tiles", 2 * (1 + t[0]->Xsize / TILE_SIZE),
            NULL) ||
```

ie. huge tiles, since poppler is *really* slow at rendering parts of a page.

jcupitt:
poppler is slow at rendering subsections of pages since it does no caching, unfortunately. 

If you render a 128x128 area on a page, it will redraw the entire page and just clip that part, and redrawing includes re-decompressing any image resources, like JPGs. You can end up repeating a lot of work for each tile, so rendering whole pages can be dramatically quicker, perhaps a factor of 100 in a bad case.

I kept this PDF around for benchmarking:

http://www.rollthepotato.net/~john/Audi_US%20R8_2017-2.pdf

It has lots of huge graphic elements, and a range of page sizes. It's *very* slow with poppler, unless you render a page at a time. It'd be interesting to test with this PR.

DarthSim:
Wow, this is hella of a PDF 🙂 I've rendered only the first 5 pages of it, but I believe that's enough:

```bash
# sequential(tile_height = 16) + VIPS_DEMAND_STYLE_FATSTRIP
/usr/bin/time -f "%e %M" vips copy /images/pdf9.pdf[n=5] /images/pdf8.png
1.96 124896

# sequential(tile_height = 128) + VIPS_DEMAND_STYLE_FATSTRIP
/usr/bin/time -f "%e %M" vips copy /images/pdf9.pdf[n=5] /images/pdf8.png
1.21 122668

# sequential(tile_height = 16) + VIPS_DEMAND_STYLE_SMALLTILE
/usr/bin/time -f "%e %M" vips copy /images/pdf9.pdf[n=5] /images/pdf8.png
2.00 143968

# sequential(tile_height = 128) + VIPS_DEMAND_STYLE_SMALLTILE
/usr/bin/time -f "%e %M" vips copy /images/pdf9.pdf[n=5] /images/pdf8.png
1.23 145340

# tilecache(tile_width = 4000, tile_height = 4000) + VIPS_DEMAND_STYLE_FATSTRIP
/usr/bin/time -f "%e %M" vips copy /images/pdf9.pdf[n=5] /images/pdf8.png
1.65 150564

# tilecache(tile_width = 4000, tile_height = 4000) + VIPS_DEMAND_STYLE_SMALLTILE
/usr/bin/time -f "%e %M" vips copy /images/pdf9.pdf[n=5] /images/pdf8.png
1.70 150528
```

Hmm, `vips_sequential` with `tile_height` of 128 and `VIPS_DEMAND_STYLE_FATSTRIP` seems to be the best option after all 🤔

jcupitt (approved):
Great!

kleisauke:
Thank you!

> Changing its arguments to ones of `FPDF_RenderPageBitmap` from this PR fixes the issue.

A PR for this would be welcome.

DarthSim:
@kleisauke Here it is: https://github.com/libvips/libvips/pull/3613

kleisauke:
> I wonder why this segv hasn't been caught by OSS-Fuzz.

FWIW, it looks like this was only reproducible with `n` other than `1` (default). Unfortunately, those optional input arguments aren't covered by the fuzzers.
https://github.com/libvips/libvips/blob/d59e8b93c1406bc43b8dde6aefe66ab979449bec/libvips/foreign/pdfiumload.c#L721-L726

DarthSim:
Exactly. This could happen only when the page's rect does not intersect with the region's rect. And this could happen only when you render multiple pages some of which are wider than the first one.

PR #3595, opened 2023-08-07
SIMD optimizations & embedless reduce

DarthSim:
Hey there 👋

Sorry for squashing two things in a single PR 😓 I developed one on top of another and it's hard to split them now. And it'd be pretty meaningless in case you're not interested in any of them 😅

## SIMD optimizations for `reduce(v/h)` and `convi`

_DISCLAIMER: I know that @kleisauke  works on bringing Highway to libvips. But since this approach is blocked by Highway's availability, I decided to try the raw SIMD way._

I optimized `reduce(v/h)` and `convi` with SSSE3 and NEON. Take note that I optimized `reduceh` for 3 and 4-band images.

### Benchmarks

I've written the following benchmark program: https://gist.github.com/DarthSim/c29e1739ad894e1b49a408963ac41325. It runs each tested operation 5000 times and measures execution time in seconds. I didn't use CLI for benchmarking to mitigate IO penalty. I ran the benchmark with [this image](https://img.darthsim.me/OFLCeY7zxR) on my MacBook Pro M1 Max (so NEON optimization was used)

```bash
# no optimizations
./vips-bench.o /images/wilber.png
reducev    62.3669
reduceh    12.5061
gaussblur  50.2628

# orc
./vips-bench.o /images/wilber.png
reducev    43.5744
reduceh    12.4838
gaussblur  64.7899

# neon
./vips-bench.o /images/wilber.png
reducev    12.7119
reduceh    6.54691
gaussblur  12.6861
```

As a result: **trice** faster `redicev`, **almost twice** faster `reduceh`, and **5 times** faster `gaussblur` (i.e. `convi`) compared to the `orc` path. Take note that `orc` `gaussblur` performance is the lowest one since the `orc` path uses the whole matrix instead of ignoring zero coeffs.

## Embedless reduce

The current approach requires using `vips_embed` to add new pixels around the image since the kernel goes out of the image borders on edge pixels. But this approach has two major disadvantages:

1. We need to copy lots of pixels during `embed`. Basically, we have to copy the whole image plus the borders.
2. When processing edge pixels, we have to do unneeded calculations processing added pixels.

The new approach involves calculating kernels for each row (`reducev`) or column (`reduceh`). Kernels for edge pixels will be smaller to not go out of image borders. So, no `embed` is needed. As a bonus, no precision penalty for all the pixel formats.

Unfortunately, I didn't manage to implement this approach for the `orc` path 😞 So it uses the current approach.

### Benchmarks

```bash
# no optimizations + embed
./vips-bench.o /images/wilber.png
reducev    62.3669
reduceh    12.5061

# no optimizations + embedless
./vips-bench.o /images/wilber.png
reducev    61.7982
reduceh    12.0303

# neon + embed
./vips-bench.o /images/wilber.png
reducev    12.7119
reduceh    6.54691

# neon + embedless
./vips-bench.o /images/wilber.png
reducev    12.4412
reduceh    6.10505
```

The results are not as impressive as the results of SIMD optimization, yet still. And remember that we calculate kernels for every row/column, so we're getting maximum quality for all the pixel formats.

Legit question: what about memory? Well, with the new approach, it's a bit lower 🙂

```bash
# with embed
/usr/bin/time -f "%M" vips reducev v1.v v2.v 1.9
49704
/usr/bin/time -f "%M" vips reduceh v1.v v2.v 1.9
28000

# without embed
/usr/bin/time -f "%M" vips reducev v1.v v2.v 1.9
49344
/usr/bin/time -f "%M" vips reduceh v1.v v2.v 1.9
25052
```


kleisauke (commented):
Thanks for working on this! The embedless feature looks interesting and would indeed ensure high-quality images for all pixel formats (and not just for `VIPS_FORMAT_DOUBLE`).

It looks like the SIMD optimizations would only work on CPUs that support AArch64+Neon or SSSE3. Without a runtime dispatch mechanism, I think this should still be opt-in by default.

jcupitt:
> We need to copy lots of pixels during embed. Basically, we have to copy the whole image plus the borders.

This isn't right -- embed only copies tiles which include edge pixels, all the centre part of the image is processed with just a pointer copy. You can see the logic here:

https://github.com/libvips/libvips/blob/8d5f33174c956623c1b7eb82c43c828c0511cc22/libvips/conversion/embed.c#L240-L254

So if the region doesn't touch the border area, the request is forwarded directly to the upstream operation. All of libvips is like this -- operations like `ifthenelse`, `insert`, etc. will pointer copy if they can. I don't think `embed` is adding any significant performance penalty, and it's much simpler and more flexible than building this into the convolver.

DarthSim:
> So if the region doesn't touch the border area, the request is forwarded directly to the upstream operation

But it does touch the borders of every line in `rediceh`, isn't it? So in `rediceh` we do copy the entire image plus the borders.

> I don't think embed is adding any significant performance penalty

`embed` itself – maybe not. But `embed` and the need to include the borders in the calculation – do. As you see, I added the calculation of all the matrices for every row/col, and this is still a little bit faster than doing `embed`.

> and it's much simpler and more flexible than building this into the convolver

With all my regards, I don't see how is it simpler or more flexible. Just having matrices for every row/col is the simplest and the most flexible thing that could be done here. The only little complication is the matrices may have different sizes, but it's not a problem either as we have input regions' boundaries. This doesn't fit into the Orc approach only because Orc is very inflexible itself.

And again, this approach gives the best quality for all the formats, not only for `double`. And with `double`, it works even faster than the current approach as we have pre-computed matrices.

DarthSim:
And... maybe it's time to let Orc go?

The only reason for Orc is optimization for different CPUs. But I believe that the vast majority of CPUs used nowadays are x64 and ARM64. Having vips running on something like PowerPC seems to be an extremely rare case. We don't optimize operations for the `float` format, yet for some reason, we optimize vips for PowerPC. You can see that `convi` "optimized" with Orc works even slower on ARM64 than its not "optimized" version. Who knows how it behaves on PPC or MIPS?

Also, Orc is very outdated: it does not support AVX2, and its NEON support is very poor. The homebrew NEON implementation of `convi` is 5 times (!!!) faster than the Orc implementation. 5 times! Hence, we can write faster code for the most popular CPUs without Orc.

And finally, Orc is very inflexible. You know better than me that it has lots of limitations and its programs should know everything compile time as they don't have loops and stuff. Its programs have to be compiled for every operation in real-time, operations usually need to allocate additional intermediate buffers for Orc, etc. And, of course, some operations can't be optimized with Orc but can be successfully optimized with raw SIMD instructions.

jcupitt:
> But it does touch the borders of every line in rediceh, isn't it?

You're right, if it's processing entire scanlines, it would have to copy. But more complex chains of calculation will almost always be running in tiled mode, so only the edges would need the copy.

> With all my regards, I don't see how is it simpler or more flexible. 

Embed has all the abyss handling for all operations, so you can pick from mirror, extend, black, white, etc., all in one place, and you can use the same abyss code for things like FFT as well.

Having said that, I've not actually read your code for this :) I should have a look.

jcupitt:
> And... maybe it's time to let Orc go?

I agree, it's an annoying, complicated and unreliable thing, and not very well maintained. We picked it 20+ years ago (I think?) and the world has moved on a very long way!

DarthSim:
> Embed has all the abyss handling for all operations, so you can pick from mirror, extend, black, white, etc., all in one place, and you can use the same abyss code for things like FFT as well.

I probably misunderstood your point 🙂 I'm not jumping on `embed`, I love it and use it here and there in imgproxy. I just believe that using it in `reduce(v/h)` is redundant, and we can gain better performance and quality without it.

kleisauke:
> So if you could show me a way to detect the CPU arch and supported intrinsics sets, I'd definitely add dispatching.

On x86 the [CPUID](https://wiki.osdev.org/CPUID) instruction can be used for this, the [__builtin_cpu_supports](https://gcc.gnu.org/onlinedocs/gcc/x86-Built-in-Functions.html#index-_005f_005fbuiltin_005fcpu_005fsupports-1) builtin function simplifies this a bit. See for example: https://github.com/kleisauke/libvips/blob/simd/libvips/iofuncs/simd.c#L89-L101

I'm not aware of anything for ARMv7. AFAIK, there is no `__builtin_cpu_supports("neon")`. You could use `#include <sys/auxv.h>` and `#include <asm/hwcap.h>` with `getauxval(AT_HWCAP) & HWCAP_NEON`, although that is specific to Linux.

> DISCLAIMER: I know that @kleisauke works on bringing Highway to libvips. But since this approach is blocked by Highway's availability, I decided to try the raw SIMD way.

I'll be opening a PR for https://github.com/libvips/libvips/compare/master...kleisauke:simd-highway upcoming week. I've kept the Orc paths as a fallback whenever [Highway](https://github.com/google/highway) is not available.

First benchmark results on my AMD Ryzen 9 7900 workstation can be found here:
https://github.com/kleisauke/vips-microbench/blob/master/results/simd-highway.md

I tested this for accuracy/speed on these targets:
https://gist.github.com/kleisauke/1f28a9fc156c753bcb1239b6fc1a2e62

For Arm64, feel free to test this on your MacBook Pro M1 Max, those Raspberry Pi's aren't great for benchmarking. :sweat_smile:

DarthSim:
> I'll be opening a PR for https://github.com/libvips/libvips/compare/master...kleisauke:simd-highway upcoming week

Welp, seems like all this noise I made here doesn't make much sense then 😅

DarthSim:
> For Arm64, feel free to test this on your MacBook Pro M1 Max, those Raspberry Pi's aren't great for benchmarking

Results are pretty much the same 👍

```
reducev    12.9944
reduceh    6.2664
gaussblur  12.8993
```

PR #3604, opened 2023-08-13
svgload: fix null-pointer dereference

kleisauke:
`g_utf8_find_next_char()` might return NULL when called with a non-NULL second argument, indicating that the end of the string has been reached.

Targets the 8.14 branch.


jcupitt (approved):
LGTM!

Issue #3588, opened 2023-08-01
segfault using custom target and vips_heifsave_target


negator:
## Bug report

Saving a heif image to a custom target segfaults when the "write" callback returns `-1` (writer closed). 

**To Reproduce**
```
#include <stdio.h>
#include <vips/vips.h>


static gint64 target_write ( VipsTargetCustom *target_custom, gpointer buffer, gint64 length, gpointer target_ptr )
{
    return -1; // always return -1
}

int
main(int argc, char **argv)
{
	VipsImage *in;	
	VipsTargetCustom *target;
	VipsImage *out;
	int code;

	if (VIPS_INIT(argv[0]))
		vips_error_exit(NULL);

	if (argc != 2)
		vips_error_exit("usage: %s infile", argv[0]);

	if (!(in = vips_image_new_from_file(argv[1], NULL)))
		vips_error_exit(NULL);

	target = vips_target_custom_new();
	g_signal_connect( target, "write", G_CALLBACK(target_write), NULL );
	code = vips_heifsave_target(in, (VipsTarget *) target, NULL);
	return code;
}
```

**Expected behavior**
Program should not crash. Instead a proper error code should be returned, as it does when using `vips_jpegsave_target`.

**Actual behavior**
`Segmentation fault: 11`

**Screenshots**
```
bash-3.2$  gcc  `pkg-config --cflags glib-2.0` heif.c `pkg-config --libs glib-2.0` -lglib-2.0 -lvips -lgobject-2.0
bash-3.2$ ./a.out tif.tif
Segmentation fault: 11
```

**Environment**
(please complete the following information)
 - OS: MacOS Ventura 13.4.1
 - Vips: 8.14.3

**Additional context**
I discovered the issue when adding streaming support to the `govips` library: https://github.com/davidbyttow/govips/pull/370#issuecomment-1659480558

kleisauke:
I was unable to reproduce this on Fedora 38 with libvips 8.14.3, libheif 1.16.1 and x265 3.5. Can you confirm if this is functional on your end?:
```bash
# heifsave
$ vips copy tif.tif x.heic
# heifsave_target
$ vips copy tif.tif .heic | cat > x.heic
```

negator:
[...]

PR #3608, opened 2023-08-14
heif{load,save}: guard against `NULL` strings

kleisauke:
This would prevent a `NULL` string from being printed with `vsnprintf` in case of a write error.

See: #3588.

Targets the 8.14 branch.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,7 +1,7 @@
 TBD 8.14.4
 
 - fix null-pointer dereference during svgload [kleisauke]
-- heifsave: avoid printing NULL strings with vsnprintf [kleisauke]
+- heif{load,save}: guard against NULL strings [kleisauke]
 
 20/7/23 8.14.3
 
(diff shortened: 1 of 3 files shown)


jcupitt (approved):
+1!

kleisauke:
I simplified/improved this with commit b3bfa34fe8b6c67734b640f5dae67b4c47e7f8f6. PTAL.

lovell (approved):

jcupitt:
C doesn't allow that, unfortunately, unless you parse the format string yourself.

kleisauke:
Indeed, it would be a bit difficult to do this in `vips_vsnprintf`. Perhaps this fix is a bit too cautious, as libheif says that this will never be `NULL`:
https://github.com/strukturag/libheif/blob/9493f9a7f9389ad15ae45ae04b119bcca4aec049/libheif/heif.h#L322

But, "better safe than sorry". :)

negator:
Ah I see, you're right this would be very clunky to NULL check. Thanks for the fast replies and fixes!

Issue #3619, opened 2023-08-22
GraphicsMagick BMP load to JPEG-XL save out of memory


dloebl:
When loading some BMPs and saving it back as JPEG-XL, vips experiences out-of-memory issues.
Interestingly, the issue is only present when vips has been configured with `magick-package=GraphicsMagick` and only occurs with JPEG-XL outputs - saving the same BMP as PNG/JPEG/etc works just fine.
As mentioned above, `magick-package=ImageMagick` is fine as well.

**Reproducer:**
[test.bmp.zip](https://github.com/libvips/libvips/files/12410295/test.bmp.zip)
```
$ file test.bmp
test.bmp: PC bitmap, Windows 3.x format, 30 x 10 x 24, image size 920, resolution 3779 x 3779 px/m, cbSize 974, bits offset 54
```

**Steps:**
```
meson setup --prefix=/usr/local -Dmagick-package=GraphicsMagick build/
meson install -C build
```

```
$ vips copy test.bmp out.jxl

/usr/local/bin/vips copy test.bmp out.jxl
zsh: killed     /usr/local/bin/vips copy test.bmp out.jxl
```

```
$ vips copy test.bmp out.png
$ file out.png

out.png: PNG image data, 30 x 10, 8-bit grayscale, non-interlaced
```

jcupitt:
Hi @dloebl,

It looks like a combination of bugs. I see:

```
$ vips copy test.bmp gm.v
$ vipsheader gm.v
gm.v: 30x10 uchar, 1 band, srgb, magickload
$ vipsheader -a im.v
im.v: 30x10 uchar, 3 bands, srgb, magickload
```

Then:

```
$ vips copy im.v x.jxl
$ vips copy gm.v x.jxl
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
Aborted (core dumped)
```

So:

- graphicsmagick load is loading a monochrome BMP as one band, but tagging it as `srgb` (should be `b_w`)
- imagemagick load is turning a monochrome BMP into RGB, which perhaps it shouldn't
- `jxlsave` is seeing a mono image tagged as `srgb` and getting badly confused

jcupitt:
I made a PR, let's continue discussion there.

PR #3620, opened 2023-08-23
Fix a crash in jxlsave

jcupitt:
Some versions of GraphicsMagick can report rgb colourspace for some mono BMPs. In turn, this can trigger a crash in jxlsave.

See https://github.com/libvips/libvips/issues/3619

@kleisauke this is against master, though since it's a crash (under some circumstances) maybe it should be cherrypicked for 8.14?

### DIFF
--- a/libvips/foreign/magick2vips.c
+++ b/libvips/foreign/magick2vips.c
@@ -67,7 +67,7 @@
  * 12/11/21
  * 	- set "orientation"
  * 26/8/22
- *      - set "magick-format"
+ *  - set "magick-format"
  * 13/3/23 MathemanFlo
  * 	- add bits per sample metadata
  */
@@ -339,6 +339,7 @@ parse_header(Read *read)
 	printf("image->rows = %zd\n", image->rows);
 #endif /*DEBUG*/
 
+	im->Coding = VIPS_CODING_NONE;
 	im->Xsize = image->columns;
 	im->Ysize = image->rows;
 	read->frame_height = image->rows;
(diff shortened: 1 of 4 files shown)


jcupitt:
:facepalm: thanks Lovell, copy-paste error :(

lovell (approved):

kleisauke:
> @kleisauke this is against master, though since it's a crash (under some circumstances) maybe it should be cherrypicked for 8.14?

I could cherry-pick this to 8.14 and make a new release if needed. However, given that we already warn users  in our documentation about the security concerns when processing untrusted images via *magick, it's probably fine to leave this fix only on the master branch (assuming this crash can only occur with GraphicsMagick and v8.15.0 is released in a couple of weeks).

libvips/foreign/magick2vips.c
```diff
@@ -385,6 +386,25 @@ parse_header(Read *read)
...
+		break;
+
+	default:
+		im->Xres = 1.0;
+		im->Yres = 1.0;
+		break;
+	}
+
+	// this can be wrong for some GM versions and must be sanity checked (see
+	// below)
```
kleisauke, line 407: I still have to get used to the fact that we now support C++ style `//` comments. :)

kleisauke (approved):

kleisauke:
Oh, huh, somehow GitHub accidentally merged this twice (see commit 029708693002f35bcaa7ce9dd5e43b454b9f839a / 4e1bf32ab7cca9b02b3d123844c83e7e5c8a3c1c).

kleisauke:
Backport to [`8.14`](https://github.com/libvips/libvips/tree/8.14) available as #3662.

Issue #3621, opened 2023-08-23
TIFF with float sample format is not scRGB


DarthSim:
Hey there 👋

During my work on 16-bit float TIFFs support, I found out that vips loads TIFFs with float/double sample format as scRGB while they have sRGB colorspace despite having a non-integer sample format. Here's an example: [tiff7.tiff.zip](https://github.com/libvips/libvips/files/12422094/tiff7.tiff.zip)

If I do `vips colourspace tiff7.tiff tiff7.png srgb`, I'm getting the wrong colors.

Here's how it should look:
![tiff7_fixed](https://github.com/libvips/libvips/assets/380856/97a44b2e-4244-43d4-b525-358f7193e1aa)

And here's how it looks:
![tiff7_bugged](https://github.com/libvips/libvips/assets/380856/c1a28560-72a2-4add-9cf4-30093a832d26)

I was going to fit this myself, but I can't come out with the right approach:

1. Apply gamma correction during loading. Sounds legit, but we'll need to add gamma correction to the saving process too. But we should probably do so anyway.
2. Multiply pixel values to 65535 and load as RGB16. We'll need to handle this during saving and divide the pixel values back.
3. Just load them as RGB16. We'll lose precision, and this doesn't seem right in general.

What do you think?

jcupitt:
Hi @DarthSim,

I wasn't sure if float TIFFs should have a gamma or not. Gamma is only really necessary for 8 bit data, so I thought probably not. And I think (though it was a long time ago) I found a program which could read and write float tiff which didn't use a gamma (radiance?).

Did you find something in the docs which clarifies this?

jcupitt:
Ah your sample image has:

```
         <photoshop:ColorMode>3</photoshop:ColorMode>
         <photoshop:ICCProfile>sRGB IEC61966-2.1 (Linear RGB Profile)</photoshop:ICCProfile>
```

So it maybe should be linear?

I tried with vipsdisp and it displays it correctly:

![image](https://github.com/libvips/libvips/assets/580843/6bf33cbf-f6ed-4718-9513-92e4698040ee)

So I think it's maybe just poor scRGB support in the thing that converts image for save. I'll have a look.

jcupitt:
nip2 displays it correctly as well, interestingly:

![image](https://github.com/libvips/libvips/assets/580843/34fc6a18-0255-4f17-85fc-cf7942fa1bb6)

jcupitt:
Ah, this works:

```
[...]

Issue #2858, opened 2022-06-12
jpeg save fails with scRGB data


jcupitt:
Probably other similar savers too.

```
$ vips colourspace k2.jpg x.v scrgb
$ vips copy x.v x.jpg 
```

Produces a black image.

`VIPS_SAVEABLE_RGB_CMYK` lets scRGB through, cast to uint8. Other savers need scRGB, so we can't just convert it. We should overhaul the SAVEABLE system, it's not very good.

jcupitt:
Actually, I've sort of bodged it in 8.13. Let's find something more interesting to work on instead.

jcupitt:
See https://github.com/libvips/libvips/pull/3623 for a fix.

PR #3623, opened 2023-08-24
Improve scRGB support

jcupitt:
Handling of scRGB has a couple of issues:

- When converting scRGB to and from sRGB and RGB16 we are changing the image gamma, so any profiles on the image will break. They need to be removed.

- Conversion for saveable RGB did not handle scRGB

See https://github.com/libvips/libvips/issues/3621

Using the [test image from the linked issue](https://github.com/libvips/libvips/files/12422094/tiff7.tiff.zip) (a float TIFF saved by photoshop) with this PR I can now run:

```
$ vips copy tiff7.tiff x.png
```

To make:

x

Though this makes a black image:

```
$ vips icc_transform tiff7.tiff x.png srgb
```

So I suppose `icc_transform` does not handle scRGB correctly. I'll mark this as a draft until this extra issue is resolved too.


kleisauke:
Issue #2858 and its test case may also be relevant here.

jcupitt:
The test case in #2858 works with this PR:

```
$ vips colourspace k2.jpg x.v scrgb
$ vips copy x.v x.jpg 
```

jcupitt:
Ah, the `icc_import` failure is because we don't support linear RGB profiles, we only have 8 and 16-bit device space. I think that this should probably be a separate issue.

jcupitt:
> requires other changes to let jxlsave encode

Ugh, this will take a while. We should get https://github.com/libvips/libvips/pull/3712 in at the same time.

Shall we park this for 8.16? It would delay release by quite a while.

kleisauke:
I had a very quick attempt with commit https://github.com/kleisauke/libvips/commit/d7daf546071a40f45cf39e300d8a298c0cf39fe3, and it appears to work; however, jxlload is unable to decode these images due to:
https://github.com/libjxl/libjxl/blob/e37516e7a7494f49613ac2cde1b58dbc06e52e62/lib/jxl/decode.cc#L2302-L2304

To decode these extra channels, you apparently also need to call `JxlDecoderExtraChannelBufferSize()` / `JxlDecoderSetExtraChannelBuffer()`. A similar issue occurs when trying to decode the CMYK + Alpha test image from the libjxl conformance repository:
https://github.com/libjxl/conformance/blob/master/testcases/cmyk_layers/input.jxl

kleisauke:
> Shall we park this for 8.16? It would delay release by quite a while.

I think this PR would still be useful to have for 8.15. How about this commit?:
https://github.com/kleisauke/libvips/commit/59e8b6c98444f4f58a2a892d87f939ec33d29531

That would mimic `VIPS_SAVEABLE_RGBA` in jxlsave and still allow it to operate on scRGB images.

> To decode these extra channels, you apparently also need to call `JxlDecoderExtraChannelBufferSize()` / `JxlDecoderSetExtraChannelBuffer()`.

On a somewhat tangent, it seems that the new streaming encode API in libjxl uses a callback-based approach, rather than a  'push'-based tile/scanline approach (used in libraries like libtiff),
https://libjxl.readthedocs.io/en/latest/api_encoder.html#_CPPv426JxlChunkedFrameInputSource

jcupitt:
OK, let's do that, and I suppose leave the jxl load / save reworking for 8.16?

kleisauke:
Indeed, let's defer the jxl load/save rework to the 8.16 cycle.

kleisauke (approved):

PR #3626, opened 2023-08-25
tiffload: add 16-bit float support

DarthSim:
Hey there 👋

This PR adds support for 16-bit float TIFFs. I didn't use `_Float16` because it's not generally available.

Here are some test images: [tiff3_f16_f32.zip](https://github.com/libvips/libvips/files/12442891/tiff3_f16_f32.zip)

### DIFF
--- a/libvips/foreign/tiff2vips.c
+++ b/libvips/foreign/tiff2vips.c
@@ -440,9 +440,9 @@ half_2_float(gushort h)
 	case -15:
 		return sign / (float) (1 << 14) * (prec / 1024.0);
 	default:
-		return exp > 0 ?
-			sign * (float) (1 << exp) * (1.0 + prec / 1024.0) :
-			sign / (float) (1 << -exp) * (1.0 + prec / 1024.0);
+		return exp > 0
+			? sign * (float) (1 << exp) * (1.0 + prec / 1024.0)
+			: sign / (float) (1 << -exp) * (1.0 + prec / 1024.0);
 	}
 }
 
--- a/libvips/foreign/tiff2vips.c
+++ b/libvips/foreign/tiff2vips.c
@@ -425,6 +425,27 @@ typedef struct _Rtiff {
 	int y_pos;
 } Rtiff;
 
+/* Convert IEEE 754-2008 16-bit float to 32-bit float
+ */
+static inline float
+half_2_float(gushort h)
+{
+	const float sign = (h >> 15) * -2 + 1;
+	const int exp = ((h & 0x7C00) >> 10) - 15;
+	const float prec = (h & 0x03FF);
+
+	switch (exp) {
+	case 16:
+		return INFINITY * sign;
+	case -15:
+		return sign / (float) (1 << 14) * (prec / 1024.0);
+	default:
+		return exp > 0 ?
+			sign * (float) (1 << exp) * (1.0 + prec / 1024.0) :
+			sign / (float) (1 << -exp) * (1.0 + prec / 1024.0);
+	}
+}
+
 /* Test for field exists.
  */
 static int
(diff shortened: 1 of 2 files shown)


kmilos:
> I didn't use `_Float16` because it's not generally available.

You could additionally introduce a dependency on [Imath ](https://github.com/AcademySoftwareFoundation/Imath) and let it handle the platform specific/intrinsic stuff if performance is the ultimate goal? And leave this non-Imath fallback if dependency is not wanted/available...

DarthSim:
@kmilos I believe it's a little bit too much for a single rare case 🙂

kmilos:
> Does anyone know if imath has runtime dispatch? Ie. the SSE2 / AVX512 / whatever path is picked at runtime?

I don't think so.

jcupitt:
I had a quick grep and I think it's probably using a look up table by default:

https://github.com/AcademySoftwareFoundation/Imath/blob/main/src/Imath/half.h#L116

You need to build with extra compiler flags to enable the use of intrinsics. Anyway, it's likely to be quicker than bit twiddling, and we probably already have this LUT inside the openexr loader, so it probably makes sense to use it.

libvips/foreign/tiff2vips.c
```diff
@@ -1229,13 +1253,37 @@ rtiff_parse_fourbit(Rtiff *rtiff, VipsImage *out)
 		} \
 	}
 
+/* GREY_LOOP implementation for 16-bit float
+ */
+#define GREY_LOOP_F16 \
+	{ \
+		gshort *p1; \
```
kmilos, line 1260: `gushort` everywhere?

kmilos:
> You need to build with extra compiler flags to enable the use of intrinsics.

I guess -march and -mcpu determine if F16C extension is available or not.

> Anyway, it's likely to be quicker than bit twiddling, and we probably already have this LUT inside the openexr loader, so it probably makes sense to use it.

Some idea of difference is already given by the header file:

https://github.com/AcademySoftwareFoundation/Imath/blob/4d4e01c8c9fb9b9dbbefa2e135b21778c9cd331e/src/Imath/half.h#L162-L177

kleisauke:
Once PR #3618 lands, another option would be to use Highway's `PromoteTo()` function, which will map to `_mm{,256,512}_cvtph_ps()` on x86.
https://github.com/google/highway/blob/b39942c49b6e6a24f9dec69b1a02143c8ad21eef/hwy/ops/x86_128-inl.h#L7942-L7945
https://github.com/google/highway/blob/b39942c49b6e6a24f9dec69b1a02143c8ad21eef/hwy/ops/x86_256-inl.h#L6025-L6034
https://github.com/google/highway/blob/b39942c49b6e6a24f9dec69b1a02143c8ad21eef/hwy/ops/x86_512-inl.h#L4624-L4633

This should be safe to use if its availability is verified at both compile- and run time with `HWY_HAVE_FLOAT16` / `hwy::HaveFloat16()`.

I might take a look at this, although 16-bit float TIFFs aren't that common.

libvips/foreign/tiff2vips.c
DarthSim: Oops, thanks. Fixed the typo

DarthSim:
> it's likely to be quicker than bit twiddling

It's definitely quicker. But since 16-bit float TIFF seems to be an extremely rare case, I decided to not go hardcore here.

> we probably already have this LUT inside the openexr loader

It uses `ImfHalfToFloatArray`. We could optionally use it if we have `HAVE_OPENEXR` defined

jcupitt:
Ah I keep forgetting about highway, you're right.

But I agree with @DarthSim --- this is not likely to be a hotspot, so just do something simple for now. We can put in a highway path if this function ever shows up on profiles.

DarthSim:
Applied some code style fixes

kleisauke (approved):

PR #3627, opened 2023-08-25
Use alpha range of 0.0-1.0 for scRGB

DarthSim:
Hey there 👋

This PR changes the alpha range of scRGB to `0.0 - 1.0` as discussed here https://github.com/libvips/libvips/issues/3621#issuecomment-1691925346

### DIFF
--- a/libvips/colour/scRGB2sRGB.c
+++ b/libvips/colour/scRGB2sRGB.c
@@ -110,7 +110,7 @@ vips_scRGB2sRGB_line_8(VipsPel *restrict q, float *restrict p,
 		q += 3;
 
 		for (j = 0; j < extra_bands; j++)
-			q[j] = VIPS_CLIP(0, p[j], UCHAR_MAX);
+			q[j] = VIPS_CLIP(0, (int) (p[j] * 255.0), UCHAR_MAX);
 		p += extra_bands;
 		q += extra_bands;
 	}
@@ -140,7 +140,7 @@ vips_scRGB2sRGB_line_16(unsigned short *restrict q, float *restrict p,
 		q += 3;
 
 		for (j = 0; j < extra_bands; j++)
-			q[j] = VIPS_CLIP(0, (int) (p[j] * 256.0), USHRT_MAX);
+			q[j] = VIPS_CLIP(0, (int) (p[j] * 65535.0), USHRT_MAX);
 		p += extra_bands;
 		q += extra_bands;
 	}
--- a/libvips/conversion/flatten.c
+++ b/libvips/conversion/flatten.c
@@ -326,9 +326,7 @@ vips_flatten_build(VipsObject *object)
 	 * interpretation.
 	 */
 	if (!vips_object_argument_isset(object, "max_alpha"))
-		if (in->Type == VIPS_INTERPRETATION_GREY16 ||
-			in->Type == VIPS_INTERPRETATION_RGB16)
-			flatten->max_alpha = 65535;
+		flatten->max_alpha = vips_interpretation_max_alpha(in->Type);
 
 	/* Is max_alpha less than the numeric range of this image? If it is,
 	 * we can get int overflow.
--- a/libvips/iofuncs/header.c
+++ b/libvips/iofuncs/header.c
@@ -238,6 +238,26 @@ vips_format_sizeof_unsafe(VipsBandFormat format)
 	return vips__image_sizeof_bandformat[format];
 }
 
+/**
+ * vips_interpretation_max_alpha:
+ * @interpretation: image interpretation
+ *
+ * Returns: the maximum alpha value for an interpretation.
+ */
+double
+vips_interpretation_max_alpha(VipsInterpretation interpretation)
+{
+	switch (interpretation) {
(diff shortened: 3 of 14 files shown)


libvips/conversion/flatten.c
```diff
@@ -325,10 +325,13 @@ vips_flatten_build(VipsObject *object)
...
 	 * interpretation.
 	 */
-	if (!vips_object_argument_isset(object, "max_alpha"))
+	if (!vips_object_argument_isset(object, "max_alpha")) {
 		if (in->Type == VIPS_INTERPRETATION_GREY16 ||
 			in->Type == VIPS_INTERPRETATION_RGB16)
 			flatten->max_alpha = 65535;
+		else if (in->Type == VIPS_INTERPRETATION_scRGB)
+			flatten->max_alpha = 1.0;
+	}
```
jcupitt, line 334: We have this code in 4 places now. How about adding something like:

```C
double
vips_image_guess_max_alpha(const VipsImage *image)
{
}
```

To `iofuncs/header.c`, just after `vips_image_guess_interpretation()`?

jcupitt (commented):
Nice! I just thought of one thing.

jcupitt:
The background here is that photoshop writes RGBA float as 0-1 linear, with alpha also 0-1 linear.

We set scRGB alpha to be 0-255 way back when there didn't seem to be a useful standard for it. Now it looks like 0-1 alpha is the thing that has been picked, we should use it too.

This is obviously a breaking change, but scRGB is (for now) such a minority thing I doubt if many people will be affected.

libvips/colour/scRGB2sRGB.c
```diff
@@ -140,7 +140,7 @@ vips_scRGB2sRGB_line_16(unsigned short *restrict q, float *restrict p,
 		q += 3;
 
 		for (j = 0; j < extra_bands; j++)
-			q[j] = VIPS_CLIP(0, (int) (p[j] * 256.0), USHRT_MAX);
+			q[j] = VIPS_CLIP(0, (int) (p[j] * 65535.0), USHRT_MAX);
```
jcupitt, line 143: Looks like `XYZ2scRGB.c` and `scRGB2XYZ.c` will need revising too, since they currently don't change the alpha range.

jcupitt (commented):
added a note about the test suite failure

libvips/conversion/flatten.c
DarthSim: Yeah, that's a good idea 👍

libvips/colour/scRGB2sRGB.c
DarthSim: So does `scRGB2BW`, my bad

DarthSim:
- Fixed alpha conversion in `XYZ2scRGB`, `scRGB2XYZ`, and `scRGB2BW`
- Fixed tests
- Extracted `max_alpha` guessing to a function

libvips/iofuncs/header.c
```diff
@@ -782,6 +802,18 @@ vips_image_guess_interpretation(const VipsImage *image)
...
 }
 
+/**
+ * vips_image_guess_max_alpha:
+ * @image: image to guess for
+ *
+ * Returns: the maximum alpha value for the image interpretation.
+ */
+double
+vips_image_guess_max_alpha(const VipsImage *image)
```
jcupitt, line 812: I'd get rid of this, I don't think we need both.

jcupitt (approved):
LGTM! I made one final comment (sorry).

libvips/iofuncs/header.c
DarthSim: Wast't sure about this too. Removed

kleisauke:
It looks like `find_trim` needs also to be adjusted.
https://github.com/libvips/libvips/blob/7ff83ba0b475ae1c9c5fa1617a05d1f2322d4e90/libvips/arithmetic/find_trim.c#L92-L101

Perhaps something along the lines of?:
```c
	if (!vips_object_argument_isset(object, "background")) {
		double max_alpha = vips_interpretation_max_alpha(find_trim->in->Type);
		if (max_alpha != 255.0) {
			vips_area_unref(VIPS_AREA(find_trim->background));
			find_trim->background = vips_array_double_newv(1, max_alpha);
		}
	}
```

kleisauke:
Actually, it looks like that it should choose an appropriate value depending on the band format in `find_trim`, I just opened PR #3631 for this.

Grep-ing through the codebase, I noticed that the OpenEXR loader scales the alpha up to 0 - 255, I think the changes in commit f4250ab7ab47a02dc57b3c28df38148c56f4ec8d should be reverted (minus the changelog notes).

kleisauke (approved):
LGTM!

kleisauke:
> Grep-ing through the codebase, I noticed that the OpenEXR loader scales the alpha up to 0 - 255, I think the changes in commit [f4250ab](https://github.com/libvips/libvips/commit/f4250ab7ab47a02dc57b3c28df38148c56f4ec8d) should be reverted (minus the changelog notes).

I just opened PR #3673 for this.

PR #3630, opened 2023-08-28
find_trim: fix a crash when enabling `line_art` mode

kleisauke:


kleisauke:
Regression on master branch. This bug did not affect any released versions, hence no ChangeLog entry or backport is required.

lovell (approved):

PR #3631, opened 2023-08-28
Improve scRGB/16-bit support in `find_trim` and `embed`

kleisauke:


jcupitt:
How about using `vips_interpretation_max_alpha()`? It does more or less what we need:

https://github.com/libvips/libvips/blob/master/libvips/iofuncs/header.c#L241-L259

jcupitt:
Embed could use this too, of course.

kleisauke:
You're right; I confused `VipsBandFormat` with `VipsInterpretation`.

kleisauke:
Here's a test case that fails on master but works on this PR:
```python
#!/usr/bin/python3

import pyvips

im = pyvips.Image.black(50, 60).colourspace(pyvips.Interpretation.SCRGB)
im = im.embed(10, 20, 200, 300, extend=pyvips.Extend.WHITE)
im.tiffsave('x.tiff')

im = pyvips.Image.new_from_file('x.tiff')
assert im.getpoint(0, 0) == [1.0, 1.0, 1.0]
assert im.getpoint(10, 20) == [0.0, 0.0, 0.0]

left, top, width, height = im.find_trim(line_art=True, threshold=0.1)
assert left == 10
assert top == 20
assert width == 50
assert height == 60
```

libvips/arithmetic/find_trim.c
```diff
@@ -92,13 +92,13 @@ vips_find_trim_build(VipsObject *object)
 	/* Is "background" unset? Default to the correct value
 	 * for this interpretation.
 	 */
-	if (!vips_object_argument_isset(object, "background"))
-		if (find_trim->in->Type == VIPS_INTERPRETATION_GREY16 ||
-			find_trim->in->Type == VIPS_INTERPRETATION_RGB16) {
+	if (!vips_object_argument_isset(object, "background")) {
+		double max_val = vips_interpretation_max_alpha(find_trim->in->Type);
+		if (max_val != 255.0) {
```
jcupitt, line 97: How about not setting `background` in `vips_find_trim_init()` and leaving it NULL? You could remove this `if()` and get rid of the 255 magic number.

libvips/conversion/embed.c
```diff
@@ -274,11 +275,14 @@ vips_embed_base_gen(VipsRegion *out_region,
 	case VIPS_EXTEND_WHITE:
 		VIPS_GATE_START("vips_embed_base_gen: work1");
 
+		ink = base->extend == VIPS_EXTEND_BLACK
+			? 0
+			: (int) vips_interpretation_max_alpha(base->in->Type);
+
```
jcupitt, line 281: Ah, much nicer!

jcupitt (approved):
LGTM!

PR #3647, opened 2023-09-02
fix crash with alpha handling for icc import and export

jcupitt:
The ICC import and export line process functions were using `SIZEOF_PEL` on the input and output images, but the image they actually handle may have had alpha removed. This could cause SEGVs with formats like CMYKA.

This PR makes them use the band count that LCMS will see.

Also bump version, and fix a test fail.

See https://github.com/libvips/libvips/discussions/3640

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,4 @@
-2/9/23 8.14.5
+TBD 8.14.5
 
 - fix a crash with alpha plus icc_import and icc_export [jcupitt]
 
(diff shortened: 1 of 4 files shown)


ChangeLog
```diff
@@ -1,3 +1,7 @@
+2/9/23 8.14.5
```
kleisauke, line 1:
```suggestion
TBD 8.14.5
```

kleisauke (approved):
LGTM!

PR #3674, opened 2023-09-21
Fix invalid UTF-8 strings instead of ignoring them

DarthSim:
Hey there 👋

libvips checks UTF-8 string validity using `g_utf8_validate` in several places and ignores invalid strings or replaces them with `<invlaid utf-8 string>`. This leads to cases where the whole string is ignored or replaced because of a single invalid character. Instead, we could use the `g_utf8_make_valid` function introduced in GLib 2.52 to replace invalid characters.

GLib 2.52 was released 6 years ago (even Debian Buster has GLib 2.58) so increasing the minimum required version to 2.52 shouldn't be a problem.

### DIFF
--- a/libvips/iofuncs/type.c
+++ b/libvips/iofuncs/type.c
@@ -563,14 +563,13 @@ vips_ref_string_new(const char *str)
 {
 	VipsArea *area;
 
-	if (!g_utf8_validate(str, -1, NULL))
-		str = "<invalid utf-8 string>";
+	char *utf8_str = g_utf8_make_valid(str, -1);
 
-	area = vips_area_new((VipsCallbackFn) vips_area_free_cb, g_strdup(str));
+	area = vips_area_new((VipsCallbackFn) vips_area_free_cb, utf8_str);
 
 	/* Handy place to cache this.
 	 */
-	area->length = strlen(str);
+	area->length = strlen(utf8_str);
 
 	return (VipsRefString *) area;
 }
--- a/meson.build
+++ b/meson.build
@@ -54,7 +54,7 @@ cpp = meson.get_compiler('cpp')
 # Prevent use of void* pointer arithmetic to support MSVC
 add_project_arguments(cc.get_supported_arguments('-Werror=pointer-arith'), language : ['cpp', 'c'])
 
-glib_dep = dependency('glib-2.0', version: '>=2.40')
+glib_dep = dependency('glib-2.0', version: '>=2.52')
 gio_dep = dependency('gio-2.0')
 gobject_dep = dependency('gobject-2.0')
 gmodule_dep = dependency('gmodule-no-export-2.0', required: get_option('modules'))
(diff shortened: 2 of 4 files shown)


libvips/iofuncs/type.c
```diff
@@ -563,10 +563,8 @@ vips_ref_string_new(const char *str)
 {
 	VipsArea *area;
 
-	if (!g_utf8_validate(str, -1, NULL))
-		str = "<invalid utf-8 string>";
-
-	area = vips_area_new((VipsCallbackFn) vips_area_free_cb, g_strdup(str));
+	area = vips_area_new((VipsCallbackFn) vips_area_free_cb,
+		g_utf8_make_valid(str, -1));
```
lovell, line 567: Do we need to check for `utf8_make_valid()` returning `NULL`?

meson.build
```diff
@@ -54,7 +54,7 @@ cpp = meson.get_compiler('cpp')
 # Prevent use of void* pointer arithmetic to support MSVC
 add_project_arguments(cc.get_supported_arguments('-Werror=pointer-arith'), language : ['cpp', 'c'])
 
-glib_dep = dependency('glib-2.0', version: '>=2.40')
+glib_dep = dependency('glib-2.0', version: '>=2.52')
```
lovell, line 57: The only active Linux I find that this might impact is Slackware 14, which still provides glib 2.46. No idea if that's a problem; there's one way to find out :)

libvips/iofuncs/type.c
DarthSim: The only case when `utf8_make_valid` returns `NULL` is when the input string is `NULL`. But `g_utf8_validate' would cause segfault with a `NULL` input, so I guess we believe that all the input strings are not `NULL` here.

meson.build
DarthSim: Is Slackware 14 still a thing? Slackware 15 provides 2.70

kleisauke: I usually pick the lowest supported RHEL version in this case, which is RHEL 7 according to https://endoflife.date/rhel. Fortunately, that provides GLib 2.56.1, so it's probably reasonable to raise this minimum required version to 2.52.

Note to self: revise these conditional guards after this PR has landed.
https://github.com/libvips/libvips/blob/5482ead701b29cbfd3eeab6e91603d972b3b8850/libvips/foreign/pforeign.h#L40
https://github.com/libvips/libvips/blob/5482ead701b29cbfd3eeab6e91603d972b3b8850/libvips/iofuncs/thread.c#L188
https://github.com/libvips/libvips/blob/5482ead701b29cbfd3eeab6e91603d972b3b8850/libvips/iofuncs/init.c#L496

libvips/iofuncs/type.c
kleisauke, line 567: I think this should be hoisted into a separate variable, as it computes the length on the original string a couple of lines below.
https://github.com/libvips/libvips/blob/75035c90713073aa74583694c99078e8a4ffc4aa/libvips/iofuncs/type.c#L571

(i.e. the length of the string should be computed on `g_utf8_make_valid(str, -1)` instead)

kleisauke (commented):
LGTM % comment

kleisauke (approved):

PR #3689, opened 2023-09-26
fix static linking when libwebp is used

kfix:
libwebp depends on symbols from its sublibs, so it must come after them on the linker command, not before, or else ld will complain about missing symbols.

static building libvips.a used to work for me with the autotools Makefile and the same libwebp, before it was replaced with meson ( https://github.com/libvips/libvips/pull/2637 & https://github.com/libvips/libvips/pull/2941 )

fwiw, pkg-config was used by autotools' `PKG_CHECK_MODULES(LIBWEBP, libwebp >= 0.6 libwebpmux >= 0.6 libwebpdemux >= 0.6)` so that got the order right then, but meson's dependency() is [supposed to use pkg-config as well](https://mesonbuild.com/Dependencies.html#dependency-detection-method). I'm not sure why that doesn't automagically get the order right for me with the current meson & pkg-config tools installed.

tested with meson setup build -Ddefault_library=static ; cd build; meson compile;

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -347,9 +347,9 @@ endif
 # insist on having both of them
 libwebp_dep = dependency('libwebp', version: '>=0.6', required: get_option('webp'))
 if libwebp_dep.found()
-    libvips_deps += libwebp_dep
     libvips_deps += dependency('libwebpmux', version: '>=0.6')
     libvips_deps += dependency('libwebpdemux', version: '>=0.6')
+    libvips_deps += libwebp_dep
     cfg_var.set('HAVE_LIBWEBP', '1')
 endif
 


kfix:
also, is tested only for a fairly old line of libwebp (1.1.0) because they've added another sublib "libsharpyuv" in more recent ones that also needs to be added in.

jcupitt:
Hi @kfix,

That's interesting. I wasn't able to reproduce this linking problem. I wondered if intransitive dependencies on webp were pulling in the sub-libraries, so I tried:

```
$ meson setup build --prefix ~/vips --libdir lib --buildtype debug \
    -Dintrospection=disabled \
    -Dmagick=disabled \
    -Dwebp=disabled \
    -Djpeg-xl=disabled \
    -Dopenjpeg=disabled \
    -Dtiff=disabled
```

ie. remove all use of libwebp, even intransitive ones. After installing, I see:

```
$ ldd $(which vips) | grep webp
$ 
```

so no use of libwebp remains. 

If I then build with:

```
$ meson setup build --prefix ~/vips --libdir lib --buildtype debug \
    -Ddefault_library=static \
    -Dintrospection=disabled \
    -Dmagick=disabled \
    -Djpeg-xl=disabled \
    -Dopenjpeg=disabled \
    -Dtiff=disabled
```

so libwebp is being pulled in just by libvips, it all works and there are no link issues.

What platform are you seeing the link problem on? I'm ubuntu 23.04, libwebp 1.2.4.

kfix:
i'm on ubuntu 20.04 😭 

After upgrading from libwebp 1.1.0 [ish](https://chromium.googlesource.com/webm/libwebp/+/refs/heads/1.1.0) to 1.2.4 [ish](https://chromium.googlesource.com/webm/libwebp/+/refs/heads/1.2.4),  libvips.a built without any issues and with without this patch.

So I'm guessing they updated their .pc data in 1.2.x so that static builds Just Works ™️ 

Maybe the minimum required version should be bumped to 1.2.0 just to prevent this issue for others?

PR #3690, opened 2023-09-27
vips2tiff: fix function pointer cast issue

kleisauke:
`g_slist_sort()` uses the same implementation as `g_slist_sort_with_data()` within GLib; hence, the function's signature must align with that of `GCompareDataFunc`.

Helps wasm-vips, where function pointer cast issues fails at runtime.

---
SO answer https://stackoverflow.com/a/14044244 might be relevant here.


jcupitt:
Argh sorry, I always forget this.

jcupitt (approved):

PR #3695, opened 2023-10-03
add threadset to the public API

jcupitt:
It's in the headers already, just not in the library binary. I'd like to use it in vipsdisp for background workers.

Does this have any consequences? I think it's pretty safe, and these functions are already documented.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -28,6 +28,7 @@
 - require C++11 as a minimum standard [kleisauke]
 - add support for SIMD via Highway [kleisauke]
 - threaded write in tiffsave for tiled JPEG and JPEG2000 [jcupitt]
+- add vips_threadset public API [jcupitt]
 
 18/9/23 8.14.5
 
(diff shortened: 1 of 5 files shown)


jcupitt:
... on second thoughts, it's probably better to rename and expose `vips_thread_execute()` so clients can schedule jobs on the libvips threadset.

ChangeLog
```diff
@@ -28,6 +28,7 @@
 - require C++11 as a minimum standard [kleisauke]
 - add support for SIMD via Highway [kleisauke]
 - threaded write in tiffsave for tiled JPEG and JPEG2000 [jcupitt]
+- add vips_thread_execute() public API [jcupitt]
```
kleisauke, line 31:
```suggestion
- add vips_thread_execute() to the public API [jcupitt]
```

kleisauke (approved):
LGTM!

PR #3698, opened 2023-10-03
Mention imgproxy in README

DarthSim:
Hey there 👋

I just noticed that one of the most popular tools that uses libvips–imgproxy–is not enlisted in README. I believe this should be fixed 😅


jcupitt (approved):
:facepalm: What an oversight! I agree, it should be in there.

PR #3711, opened 2023-10-11
Revise docs for Highway

kleisauke:


lovell:
Is this the moment we need to settle on optimise vs optimize? ;)

(jpegsave has `optimize_coding`, gifsave has/had `reoptimise`).

jcupitt:
> (jpegsave has optimize_coding, gifsave has/had reoptimise).

Oh heh. I think the jpeg one is exposing the `optimize_coding` setting in libjpeg, so it's US spelling, but the GIF one is our own, so it's UK spelling. But you're right, they should match.

It shloud probably have been US spelling everywhere way back when that choice was made, but it's too late now and swapping everything to US spelling sounds annoying, so I'd suggest deprecating and renaming the jpeg one.

lovell (approved):

PR #3720, opened 2023-10-18
Fix saving EXIF strings with invalid UTF8 encoding

DarthSim:
Hey there! 👋

https://github.com/libvips/libvips/pull/3674 led to a bug in a case where an EXIF string contains an invalid UTF8 character:

1. When vips loads an EXIF string, its invalid characters are fixed here https://github.com/libvips/libvips/blob/c1e4d031fba77b76d4ce5a104ac8608ffa8f1f18/libvips/iofuncs/type.c#L566
2. When vips updates EXIF during saving, it compares the string from the image's meta with the output of `vips_exif_to_s`. And since vips fixed the string during the load, the strings become different, so vips replaces the EXIF string with one from the image's meta which includes a human-friendly representation, size, components number, etc.

This PR adds `g_utf8_make_valid` to `vips_exif_to_s` so it returns a fixed UTF8 string.

### DIFF
--- a/libvips/foreign/exif.c
+++ b/libvips/foreign/exif.c
@@ -296,7 +296,10 @@ vips_exif_to_s(ExifData *ed, ExifEntry *entry, VipsBuf *buf)
 	int iv;
 	ExifRational rv;
 	ExifSRational srv;
-	char txt[256];
+	char txt[256], *value;
+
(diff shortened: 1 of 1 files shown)


kleisauke (approved):
Thanks!

PR #3725, opened 2023-10-20
heifload: don't allocate meta/ICC memory as local to out

DarthSim:
Hey there 👋

I found a nasty bug in `heifload` that leads to segfault in specific cases.

`heifload` allocates memory for meta (EXIF, XMP, etc) and ICC profile as local to the output image. If you close the image while having its descendants alive, you'll get a segfault when one of its descendants try to read ICC/meta.

Here's a reproduction example:

```c
#include <vips/vips.h>

int main( int argc, char *argv[] )  {
  if( argc < 2 ) {
    printf("src image path is missing\n");
    return 1;
  }

  vips_init("test");

  VipsImage *im1, *im2, *im3;

  if (!(im1 = vips_image_new_from_file(argv[1], "access", VIPS_ACCESS_SEQUENTIAL, NULL))) {
    printf("Failed to load image: %s", vips_error_buffer());
    return 1;
  }

  im2 = vips_image_copy_memory(im1);

  g_clear_object(&im1);

  vips_icc_import(im2, &im3, NULL);

  g_clear_object(&im2);
  g_clear_object(&im3);

  vips_shutdown();
}
```

It doesn't always cause segfault, so build it with `-fsanitize=address`

### DIFF
--- a/libvips/foreign/heifload.c
+++ b/libvips/foreign/heifload.c
@@ -514,11 +514,12 @@ vips_foreign_load_heif_set_header( VipsForeignLoadHeif *heif, VipsImage *out )
 
 		if( !length )
 			continue;
-		if( !(data = VIPS_ARRAY( out, length, unsigned char )) )
+		if( !(data = VIPS_ARRAY( NULL, length, unsigned char )) )
 			return( -1 );
 		error = heif_image_handle_get_metadata( 
 			heif->handle, id[i], data );
 		if( error.code ) {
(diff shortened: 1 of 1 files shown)


jcupitt (approved):
Ooof that was a bad brainfart on my part. Well done for finding this!

I'll make a PR for master as well.

jcupitt:
I applied this to 8.15 and master as well.

kleisauke (approved):

PR #3731, opened 2023-10-23
heifsave: set default nclx profile if ICC profile is not provided

DarthSim:
Hey there 👋

Since `1.17.0`, libheif saves files with the default NCLX profile `6,6,6` (Rec. 601 aka NTSC) which is pretty far from sRGB. This leads to incorrect colors in most viewers. This PR makes `heifsave` use NCLX `13,1,1` (Rec. 709 with IEC 61966-2-1 transfer function) if no ICC profile is provided. NCLX `13,1,1` is basically sRGB.

### DIFF
--- a/libvips/foreign/heifsave.c
+++ b/libvips/foreign/heifsave.c
@@ -261,6 +261,33 @@ vips_foreign_save_heif_add_orig_icc(VipsForeignSaveHeif *heif)
 
 	return 0;
 }
+
+static int
+vips_foreign_save_heif_add_default_nclx(VipsForeignSaveHeif *heif)
+{
+	struct heif_color_profile_nclx *nclx;
+	struct heif_error error;
+
+	if (!(nclx = heif_nclx_color_profile_alloc())) {
+		g_warning("heifsave: can't allocate nclx profile");
+		return 0;
+	}
+
+	nclx->color_primaries = heif_color_primaries_ITU_R_BT_709_5;
+	nclx->transfer_characteristics = heif_transfer_characteristic_IEC_61966_2_1;
+	nclx->matrix_coefficients = heif_matrix_coefficients_ITU_R_BT_709_5;
+
+	error = heif_image_set_nclx_color_profile(heif->img, nclx);
+	if (error.code) {
+		heif_nclx_color_profile_free(nclx);
+		vips__heif_error(&error);
+		return -1;
+	}
+
+	heif_nclx_color_profile_free(nclx);
+
+	return 0;
+}
 #endif /*HAVE_HEIF_COLOR_PROFILE*/
 
 static int
@@ -283,6 +310,8 @@ vips_foreign_save_heif_write_page(VipsForeignSaveHeif *heif, int page)
 		if (vips_foreign_save_heif_add_orig_icc(heif))
 			return -1;
 	}
+	else
+		vips_foreign_save_heif_add_default_nclx(heif);
 #endif /*HAVE_HEIF_COLOR_PROFILE*/
 
 	options = heif_encoding_options_alloc();


kleisauke:
> Since `1.17.0`, libheif saves files with the default NCLX profile `6,6,6` (Rec. 601 aka NTSC) which is pretty far from sRGB.

FWIW, with libheif v1.16.2, I see:
```console
$ curl -LO https://wsrv.nl/zebra.jpg
$ vips copy zebra.jpg x.avif[keep=none]
$ exiftool -MatrixCoefficients x.avif
Matrix Coefficients             : BT.601
```

So, that also saves with a Rec. 601 matrix.
https://github.com/strukturag/libheif/blob/v1.16.2/libheif/box.cc#L1815

Looking at https://github.com/strukturag/libheif/compare/v1.16.2...v1.17.0, I think this issue was introduced in commit https://github.com/strukturag/libheif/commit/062b5284216748925cdbf67f9eb8a12bbda503e1.

DarthSim:
1.16.2 leaves color primaries and transfer function unspecified. It seems like most viewers use the Rec. 709 color primaries and the IEC 61966-2-1  transfer function by default.

Yes, you,ve found the right commit. As you see, it adds setting the default values for the undefined CiCp fields

kleisauke:
With libheif v1.16.2 and v1.17.1, I see:
```console
$ vips copy zebra.jpg x.avif[keep=none]
$ exiftool -ColorPrimaries -TransferCharacteristics -MatrixCoefficients x.avif 
Color Primaries                 : Unspecified
Transfer Characteristics        : Unspecified
Matrix Coefficients             : BT.601
```

So, it writes both color primaries and transfer characteristics as unspecified, so I don't quite understand why viewers would interpret it differently when the image was saved with libheif v1.17.1.

@farindk Sorry to ping you, do you have any thoughts on this? Perhaps the issue is that libheif's internal color conversions now assume Rec. 601 color primaries and transfer characteristics, but still write them as unspecified, causing a mismatch? If so, perhaps commit https://github.com/kleisauke/libheif/commit/a7ad14c1c30ddad2dd5b8c421c4c8f0cccf43f03 would help with this.

DarthSim:
If all the CiCp values are enforced to Rec. 601, the coloring is still wrong since Rec. 601 represents NTSC, not sRGB.

I believe that libvips should've been setting NCLX profile with v1.16.2 too. It's just a matter of luck that viewers are interpreting unspecified CiCp values as sRGB ones 🙂

PS. I'm working on adding NCLX support to libvips, so I read a thing or two about CiCp/NCLX. And I can assure you that `2,2,6` is the wrong profile for sRGB.

DarthSim:
Here's a picture for tests:

hdr

`vips icc_transform hdr.jpg hdr.avif[strip=1] srgb` will give you this (compressed since GitHub doesn't support AVIF attaches):

[hdr.avif.zip](https://github.com/libvips/libvips/files/13120109/hdr.avif.zip)

kleisauke:
Thanks for the reproducer! I was able to reproduce the color distortion on:
- Firefox (both on Windows and Fedora);
- Windows default image viewer.

Google Chrome and Eye of GNOME does not appear to have this issue. I can also confirm this is a regression in libheif v1.17.0+, as I didn't encounter this issue when re-saving the image with libheif v1.16.2. I've attached this image here for reference:
[hdr2.zip](https://github.com/libvips/libvips/files/13198212/hdr2.zip)

> If all the CiCp values are enforced to Rec. 601, the coloring is still wrong since Rec. 601 represents NTSC, not sRGB.

I think you're right. I was only triggered on the fact that this is only reproducible when saving the image with libheif v1.17.0+, which sounds like it should be fixed upstream instead.

> It's just a matter of luck that viewers are interpreting unspecified CiCp values as sRGB ones

IIUC (correct me if I'm wrong), libheif v1.17.0+ still saves the image with both color primaries and transfer characteristics as unspecified (i.e. `2,2,6` as [identified by exiftool](https://github.com/libvips/libvips/pull/3731#issuecomment-1777416281) and not `6,6,6` as stated in the PR description), so that doesn't seem to be the issue.

DarthSim:
> IIUC (correct me if I'm wrong), libheif v1.17.0+ still saves the image with both color primaries and transfer characteristics as unspecified (i.e. 2,2,6 as https://github.com/libvips/libvips/pull/3731#issuecomment-1777416281 and not 6,6,6 as stated in the PR description), so that doesn't seem to be the issue.

If it wasn't the issue, this PR wouldn't fix it :)

I was confused with this, too. I'm not completely sure, but it seems like NCLX profile is saved in two places: the `colr` HEIF box and the AV1 stream header. And it seems like libheif writes the `colr` box before "[fixing](https://github.com/strukturag/libheif/commit/062b5284216748925cdbf67f9eb8a12bbda503e1#diff-52479e0a0a104437084235c078e8dc902c5c866975de952b1cf1abc6d90d8b88R505-R515)" unspecified values. If it would "fix" the profile before writing the `colr` box, ExifTool would show `6,6,6`.

A couple of things that confirm this theory:
1. Photoshop opens AVIFs created with `1.16.2` but doesn't open ones created with `1.17.0`. I guess this is because the profile from the `colr` box and one from the AV1 stream header don't match.
2. Apple Preview identifies the profile of the image saved with `1.16.2` as sRGB, but it identifies the profile of the image saved with `1.17.0` as NTSC. I believe that this is because it uses the profile from the AV1 stream header.
3. This also explains why some viewers show distorted colors and some don't: ones that use the profile from the `colr` box show correct colors, and ones that use the profile from the AV1 stream header show distorted colors. So if libheif would write the "fixed" profile to the `colr` box, all the viewers would show distorted colors.

kleisauke:
I still need to polish the commit messages, but I think this revision range would fix both issues upstream:
https://github.com/strukturag/libheif/compare/master...kleisauke:rec-709-default

DarthSim:
`transfer_characteristics` should be `heif_transfer_characteristic_IEC_61966_2_1` to match sRGB. `heif_transfer_characteristic_ITU_R_BT_709_5` gives a bit distorted colors.

farindk:
> I'm not completely sure, but it seems like NCLX profile is saved in two places: the `colr` HEIF box and the AV1 stream header. And it seems like libheif writes the `colr` box before "[fixing](https://github.com/strukturag/libheif/commit/062b5284216748925cdbf67f9eb8a12bbda503e1#diff-52479e0a0a104437084235c078e8dc902c5c866975de952b1cf1abc6d90d8b88R505-R515)" unspecified values. If it would "fix" the profile before writing the `colr` box, ExifTool would show `6,6,6`.

Yes, the nclx profile can be stored in the encoded bitstream or the HEIF header. libheif writes the same nclx data in both places (unless there is a bug).
A lot of these problems arise from version 1 of the HEIF standard that said that there may only be one `colr`box. Now, if the file contains an ICC profile, it wouldn't be legal to also write another nclx into the HEIF container. Thus, libheif writes the nclx to the bitstream only and keeps the ICC in the HEIF.
With HEIF version 2, the standard now says that there may only be one `colr` for *each profile type*. Thus libheif may now save `nclx` also to the container even when there is ICC information. To enable this behavior, one has to set `heif_encoding_options::save_two_colr_boxes_when_ICC_and_nclx_available=true`.

kleisauke:
This should now be fixed upstream via https://github.com/strukturag/libheif/compare/c845ffa869cb3e6ce7814b4568e04f6381105690...7cd873f11796f0d5c374a6a0b4b3f0d42f418ed3. Therefore, I'm not entirely sure if it's still necessary to land this PR (assuming that this revision range will soon be released in a patch release of libheif).

If landing is still necessary, we should consider adjusting the matrix coefficients to make them consistent with upstream:
https://github.com/libvips/libvips/blob/4981107c581799cf1667ce8ae924184fe2f22817/libvips/foreign/heifsave.c#L278
https://github.com/strukturag/libheif/blob/7cd873f11796f0d5c374a6a0b4b3f0d42f418ed3/libheif/nclx.cc#L301

DarthSim:
Honestly, I'm not sure, too. On one hand, this PR will secure us from any changes in the default NCLX profile in libheif. It was pure luck that I noticed this, so I'd rather be safe than sorry. On the other hand, this is another piece of code to maintain, but no code is better than no code (imgproxy's moto 🙂).

I'd probably like this to be merged, but this is definitely not necessary.

kleisauke:
[libheif v1.17.2](https://github.com/strukturag/libheif/releases/tag/v1.17.2) has just been released, I'll close.

PR #3735, opened 2023-10-24
magick: prefer use of `GetMagicInfo()`

kleisauke:
This makes our magick sniffer aware of the various magic signatures defined with the `MagickCoderHeader` macro in ImageMagick.

See for context #3734, targets the 8.15 branch.

### DIFF
--- a/libvips/foreign/magick.c
+++ b/libvips/foreign/magick.c
@@ -79,6 +79,14 @@
 static const char *
 magick_sniff(const unsigned char *bytes, size_t length)
 {
+	if (length >= 5 &&
+		bytes[0] == 0 &&
+		bytes[1] == 1 &&
+		bytes[2] == 0 &&
+		bytes[3] == 0 &&
+		bytes[4] == 0)
+		return "TTF";
+
 	if (length >= 6 &&
 		bytes[0] == 0 &&
 		bytes[1] == 0 &&
--- a/libvips/foreign/magick.c
+++ b/libvips/foreign/magick.c
@@ -106,7 +106,7 @@ magick_sniff(const unsigned char *bytes, size_t length)
 	const MagicInfo *magic_info = GetMagicInfo(bytes, length, exception);
 	magick_destroy_exception(exception);
 
-	if (magic_info != NULL)
+	if (magic_info)
 		return GetMagicName(magic_info);
 
 	return NULL;
--- a/libvips/foreign/magick.c
+++ b/libvips/foreign/magick.c
@@ -100,6 +100,7 @@ magick_sniff(const unsigned char *bytes, size_t length)
 		memcmp(bytes + 4, "ftyp", 4) != 0)
 		return "TGA";
 
+#if defined(HAVE_GETMAGICINFO) || defined(HAVE_MAGICK7)
 	/* Try to search the internal magic list for a match.
 	 */
 	ExceptionInfo *exception = magick_acquire_exception();
@@ -108,6 +109,7 @@ magick_sniff(const unsigned char *bytes, size_t length)
 
 	if (magic_info)
 		return GetMagicName(magic_info);
+#endif
 
 	return NULL;
 }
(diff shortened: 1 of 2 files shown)


libvips/foreign/magick.c
```diff
@@ -88,14 +87,6 @@ magick_sniff(const unsigned char *bytes, size_t length)
 		(bytes[4] != 0 || bytes[5] != 0))
 		return "ICO";
 
-	if (length >= 5 &&
-		bytes[0] == 0 &&
-		bytes[1] == 1 &&
-		bytes[2] == 0 &&
-		bytes[3] == 0 &&
-		bytes[4] == 0)
-		return "TTF";
```
kleisauke, line 97: This was removed as it would be covered by:
https://github.com/ImageMagick/ImageMagick/blob/9b109732da21e1604ec28d32926c176bffb9d3e2/coders/ttf.h#L20

libvips/foreign/magick.c
```diff
@@ -109,37 +100,14 @@ magick_sniff(const unsigned char *bytes, size_t length)
...
-		ifd_offset = ((guint32) bytes[7] << 24) |
-			((guint32) bytes[6] << 16) |
-			((guint32) bytes[5] << 8) |
-			bytes[4];
-
-	if (length >= 10 &&
-		ifd_offset >= 6 &&
-		bytes[8] == 'C' &&
-		bytes[9] == 'R')
-		return "CR2";
```
kleisauke, line 142: This was removed as it would be covered by:
https://github.com/ImageMagick/ImageMagick/blob/9b109732da21e1604ec28d32926c176bffb9d3e2/coders/dng.h#L20-L21

libvips/foreign/magick.c
```diff
@@ -109,37 +100,14 @@ magick_sniff(const unsigned char *bytes, size_t length)
...
+	ExceptionInfo *exception = magick_acquire_exception();
+	const MagicInfo *magic_info = GetMagicInfo(bytes, length, exception);
```
jcupitt, line 106: I don't think this will work with graphicsmagick -- they have:

```C
MagickExport const MagickInfo *
GetMagickInfo(const char *name,ExceptionInfo *exception)
```

jcupitt: (where `name` is the name of a format, eg. `"GIF"`)

jcupitt: Ah! They have:

```C
MagickExport const char *
GetImageMagick(const unsigned char *magick,const size_t length)
```

To find the format name from a string, then use `GetMagickInfo()` to get the `MagickInfo` for that format.

kleisauke: Good point, I excluded this for GraphicsMagick with commit bc31f1b536b5794b86c25030d0fff2154a3f4997.

We already use `GetImageMagick()` here:
https://github.com/libvips/libvips/blob/df8b8e247ed410138fcacec2f43f0c5c16aecc37/libvips/foreign/magick.c#L663-L666

kleisauke: ... `GetImageMagick()` would only work for formats that define the `->magick` function (see e.g. the JPEG case [here](https://github.com/ImageMagick/ImageMagick/blob/4d18167db34ea2fd199acd99420d61192d708f06/coders/jpeg.c#L1773)). The various DNG/RAW formats doesn't use this, so you'll still need to use `GetMagicInfo()`.

libvips/foreign/magick.c
kleisauke: Ah,  we still need the TTF sniffer as the TGA one is too greedy. Commit 95cdc5be014d70b3b4a06aa0267f41894fb9e1d0 re-adds this.

jcupitt (approved):
LGTM!

jcupitt:
I tested with GM 1.3.34 and IM 6.9.11 and it seems to work well.

Issue #3749, opened 2023-11-08
8.15.0 RC2 doesn't compile on i386


gcsideal:
**Conversion error on i386**
Just do a normal compilation on a i386 machine.

**To Reproduce**
1. Version 8.15.0 RC2 is used
2. Configured generic as I can tell, in a build subdirectory 'meson setup .. --buildtype=plain' was used.
3. For libvips/conversion/composite.cpp I get an error:
In function ‘void vips_composite_base_blend3(VipsCompositeSequence*, VipsBlendMode, v4f&, T*)’:
../libvips/conversion/composite.cpp:867:31: error: conversion of scalar ‘double’ to vector ‘v4f’ {aka ‘__vector(4) float’} involves truncation
  867 |                         f = B <= 0.5
      |                             ~~^~~~~~
../libvips/conversion/composite.cpp:893:31: error: conversion of scalar ‘double’ to vector ‘v4f’ {aka ‘__vector(4) float’} involves truncation
  893 |                         f = A <= 0.5
      |                             ~~^~~~~~

**Expected behavior**
Compile the library normally as on other architectures.

**Actual behavior**
You can see the full build log at: https://buildd.debian.org/status/fetch.php?pkg=vips&arch=i386&ver=8.15.0%7Erc2-1&stamp=1699387234&raw=0

**Environment**
 - OS: Debian GNU/Linux Sid, up-to-date
 - Vips: 8.15.0 RC2

**Additional context**
8.14.5 was recently compiled correctly on i386, so the above bug is a new regression: https://buildd.debian.org/status/fetch.php?pkg=vips&arch=i386&ver=8.14.5-1%2Bb1&stamp=1698013737&raw=0

kleisauke:
Thanks for testing and reporting this! This reminds me of commit https://github.com/google/highway/commit/76e6863ce2fe72c139eaa6fc8654f470b3092cb4, does it work if you compile with `-fexcess-precision=standard`? If not, could you try applying this commit as patch?: https://github.com/kleisauke/libvips/commit/5dd9e169b4cc25287f8a91ded662429d8535668a.

kleisauke:
As an aside, I noticed this in the log file:
```
...
Run-time dependency libarchive found: NO (tried pkgconfig and cmake)
...
image pyramid save with libarchive: NO
````

You might want to replace `libgsf-1-dev` with `libarchive-dev` and `liborc-0.4-dev` with `libhwy-dev` in the control file:
[...]

PR #3750, opened 2023-11-09
composite: ensure compilation on i386

kleisauke:
Curiously, this only seems to happen with `-std=c++11` (commit 5482ead701b29cbfd3eeab6e91603d972b3b8850 / c264c54b351c8faac8e6f2db42413226a8428e16), it compiles fine when that's omitted.

Resolves: #3749.

Targets the 8.15 branch.


jcupitt (approved):
Looks good here

PR #3753, opened 2023-11-11
Bump version to 8.16 for the next release cycle

kleisauke:
(My spare time will be limited during this release cycle)


jcupitt:
I'm planning a six-month sabbatical starting maybe in January, so I can handle 8.16 if no one else wants to.

PR #3756, opened 2023-11-13
Add VIPS_META_INDEXED meta

DarthSim:
Hey there 👋

`palette-bit-depth` was a handy way to detect if an image has a palette and get its palette size. Now it's being replaced with `VIPS_META_BITS_PER_SAMPLE`, but we can't rely on it since 8-bit RGB images and indexed images with 256-color palette have the same `VIPS_META_BITS_PER_SAMPLE` value.

This PR adds the `VIPS_META_INDEXED` header that explicitly indicates if an image is indexed.


jcupitt:
Good idea @DarthSim.

I have an annoying comment -- we use the flag `palette` to select palette mode, but use `indexed` as the metadata item to indicate it. 

We should probably use the same word for consistency. What do you think?

DarthSim:
Yeah, `palette` is probably better. Changed the name

jcupitt (approved):

jcupitt:
LGTM! Thank you, @DarthSim

PR #3758, opened 2023-11-18
heifsave: set `threads` to vips_concurrency_get()

DarthSim:
Hey there 👋

Some AVIF encoders in libheif support the `threads` parameter. Its default value is `4` by default but it's almost never optimal. This PR sets it to the `vips_concurrency_get()` since it is the level of parallelism a Vips user wants.

On my machine (Apple M1, 10 cores), this speeds up the conversion of a 7360x4912 JPEG to AVIF from 11s to 6s.

### DIFF
--- a/libvips/foreign/heifsave.c
+++ b/libvips/foreign/heifsave.c
@@ -490,6 +490,7 @@ vips_foreign_save_heif_build(VipsObject *object)
 	struct heif_writer writer;
 	char *chroma;
 	const struct heif_encoder_descriptor *out_encoder;
+	const struct heif_encoder_parameter *const *param;
 
 	if (VIPS_OBJECT_CLASS(vips_foreign_save_heif_parent_class)->build(object))
 		return -1;
@@ -578,6 +579,31 @@ vips_foreign_save_heif_build(VipsObject *object)
 		error.subcode != heif_suberror_Unsupported_parameter) {
 		vips__heif_error(&error);
 		return -1;
+	}
+
+	for (param = heif_encoder_list_parameters(heif->encoder); *param; param++) {
+		int have_minimum;
+		int have_maximum;
+		int minimum;
+		int maximum;
+
+		if (strcmp(heif_encoder_parameter_get_name(*param), "threads") != 0)
+			continue;
+
+		error = heif_encoder_parameter_get_valid_integer_values(*param,
+			&have_minimum, &have_maximum, &minimum, &maximum, NULL, NULL);
+		if (error.code) {
+			vips__heif_error(&error);
+			return -1;
+		}
+
+		error = heif_encoder_set_parameter_integer(heif->encoder,
+			"threads", VIPS_CLIP(minimum, vips_concurrency_get(), maximum));
+		if (error.code &&
+			error.subcode != heif_suberror_Unsupported_parameter) {
+			vips__heif_error(&error);
+			return -1;
+		}
 	}
 
 	/* TODO .. support extra per-encoder params with


lovell:
Allowing control over the number of threads that the various libheif encoders can spawn sounds like a good thing to me. It would be interesting to see some performance comparisons at the lower end of the scale such as on machines with fewer than 4 cores, as well as in virtualised environments where the value returned by the underlying `g_get_num_processors` doesn't reflect reality (e.g. cgroups settings).

libvips/foreign/heifsave.c
```diff
@@ -580,6 +580,14 @@ vips_foreign_save_heif_build(VipsObject *object)
 		return -1;
 	}
 
+	error = heif_encoder_set_parameter_integer(heif->encoder,
+		"threads", vips_concurrency_get());
```
kleisauke, line 584: It looks like libheif does not allow more than 16 threads and will fail with the `heif_suberror_Invalid_parameter_value` subcode if you exceed that limit.
https://github.com/strukturag/libheif/blob/c5ea21eb51d8cd14f592cff64f8b179707e45592/libheif/plugins/encoder_aom.cc#L244
https://github.com/strukturag/libheif/blob/c5ea21eb51d8cd14f592cff64f8b179707e45592/libheif/plugins/encoder_rav1e.cc#L122
```suggestion
		"threads", VIPS_MIN(vips_concurrency_get(), 16));
```
(this was noticed on my 12-core/24-thread workstation)

DarthSim: Good catch, thank you! Fixed

DarthSim:
Actually, reducing the number of threads was the main reason for this PR. In imgproxy, we set Vips concurrency to 1 because imgproxy usually processes multiple images simultaneously, and using more threads than the CPU can run simultaneously slows down the processing dramatically. Same thing with libheif. In our Docker image, we were patching libheif to use a single thread by default, and this was speeding up saving AVIFs a lot. But I believe it's better to do so on the Vips level.

Here are some benchmaring results:

```bash
# 10 cores, master
/usr/bin/time -f "%e %M" vips copy /images/wat-arun.jpg /images/wat-arun.avif
11.15 844364

# 10 cores, this PS
/usr/bin/time -f "%e %M" vips copy /images/wat-arun.jpg /images/wat-arun.avif
5.90 841172

# 2 cores (cgroups), master
/usr/bin/time -f "%e %M" vips copy /images/wat-arun.jpg /images/wat-arun.avif
22.50 838812

# 2 cores (cgroups), this PR
/app# /usr/bin/time -f "%e %M" vips copy /images/wat-arun.jpg /images/wat-arun.avif
38.96 837752

# 2 cores (cgroups), VIPS_CONCURRENCY=2, master
VIPS_CONCURRENCY=2 /usr/bin/time -f "%e %M" vips copy /images/wat-arun.jpg /images/wat-arun.avif
22.24 802616

# 2 cores (cgroups), VIPS_CONCURRENCY=2, this PR
VIPS_CONCURRENCY=2 /usr/bin/time -f "%e %M" vips copy /images/wat-arun.jpg /images/wat-arun.avif
20.88 800136

# 2 cores (real), master
/usr/bin/time -f "%e %M" vips copy /images/wat-arun.jpg /images/wat-arun.avif
40.76 829960

# 2 cores (real), this PR
/usr/bin/time -f "%e %M" vips copy /images/wat-arun.jpg /images/wat-arun.avif
40.69 827696
```

So the only case where this PR slows down saving AVIF is when CPU usage is limited with cgroups and `VIPS_CONCURRENCY` is not set explicitly. However I believe that it is a user's responsibility to set `VIPS_CONCURRENCY` according to the cgroups limitations. Otherwise, vips itself won't work optimally.

lovell (approved):
LGTM, thank you very much for generating more performance test results.

jcupitt:
We could get the max threads from libheif, eg.:

```C
    int n_threads = vips_concurrency_get();

    const struct heif_encoder_parameter* const* p;
    for (p = heif_encoder_list_parameters(heif->encoder); *p; p++) {
        int have_minimum; 
        int have_maximum,
        int minimum; 
        int maximum;

        if (strcmp(heif_encoder_parameter_get_name(*p), "threads") != 0)
            continue;

        error = heif_encoder_parameter_get_valid_integer_values(*p,
            &have_minimum, &have_maximum, &minimum, &maximum, NULL, NULL);
        if (error != heif_error_success) {
            vips__heif_error(&error);
            return -1;
        } 

        if (have_minimum && 
            have_maximum)
            n_threads = VIPS_CLIP(minimum, n_threads, maximum);
    }

    error = heif_encoder_set_parameter_integer(heif->encoder,
        "threads", n_threads);
    if (error.code &&
        error.subcode != heif_suberror_Unsupported_parameter) {
        vips__heif_error(&error);
        return -1;
    }
```

It'd avoid having the magic number 16 in the code.

DarthSim:
Added detection of min/max `threads` value.

> It'd avoid having the magic number 16 in the code.

Well, we have the magic number 9 for `speed` 🙂

PR #3759, opened 2023-11-18
Fix build with PDFium

kleisauke:
Noticed on OSS-Fuzz: https://oss-fuzz-build-logs.storage.googleapis.com/index.html#libvips


jcupitt:
Ooop! Sorry and thanks.

PR #3760, opened 2023-11-18
heifsave: disable chroma subsampling by default in lossless mode

kleisauke:


jcupitt:
Ah, I wondered why lossless wasn't lossless! Of course.

PR #3763, opened 2023-11-22
reduceh: fix HWY path on SSE2

DarthSim:
Hey there 👋

`TableLookupBytes` doesn't work on SSE2 targets. `TableLookupBytesOr0` is a wrapper over `TableLookupBytes`. On targets higher than SSE2, it just calls `TableLookupBytes`, so there's no speed loss. On SSE2, it adds some code for zeroing. https://github.com/google/highway/blob/1.0.7/hwy/ops/x86_128-inl.h#L1590-L1602

Fixes https://github.com/imgproxy/imgproxy/issues/1223

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+TBD 8.15.1
+
+- reduceh: fix Highway path on SSE2 [DarthSim]
+
 11/11/23 8.15.0
 
 - add support for target_clones attribute [lovell]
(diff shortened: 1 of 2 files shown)


kleisauke (approved):
Thanks! It looks like I assumed that out-of-range indices were always zeroed in the output vector, but this zeroing behavior is only guaranteed with `TableLookupBytesOr0`. 

Fortunately, this function also has zero cost on Arm compared to the previous one. It may, however, incur a small penalty on SVE and RVV.

DarthSim:
> It looks like I assumed that out-of-range indices were always zeroed in the output vector

Well, they are zeroed in `_mm_shuffle_epi8`:  https://github.com/DarthSim/libvips/blob/feature/simd-optimizations-embedless/libvips/resample/reduceh_sse.cpp#L28-L38, but `_mm_shuffle_epi8` is an `SSSE3` instruction. So I understand your confusion here, `TableLookupBytes` was literally the last thing I checked 😅

jcupitt:
Thank you!

PR #3778, opened 2023-12-12
Fix jxlsave.c description

louie-github:
There's currently an error in the documentation for `vips jxlsave`:
```
save image in JPEG-XL format
usage:
   jxlsave in filename [--option-name option-value ...]
where:
   filename     - Filename to load from, input gchararray
```

"Filename to load from" should be "Filename to save to", just like with `pngsave`.

### DIFF
--- a/libvips/foreign/jp2ksave.c
+++ b/libvips/foreign/jp2ksave.c
@@ -1058,7 +1058,7 @@ vips_foreign_save_jp2k_file_class_init(VipsForeignSaveJp2kFileClass *class)
 
 	VIPS_ARG_STRING(class, "filename", 1,
 		_("Filename"),
-		_("Filename to load from"),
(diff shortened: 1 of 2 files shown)


louie-github:
On another note, I think "JPEG-XL" should be "JPEG XL" without the hyphen, as in https://jpeg.org/jpegxl/. Though this is very minor.

kleisauke:
Good catch! It looks like `jp2ksave` is also having the same issue.
https://github.com/libvips/libvips/blob/e95a169a3c66e36c5062736f3c622aa06f924d02/libvips/foreign/jp2ksave.c#L1061

jcupitt:
:facepalm: thanks @louie-github

kleisauke (approved):

PR #3790, opened 2023-12-19
fix(pdfium): render pdf annotations

project0:
Currently annotations are not rendered, this is very unexpected behavior as with other tools (imagemagick) or dependencies (poppler) it will be drawn. 

The flag `FPDF_ANNOT` enables render of annotations, this results in the same behavior as with poppler.

I used the annotation example files from pdfium to test it: https://pdfium.googlesource.com/pdfium_tests/+/refs/heads/main/pdfium/annots/


jcupitt:
Hi @project0,

I agree, this is a good fix. Thanks for pointing it out! I'll add a note to the changelog and credit you.

SpringMT:
When will this patch be released?

jcupitt:
In 8.16, so some time this summer.

PR #3791, opened 2023-12-21
fix: punctuation in readme

pReya:
There shouldn't be a space in front of a colon. This practice even has a name ("Plenken"), see: https://en.m.wikipedia.org/wiki/Plenken


jcupitt:
Heh, ok, let's change it!

(python RST uses space in front of colon for syntax :( so it's easy to fall into bad habits)

PR #3826, opened 2024-01-23
Fix typo to appease codespell

lovell:
Introduced in commit 6f7e43b

https://github.com/libvips/libvips/actions/runs/7603275230


jcupitt:
oops!

PR #3827, opened 2024-01-23
svgload: allow smaller scale factor

lovell:
By allowing smaller scale factors, this PR normalises the accepted range for `scale` from 1E-5 to 1E+5.

It also adds tests cases for both scale-up and scale-down (the latter would previously have failed).

Relates to https://github.com/lovell/sharp/issues/3964, where an input dimension of 333333 and target dimension of 250 results in a scale factor of 0.000750.


kleisauke:
A variant of this PR (+ changelog entry) for inclusion in [`8.15`](https://github.com/libvips/libvips/tree/8.15) is available as #3833.

Issue #3830, opened 2024-01-27
Regression in icc_transform for rgb16 -> srgb conversions


turtletowerz:
## Bug report

**Describe the bug**
Commit f807d7b1aaa09d10e4b38583fb417fb7f74b5e3c seems to have introduced a regression from `v8.14.5` where when inputting certain RGB16 images (it doesn't seem to happen with *all* of them) and attempting to export to sRGB will stay RGB16. The output image will remain RGB16 with no debug information or errors indicating any issues. I used `git bisect` to narrow it down to this commit. This breaks behavior of a project I'm working on which expects 8-bit images only, and I cannot find a way to convert certain images from 16 to 8 because of this.

**To Reproduce**
1. Run `vipsthumbnail input.png -v --vips-info --size=9x9! --export-profile=srgb -o out.png` on [this image](https://cdn.donmai.us/original/3c/95/3c956fce56dd1262cb509a0503154d1a.png)
2. Run `vipsheader -v --vips-info -a out.png`
3. Observe the differing interpretations between `8.14.5` and `8.15.1`

**Expected behavior**
The outputted image should have an `srgb` interpretation

**Actual behavior**
On version `8.15.0` and above `rgb16` is outputted

**Screenshots**
Here is `8.14.5` behavior (the version is printed in each screenshot)
![image](https://github.com/libvips/libvips/assets/29441515/35f89e28-727f-422e-8bdd-048f22269d05)
And here's the `8.15.1` behavior
![image](https://github.com/libvips/libvips/assets/29441515/c6830da0-4b31-4c25-9474-05aa37befeff)

**Environment**
 - OS: [Arch Kernel 6.7.0]
 - Vips: [tested with 8.14.5 (jxl disabled for this version), 8.15.0 and 8.15.1]

jcupitt:
Hi @turtletowerz,

I think the commit is correct (`icc_transform` should not turn 16 bit images to 8 by default), but I agree we need some more code in `vipsthumbnail` to allow for this change. We didn't think of this case. I'll have a go at a PR.

Thanks for reporting!

jcupitt:
I made a PR and credited you (hope that's OK). Let's move any discussion there.

PR #3831, opened 2024-01-27
thumbnail should always export as 8 bit

jcupitt:
After #f807d7b, `icc_export` and `icc_transform` would keep 16 bit images as 16-bit images. We didn't make a matching change to `thumbnail`, so 8.15 started making 16-bit thumbnails of 16-bit images, with no way to request 8 bits.

This PR adds `depth 8` to all `icc_*()` to make the behaviour match 8.14 (and tries to improve line breaks).

See https://github.com/libvips/libvips/issues/3830

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,7 @@ date-tbd 8.15.2
 
 - fix deflate compression of tiff pyramids [manthey]
 - better no-chroma-subsample switching for jpeg in tiff [kleisauke]
+- thumbnail always writes 8-bit thumbnails [turtletowerz]
 
 18/12/23 8.15.1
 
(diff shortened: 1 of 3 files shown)


jcupitt:
Testing and comments welcome! I think this covers all cases (ie. we will *always* write an 8-bit thumbnail), but I'm often wrong, of course.

turtletowerz:
Testing this branch on my machine confirms that the old functionality has been restored. I can't think of any reason thumbnails should have an option to set depth so I think this change is appropriate. I appreciate the quick reply and fix!

kleisauke (approved):
LGTM! 

(I changed the target branch to [`8.15`](https://github.com/libvips/libvips/tree/8.15) to ensure its inclusion in v8.15.2)

jcupitt:
Ooops! Yes, I meant to target 8.15, thanks. I'll update master too.

PR #3834, opened 2024-01-28
heifload: don't warn on images with nclx profiles

kleisauke:
Since this message is emitted during an AVIF roundtrip, its severity does not warrant a warning. So, let's downgrade it to a `g_info()`.

See: #2734 and kleisauke/wasm-vips#62.

Targets the 8.15 branch.


jcupitt:
Good idea, the warning was annoying.

Issue #3835, opened 2024-01-29
Attempting to vertically arrayjoin a large amount of images results in a segfault


TheEssem:
## Questions, enhancements, tips, etc.

Please use libvips discussions for other topics.

https://github.com/libvips/libvips/discussions

## Bug report

**Describe the bug**
It appears that trying to vertically append many images in an array that have been cropped/modified from an original source image with many similar pages results in libvips crashing with a segfault.

**To Reproduce**
Run the following Python script (`./mirror.py input.gif output.gif`) with the following image as input:
![flashtest](https://github.com/libvips/libvips/assets/21112968/75baecd0-80f6-4e40-96a1-188ba5f51fb5)
```py
#!/usr/bin/python3

import sys
import pyvips

original = pyvips.Image.new_from_file(sys.argv[1] + "[n=-1]")
page_height = original.get_page_height()
n_pages = original.get_n_pages()

img = []
for i in range(0, n_pages):
  x = i * page_height
  cropped = original.crop(0, x, original.width, page_height / 2)
  flipped = cropped.flip('vertical')
  img.append(cropped)
  img.append(flipped)

image = pyvips.Image.arrayjoin(img, across=1)
image.set('page-height', page_height - (page_height % 2 if 1 else 0))
image.write_to_file(sys.argv[2])
```

**Expected behavior**
The script should output a GIF with the top half of each frame mirrored onto the bottom half (or, in the case of the provided image, a nearly identical output compared to the input).

**Actual behavior**
The script will exit with a "segmentation fault (core dumped)" error.

**Environment**
(please complete the following information)
 - OS: EndeavourOS Linux (6.6.12-273-tkg-linux-tkg-eevdf-generic_v3)
 - Vips: 8.15.1

**Additional context**
I've provided a Python script for testing, but I've tested this using the C++ library and it occurs there as well.

jcupitt:
I found a workaround:

```
$ VIPS_DISC_THRESHOLD=1g ./mirror.py ~/pics/flash.gif x.gif
$
```

It looks like a bug in the way libvips is flipping between memory and temporary files for intermediates.

jcupitt:
I think found a proper fix, let's move any discussion and testing to the PR.

PR #3839, opened 2024-01-29
fix arrayjoin with some pipelines

jcupitt:
If the operation before arrayjoin was something that just forwarded a pointer, we could be left with dangling memory references in some cases.

Also some tiny comment adjustment and formatting changes.

thanks TheEssem

see https://github.com/libvips/libvips/issues/3835

### DIFF
--- a/libvips/conversion/arrayjoin.c
+++ b/libvips/conversion/arrayjoin.c
@@ -6,6 +6,8 @@
  * 	- minmise inputs once we've used them
  * 29/12/22
  *	- much faster with large arrays
+ * 29/1/24
+ *	- render and don't forward pixels for complete subregions
  */
 
 /*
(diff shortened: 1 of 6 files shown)


jcupitt:
Test program:

```python
#!/usr/bin/python3

import sys
import pyvips

original = pyvips.Image.new_from_file(sys.argv[1])
page_height = original.get_page_height()
n_pages = original.get_n_pages()

img = []
for i in range(0, n_pages):
  x = i * page_height
  cropped = original.crop(0, x, original.width, page_height // 2)
  flipped = cropped.flip('vertical')
  img.append(cropped)
  img.append(flipped)

image = pyvips.Image.arrayjoin(img, across=1)
image.set('page-height', 2 * (page_height // 2))
image.write_to_file(sys.argv[2])
```

Run with eg.:

```
$ ./mirror.py ~/pics/3198.gif[n=-1] x.gif
```

With larger GIFs (over 100mb fully decompressed) you'll get a segv without this PR.

libvips/conversion/arrayjoin.c
```diff
@@ -112,16 +114,24 @@ vips_arrayjoin_gen(VipsRegion *out_region,
...
 		i = VIPS_MIN(n - 1, left + top * join->across);
 
-		reg = vips_region_new(in[i]);
+		/* The part of in[i] we need.
+		 */
+		need = out_region->valid;
+		need.left -= join->rects[i].left;
+		need.top -= join->rects[i].top;
 
-		if (vips__insert_just_one(out_region, reg,
```
kleisauke, line 119: It looks like `vips__insert_just_one()` can be made static after this.
https://github.com/libvips/libvips/blob/3698b934969b00e8d7497186d64f8f18a49b68cf/libvips/conversion/insert.c#L118-L123

kleisauke (approved):
LGTM! I left a minor comment inline, but that can also be addressed after this PR lands.

libvips/conversion/arrayjoin.c
jcupitt: Ah good point, I'll do that too.

PR #3841, opened 2024-01-30
add some more notes to the dev checklist

jcupitt:
Sorry to ping you @kleisauke, I added some more dev notes, what do you think? You have much more experience around this than me! Can you think of any other related things we could add? 

The order of items should probably be adjusted now as well.

Maybe there should be a section just for web server image handling?

### DIFF
--- a/doc/Developer-checklist.md
+++ b/doc/Developer-checklist.md
@@ -139,13 +139,18 @@ $
    on uploading in this terrible format, push them to a separate low-priority
    queue on a special container. Keep them away from your main image path.
 
-## Use a different memory allocator
+## Linux memory allocator
 
-Consider using a different memory allocator, such as jemalloc, to prevent
-memory fragmentation in long-running, multi-threaded, glibc-based Linux
-processes. See [Leak on Linux?](https://github.com/lovell/sharp/issues/955) 
-to learn more about memory allocation and why the choice of a memory
-allocator is important.
+The default memory allocator on most glibc-based Linux systems (e.g.
+Debian, Red Hat) is unsuitable for long-running, multi-threaded processes
+that involve lots of small memory allocations.
+
+To help avoid fragmentation and improve performance on these systems,
+the use of an alternative memory allocator such as [jemalloc](
+https://github.com/jemalloc/jemalloc) is recommended.
+
+Those using musl-based Linux (e.g. Alpine) and non-Linux systems are
+unaffected.
 
 ## Disable the libvips operation cache if you don't need it
 
--- a/doc/Developer-checklist.md
+++ b/doc/Developer-checklist.md
@@ -87,57 +87,40 @@ If you can, aim for long pipelines of processing operations.
 If you can, put large resizes right at the start (see `thumbnail` above),
 then area filters (sharpen, for example), and finally any point operations.
 
-## Disable or adjust imagemagick
+## Only enable the load libraries you need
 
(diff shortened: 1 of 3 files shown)


kleisauke (approved):
Sorry for the delay, LGTM!

Regarding web server image handling, perhaps a section on addressing memory fragmentation and sanity checks for image dimensions would be useful? I wrote some notes at https://github.com/kleisauke/net-vips/issues/96#issuecomment-706692247 some time ago.

jcupitt:
Ah nice! I've tried to incorporate your notes, does it look ok?

doc/Developer-checklist.md
```diff
@@ -87,3 +87,67 @@ If you can, aim for long pipelines of processing operations.
...
+   on uploading in this terrible format, push them to a separate low-priority
+   queue on a special container. Keep them away from your main image path.
+
+## Use a different memory allocator
+
+Consider using a different memory allocator, such as jemalloc, to prevent
+memory fragmentation in long-running, multi-threaded, glibc-based Linux
+processes. See [Leak on Linux?](https://github.com/lovell/sharp/issues/955) 
+to learn more about memory allocation and why the choice of a memory
+allocator is important.
```
kleisauke, line 148: Perhaps rephrasing this as:
```suggestion
## Linux memory allocator

The default memory allocator on most glibc-based Linux systems (e.g.
Debian, Red Hat) is unsuitable for long-running, multi-threaded processes
that involve lots of small memory allocations.

To help avoid fragmentation and improve performance on these systems,
the use of an alternative memory allocator such as [jemalloc](
https://github.com/jemalloc/jemalloc) is recommended.

Those using musl-based Linux (e.g. Alpine) and non-Linux systems are
unaffected.
```

Would be better? 
(borrowed from https://sharp.pixelplumbing.com/install#linux-memory-allocator)

doc/Developer-checklist.md
```diff
@@ -87,3 +87,67 @@ If you can, aim for long pipelines of processing operations.
...
+
+## Disable the libvips operation cache if you don't need it
+
+The libvips' operation cache is not useful for image proxies (i.e.  processing
```
kleisauke, line 152:
```suggestion
The libvips' operation cache is not useful for image proxies (i.e. processing
```

doc/Developer-checklist.md
```diff
@@ -87,3 +87,67 @@ If you can, aim for long pipelines of processing operations.
...
+   is in memory, which prevents any streaming processing and hugely increases
+   memory use. For example:
+
+```
+$ /usr/bin/time -f %M:%e vipsthumbnail big-progressive.jpg
+3732224:4.23
+$ vips copy big-progressive.jpg x.jpg
+$ /usr/bin/time -f %M:%e vipsthumbnail x.jpg
+72448:0.26
+$
```
kleisauke, line 131: This stray cmd line can be removed
```suggestion
```
(perhaps this entire cmd block should also be indented with 3 spaces?)

kleisauke:
We might also need to regenerate the Docbook file at [`doc/Developer-checklist.xml`](https://github.com/libvips/libvips/blob/master/doc/Developer-checklist.xml) after this PR lands. It's a bit unfortunate that this is necessary. :(

jcupitt:
I updated all the doc xml too, thanks for pointing that out.

OK, merging!

Issue #3857, opened 2024-02-16
`exif.c` has a 256 character limit for some values


jcupitt:
### Discussed in https://github.com/libvips/libvips/discussions/3853

<div type='discussions-op-text'>

<sup>Originally posted by **MarcosAndre** February 15, 2024</sup>
Hello,
I'm using the vips_image_get_as_string to extract copyright from the images, but there are some values that are truncated when the they are too long. I think that the [vips_image_get_as_string](https://github.com/libvips/libvips/blob/67647a6968c263d90f508346cb5c9ac7f3ddb13d/libvips/foreign/exif.c#L293) function only gets the first 256 characters. 

Is it possible to get the entire string using vips? Can you increase the string length?

Thank you in advance.
Best regars,
Marcos André 

</div>

jcupitt:
For example, with this PNG suggested by @MarcosAndre:

![img-with-lots-of-metadata](https://github.com/libvips/libvips/assets/580843/18c82936-f4a0-449a-8795-4f56619c5450)

```
$ vipsheader -f exif-ifd0-XPComment img-with-lots-of-metadata.png | wc
     13      34     252
```

We see 252 bytes, but there should be 292.

Swapping all the 256 in `exif.c` for eg. 4096 fixes this, but we should probably do something better.

jcupitt:
I made a PR which adds long exif value support for 8.16, let's move discussion there.

https://github.com/libvips/libvips/pull/3858

MarcosAndre:
Sorry for the late feedback, I've tested the new version and it is working! Thank you so much @jcupitt 🙏🏼

PR #3858, opened 2024-02-18
support long exif values

jcupitt:
We were using static fixed length buffers to render i18n strings from libexif. Fields like "copyright" and "artist" can have extremely long values, and these were being truncated.

This PR adds g_auto as a hard requirement for libvips. This is supported by all recent versions of clang and gcc, so hopefully it's OK. openslide, for example, requires this.

See https://github.com/libvips/libvips/issues/3857

Thanks MarcosAndre

### DIFF
--- a/libvips/foreign/exif.c
+++ b/libvips/foreign/exif.c
@@ -87,7 +87,7 @@ entry_to_s(ExifEntry *entry)
 	 */
 	int size = VIPS_MIN(entry->size, 10000);
 	int max_size = size * 5;
-	g_autofree char *text = VIPS_MALLOC(NULL, max_size + 1);
+	char *text = VIPS_MALLOC(NULL, max_size + 1);
 
 	// this renders floats as eg. "12.2345", enums as "Inch", etc.
 	exif_entry_get_value(entry, text, max_size);
@@ -97,7 +97,11 @@ entry_to_s(ExifEntry *entry)
 	if (entry->format == EXIF_FORMAT_ASCII)
 		text[size] = '\0';
 
-	return g_utf8_make_valid(text, -1);
+	char *utf8 = g_utf8_make_valid(text, -1);
+
+	g_free(text);
+
+	return utf8;
 }
 
 #ifdef DEBUG_VERBOSE
@@ -133,7 +137,7 @@ show_tags(ExifData *data)
 static void
 show_entry(ExifEntry *entry, void *client)
 {
-	g_autofree char *text = entry_to_s(entry);
+	char *text = entry_to_s(entry);
 
 	printf("%s", exif_tag_get_title(entry->tag));
 	printf("|");
@@ -143,6 +147,8 @@ show_entry(ExifEntry *entry, void *client)
 	printf("|");
 	printf("%d bytes", entry->size);
 	printf("\n");
+
+	g_free(text);
 }
 
 static void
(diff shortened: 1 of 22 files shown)


jcupitt:
@kleisauke, @lovell this boring PR fixes a tiny thing in our terrible exif handling, which ought to be replaced anyway, but it also adds `g_auto` as a hard requirement. What do you think?

`g_auto` is the glib wrapper around the gcc "cleanup" attribute. It lets you write code like this:

```C
{
    g_autoptr(VipsImage) image = vips_image_new_from_file("xxx", NULL);

   ... do something with image
}
```

And on all exit paths from the block, `image` will be automatically unreffed. If you have a path where you don't unref (eg. successful return?) you can use eg. `return g_steal_pointer(&image);`. There's no performance cost.

I've not used it for much in this PR, but I think it could be a useful improvement in safety, readability and leak correctness. We could perhaps phase in use of this feature as we work on bits of libvips over the next few years.

gcc and clang support g_auto, but I think the intel compiler does not. I imagine emscripten does (being based on llvm), but I've not checked. 

Any thoughts?

libvips/foreign/exif.c
```diff
@@ -78,6 +78,27 @@
 
 #include "pforeign.h"
 
+// render an entry as a utf8 string
+static char *
+entry_to_s(ExifEntry *entry)
+{
+	// some extra space for conversion to string ... this can be quite a bit
+	// for formats like float
+	size_t max_size = entry->size * 5;
```
lovell, line 87: Can we trust `entry->size`? Perhaps worth adding a limit of our own.

lovell (commented):
Use of `g_auto` would rule out MSVC support; I'm unsure if this is a problem. (We did some work last year as part of https://github.com/libvips/libvips/issues/3378 to allow it.)

jcupitt:
> Use of g_auto would rule out MSVC support; I'm unsure if this is a problem.

Ah I hadn't thought of that. Maybe this is a showstopper.

libvips/foreign/exif.c
jcupitt: Good idea!

kleisauke:
Dropping MSVC support would mean that [ConanCenter](https://conan.io/center/recipes/libvips) would stop distributing libvips for Windows. I'm unsure how many users would be affected by this.

I think that's also the reason why GLib/GTK doesn't use `g_auto` internally, expect for the `G_DEFINE_AUTOPTR_CLEANUP_FUNC` and related macros (this allows dependents to still use `g_auto` and friends).

jcupitt:
Yes, I should have though of that, I'm actually [on a conan bug now](https://github.com/conan-io/conan/issues/15578) and hadn't made the connection haha.

I'll remove the `g_auto` stuff, but leave those few `G_DEFINE_AUTOPTR_CLEANUP_FUNC` definitions, they are handy for libs and apps (eg. nip4, vipsdisp) downstream from libvips.

libvips/iofuncs/type.c
```diff
@@ -217,6 +217,14 @@ vips_area_unref(VipsArea *area)
 		g_mutex_unlock(area->lock);
 }
 
+/* autoptr needs typed versions of functions for free.
+ */
+void
+VipsArrayDouble_unref(VipsArrayDouble *array)
+{
+    vips_area_unref(VIPS_AREA(array));
```
kleisauke, line 225:
```suggestion
	vips_area_unref(VIPS_AREA(array));
```
This would (hopefully) appease clang-format.

kleisauke (approved):
LGTM!

jcupitt:
Thanks!

PR #3860, opened 2024-02-19
Highway improvements

kleisauke:
Commit 4246c062f290f1c02b54dfd769b37254b4a6f843 removes the restriction of not crossing the 128-bit block boundaries on RVV/SVE (as mentioned in https://github.com/libvips/libvips/pull/3618#discussion_r1345598849).

Commit 4c32343894c15e2132cb354ce018f2757a2c56c5 prefers use of `Dup128VecFromValues` instead of `LoadDup128` to improve the codegen on some targets.

Both improvements are conditionally guarded for Highway 1.1.0 to avoid having to bump our minimum required version.


jan-wassenberg:
Nice :)

jcupitt (approved):
LGTM!

PR #3862, opened 2024-02-19
Prefer use of portable/thread-safe `g_strerror()`

kleisauke:


jcupitt:
You're right, I should have swapped those too. Thanks!

PR #3864, opened 2024-02-20
tiffsave: ensure large file support (>2GB) on MSVC

kleisauke:
On MSVC, `off_t` and `lseek()` are limited to signed 32-bit offsets, which limits their capability to handle files larger than 2GB.

To support saving TIFF files that exceed this size, replace their use with `gint64` and `vips__seek()`, where the latter uses `_lseeki64()` internally.

See: https://github.com/conan-io/conan/issues/15578.

Targets the 8.15 branch.


jcupitt (approved):
Great!

jcupitt:
Wow I must have been asleep when I added target seek. Why didn't I just copy the source seek stuff? Ah well, nice job fixing this.

jcupitt:
I updated master too.

Issue #3880, opened 2024-03-07
jxlsave fails with "jxlsave: error JxlEncoderSetBasicInfo" for a 32-bit (?) CMYK colour TIFF


JinEnMok:
## Bug report

Trying to convert any of the TIFF images from NASA's ["Visions of the Future" gallery](https://www.jpl.nasa.gov/galleries/visions-of-the-future) fails with `jxlsave: error JxlEncoderSetBasicInfo`. Since this seems to be a 32-bit colour (see below for ImageMagick's identify) image, this appears to be an issue not dissimilar to  #2830, although that one's marked resolved.

`pngsave` and `jpegsave` appear to work as intended on the same file.

**To Reproduce**
1. Use any of the TIFF images from ["here"](https://www.jpl.nasa.gov/galleries/visions-of-the-future), the smallest one (24MB) is [this](https://nasa-external-ocomm.app.box.com/s/ef3zm7plvks3q0fdba9nkd1leold4j3i)
2. try `vips jxlsave` with any quality/effort setting
3. See error

**Expected behavior**
VIPS converts the TIFF to JXL without any hiccups.

**Actual behavior**
VIPS produces an error (`jxlsave: error JxlEncoderSetBasicInfo`) and converts nothing.

**Environment**
 - OS: Windows 10/[Arch]Linux
 - Vips: 8.15.1
 - libjxl: 0.10.1

<details>
<summary>
<code>vipsheader --all </code>
</summary>

.\kepler186f.tif: 8175x11775 uchar, 4 bands, cmyk, tiffload
width: 8175
height: 11775
bands: 4
format: uchar
coding: none
interpretation: cmyk
xoffset: 0
yoffset: 0
xres: 11,811
yres: 11,811
filename: .\kepler186f.tif
vips-loader: tiffload
n-pages: 1
icc-profile-data: 557168 bytes of binary data
xmp-data: 14642 bytes of binary data
iptc-data: 220 bytes of binary data
photoshop-data: 8400 bytes of binary data
resolution-unit: in
bits-per-sample: 8
orientation: 1

</details>

<details>
<summary><code>magick identify -verbose </code></summary>

```
Image:
  Filename: .\kepler186f.tif
  Permissions: rw-rw-rw-
  Format: TIFF (Tagged Image File Format)
  Mime type: image/tiff
  Class: DirectClass
  Geometry: 8175x11775+0+0
  Resolution: 300x300
  Print size: 27.25x39.25
  Units: PixelsPerInch
[...]

PR #3883, opened 2024-03-08
Add jxl cmyk support

jcupitt:
I had a quick go, but it fails with:

```
$ vips copy cmyk.jpg x.jxl
attaching 961644 bytes of ICC
attaching exif-data ..
JxlBasicInfo:
    have_container = 0
    xsize = 1450
    ysize = 2048
    bits_per_sample = 8
    exponent_bits_per_sample = 0
    intensity_target = 0
    min_nits = 0
    relative_to_max_display = 0
    linear_below = 0
    uses_original_profile = 0
    have_preview = 0
    have_animation = 0
    orientation = 1
    num_color_channels = 3
    num_extra_channels = 1
    alpha_bits = 8
    alpha_exponent_bits = 0
    alpha_premultiplied = 0
    preview.xsize = 0
    preview.ysize = 0
    animation.tps_numerator = 10
    animation.tps_denominator = 1
    animation.num_loops = 0
    animation.have_timecodes = 0
JxlPixelFormat:
    num_channels = 4
    data_type = JXL_TYPE_UINT8
    endianness = 0
    align = 0
JxlEncoderFrameSettings:
    tier = 0
    distance = 2.35
    effort = 7
    lossless = 0
./lib/jxl/encode.cc:820: Extra channel 0 is not initialized
JXL_ENC_ERROR
```

ie. we are passing a 4 channel buffer with 1 extra channel. 

We use `JxlEncoderSetExtraChannelInfo()` to tag the extra channel as black. I'd think this ought to work, how do we encode an interleaved CMYK image?

### DIFF
--- a/libvips/foreign/jxlsave.c
+++ b/libvips/foreign/jxlsave.c
@@ -34,8 +34,8 @@
  */
 
 /*
-#define DEBUG
  */
+#define DEBUG
 
 #ifdef HAVE_CONFIG_H
 #include <config.h>
@@ -251,12 +251,8 @@ vips_foreign_save_jxl_print_status(JxlEncoderStatus status)
 		printf("JXL_ENC_NEED_MORE_OUTPUT\n");
 		break;
 
-	case JXL_ENC_NOT_SUPPORTED:
-		printf("JXL_ENC_NOT_SUPPORTED\n");
-		break;
-
 	default:
-		printf("JXL_ENC_<unknown>\n");
+		printf("JXL_ENC_<unknown %d>\n", status);
 		break;
 	}
 }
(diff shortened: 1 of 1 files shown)


jcupitt:
This pr is for https://github.com/libvips/libvips/issues/3880

jcupitt:
I asked on the libjxl discord and they say CMYK needs to be encoded as CMY plus a separate K channel, so we'll need quite a bit of code to break off the K and compress the extra channel. Let's park this PR for now.

jcupitt:
Thinking a little more, we'll need to:

- render the whole CMYK (and possibly alpha?) image to a huge memory array
- break it into separate CMY and KA arrays
- encode the CMY, then encode the KA

So that's at least 2x the memory use. We should probably do progressive write before returning to this PR, we might be able to avoid the extra huge memory allocations.

kleisauke:
I remembered a similar issue when loading a CMYK + Alpha image, see: https://github.com/libvips/libvips/pull/3623#issuecomment-1758141026.

Furthermore, it looks like `cjxl` also fails to encode CMYK images: https://github.com/libjxl/libjxl/issues/2126.

PR #3884, opened 2024-03-09
Make `vips_addalpha` a VipsOperation

RiskoZoSlovenska:
This PR turns `vips_addalpha` into a fully-fledged VipsOperation so that bindings may call it using the same mechanisms they use to call other operations. See https://github.com/libvips/lua-vips/pull/46#issuecomment-1983202875 for a discussion as to why this may be desired.

I'm largely new to C and the GObject system, so I'm not sure whether this is the correct (let alone the best) way to implement this change. For example, I'm not sure whether the new operation should subclass VipsBandary or VipsConversion — on one hand, it is an operation of the same type as the other bandary operations (it simply manipulates bands); on the other, it doesn't use any of the machinery defined by VipsBandary and subclassing VipsConversion would be sufficient.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -13,6 +13,7 @@ master 8.16
 - add configurable max coordinate and vips_max_coord_get()
 - improve kill handling
 - PFM save and load now uses scRGB (ie. linear 0-1) [NiHoel]
+- turn `vips_addalpha` into a VipsOperation [RiskoZoSlovenska]
 
 date-tbd 8.15.2
 
--- a/libvips/conversion/addalpha.c
+++ b/libvips/conversion/addalpha.c
@@ -1,0 +1,122 @@
+/* Add an appropriate alpha band.
+ */
+
+/*
+
+	This file is part of VIPS.
+
+	VIPS is free software; you can redistribute it and/or modify
+	it under the terms of the GNU Lesser General Public License as published by
+	the Free Software Foundation; either version 2 of the License, or
+	(at your option) any later version.
+
+	This program is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU Lesser General Public License for more details.
+
+	You should have received a copy of the GNU Lesser General Public License
+	along with this program; if not, write to the Free Software
+	Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+	02110-1301  USA
+
+ */
+
+/*
+
+	These files are distributed with VIPS - http://www.vips.ecs.soton.ac.uk
+
+ */
+
+#ifdef HAVE_CONFIG_H
+#include <config.h>
+#endif /*HAVE_CONFIG_H*/
+#include <glib/gi18n-lib.h>
+
+#include <stdio.h>
+#include <stdlib.h>
+
+#include <vips/vips.h>
+#include <vips/internal.h>
+
+#include "pconversion.h"
+
+typedef struct _VipsAddAlpha {
+	VipsConversion parent_instance;
+
+	VipsImage *in;
+
+} VipsAddAlpha;
+
+typedef VipsConversionClass VipsAddAlphaClass;
+
+G_DEFINE_TYPE(VipsAddAlpha, vips_addalpha, VIPS_TYPE_CONVERSION);
+
+static int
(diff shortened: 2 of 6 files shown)


jcupitt:
Hi again, thanks for doing this. 

- I'd make it a separate source code file since it's a new class
- It's maybe most like `falsecolour` (that just calls a few libvips operations too), so maybe subclass `conversion`
- add a line to changelog as well and credit yourself

RiskoZoSlovenska:
Alright. Just a quick question: I've noticed that some functions (e.g. `vips_falsecolour_build`) use a macro like `VIPS_CONVERSION()`, while others (e.g. `vips_bandjoin_const_buffer`) use a cast like `(VipsConversion *)`. Is there any difference, and which one should I use?

RiskoZoSlovenska:
Actually, got a couple more questions:
* Should I keep the `VIPS_DEBUG_MSG("vips_addalpha_class_init\n");` call?
* Do I need to do anything else? Looking at `falsecolour` as an example, I see that it pops up in files such as `VImage8.h`, `vips7compat.c` and `POTFILES.in`, among others. Do I need to do something similar with `addalpha`? If so, what exactly?

jcupitt:
> VIPS_CONVERSION(), while others (e.g. vips_bandjoin_const_buffer) use a cast like (VipsConversion *)

Just laziness and history. Recent glib type macros define PACKAGE_TYPENAME() macros for you, but you used to have to write a block of boilerplate in a header file.

The best thing now is to use `G_DECLARE_FINAL_TYPE()` to emit the header parts of a type definition, and `G_DEFINE_FINAL_TYPE()` to emit the code parts. libvips ought to use these two everywhere, but does not yet.

For your operation, you can put both of them at the top of the file, since you don't want the "classiness" of `addalpha` to be visible elsewhere.

> Do I need to do anything else? 

You need to init the class in `conversion/conversion.c` by calling the `_get_type()` function (this will indirectly call `_class_init()` and add it to the gobject class hierarchy), and of course you'll need to add the filename to `conversion/meson.build`. You'd normally add a decl for the C API to `include/vips/conversion.h`, but in this case that's already present, of course.

Everything else (docs, C++ headers, i18n, etc.) is generated automatically by various scripts which we run by hand occasionally, no need to do that until we get close to a release.

> printf() debugging calls

I do this sometimes to get a trace of execution (can be very handy for seeing races etc.), but there's no need for this operation since it's so simple.

RiskoZoSlovenska:
> The best thing now is to use `G_DECLARE_FINAL_TYPE()` to emit the header parts of a type definition including the cast macros, and `G_DEFINE_FINAL_TYPE()` to emit the code parts. libvips ought to use these two everywhere, but does not yet.

I've had a go at trying to use these, but the compiler started complaining about an undefined `glib_autoptr_clear_VipsConversion`. I tried using a similar macro (`G_DECLARE_DERIVABLE_TYPE()`) in `VipsConversion`, but then things started breaking. If I understand it correctly, `G_DECLARE_DERIVABLE_TYPE()` makes it so that the public struct (`VipsConversion` in this case) cannot have additional public members (e.g. `out`). Either way, I am either severely misunderstanding how to use these macros, or this is a can of worms I'm not quite ready for yet.

jcupitt:
Ooop, sorry, bad advice, now I actually try `G_DEFINE_FINAL_TYPE`, I remember it makes you keep the struct private and that's why we didn't use it. 

Try just https://github.com/libvips/libvips/commit/4a8f5c5b610da26e08e9e3fe02facb97f5c7be69 ... although that macro is so trivial it's hardly worth it, a simple cast is arguably better.

RiskoZoSlovenska:
> Try just https://github.com/libvips/libvips/commit/4a8f5c5b610da26e08e9e3fe02facb97f5c7be69 ... although that macro is so trivial it's hardly worth it, a simple cast is arguably better.

For that to work, we'd also have to define `VIPS_TYPE_ADDALPHA`. Doing that isn't hard, but the boilerplate starts to add up. I'm more inclined to keep the simple cast.

RiskoZoSlovenska:
@jcupitt Is this ready to be merged?

jcupitt:
Ooop, sorry, I  was looking at other stuff. Just a line in the changelog left to do.

RiskoZoSlovenska:
That's done, unless you'd prefer it in a separate commit?

jcupitt:
Ooop, sorry, I scrolled right past it for some reason.

Thanks for doing this!

Issue #3885, opened 2024-03-09
Conversion to PFM gives incorrect output


NiHoel:
**Describe the bug**
Converting a floating point linear SRGB image (e.g. EXR or HDR) to PFM results in a file with incorrect orientation (e.g. bottom left corner is top right), clipped highlights and squeezed colors (rougly multiplied by a factor of 2^-20)

**To Reproduce**
1. Use Image https://drive.google.com/file/d/1I9J3iyCnu7LX1K9eXAz_dF6cjq1rJU9t/view?usp=sharing or any other EXR or HDR image file (preferably with intensity values > 1).
2. Run `vips copy sun.hdr sun.pfm`
3. Open sun.pfm in gimp. Everything is just white. By lowering exposure some of the original image becomes recognizable.

The indicated byte order is incorrect: The file is written big-endian but indicates little-endian encoding (-1 in third line). Changing the sign in a hex editor fixes the issue.

**Screenshot**
![grafik](https://github.com/libvips/libvips/assets/28017252/bc05f09d-1769-4652-8741-b762e5d603ac)
left: vips produced file including fix (marked location); right: original pfm file (before conversion to hdr using gimp and to pfm using vips)

**Environment**
 - OS: Win10, i7-9700
 - Vips: 8.15.1

jcupitt:
Hi @NiHoel,

You're right, `ppmsave` treats PFM as regular sRGB, just with float pixels. It should use linear 0-1 (scRGB). I'll make a PR.

Thanks for the report!

jcupitt:
I think this is fixed, I'll close.

PR #3887, opened 2024-03-10
Improve scrgb handling for ppmsave

jcupitt:
Saving a scrgb image as `x.pfm` ought to save as scrgb.

Referring issue https://github.com/libvips/libvips/issues/3885

See also https://github.com/libvips/libvips/pull/3623 for when we improved scrgb for `tiffsave`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,7 +116,9 @@ jobs:
         run: meson compile -C build
 
       - name: Check libvips
-        run: meson test -C build --print-errorlogs
+        run: |
+          meson test -C build \
+          || (cat build/meson-logs/testlog.txt && exit 1)
 
       - name: Install libvips
         run: sudo meson install -C build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,7 +110,9 @@ jobs:
         run: meson compile -C build
 
       - name: Check libvips
-        run: meson test -C build
+        run: |
+          meson test -C build \
+          || (cat build/meson-logs/testlog.txt && exit 1)
 
       - name: Install libvips
         run: sudo meson install -C build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,9 +116,7 @@ jobs:
         run: meson compile -C build
 
       - name: Check libvips
-        run: |
-          meson test -C build \
-          || (cat build/meson-logs/testlog.txt && exit 1)
+        run: meson test -C build --print-errorlogs
 
       - name: Install libvips
         run: sudo meson install -C build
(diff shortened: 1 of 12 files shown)


jcupitt:
OK, I think this is done. @NiHoel, I credited you in the changelog, I hope that's OK.

Would you have time to test it?

NiHoel:
It will take some time to test since I need to build it from source my Windows computer.

jcupitt:
OK, I'll beef up the HDR tests and merge. Thanks again for reporting this thing.

Could I add your `sun.hdr` image to the libvips test suite? It'd be handy for testing the various HDR paths.

.github/workflows/ci.yml
```diff
@@ -116,7 +116,9 @@ jobs:
         run: meson compile -C build
 
       - name: Check libvips
-        run: meson test -C build
+        run: |
+          meson test -C build \
+          || (cat build/meson-logs/testlog.txt && exit 1)
```
kleisauke, line 121: This can be simplified using the `--print-errorlogs` argument.
```suggestion
        run: meson test -C build --print-errorlogs
```

jcupitt: Oh nice, I didn't know about that.

jcupitt: ... I swapped it back, `--print-errorlogs` only prints the last 100 lines and we need more than that in many cases.

kleisauke: > `--print-errorlogs` only prints the last 100 lines and we need more than that in many cases.

Ah, indeed. It seems that also passing `-v` would print the entire log.
https://github.com/mesonbuild/meson/blob/1.3.2/mesonbuild/mtest.py#L660-L661

jcupitt:
Hey @kleisauke, can I pick your brains?

I've just realised that PFM images can have any byte alignment, since the header part of the file can be any size.

How will this interact with things like highway? I'm unsure how that handles alignment. Would padding the PFM header with extra bytes to force alignment give a speedup?

jcupitt:
... I added a thing to secretly pad ppm headers to get the data aligned properly.

kleisauke:
> How will this interact with things like highway? I'm unsure how that handles alignment.

The uchar-specific Highway paths handles the alignment in `buffer.c`:
https://github.com/libvips/libvips/blob/697ee32e9eaa26e58cd1cb55428ddee6b9fef0c4/libvips/iofuncs/buffer.c#L483-L493

Perhaps adding `im->BandFmt == VIPS_FORMAT_FLOAT` there would help (untested)? A better approach would be to let operations hint the required buffer alignment. This way, we don't have to allocate that extra  `alignment - 1` bytes in all cases (but perhaps this is negligible).

As an aside, it looks like Highway bumped `HWY_ALIGNMENT` from `64` to `128` in version 1.1.0.
https://github.com/google/highway/blob/1.1.0/hwy/aligned_allocator.h#L35-L39

> Would padding the PFM header with extra bytes to force alignment give a speedup?

Padding the PFM header with extra bytes to force alignment doesn't give a speedup in the Highway paths, as these paths are exclusively enabled for uchar images.

Issue #3876, opened 2024-03-05
Vips 8.15.1 build failure on MacOS x86_64


alucryd:
## Questions, enhancements, tips, etc.

Please use libvips discussions for other topics.

https://github.com/libvips/libvips/discussions

## Bug report

**Describe the bug**

My x86_64 MacOS builds started failing with vips 8.15.1, it was previously working fine with 8.14.5.

**To Reproduce**
1. Create an osxcross toolchain using MacOS SDK 10.15, targeting MacOS 10.14
2. Try building vips (and all its dependencies) with the toolchain
3. Notice some undefined symbols

**Expected behavior**
The build should complete like it did with 8.14.5.

**Actual behavior**
The build fails because of some undefined symbols. I'm not sure where they come from as I can't find any reference to them in my entire build environment.

**Screenshots**
```
FAILED: libvips/libvips.42.dylib 
[...]

PR #3888, opened 2024-03-10
Check linker supports target_clones attribute

lovell:
Fixes https://github.com/libvips/libvips/issues/3876

This allows libvips to be built with buggy versions of LLVM where the compiler but not linker supports this feature, and therefore prevents an "undefined symbol: __cpu_model" error at link time.

CI before:

- Linux/gcc: Checking if "Has target_clones attribute" compiles: YES 
- Linux/clang14: Checking if "Has target_clones attribute" compiles: YES 
- macOS/xcode14: Checking if "Has target_clones attribute" compiles: NO 

CI after:

- Linux/gcc: Checking if "Has target_clones attribute" : links: YES 
- Linux/clang14: Checking if "Has target_clones attribute" : links: YES 
- macOS/xcode14: Checking if "Has target_clones attribute" : links: NO 

Targets the 8.15 branch.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -8,6 +8,7 @@ date-tbd 8.15.2
 - fix arrayjoin with some pipelines [TheEssem]
 - fix high Q mono JPEG TIFF write with mozjpeg [cavenel]
 - tiffsave: ensure large file support (>2GB) on MSVC [kleisauke]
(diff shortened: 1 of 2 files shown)


jcupitt:
Nice!

PR #3908, opened 2024-03-23
Jpegload, put the max shrink value to 8 instead of 16

quentin-bettoum:
Hello,

The `jpegload` doc displayed in CLI states that the max shrink value is 16
```bash
   shrink       - Shrink factor on load, input gint
			default: 1
			min: 1, max: 16
```

Trying to put a shrink factor > 8 and <= 16 triggers a `error buffer: VipsFormatLoadJpeg: bad shrink factor 16`

From other parts of the doc, it seems that the max shrink factor should be 8.


jcupitt:
Hi @quentin-bettoum,

Thanks for pointing this out, it's been wrong for a loooong time, looking at git blame.

PR #3917, opened 2024-04-01
Increase EXIF size limit to 32Mb

DarthSim:
Hey there 👋

Some of our users showed us AI-generated images that have pretty large EXIF data (around 4 Mb). All-in-all the images are ok, and their EXIF can be decoded successfully with higher limits, so I'd offer to increase the EXIF size limit to 32 Mb.

### DIFF
--- a/libvips/foreign/exif.c
+++ b/libvips/foreign/exif.c
@@ -193,13 +193,13 @@ vips_exif_load_data_without_fix(const void *data, size_t length)
 	ExifData *ed;
 
 	/* exif_data_load_data() only allows uint for length. Limit it to less
-	 * than that: 2**20 should be enough for anyone.
+	 * than that: 2**23 should be enough for anyone.
 	 */
 	if (length < 4) {
 		vips_error("exif", "%s", _("exif too small"));
 		return NULL;
 	}
-	if (length > 1 << 20) {
+	if (length > 1 << 23) {
 		vips_error("exif", "%s", _("exif too large"));
 		return NULL;
 	}


libvips/foreign/exif.c
```diff
@@ -193,13 +193,13 @@ vips_exif_load_data_without_fix(const void *data, size_t length)
 	ExifData *ed;
 
 	/* exif_data_load_data() only allows uint for length. Limit it to less
-	 * than that: 2**20 should be enough for anyone.
+	 * than that: 2**25 should be enough for anyone.
```
lovell, line 196: Perhaps we should limit this to `2**23`, which is _only_ an ~8x increase on the current limit :sweat_smile: and will easily allow the example 4MB? I ask as the proposed `2**25` is ~32x the current limit.

fenos: Hi @lovell any thoughts about this?

lovell: @fenos No further thoughts from me beyond what I've provided above, happy to wait and see what other maintainers think.

jcupitt: Hello everyone, sorry, I've been on holiday.

My instinct is horror and revulsion at the thought of 32mb of EXIF data haha, but of course if there really are useful images like that in the wild, then we should support them.

Let's do 8mb for now, we can always raise it again.

jcupitt: > It doesn't seem to be a DDoS surface 

I suppose the issue would be that it could potentially add a lot of time and memory to the open-image budget. 

libvips has tried to make `_new_from_file()` a cheap operation, but if it is forced to parse 32mb of EXIF, it'll be quite a bit less cheap. I've not tried to measure this, perhaps libexif can do it very quickly?

@fenos do you have a link explaining what this huge exif data block is used for? Do they put lots of other metadata into exif comments?

DarthSim:
Alrighty, I lowered the limit to 8mb since everyone isn't brave (or crazy) enough to accept 32mb one 😅

lovell (approved):

jcupitt:
Thanks everyone!

Issue #3922, opened 2024-04-03
Subsample mode not working with jpegli


Snafuh:
Hello,

I've build windows binaries with the jpegli flag and was trying different setting combinations. 
The subsampling flag does not work correctly when using jpegli and there is no subsampling used when compressing images. File sizes are the same with VIPS_FOREIGN_SUBSAMPLE_ON, VIPS_FOREIGN_SUBSAMPLE_OFF and VIPS_FOREIGN_SUBSAMPLE_AUTO in various quality modes (tested 70, 80, 85,90) and imagemagick repots 444 for all images.

I've tested the prebuild windows binaries (mozjpeg) with the same settings, and OFF and ON are correctly working, and AUTO switches subsampling as expected depending on the Quality.

jpegli itself does not seem to feature and auto mode, but the cli tool shipped with the libjxl binaries supports the a _--chroma_subsampling_ parameter with 444|440|422|420 as valid parameters. These are all working correctly. 

**To Reproduce**
1. Build libvips with jpegli
2. Compress images via 
vips.exe resize test.png out-on.jpg[strip,Q=80,subsample_mode=VIPS_FOREIGN_SUBSAMPLE_ON] 1
vips.exe resize test.png out-auto.jpg[strip,Q=80,subsample_mode=VIPS_FOREIGN_SUBSAMPLE_AUTO] 1
vips.exe resize test.png out-off.jpg[strip,Q=80,subsample_mode=VIPS_FOREIGN_SUBSAMPLE_OFF] 1
4. Compare image size

**Environment**

 - OS: Windows 10
 - Vips: 8.15.2

From what I've understood, there were no adjustments made for jpegli as it is compatible as drop in replacement with mozjpeg/libjpeg-turbo. So I am not entirely sure how such differences would be handled by libvips, or if jpegli should conform to different standards. I'd be happy to open an issue over at libjxl if it's the latter. 

Cheers,
Kim

Snafuh:
From what I can see libvips sets  h_samp_factor and v_samp_factor only ever to 1. 
Mozjpeg uses 4:2:0 (h_samp_factor = 2; v_samp_factor  = 2) as default.
Jpegli might use 4:4:4 as default. Therefore we never set any other values in libvips and subsampling is always off.

[...]

PR #3924, opened 2024-04-04
jpegsave: explicitly use 4:2:0 subsampling by default

kleisauke:
Unless saving with a quality setting of 90 or higher, or when chroma subsampling is explicitly disabled (`subsample_mode=off`).

This ensures compatibility with jpegli, which uses 4:4:4 by default.

See: https://github.com/libvips/libvips/issues/3922.

Targets the 8.15 branch.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,7 @@
 26/3/24 8.15.3
 
 - fix dzsave of >8-bit images to JPEG
+- jpegsave: fix chrominance subsampling mode with jpegli [kleisauke]
 
 12/3/24 8.15.2
 
--- a/libvips/foreign/vips2jpeg.c
+++ b/libvips/foreign/vips2jpeg.c
@@ -667,12 +667,26 @@ set_cinfo(struct jpeg_compress_struct *cinfo,
 	if (progressive)
 		jpeg_simple_progression(cinfo);
 
-	if (subsample_mode == VIPS_FOREIGN_SUBSAMPLE_OFF ||
+	if (in->Bands != 3 ||
+		subsample_mode == VIPS_FOREIGN_SUBSAMPLE_OFF ||
 		(subsample_mode == VIPS_FOREIGN_SUBSAMPLE_AUTO &&
-			qfac >= 90)) {
-		int i;
+			qfac >= 90))
+		/* No chroma subsample.
+		 */
+		for (int i = 0; i < in->Bands; i++) {
+			cinfo->comp_info[i].h_samp_factor = 1;
+			cinfo->comp_info[i].v_samp_factor = 1;
+		}
+	else {
+		/* Use 4:2:0 subsampling, we must set this explicitly, since some
+		 * jpeg libraries do not enable chroma subsample by default.
+		 */
+		cinfo->comp_info[0].h_samp_factor = 2;
+		cinfo->comp_info[0].v_samp_factor = 2;
 
-		for (i = 0; i < in->Bands; i++) {
+		/* Rest should have sampling factors 1,1.
+		 */
+		for (int i = 1; i < in->Bands; i++) {
 			cinfo->comp_info[i].h_samp_factor = 1;
 			cinfo->comp_info[i].v_samp_factor = 1;
 		}
--- a/libvips/foreign/vips2tiff.c
+++ b/libvips/foreign/vips2tiff.c
@@ -682,21 +682,32 @@ wtiff_compress_jpeg_header(Wtiff *wtiff,
 
 	jpeg_set_defaults(cinfo);
 
-	// disable chroma subsample for high Q
-	if (wtiff->Q >= 90)
+	/* Set compression quality.
+	 */
+	jpeg_set_quality(cinfo, wtiff->Q, TRUE);
+
+	if (image->Bands != 3 ||
+		wtiff->Q >= 90)
+		/* No chroma subsample.
+		 */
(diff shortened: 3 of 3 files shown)


kleisauke:
FWIW, I haven't tested this properly yet, I'll do so later today.

kleisauke:
Marking as draft due to this error:
```
E           pyvips.error.Error: unable to write to image
E             VipsJpeg: lib\jpegli\encode.cc:293: Sampling is not supported for simgle component image.
```
https://github.com/libvips/libvips/blob/009afbc00388eb46320187f3b556afaf1236f151/test/test-suite/test_foreign.py#L159

FWIW, `test_jpegsave` seems to pass now, it failed previously with:
```
>       assert len(q10_subsample_off) > len(q10)
E       assert 9459 > 9459
E        +  where 9459 = <snip />
E        +  and   9459 = <snip />

test\test-suite\test_foreign.py:292: AssertionError
```
https://github.com/libvips/libvips/blob/009afbc00388eb46320187f3b556afaf1236f151/test/test-suite/test_foreign.py#L292

Snafuh:
I can confirm this fixes the issue with jpegli. ON/OFF is working as expected (4:2:0/4:4:4) and AUTO goes for 4:4:4 for Q>=90.

kleisauke:
> Marking as draft due to this error:
> 
> ```
> E           pyvips.error.Error: unable to write to image
> E             VipsJpeg: lib\jpegli\encode.cc:293: Sampling is not supported for simgle component image.
> ```
> 
> https://github.com/libvips/libvips/blob/009afbc00388eb46320187f3b556afaf1236f151/test/test-suite/test_foreign.py#L159

This is no longer an issue with the latest revision of this PR which uses only 4:2:0 subsampling for multiband images.

I also opened PR https://github.com/libjxl/libjxl/pull/3457 to mirror this logic in jpegli.

jcupitt:
```C
    if (subsample_mode == VIPS_FOREIGN_SUBSAMPLE_OFF ||
        (subsample_mode == VIPS_FOREIGN_SUBSAMPLE_AUTO &&
            qfac >= 90))
        for (int i = 0; i < in->Bands; i++) {
            cinfo->comp_info[i].h_samp_factor = 1;
            cinfo->comp_info[i].v_samp_factor = 1;
        }
    else if (in->Bands > 1) {
        /* Use 4:2:0 subsampling, but only if the image has more than
         * 1 band, since sampling factors are actually meaningless with
         * greyscale JPEG files.
         */
        cinfo->comp_info[0].h_samp_factor = 2;
        cinfo->comp_info[0].v_samp_factor = 2;

        /* Rest should have sampling factors 1,1.
         */
        for (int i = 1; i < in->Bands; i++) {
            cinfo->comp_info[i].h_samp_factor = 1;
            cinfo->comp_info[i].v_samp_factor = 1;
        }
    }
```

It looks odd to not have the bands=1 case covered.

How about:

```C
    if (subsample_mode == VIPS_FOREIGN_SUBSAMPLE_OFF ||
        in->Bands != 3 ||
        (subsample_mode == VIPS_FOREIGN_SUBSAMPLE_AUTO &&
            qfac >= 90))
        // no chroma subsample
        for (int i = 0; i < in->Bands; i++) {
            cinfo->comp_info[i].h_samp_factor = 1;
            cinfo->comp_info[i].v_samp_factor = 1;
        }
    else {
        // chroma subsample
        cinfo->comp_info[0].h_samp_factor = 2;
        cinfo->comp_info[0].v_samp_factor = 2;

        for (int i = 1; i < in->Bands; i++) {
            cinfo->comp_info[i].h_samp_factor = 1;
            cinfo->comp_info[i].v_samp_factor = 1;
        }
    }
```

That would make it clear what happens for CMYK JPEG as well.

Though looking at the libjpeg docs they say:

```
int h_samp_factor
int v_samp_factor
        Horizontal and vertical sampling factors for the component; must
        be 1..4 according to the JPEG standard.  Note that larger sampling
        factors indicate a higher-resolution component; many people find
        this behavior quite unintuitive.  The default values are 2,2 for
        luminance components and 1,1 for chrominance components, except
        for grayscale where 1,1 is used.

```

So the `jpeg_set_defaults(cinfo);` should have set 2,1,1 (for three bands images), 1 for greyscale, and 1,1,1,1 for CMYK. Doesn't that mean we don't need the `else{}` case here?

jcupitt:
Ah I think I see, the `else{}` case is for jpegli which strangely does not match libjpeg behaviour for default chroma subsampling.

kleisauke:
I've incorporated your suggestion. I'm not sure why jpegli doesn't match libjpeg's behaviour for default chroma subsampling. Perhaps that was done intentionally?

Of course, this PR might not be needed if jpegli uses 4:2:0 subsampling by default for 3 band images.

kleisauke:
> I also opened PR [libjxl/libjxl#3457](https://github.com/libjxl/libjxl/pull/3457) to mirror this logic in jpegli.

This PR has just been merged. I think the `in->Bands != 3` condition is still necessary for CMYK save.

jcupitt:
Ooop! Sorry I've been distracted with other stuff.

Thank you for fixing this!

kleisauke:
> Of course, this PR might not be needed if jpegli uses 4:2:0 subsampling by default for 3 band images.

https://github.com/libjxl/libjxl/pull/3507 :sweat_smile:

jcupitt:
I saw! But it's probably safest for us to set it anyway.

PR #3928, opened 2024-04-08
pngload: disable ADLER32/CRC checking in non-fail mode

kleisauke:
The condition was inverted for spng, also sync libpng load while we are here.

Targets the 8.15 branch.


lovell (approved):
:+1:

jcupitt:
Nice!

PR #3935, opened 2024-04-16
fix some int32 limits on windows

jcupitt:
write() on windows uses int for byte counts, leading to errors with very wide (more than 1.5m pixels wide) images.

libvips was assuming posix:

```C
ssize_t write(int fd, const void buf[.count], size_t count);
```

But windows has:

```C
int _write(int fd, const void *buffer, unsigned int count);
```

https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/write?view=msvc-170

We probably have some other int32 limits still :( we should make some test cases and do more investigation, especially on windows.

Referring discussion:

https://github.com/libvips/libvips/discussions/3918#discussioncomment-9125646

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -5,7 +5,8 @@
 - pngload: disable ADLER32/CRC checking in non-fail mode [kleisauke]
 - improve target_clones support check [kleisauke]
 - fix pipe read limit
-- fix a rare crash on windows in highly threaded applications [Julianiolo]
+- fix a rare crash on Windows in highly threaded applications [Julianiolo]
+- vipssave: fix infinite loop on Windows with large images [pdbourke]
 
 12/3/24 8.15.2
 
--- a/libvips/deprecated/package.c
+++ b/libvips/deprecated/package.c
@@ -323,7 +323,7 @@ static int
 getext_vec(im_object *argv)
 {
 	void **out = (void **) &argv[1];
-	int size;
+	size_t size;
 
 	/* void/char confusion is fine.
 	 */
--- a/libvips/foreign/rawsave.c
+++ b/libvips/foreign/rawsave.c
@@ -93,16 +93,12 @@ vips_foreign_save_raw_write(VipsRegion *region, VipsRect *area, void *a)
 {
 	VipsForeignSave *save = (VipsForeignSave *) a;
 	VipsForeignSaveRaw *raw = (VipsForeignSaveRaw *) a;
-	int i;
 
-	for (i = 0; i < area->height; i++) {
-		VipsPel *p =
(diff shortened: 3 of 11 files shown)


jcupitt:
`size_t` is always correct for the size of memory objects, but it's 32-bits on win32 32-bit builds, which means it's too small for the size of filesystem objects. 

`size_t` is 64 bits on 32-bit *nix systems as long as large file support is enabled (I think it always is).

Perhaps we should just use `gint64` for all functions related to the filesystem? It'd be (maybe?) less confusing.

kleisauke:
This looks good! I'll test this out on Windows later.

> We probably have some other int32 limits still

I think clang-tidy's [`cppcoreguidelines-narrowing-conversions`](https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/narrowing-conversions.html) check might catch such things. Here's the output of that check over the entire project:
https://gist.github.com/kleisauke/9025c3564d3aec17496bd21d675908fb

Though, that could also generate false-positives, for example:
https://github.com/libvips/libvips/blob/f1e1407c49acd4629c62f0fdbd980ab60a5b0c5f/libvips/foreign/archive.c#L180
```
../libvips/foreign/archive.c:180:33: warning: narrowing conversion from 'int' to signed type 'char' is implementation-defined [cppcoreguidelines-narrowing-conversions]
  180 |         char compression_string[2] = { '0' + compression, 0 };
      |                                        ^
```
`compression` is already limited to `[0,9]` when it reaches there.

kleisauke:
It looks like `vipssave` uses `write_vips` in `generate.c` instead.
https://github.com/libvips/libvips/blob/0964d602889a8de333e3e31cbf96706fcc29f851/libvips/iofuncs/generate.c#L626-L629

Applying the same fix on that function (see e.g. commit https://github.com/kleisauke/libvips/commit/af7c7bcb62b7e6636712b24e113c283f321d67f8) fixes the test case mentioned in https://github.com/libvips/libvips/discussions/3918#discussioncomment-9129000 on my Windows 11 PC.
```console
$ vips black x.v 1400000 10000 --bands 3 --vips-progress
vips x.v: 1400000 x 10000 pixels, 24 threads, 1400000 x 1 tiles, 512 lines in buffer
vips x.v: done in 18.5s
$ (Get-Item x.v).Length
42000002024
```

BTW, should this PR target the [`8.15`](https://github.com/libvips/libvips/tree/8.15) branch instead?

jcupitt:
Ah thank you for fixing this Kleis. It looks great, let's merge.

jcupitt:
I merged to master as well.

PR #3942, opened 2024-04-21
Rework rawsave and add rawsave_target

### LINKED ISSUE #3941: Add target write to `rawsave`
### Discussed in https://github.com/libvips/libvips/discussions/3940

<sup>Originally posted by **akash-akya** April 18, 2024</sup>
Does it makes sense to create `VipsForeignSaveRawTarget` to support streaming raw pixels in a unified way (`vips_image_write_to_target(in, ".raw", target)`)?
I know there is already `VipsForeignSaveRawFd` which can be used to create similar interface, but in our case we have a unified interface to do streaming writes and this would help us to avoid creating special case to write raw pixels.

If this makes sense then I can try implementing it</div>

akash-akya:
Fixes #3941

### DIFF
--- a/doc/gen-function-list.py
+++ b/doc/gen-function-list.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 
 # walk vips and generate a list of all operators and their descriptions
 # for docs
--- a/libvips/foreign/rawsave.c
+++ b/libvips/foreign/rawsave.c
@@ -121,6 +121,9 @@ vips_foreign_save_raw_build(VipsObject *object)
 		vips_sink_disc(save->in, vips_foreign_save_raw_block, raw))
 		return -1;
 
+	if (vips_target_end(raw->target))
+		return -1;
+
 	return 0;
 }
 
--- a/libvips/foreign/rawsave.c
+++ b/libvips/foreign/rawsave.c
@@ -7,6 +7,7 @@
  * 21/04/24
  * 	- reworked based on ppmsave.c
  * 	- added save to target
+ * 	- added save to buffer
  * 10/06/08 JF
  * 	- initial code based on im_vips2ppm()
  * 04/07/08 JF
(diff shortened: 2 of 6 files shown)


akash-akya:
@jcupitt need your help to review the changes.

I kept `vips_rawsave_fd` function but removed `rawsave_fd` operation in favor of `rawsave_target`. I am not sure if this breaks something in the downstream. Do we have to retain the operation?

How to regenerate `doc/function-list.xml` file?

jcupitt:
It's probably OK to remove `rawsave_fd`, I doubt if anyone was using it. We'll need to keep the C API point `vips_rawsave_fd()` to avoid link failures, but we can tag it as deprecated.

You need `gen-function-list.py` in the `doc/` directory.

I'll do a very quick review for you.

libvips/foreign/rawsave.c
```diff
@@ -116,49 +117,149 @@ vips_foreign_save_raw_build(VipsObject *object)
 	if (VIPS_OBJECT_CLASS(vips_foreign_save_raw_parent_class)->build(object))
 		return -1;
 
-	if ((raw->fd = vips__open_image_write(raw->filename, FALSE)) < 0 ||
-		vips_image_pio_input(save->in) ||
-		vips_sink_disc(save->in, vips_foreign_save_raw_write, raw))
+	if (vips_image_pio_input(save->in) ||
+		vips_sink_disc(save->in, vips_foreign_save_raw_block, raw))
 		return -1;
```
jcupitt, line 123: You'll need to `_end()` the target after write, eg, add:

```C
    if (vips_target_end(raw->target))
        return -1;
```

jcupitt (commented):
Looks fine, other than one small thing. Nice!

libvips/foreign/rawsave.c
jcupitt: I think this will fix the testsuite failures as well.

jcupitt:
We could maybe add a ` _buffer` subclass too, for completeness.

libvips/foreign/rawsave.c
akash-akya: Corrected now, thanks!

akash-akya:
Facing an issue with generating the docs. I am not able to install pyvips due to compilation issue. I think my local env is messed up because of multiple vips I have installed

```
 build/temp.macosx-14.3-x86_64-cpython-312/_libvips.c:6208:27: error: incompatible function pointer types passing 'void (*)(void *)' to parameter of type 'VipsCallbackFn' (aka 'int (*)(void *, void *)') [-Wincompatible-function-pointer-types]
        vips_value_set_blob(x0, x1, x2, x3);
```

So `vips_rawsave_buffer` is very similar to `vips_image_write_to_memory` but only added for completeness sake correct?

doc/gen-function-list.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
```
akash-akya, line 1: Changed to pick python from PATH. This solved one of the issue I faced while running.

akash-akya:
Hi @jcupitt, can you take a look at this? Thanks

jcupitt:
Sorry, I was distracted on other projects, I'll reread this now.

jcupitt:
It looks great! Thank you for doing this work, @akash-akya!

PR #3946, opened 2024-04-23
cpp: Fix build in Qt-using projects without changes

ximion:
Hi!

Currently, when including libvips in a Qt-Using C++ project, you may get errors like:
``` 
/usr/include/glib-2.0/gio/gdbusintrospection.h:155:24: error: expected ‘;’ at end of member declaration
   GDBusSignalInfo     **signals;
                        ^
/usr/include/glib-2.0/gio/gdbusintrospection.h:155:32: error: expected ‘:’ before ‘;’ token
   GDBusSignalInfo     **signals;
                                ^
``` 

This is due to Qt enabling special keywords which GLib uses as variable names. It can easily worked around with some preprocessor magic, but it's a bit jarring and also forces the addition of either a custom header to include VIPS, or to disable keywords entirely for the Qt project.

With this patch, libvips detects Qt when the C++ header is used and temporarily disables the keyword, just to enable it again after inclusion. It should have no drawbacks for non-Qt projects, but yield fewer surprises if someone wants to use VIPS from a Qt project.

Thanks for creating this useful library and considering this patch :-)

### DIFF
--- a/cplusplus/include/vips/vips8
+++ b/cplusplus/include/vips/vips8
@@ -3,7 +3,7 @@
 /*
 
     This file is part of VIPS.
-    
+
     VIPS is free software; you can redistribute it and/or modify
     it under the terms of the GNU Lesser General Public License as published by
     the Free Software Foundation; either version 2 of the License, or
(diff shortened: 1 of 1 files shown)


jcupitt:
Hi @ximion,

Ah interesting, I didn't know about the `signal` define for Qt. Thank you!

huskier:
@ximion Thank you for the patch, which has really solved my weird compiling errors when the including file "#include <vips/vips8>" is not put on top of other including files.

ximion:
It's really obvious when you are familiar with both Qt and GLib, but otherwise it makes absolutely no sense and is a really hard to grasp error...
It's nice that it's now resolved centrally and hopefully nobody will run into it again :-)

PR #3949, opened 2024-04-30
rawsave: add missing suffix to buffer saver

kleisauke:
Resolves this warning:
```
no suffix defined for rawsave_buffer
```
https://github.com/libvips/libvips/actions/runs/8861347308/job/24333121514#step:14:285

Also, while we are here, sort the changelog notes in chronological order.


jcupitt:
Ahh thanks!

PR #3964, opened 2024-05-17
fix(pdfium): render annotations

SpringMT:
related to: https://github.com/libvips/libvips/pull/3790

Rendering all annotations is also required `FPDF_ANNOT` flags at FPDF_RenderPageBitmap.

Ref: https://pdfium.googlesource.com/pdfium/+/main/public/fpdfview.h#899


jcupitt:
Thank you! This will be in 8.16.

Issue #3977, opened 2024-06-06
8.15.2 crashes Erlang VM on NetBSD


xorrvin:
## Bug report

**Describe the bug**
Mostly outlined in https://github.com/erlang/otp/issues/8550. To summarize, recent (ec432a57968b8c9c240cdef0343437eb59571929) master branch commit produces working library, while 8.15.2 release crashes Erlang VM.

**To Reproduce**
Try to compile https://github.com/akash-akya/vix against both versions. Crash occurs when Erlang code tries to open a derivative library (`vix.so`).

libvips is configured with `-Dintrospection=disabled -Darchive=disabled`, final library is linked like this:
```
netbsd# ldd /opt/pkg/lib/libvips.so.42
/opt/pkg/lib/libvips.so.42:
	-lglib-2.0.0 => /opt/pkg/lib/libglib-2.0.so.0
	-liconv.2 => /opt/pkg/lib/libiconv.so.2
	-lc.12 => /usr/lib/libc.so.12
	-lintl.8 => /opt/pkg/lib/libintl.so.8
	-lm.0 => /usr/lib/libm.so.0
	-lpcre2-8.0 => /opt/pkg/lib/libpcre2-8.so.0
	-lpthread.1 => /usr/lib/libpthread.so.1
	-lgio-2.0.0 => /opt/pkg/lib/libgio-2.0.so.0
	-lgobject-2.0.0 => /opt/pkg/lib/libgobject-2.0.so.0
	-lffi.8 => /opt/pkg/lib/libffi.so.8
	-lgmodule-2.0.0 => /opt/pkg/lib/libgmodule-2.0.so.0
	-lz.1 => /opt/pkg/lib/libz.so.1
	-lexpat.1 => /opt/pkg/lib/libexpat.so.1
	-lexif.12 => /opt/pkg/lib/libexif.so.12
	-ljpeg.9 => /opt/pkg/lib/libjpeg.so.9
	-lpng16.16 => /opt/pkg/lib/libpng16.so.16
```

**Expected behavior**
Compilation succeeds

**Actual behavior**
Erlang VM segfaults on 8.15.2

**Environment**
 - OS: NetBSD 10.0 amd64
 - Vips: 8.15.2

kleisauke:
Could you also test the [`8.15`](https://github.com/libvips/libvips/commits/8.15) branch? The gdb backtrace includes a reference to `ifunc`, so perhaps this was fixed via PR #3953.

[...]

PR #3980, opened 2024-06-08
Ensure `target_clones` support is checked at runtime

kleisauke:
```console
$ cat reduced.c
static int __attribute__((target_clones("default,avx")))
has_target_clones(void) {
    return 0;
}

int main(void) {
    int (*func)(void) = has_target_clones;
    return func();
}
$ cc reduced.c -o repro -Wl,-z,now,-z,relro
$ ./repro
[1]   Segmentation fault (core dumped) ./repro
$ uname -mrs
NetBSD 10.0 amd64
```

Resolves: #3977
Upstream issue: https://gnats.netbsd.org/57792

Targets the 8.15 branch.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -147,7 +147,13 @@ int main(void) {
     return func();
 }
 '''
-if cc.links(target_clones_check, args: '-Werror', name: 'Has target_clones attribute')
+if meson.can_run_host_binaries()
+    rres = cc.run(target_clones_check, args: '-Werror', name: 'Has target_clones attribute')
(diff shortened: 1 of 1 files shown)


lovell (approved):
LGTM, perhaps add a changelog entry for this too.

kleisauke:
The idea was to reuse this existing changelog entry:
https://github.com/libvips/libvips/blob/a4be03b592204a95dec1b6eae4bf29e5efdabf2a/ChangeLog#L6

Issue #3984, opened 2024-06-10
Crash trying to read/decode large Big TIFF File


snowzach:
## Bug report

**Describe the bug**
When trying to thumbnail or get the headers for a very large BTF file it give an error and crashes/exits.

**To Reproduce**

Download this huge BTF file (warning, 11-12ish gig)
https://cf.10xgenomics.com/samples/spatial-exp/3.0.0/Visium_HD_Human_Colon_Cancer/Visium_HD_Human_Colon_Cancer_tissue_image.btf

Try to use vipsheader (or vipsthumbnail) on it by piping..
```
cat Visium_HD_Human_Colon_Cancer_tissue_image.btf | vipsheader -a stdin

(vipsheader:28): GLib-ERROR **: 16:00:14.638: adding 4096 to array would overflow
Trace/breakpoint trap (core dumped)
```

**Expected behavior**
Work as expected and print out information or thumbnail.

**Actual behavior**
Traps and coredumps.

**Environment**
Tried in ubuntu 22 with 8.12.1 and Alpine latest with 8.15.0-r0

 - OS: [e.g. Ubuntu 16.04, MacOS 10.13]
 - Vips: [e.g. 8.8, 8.9, etc]

**Additional context**

jcupitt:
Hi @snowzach, wow what a monster. I'll have a look.

jcupitt:
It's working fine in git master libvips:

```
$ vipsheader Visium_HD_Human_Colon_Cancer_tissue_image.btf 
Visium_HD_Human_Colon_Cancer_tissue_image.btf: 75250x48740 uchar, 3 bands, srgb, tiffload
```

I'll try some older versions.

I can view it in the current vipsdisp binary:

![image](https://github.com/libvips/libvips/assets/580843/d87f746d-b764-48be-9c34-3cffde870fa4)

See https://github.com/jcupitt/vipsdisp, you can install the current version with `flatpak install flathub org.libvips.vipsdisp`.

snowzach:
It seems to work okay when pointing it right to the file. I want to try to use thumbnail and header from an AWS Lambda which doesn't have enough disk space to store it. So I am trying to pipe it via stdin. It doesn't seem to like that. That's what's giving the error.

jcupitt:
Ahhhh I see. I'll try again.

jcupitt:
[...]

PR #3985, opened 2024-06-10
fix pipe read limit

jcupitt:
- We were not checking `VIPS_PIPE_READ_LIMIT` for huge single reads
- Use vips__parse_size() to parse the size of the read limit, so eg. "2g" now works for a 2gb read limit
- vipsheader was not always displaying the error buffer correctly

see https://github.com/libvips/libvips/issues/3984


jcupitt:
Values of `VIPS_PIPE_READ_LIMIT` over 4gb will cause strange errors. With a 12gb test file:

```
$ export VIPS_PIPE_READ_LIMIT=4g
$ cat Visium_HD_Human_Colon_Cancer_tissue_image.btf | vipsheader -a stdin

(vipsheader:169956): GLib-ERROR **: 19:13:03.366: adding 4096 to array would overflow
Trace/breakpoint trap (core dumped)
```

Since GByteArray is limited to 4gb. Perhaps we should enforce 2**32-1 as an upper bound for `VIPS_PIPE_READ_LIMIT`?

jcupitt:
... in case anyone wants to test this, you can gdb with a piped input with:

```
$ mkfifo my-fifo
$ cat hugefile > my-fifo &
$ gdb vipsheader
(gdb) run < my-fifo
```

(took me a while to work out)

ChangeLog
```diff
@@ -20,6 +20,7 @@ master 8.16
 - webpsave has @passes parameter to set number of passes to achieve
   desired target_size [john-parton]
 - revise sRGB -> B_W coefficients [daniellovera]
+- fix pipe read limit
```
lovell, line 23: Is it worth mentioning that the previous default limit of 1GB has been removed (and must now be set explicitly if required)?

kleisauke: It looks like the default is still 1GiB, unless I'm overlooking something:
https://github.com/libvips/libvips/blob/9de5f24dab4ea2064666b351960a9719f27a4ac5/libvips/iofuncs/source.c#L79-L84

lovell: D'oh, yes it is, thank you Kleis.

lovell (approved):

kleisauke:
Should this PR target the [`8.15`](https://github.com/libvips/libvips/tree/8.15) branch instead? FWIW, version 8.15.3 has not been released yet; the release date in the changelog could be a bit misleading.
https://github.com/libvips/libvips/blob/a4be03b592204a95dec1b6eae4bf29e5efdabf2a/ChangeLog#L1

jcupitt:
Ah true. I was imagining releasing 8.16 soon (it's been six months), but you're right that 8.15 has a couple of useful things that should go out before then. I'll retarget it.

jcupitt:
Haha I completely failed to retarget it, I wonder how your are supposed to do this? Anyway, I've made a new pr for 8.15.

https://github.com/libvips/libvips/pull/3990

kleisauke:
> I wonder how your are supposed to do this?

It's a bit hidden in GitHub's UI:
Change the base branch of a PR
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request

jcupitt:
Ahh, I missed that. Thanks!

PR #3991, opened 2024-06-12
jxlload: guard against NaN values

kleisauke:
Both `tps_denominator` and `tps_numerator` will be zero for non-animated images.

Found by UBSan:
https://github.com/kleisauke/libvips/actions/runs/9486292772/job/26140300986#step:14:327


jcupitt:
Nice!

PR #4000, opened 2024-06-15
CI: upgrade macOS runner to version 14 (Sonoma)

kleisauke:
This resolves the current CI failure caused by GitHub Actions installing a non-Homebrew Python into the Homebrew prefix. On `macos-14` runners, which use Apple Silicon/ARM, this issue does not occur because Homebrew uses `/opt/homebrew` as its install prefix on these systems.


jcupitt:
Great!

Issue #4003, opened 2024-06-15
Meson does not find vsnprintf on msvc


Julianiolo:
## Bug report

Meson fails to find vsnprintf when compiling on msvc.
```
...
Checking for function "vsnprintf" with dependency -lm: NO 
...
```

I think this can be resolved by adding `prefix: "#include <stdio.h>"` to the [`cc.has_function`](https://github.com/libvips/libvips/blob/ed700dcd38957d627347a2fe01d9d12bbc0f73d9/meson.build#L178) call.

The underlying issue seems to be that msvc only defines the `vsnprintf` function in the stdio.h header (as supposed to having it in a system library and only having the function signature in the header). This makes meson not find the vsnprintf symbol without the prefix parameter, as it then just tries to define the prototype in a  file and see if it links.

**Environment**
 - OS: Windows 10
 - Compiler: msvc 193
 - Vips: 8.15.2

jcupitt:
Hi @Julianiolo, this is a great find, this must have been causing rare problems on windows (I suppose?) since we switched to meson. I'll make a PR for these two issues.

lovell:
Could we switch to the glib-provided [g_vsnprintf](https://docs.gtk.org/glib/func.vsnprintf.html)?

jcupitt:
Yup, our terrible home-made fallback pre-dates libvips being used in heavily threaded applications. The glib one is a great upgrade.

kleisauke:
I just opened PR #4004 for this.

jcupitt:
I made https://github.com/libvips/libvips/pull/4005 for 8.15 -- it's a minimal PR that just changes the fallback and doesn't touch most of the code. Let's move any discussion there.

Let's use Kleis' much more comprehensive one for master.

I credited you in the changelog @Julianiolo, and thanks again for finding this terrible thing!

Issue #4001, opened 2024-06-15
vips_vsnprintf own implementation is not thread safe


Julianiolo:
## Bug report

**Describe the bug**
The implementation of [`vips_vsnprintf`](https://github.com/libvips/libvips/blob/ed700dcd38957d627347a2fe01d9d12bbc0f73d9/libvips/iofuncs/util.c#L456) that is used when vsnprintf is not present is not thread safe, as it uses a [buffer](https://github.com/libvips/libvips/blob/ed700dcd38957d627347a2fe01d9d12bbc0f73d9/libvips/iofuncs/util.c#L467) with static storage duration.
I noticed this while trying to load thumbnails from multiple threads which spontaneously failed due to the `vips_vsnprintf` in [`vips_object_set_from_string`](https://github.com/libvips/libvips/blob/ed700dcd38957d627347a2fe01d9d12bbc0f73d9/libvips/iofuncs/object.c#L2586) sometimes producing garbage data in the buffer (for me this was "t]" instead of the expected "[]").
I guess the buffer has this static storage duration to make overflows less likely to mess up the stack? If that is the case, maybe mallocing or a thread local buffer (if that is even available in c99?) would be better?

The weird part is this happened because meson does not seem to recognise msvc having vsnprintf, which I still have to investigate...

**Environment**
 - OS: Windows 10
 - Compiler: msvc 193
 - Vips: 8.15.2

Julianiolo:
I found out why meson does not find vsnprintf on msvc: #4003

jcupitt:
Fixed with the PR, let's move any discussion there. Thank you!

PR #4005, opened 2024-06-16
fix vips_vsnprintf()

jcupitt:
- vsnprintf() detection with meson was not working on windows
- the vips_vsnprintf() fallback was not threadsafe

This PR changes the fallback to g_vsnprintf() and improves detection in meson.

See https://github.com/libvips/libvips/issues/4003 and https://github.com/libvips/libvips/issues/4001

Thanks Julianiolo

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -5,6 +5,7 @@
 - pngload: disable ADLER32/CRC checking in non-fail mode [kleisauke]
 - improve target_clones support check [kleisauke]
 - fix pipe read limit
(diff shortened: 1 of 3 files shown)


jcupitt:
I tried to be unobtrusive for the stable branch and just changed the fallback.

For 8.16 I think we should get rid of `vips_snprintf()` and just use `g_snprintf()` everywhere. We'll need to keep tiny `vips_*snprintf()` in `deprecated/` for compat.

kleisauke (approved):
LGTM!

Issue #3999, opened 2024-06-15
Add an `sdf` operator


jcupitt:
We could add something like eg.:

```python
image = pyvips.Image.sdf("circle", 1000, 1000, x=100, y=200, r=20)
```

Rather like `xyz`, but it would generate a signed distance field rather than simply (x, y) coordinates. These are easier to work with and combine.

https://iquilezles.org/articles/distfunctions2d/

Has a handy list of basic shapes that could be implemented easily.

See https://github.com/libvips/libvips/discussions/3998

jcupitt:
... I made a draft PR with circles implemented. Let's move any discussion there.

PR #4008, opened 2024-06-16
add vips_sdf()

jcupitt:
see https://github.com/libvips/libvips/issues/3999


jcupitt:
Run with eg.

```
$ vips sdf x.v 1000 1000 circle --cx 500 --cy 500 --r 300
```

to make a circular SDF centred at (500, 500), radius 300.

image

Handy list of other functions we could implement easily: https://iquilezles.org/articles/distfunctions2d/

libvips/create/sdf.c
```diff
@@ -0,0 +1,234 @@
...
+};
+
+typedef VipsCreateClass VipsSdfClass;
+
+G_DEFINE_TYPE(VipsSdf, vips_sdf, VIPS_TYPE_CREATE);
+
+static float
+vips_sdf_circle(VipsSdf *sdf, int x, int y)
+{
+	return sqrt(x * x + y * y) - sdf->r;
```
andreas-kupries, line 79:
```suggestion
	return hypot(x, y) - sdf->r;
```

jcupitt: I always forget about `hypot()`. We're C99 now, so it's even easier.

kleisauke: According to: https://github.com/libvips/libvips/blob/312ec1c2671b080a81cfca09b77afb7314553e69/libvips/create/worley.c#L197-L203 
`sqrt(x * x + y * y)` might be actually faster for int arguments.

jcupitt: That's true, though `x * x` will overflow for ints over 32767, unfortunately :(

libvips/include/vips/create.h
```diff
@@ -79,6 +79,8 @@ int vips_sines(VipsImage **out, int width, int height, ...)
 VIPS_API
 int vips_zone(VipsImage **out, int width, int height, ...)
 	G_GNUC_NULL_TERMINATED;
+VIPS_API
+int vips_sdf(VipsImage **out, int width, int height, const char *name, ...);
```
kleisauke, line 83:
```suggestion
int vips_sdf(VipsImage **out, int width, int height, const char *name, ...)
	G_GNUC_NULL_TERMINATED;
```

jcupitt: Ooop! Nice.

jcupitt:
`rounded-box` works well. You can run eg.

```
$ vips sdf x.v 1000 1000 rounded-box --sx 400 --sy 200 --corners "100 0 0 0" 
```

To make:

image

Cute!

I think this will be a useful fast and low-memory replacement for SVG load for many large masks.

libvips/arithmetic/clamp.c
```diff
@@ -0,0 +1,213 @@
...
+ * This operation clamps pixel values to a range, by default 0 - 1.
+ *
+ * Use @min and @max to change the range.
+ *
+ * See also: vips_sign(), vips_abs(), vips_sdf().
+ *
+ * Returns: 0 on success, -1 on error
+ */
+int
+vips_clamp(VipsImage *in, VipsImage **out, ...)
```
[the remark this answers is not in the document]
jcupitt: Ooop! Fixed, thanks

jcupitt:
It works well with python too:

```python
#!/usr/bin/env python3

import sys
import pyvips

# generate the SDF
box1 = pyvips.Image.sdf(1000, 1000, "rounded-box",
                        sx=400, sy=200,
                        corners=[100, 0, 0, 0])

# subtracting a constant expands the object by that many pixels
box2 = box1 - 50

# clamp and subtract to make a thick outline
outline = box1.clamp() - box2.clamp()
(255 * outline).write_to_file(sys.argv[1])
```

To render a high-quality anti-aliased outline:

x

jcupitt:
I've added some docs and tests, so I think this is probably ready for review.

It only has four SDF generators right now, but it'd be easy to add more in later PRs. 

I added `maxpair` and `minpair` as well, they are handy for combining SDFs, and quicker than `(a > b).ifthenelse(a, b)`.

libvips/create/sdf.c
```diff
@@ -0,0 +1,396 @@
...
+		1, VIPS_MAX_COORD, 1);
+
+	VIPS_ARG_INT(class, "height", 3,
+		_("Height"),
+		_("Image height in pixels"),
+		VIPS_ARGUMENT_REQUIRED_INPUT,
+		G_STRUCT_OFFSET(VipsSdf, height),
+		1, VIPS_MAX_COORD, 1);
+
+	VIPS_ARG_STRING(class, "name", 6,
```
kleisauke, line 308: A `VIPS_ARG_ENUM` would probably be neater for strongly typed language bindings. For example, in NetVips (using C# 12 and later), this would result in:
```csharp
using var image = Image.Sdf(1000, 1000, Enums.Sdf.RoundedBox, a: [200, 100], b: [500, 200], corners: [0, 0, 0, 50]);
```
instead of:
```csharp
using var image = Image.Sdf(1000, 1000, "rounded-box", a: [200, 100], b: [500, 200], corners: [0, 0, 0, 50]);
```

kleisauke (commented):
This looks great! The videos linked at https://iquilezles.org/articles/distfunctions2d/ are also very interesting and provide some helpful background information.

I only have a small comment regarding the type of `@name`.

libvips/create/sdf.c
jcupitt: Yes, I was unsure about that. We could potentially end up with a lot of these, so I thought maybe a string was easier.

andreas-kupries:
Notes on combining SDFs

Already seen:

 - `maxpair` - This is `intersection`, `and`, `A*B`
 - `minpair` - This is `union`, `or`, `A+B`.

Not yet seen:

 - Negation, i.e. `f(x) = -x` is `inversion` / `not`.
 - The difference `A - B` is then definable as `max(A, -B)`, i.e. `A and not B`,
 - Symmetric difference is `(A + B) - (A * B)` = `max(min(A,B),-max(A,B))`

This is all in section `Primitive combinations` of https://iquilezles.org/articles/distfunctions/ which is more about 3D SDF.
Still useful to 2D. Also useful the notes about `smooth +, *, -` immediately after the strict combinators.

jcupitt:
I flipped it to using an enum ... I went for `VipsSdfShape`, with members eg. `VIPS_SDF_SHAPE_ROUNDED_BOX`.

I tried generating the docs and they seem fine.

Another test program:

```python
#!/usr/bin/env python3

import sys
import pyvips

box = pyvips.Image.sdf(1000, 1000, "rounded-box", 
                       a=[300, 400],
                       b=[700, 600],
                       corners=[100, 0, 0, 0])

circle = pyvips.Image.sdf(1000, 1000, "circle", 
                          a=[500, 300],
                          r=100)

line = pyvips.Image.sdf(1000, 1000, "line", 
                        a=[500, 500], 
                        b=[600, 900])

# union
sdf = box.minpair(circle).minpair(line)

# make annular
sdf = sdf.abs() - 15

# render as an antialiased image
(255 * (1 - sdf.clamp())).write_to_file(sys.argv[1])
```

To make:

x


Someone rowing a boat, obviously.

libvips/create/sdf.c
jcupitt: I changed it an enum ... thanks!

andreas-kupries:
> # make annular
> sdf = sdf.abs() - 15
> 
> # render as an antialiased image
> (255 * (1 - sdf.clamp())).write_to_file(sys.argv[1])
> ```
> Someone rowing a boat, obviously.

I thought it was some kind (railroad) sign, although without content. :smile: 

Regarding annular and render, I packaged this up for my own convenience:

https://core.tcl-lang.org/akupries/aktive/file?ci=tip&name=etc/generator/virtual/sdf.tcl&ln=174-233

and

https://core.tcl-lang.org/akupries/aktive/file?ci=tip&name=etc/generator/virtual/sdf.tcl&ln=235-273

Primitives at
- https://core.tcl-lang.org/akupries/aktive/dir?ci=tip&name=etc/generator/virtual/sdf/coding, plus
- https://core.tcl-lang.org/akupries/aktive/file?name=op/sdf.h&ci=tip, and
- https://core.tcl-lang.org/akupries/aktive/file?name=op/sdf.c&ci=tip for the underlying C functions.

kleisauke (approved):

PR #4053, opened 2024-07-18
Test: regenerate huge AVIF fixture

kleisauke:
This is required since PR strukturag/libheif#1204 (libheif 1.18.0), which limits the total image size instead of the width and height individually.

New fixture was generated with:
```console
$ vips black x.png 17000 17000
$ avifenc -q 0 -s 0 --ignore-exif x.png 17000x17000.avif
```


jcupitt:
:+1:

PR #4054, opened 2024-07-18
matrixsave: ensure `matrixprint` will write to stdout

kleisauke:
The integer file descriptors associated with the streams stdin, stdout, and stderr are 0, 1, and 2, respectively.


jcupitt:
:facepalm: thanks!

Issue #1162, opened 2018-11-10
segfault in python or libvips during approximate convolution of a short or ushort image


erdmann:
Hi John,

I was experimenting with the approximate convolution you mentioned in #1160 and I came across a persistent segfault but only when convolving ushort images.  I can make the segfault happen either in python or libvips depending on the size of the image, so I suppose this is might be a double free.  Here is a test program with some experiments to see where the problem happens.  Below that is a stack trace of the segfault location inside libvips.so.42.9.0 (without using a debugging build at the moment, so maybe not so helpful).

I'm running vips `vips-8.7.0-Sun Jan 28 17:23:05 UTC 2018` using the pyvips interface under python `Python 2.7.15 :: Anaconda custom (64-bit)`

Thanks in advance.

Note: actually a .py file:
[segfault.txt](https://github.com/libvips/libvips/files/2568715/segfault.txt)

```
# /usr/bin/env python
import pyvips

slab = pyvips.Image.new_from_array([[1]*10]*10, scale=100)  # 10x10 moving average

n = 30000 # also segfaults when n=30000.  See note below.

# The following are all fine
uchar_im = pyvips.Image.black(n, n).cast('uchar')  # just to be sure it's uchar
uchar_im.conv(slab, precision='approximate', layers=1).write_to_file('uchar_conv.v')

float_im = pyvips.Image.black(n, n).cast('float')
float_im.conv(slab, precision='approximate', layers=1).write_to_file('float_conv.v')


# EVERY CONVOLUTION BELOW THIS LINE SEGFAULTS

# if n=1, completes by writing result to file, then segfaults in python:
# *** Error in `python': free(): invalid pointer: 0x00005651d79154c0 ***

# if n=30000, does not complete and segfaults in libvips
# [1049087.739802] worker[599]: segfault at 7f4cd631a000 ip 00007f4ce7f93318 sp 00007f4cd2caea70 error 4 in libvips.so.42.9.0[7f4ce7eb6000+1f5000]

ushort_im = pyvips.Image.black(n, n).cast('ushort')
ushort_im.conv(slab.cast('ushort'), precision='approximate', layers=1).write_to_file('ushort_conv.v')  # segfault happens here

ushort_im = pyvips.Image.black(n, n).cast('ushort')
ushort_im.conv(slab, precision='approximate', layers=1).write_to_file('ushort_conv.v')  # segfault would also happen here

short_im = pyvips.Image.black(n, n).cast('short')
[...]

PR #4055, opened 2024-07-18
conva: fix crash with {u,}{short,int} images

kleisauke:
{u,}{short,int} images should use {u,}short as an intermediate type when the length of the longest line is below 256.

Resolves: #1162.


lovell:
Should we target the 8.15 branch for this (assuming a possible 8.15.3 release)?

kleisauke:
Good idea. Targeted against the [`8.15`](https://github.com/libvips/libvips/tree/8.15) branch and added a changelog note.

lovell (approved):

kleisauke:
CI will be fixed via PR #4057.

jcupitt:
Ah nice. Yes, I lost enthusiasm for approximate convolution and never really finished it :( Thanks for picking this up!

PR #4057, opened 2024-07-18
heifsave: ensure true lossless with libheif < v1.17.2

kleisauke:
In accordance with commit https://github.com/strukturag/libheif/commit/136cf0173ffb52a92842736fec47c83e69475327.

However, it still requires libheif >= 1.13.0 for true lossless, so update `test_avifsave_lossless` accordingly to fix CI on the 8.15 branch (which still uses Ubuntu 22.04 as the runner image).

Targets the 8.15 branch.


lovell (approved):

jcupitt:
I added all these changes to master as well.

PR #4058, opened 2024-07-20
composite: fix dest-atop blend mode

kleisauke:
The dest-atop blend mode should function like atop, but with the drawing operations reversed. Therefore, t1 must be `1 - aB`.

Found using this test script:
```python
import pyvips

base = (pyvips.Image.black(120, 90)
        .copy(interpretation='srgb')
        .new_from_image([179, 0, 0, 204])
        .embed(0, 0, 160, 120))

overlay = (pyvips.Image.black(120, 190)
           .copy(interpretation='srgb')
           .new_from_image([0, 0, 230, 204])
           .embed(40, 30, 160, 120))

im = base.composite2(overlay, 'dest-atop')
im.write_to_file('vips.png')
```

| Expected image | Actual image |
| :--: | :--: |
| vips |vips-before |

Targets the 8.15 branch.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -10,6 +10,7 @@
 - conva: fix a crash with {u,}{short,int} images [erdmann]
 - fix vips_image_get_string
 - heifsave: fix lossless mode [kleisauke]
+- composite: fix dest-atop blend mode [kleisauke]
 
 12/3/24 8.15.2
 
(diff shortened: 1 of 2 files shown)


lovell (approved):
This matches with the canonical logic of https://www.cairographics.org/operators/#dest_atop

PR #4070, opened 2024-07-27
Fix CI failures

kleisauke:
```
==================================== ERRORS ====================================
________________ ERROR at setup of TestIofuncs.test_revalidate _________________
Error evaluating 'skipif': you need to specify reason=STRING when using booleans as conditions.
```
https://github.com/libvips/libvips/actions/runs/10123138814/job/27996075308#step:14:671


jcupitt:
Argh! Thank you.

PR #4075, opened 2024-07-30
jpeg2vips: prefer use of `cinfo->output_scanline`

kleisauke:


lovell (approved):
This is a nice simplification, thanks Kleis.

PR #4076, opened 2024-07-30
tilecache: prefer use of `g_hash_table_size()`

kleisauke:


lovell (approved):
Do we think this might have been the cause of any errors?

kleisauke:
This ought to be a non-functional change. The only difference is that `g_hash_table_size()` uses `guint` instead of `int` for its size. This shouldn't cause any errors, as we create tiles up to a maximum of `cache->max_tiles` and otherwise reuse old tiles as needed.

jcupitt:
Thanks! 

I was thinking in the shower this morning, perhaps the performance issue with `tile_cache` is related to the way that threads are forced to queue on the results of other threads? This effectively pushes the system into something like a single-threaded mode.

Instead, we should have a queue of tiles awaiting computation, and rather than blocking, threads should cooperatively take tiles from this list, rather like the other sinks.

I suppose the tricky part would be efficiently deciding when the tiles that a thread needs to fulfil its request have been computed and it should change from computing regions in the source image to returning a result to its caller.

(erm this is from memory, I've not looked at the code carefully for a long time, perhaps I'm mistaken)

jcupitt:
Background (`tile_cache` is very complicated and it's not obvious how it works):

- a cached image has `max_tiles` regions of pixels in a hash table indexed by the tile coordinate
- when a request arrives, the thread (thread A) makes and locks all the tiles it needs -- some tiles might be there already, some might not have been computed yet
- it starts computing the missing tiles
- while it's doing that, thread B arrives with another request ... if it's for a separate set of tiles, it just sets to work exactly as thread A 
- however if the two requests overlap (this is the most likely case when processing sequentially), thread B waits for thread A to complete
- and that means we're wasting time

(again, from memory, I should read the code again)

kleisauke:
Indeed, `tile_cache` is a bit complicated. I found these non-functional changes while fiddling with the [`remove-seq-from-reduce`](https://github.com/libvips/libvips/tree/remove-seq-from-reduce) branch.

The idea was that `reducev` / `shrinkv` could somehow hint the number of tiles to cache. For example, on that branch, I see:
```console
$ vips black x.jpg 10000 10000
$ VIPS_STALL=1 vips reducev x.jpg x.v 2 --vips-concurrency 12
vips_line_cache_build: n_lines = 256
vips_line_cache_build: max_tiles = 128
vips_line_cache_build: tile_height = 8
vips_line_cache_build: max size = 9.76562 MB
vips_worker_work_unit: stall done, releasing y = 0 ...
vips_tile_cache_ref: bumped max_tiles from 128 to 130

(vips:30778): VIPS-WARNING **: 14:18:35.803: error in tile 0 x 504

(vips:30778): VIPS-WARNING **: 14:18:35.803: error in tile 0 x 512
vips_worker_work_unit: stall done, releasing y = 0 ...
VipsJpeg: out of order read at line 1544
```

In this example, 2 old tiles are being reused too early. By the time `max_tiles` is bumped, it's too late because the old tiles are already gone. Forcing the value to `130` here:
https://github.com/libvips/libvips/blob/73e91aa8acb96bc6e5cb36b14e482855f6588429/libvips/conversion/tilecache.c#L929-L930

makes this example work correctly.
```console
$ VIPS_STALL=1 vips reducev x.jpg x.v 2 --vips-concurrency 12
vips_line_cache_build: n_lines = 256
vips_line_cache_build: max_tiles = 130
vips_line_cache_build: tile_height = 8
vips_line_cache_build: max size = 9.91821 MB
vips_worker_work_unit: stall done, releasing y = 0 ... [repeated 19 times]
```

PR #4077, opened 2024-07-30
tilecache: only call `vips_tile_move()` when reusing a tile

kleisauke:
Avoids an unnecessary call to `g_hash_table_{steal,insert}()` when creating a new tile.


jcupitt:
Sorry, took  me a while.

Issue #3890, opened 2024-03-10
Improve `tilecache` performance


jcupitt:
This test prog:

```python
#!/usr/bin/env python3

import sys
import pyvips

image = pyvips.Image.new_from_file(sys.argv[1])

image = image.tilecache(tile_width=image.width,
                        tile_height=256,
                        max_tiles=1,
                        persistent=True)

patches = [image.crop(x, y, min(256, image.width - x), min(256, image.height - y))                    
           for y in range(0, image.height, 256)
           for x in range(0, image.width, 256)]
           
for patch in patches:
    pixel_array = patch.numpy()
```

With this (7 gb) test tiff:

```
$ vips bandjoin "st-francis.jpg st-francis.jpg st-francis.jpg" x.tif[tile]
```

I see:

```
$ VIPS_CONCURRENCY=1 /usr/bin/time -f %M:%e ./readtile.py x.tif
472656:13.37
```

If I change this line:

```python
image2 = image.tilecache(tile_width=image.width,
```

ie. create the `tilecache` but never use it, I see:

```
$ VIPS_CONCURRENCY=1 /usr/bin/time -f %M:%e ./readtile.py x.tif
143064:4.51
```

It looks like the tilecache alone is taking almost 10s of runtime. Can we improve this?

jcupitt:
Test PR made, let's move any discussion there.

PR #4081, opened 2024-08-05
revise tilecache

jcupitt:
quite a bit simpler, perhaps faster

see https://github.com/libvips/libvips/issues/3890

### DIFF
--- a/libvips/conversion/tilecache.c
+++ b/libvips/conversion/tilecache.c
@@ -67,6 +67,7 @@
 /*
 #define VIPS_DEBUG_RED
 #define VIPS_DEBUG
+#define VIPS_SANITY
  */
 
 #ifdef HAVE_CONFIG_H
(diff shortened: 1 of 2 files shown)


jcupitt:
Needs quite a bit of testing and benchmarking! But at least simple cases seem to work.

jcupitt:
This PR makes threads that request tiles compute downstream tiles cooperatively. It should (I hope) handle overlapping requests.

Threads loop every tile to see if their set of pending tiles have been computed, we could get rid of this, though I doubt if it's a bottleneck. But who knows!

Other benchmarks would be useful for testing.

jcupitt:
I've not found a case where it's faster yet :( In fact overall it's slightly slower.

I'll get the test suite to pass, perhaps I missed something.

kleisauke:
Either removing `tile-width=256` here:
https://github.com/libvips/libvips/blob/566ce6d470faaa70b6ff9c7d02bbbc219576e2be/test/test-suite/test_foreign.py#L592-L593

Or disabling threaded access here:
https://github.com/libvips/libvips/blob/566ce6d470faaa70b6ff9c7d02bbbc219576e2be/libvips/foreign/tiff2vips.c#L2614

Makes the test suite pass, fwiw.

libvips/conversion/tilecache.c
```diff
@@ -602,156 +670,59 @@ static int
...
-				break;
-			}
+	/* Loop until all our tiles are ready, or there's an error of some sort,
+	 * perhaps a loader hits a bad tile.
+	 */
+	GSList *todo;
+	todo = g_slist_copy(work);
+	while (todo) {
+		VipsTile *tile = (VipsTile *) todo->data;
+		todo = g_slist_remove(todo, tile);
```
kleisauke, line 704: I think we could avoid the (possible expensive) `g_slist_copy()` here.

```suggestion
	for (GSList *todo = work; todo; todo = todo->next) {
		VipsTile *tile = (VipsTile *) todo->data;
```

libvips/conversion/tilecache.c
```diff
@@ -602,156 +670,59 @@ static int
...
+
+		if (tile->dirty &&
+			vips_tile_cache_compute(in_region, tile)) {
+			*stop = TRUE;
+			break;
 		}
+	}
+	VIPS_FREEF(g_slist_free, todo);
```
kleisauke, line 712: `g_slist_free()` is not needed if the other suggested change is accepted.

```suggestion
```

PR #4090, opened 2024-08-18
OSS-Fuzz: add generic buffer saver

kleisauke:
Introduce a generic buffer saver that attempts to save images in the format matching the input. This ensures that our fuzzers covers the following savers: `heifsave`, `ppmsave`, `radsave` and `tiffsave`.

### DIFF
--- a/fuzz/generic_buffer_fuzzer.cc
+++ b/fuzz/generic_buffer_fuzzer.cc
@@ -1,0 +1,64 @@
+#include <vips/vips.h>
+
+extern "C" int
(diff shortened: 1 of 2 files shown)


jcupitt:
Oh, nice! 

Could you pass in the save format as a CLI parameter or perhaps an env var? It would avoid the weirdness getting the format from the loader name. Maybe the fuzzer doesn't allow that, I can't remember.

kleisauke:
Indeed, this is a bit odd. Closing in favor of #4092.

PR #4092, opened 2024-08-19
OSS-Fuzz: expand coverage of buffer save fuzzers

kleisauke:
Ensures that the following buffer savers are covered in OSS-Fuzz: `csvsave`, `heifsave`, `matrixsave`, `ppmsave`, `radsave`, `rawsave` and `tiffsave`.

Supersedes: #4090.

PR #4096, opened 2024-08-20
OSS-Fuzz: fix seed corpus link

kleisauke:
After PR #4092, the buffer save fuzzers source file was generalized, so we need to ensure that the seed corpus remains correctly linked to each individual fuzz target.


kleisauke:
This also fixes the failing introspector build:
https://introspector.oss-fuzz.com/project-profile?project=libvips
https://oss-fuzz-build-logs.storage.googleapis.com/log-8dfb51f3-521d-4fda-bf58-1a9b70a82416.txt

jcupitt:
:+1:

PR #4103, opened 2024-08-21
Add buffer with args fuzzer

dloebl:
For reference: #4095
Allows the fuzzer to vary  `option_string` and `suffix`.
 `option_string` and `suffix` are appended at the beginning. For example:
```
[n=-1]
.jpg[Q=85]
<image-data>
```
I started to add new inputs to the seed corpus, but some loaders/savers are still missing

PR #4107, opened 2024-08-23
Fix potential memory leaks in webpsave

dloebl:
Found while fuzzing locally with #4103:
```
==14==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 1296 byte(s) in 1 object(s) allocated from:
    #0 0x56296b629ad8 in calloc /src/llvm-project/compiler-rt/lib/asan/asan_malloc_linux.cpp:77:3
    #1 0x56296c1fa633 in WebPAnimEncoderNewInternal /src/libwebp/src/mux/anim_encode.c:250:27
    #2 0x56296b79958a in WebPAnimEncoderNew /work/include/webp/mux.h:481:10
    #3 0x56296b79958a in vips_foreign_save_webp_init_anim_enc /src/libvips/build/../libvips/foreign/webpsave.c:658:14
    #4 0x56296b798c63 in vips_foreign_save_webp_build /src/libvips/build/../libvips/foreign/webpsave.c:780:7
    #5 0x56296b79c5c1 in vips_foreign_save_webp_target_build /src/libvips/build/../libvips/foreign/webpsave.c:979:6
```

**To Reproduce**
```
vips copy test/test-suite/images/invalid_multiframe.gif[n=-1] out.webp
```

**ToDo:**
- [x] Add a test


jcupitt (commented):
How about adding a call to `vips_foreign_save_webp_unset()` to `vips_foreign_save_webp_dispose()`? Then we could get rid of all these other calls on the error paths, since the operation will always be disposed pretty rapidly during error handling. 

We could leave a call at the end of `_build()`, since it can potentially free a useful amount of memory that won't otherwise get disposed for a long time.

dloebl:
Sounds good! I've simplified the cleanup handling with a36c89a44d22f3e001500f4bd2594ace737ed066

jcupitt (approved):
Ahhh much safer and prettier!

kleisauke:
Perhaps this should target the [`8.15`](https://github.com/libvips/libvips/tree/8.15) branch instead?

kleisauke (approved):

test/test-suite/test_foreign.py
kleisauke, line 1: It seems that an import for `GIF_ANIM_FILE_INVALID` is missing at the beginning of this file.

kleisauke: Ah, it looks like this is due to:
https://github.com/libvips/libvips/commit/6f7e43bcf408b029cbde736e654074e8c9aa81ba#diff-787c730538992dd0c53f7fcd19efc49bb412a43595a38d3082441437715265d0

dloebl: I've added 555121866eb96455d617bce1c36aa8f576d2a58a to make the test pass again

dloebl:
Hmm strange, I wonder why the macOS test-suite step is failing now.. it passed in my workflow: https://github.com/dloebl/libvips/actions/runs/10540080027/job/29204347884
> pyvips.error.Error: unable to call webpsave_buffer

Maybe a retry helps

kleisauke:
I could reproduce that failure with libwebp 1.3.2 using this test script:

`check-regression.sh`:
```bash
#!/usr/bin/env bash
set -ex

n=$1
exit_code=0

  echo "run $i"
  vips webpsave_buffer test/test-suite/images/invalid_multiframe.gif[n=-1] || exit_code=$?
  if [ $exit_code -ne 0 ]; then
    echo "vips exited with non-zero exit code ($exit_code) after $i runs" >&2
    exit $exit_code
  fi
done
```
```console
$ ./check-regression.sh 1024
[...]
vips exited with non-zero exit code (1) after 63 runs
```

Curious.

dloebl:
Yeah, it's strange.. the test is definitely flaky.
It fails with `gifload: Invalid frame data` from time to time.

dloebl:
Maybe some sort timing issue with threads?
It always fails on the first run with `--vips-concurrency=1`:
```
vips --vips-concurrency=1 webpsave_buffer 'test/test-suite/images/invalid_multiframe.gif[n=-1]'
```

dloebl:
I'll remove the test. There is a test case in #4103 that also provides coverage: 1228b78aaaf2e0eb435a5427be0c4dd30b05d698
Additionally, I think that the test won't even trigger in CI because libwebp isn't build with ASAN

kleisauke (approved):
Thank you! Flaky test will be fixed via PR #4116.

PR #4116, opened 2024-08-24
threadpool: fix a race condition in error handling

kleisauke:
Fixes a regression introduced in commit 9248ab7 (PR #3105).

Targets the 8.15 branch.

Context: https://github.com/libvips/libvips/pull/4107#issuecomment-2308478238.


jcupitt:
Ahhhh! Great spot, yes this was a dumb race condition.

Returning the error from `vips_threadpool_free()` feels like two ideas are being conflated. How about adding `vips_threadpool_wait()` to wait for all threads to finish, and calling that from `vips_threadpool_free()`? Then the end of `vips_threadpool_run()` can wait for all threads to finish, get any error, and free the threadpool. You wouldn't need to handle errors in the free function.

kleisauke:
Good idea. Addressed in the latest revision.

jcupitt:
Do you need the `wait` param? It's fine to call `vips_threadpool_free()` many times, I think.

kleisauke:
I wasn't sure about that one, let me try to remove it. It was inspired by [`g_thread_pool_free()`](https://docs.gtk.org/glib/method.ThreadPool.free.html).

kleisauke:
... you're right, it doesn't matter that `vips_threadpool_wait()` is now called twice. It makes the diff also easier to read now. :)

jcupitt (approved):

jcupitt:
Looks great!

PR #4122, opened 2024-08-26
text: reformat and polish

kleisauke:
Non-functional change.

### DIFF
--- a/libvips/create/text.c
+++ b/libvips/create/text.c
@@ -391,10 +391,10 @@ vips_text_build(VipsObject *object)
 	VipsObjectClass *class = VIPS_OBJECT_GET_CLASS(object);
 	VipsCreate *create = VIPS_CREATE(object);
 	VipsText *text = (VipsText *) object;
-	VipsImage **t = (VipsImage **) vips_object_local_array(object, 3);
+	VipsImage **t = (VipsImage **) vips_object_local_array(object, 2);
 
 	VipsRect extents;
-	VipsImage *image;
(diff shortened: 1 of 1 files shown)


kleisauke:
I converted it into a polish"PR targeting the `master` branch. This is now ready for review.

libvips/create/text.c
```diff
@@ -512,29 +512,25 @@ vips_text_build(VipsObject *object)
...
-		if (vips_extract_band(image, &t[1], 3, NULL) ||
-			vips_copy(t[1], &t[2],
-				"interpretation", VIPS_INTERPRETATION_MULTIBAND,
-				NULL))
+		if (vips_extract_band(in, &t[1], 3, NULL))
 			return -1;
-		image = t[2];
+		in = t[1];
+
+		in->Type = VIPS_INTERPRETATION_MULTIBAND;
```
jcupitt, line 530: Is there a danger that `t[1]` could be a shared image from the cache? I think this is why we originally had a `copy()` here (the official safe way to change `Type`).

kleisauke: Hmm, good point. Perhaps we should investigate other occurrences of this as well, e.g.:
https://github.com/libvips/libvips/blob/d52a983feb02c955f6422d862b27481b82e18fd6/libvips/create/point.c#L111-L113

But let's do that as follow-up. Reverted with commit ce89e78a1d2ef438c0ee9c03a4dfe2fd65fba15e.

PR #4123, opened 2024-08-27
threadpool: prefer use of dedicated GLib atomic operations

kleisauke:
That is:
`g_atomic_int_add(..., 1)` -> `g_atomic_int_inc(...)`
`g_atomic_int_add(..., -1)` -> `g_atomic_int_dec_and_test(...)`

Should be considered as a non-functional change.

PR #4129, opened 2024-08-29
threadpool: fix a race condition during downsizing

kleisauke:
Under certain conditions, particularly with short-lived pipelines and high thread concurrency, a race condition could occur between decrementing and restoring the `exit` flag atomically, potentially causing more threads to cooperatively exit than intended.

Fix this race by using a single atomic compare-and-exchange, which only resets the flag when it is set. Additionally, the flag's type has been changed from `int` to `gboolean` for clarity.

Context: #4069.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -5,6 +5,7 @@
 - webpsave: fix memory leak on error [dloebl]
 - heifsave: ensure NCLX profile is freed in lossless mode [kleisauke]
 - threadpool: fix a race condition in error handling [kleisauke]
+- threadpool: fix a race condition during downsizing [kleisauke]
 
 11/8/24 8.15.3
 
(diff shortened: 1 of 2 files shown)


kleisauke:
Upon a closer look, I don't think this is a race condition. The timing of the exit count increment in `vips_threadpool_run()` likely doesn't matter, even if it occurs between the decrement and increment of the exit count in `vips_worker_work_unit()`.

I do like the `g_atomic_int_compare_and_exchange()` and `gboolean` simplification though, so perhaps this PR should turn into a polishing PR. However, I'm uncertain whether it should target the `8.15` branch in that case.

jcupitt:
Good idea, let's make this a polish on master instead.

kleisauke:
Great! Closing in favor of PR #4123.

PR #4128, opened 2024-08-29
Silence a clang warning

kleisauke:
<details>
  <summary>Details</summary>

```console
[371/416] Compiling C object tools/vipsheader.p/vipsheader.c.o
../tools/vipsheader.c:229:8: warning: variable 'image' is used uninitialized whenever '||' condition is true [-Wsometimes-uninitialized]
  229 |                         if (!(source = vips_source_new_from_descriptor(0)) ||
      |                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../tools/vipsheader.c:241:7: note: uninitialized use occurs here
  241 |                 if (image &&
      |                     ^~~~~
../tools/vipsheader.c:229:8: note: remove the '||' if its condition is always false
  229 |                         if (!(source = vips_source_new_from_descriptor(0)) ||
      |                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../tools/vipsheader.c:221:19: note: initialize the variable 'image' to silence this warning
  221 |                 VipsImage *image;
      |                                 ^
      |                                  = NULL
1 warning generated.
```
</details>

Targets the 8.15 branch in case someone compiles with `-Werror`.

PR #4131, opened 2024-08-29
CI: build without *magick

kleisauke:
In line with commit 23998b42ce25987714e117d480bb30d718767529. This is pre-installed only on `macos-12` and `ubuntu-22.04`. Newer images no longer include it due to maintenance concerns, see:
https://github.com/actions/runner-images/issues/7899#issuecomment-1632115075

Context: https://github.com/libvips/libvips/pull/4128#issuecomment-2317374277.

Targets the 8.15 branch.


jcupitt:
Nice!

PR #4134, opened 2024-08-31
Simplify `GPrivate` initialisation

kleisauke:
Revise the code to move the initialisation to the global file scope, aligning with the example outlined in the GLib documentation: https://docs.gtk.org/glib/func.PRIVATE_INIT.html

This should be considered a non-functional change, at least on Linux, on Windows I'm not sure how thread-local storage should be initialised, so it targets the 8.15 branch just to be sure.

### DIFF
--- a/libvips/iofuncs/buffer.c
+++ b/libvips/iofuncs/buffer.c
@@ -89,7 +89,9 @@ static const int buffer_cache_max_reserve = 2;
 /* Workers have a BufferThread (and BufferCache) in a GPrivate they have
  * exclusive access to.
  */
-static GPrivate *buffer_thread_key = NULL;
+static void buffer_thread_destroy_notify(gpointer data);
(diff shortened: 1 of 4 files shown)


jcupitt:
I'd forgotten we'd bumped the glib version so high now. Sure, let's change this.

For 8.16 I guess we could remove `vips_g_mutex_new()` and friends too.

kleisauke:
Ah, indeed, it looks like the `GCond` and `GMutex` compat layer can also be removed/deprecated.

Note that when you merge this back into the `master` branch, this line should also be updated:
https://github.com/libvips/libvips/blob/7bc895bbc29bf7354c36e9ef1a68ab62f3b4ec00/libvips/iofuncs/threadpool.c#L472

jcupitt:
Ooo good point, thanks!

PR #4144, opened 2024-09-08
Use release version for C++ shared library SONAME

lovell:
A possible approach for the discussion at https://github.com/libvips/libvips/discussions/4141.

Where possible (i.e. not macOS), the C ABI version is retained as a prefix to ensure backwards compatibility when ordering against previous versions.

The project version is used as the suffix to ensure each release has a distinct `SONAME`.

```sh
$ readelf -a build/cplusplus/libvips-cpp.so | grep SONAME
 0x000000000000000e (SONAME)             Library soname: [libvips-cpp.so.42.8.15.4]
```
Targets the 8.15 branch.


jcupitt:
You're right, C++ ABI compat is just terrible -- I've been bitten by ABI breakages caused by minor compiler version changes. Maybe we should include the compiler name and version in the soname, so `libvips-cpp.so.42.17.4.8.15.4.clang-11`?

(hehe only joking)

I agree with this PR, anything to reduce the chances of errors like this.

jcupitt:
OK, done!

Issue #3336, opened 2023-02-15
Broken table formatting in documentation


andreas-kupries:
## Bug report

**Describe the bug**

Table formatting at https://www.libvips.org/API/current/libvips-create.html#vips-buildlut is broken.
It shows as
```
For example, consider this 2 x 2 matrix of (x, y) coordinates:

<tgroup cols='2' align='left' colsep='1' rowsep='1'> <tbody> <row> <entry>0</entry> <entry>0</entry> </row> <row> <entry>255</entry> <entry>100</entry> </row> </tbody> </tgroup>

We then generate:

<tgroup cols='2' align='left' colsep='1' rowsep='1'> <thead> <row> <entry>Index</entry> <entry>Value</entry> </row> </thead> <tbody> <row> <entry>0</entry> <entry>0</entry> </row> <row> <entry>1</entry> <entry>0.4</entry> </row> <row> <entry>...</entry> <entry>etc. by linear interpolation</entry> </row> <row> <entry>255</entry> <entry>100</entry> </row> </tbody> </tgroup>

This is then written as the output image, with the left column giving the index in the image to place the value.
```

This looks to me as if the table tags got overquoted to make them look like ordinary text.

**To Reproduce**

Referenced page was visited and read with a browser (firefox, although likely not brower-dependent).

**Expected behavior**

To see a proper table

**Actual behavior**

Table formatting shows as ordinary text

**Screenshots**

n/a

**Environment**

n/a, docs issue

**Additional context**

n/a

andreas-kupries:
Ditto https://www.libvips.org/API/current/libvips-create.html#vips-invertlut

jcupitt:
Argh, the old docbook markup is no longer supported in current gtkdoc. I've redone them in ascii art, thanks!

jcupitt:
There are a few others that need fixing too:

```
~/GIT/libvips/libvips (master) $ grep -l tgroup */*.c
arithmetic/add.c
arithmetic/arithmetic.c
arithmetic/divide.c
arithmetic/multiply.c
arithmetic/subtract.c
arithmetic/sum.c
deprecated/im_lab_morph.c
```

andreas-kupries:
Just seen, the `See Also` section of
https://www.libvips.org/API/current/libvips-generate.html#vips-image-pipelinev

PR #4149, opened 2024-09-12
Fix markup in generated gtk-doc

kleisauke:
Resolves: #3336.

Targets the 8.15 branch.

Context: https://github.com/libvips/libvips/discussions/4146#discussioncomment-10626247.

### DIFF
--- a/doc/meson.build
+++ b/doc/meson.build
@@ -51,7 +51,7 @@ private_libvips_headers = [
 
 private_headers = []
 foreach private_libvips_header : private_libvips_headers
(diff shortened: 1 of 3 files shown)


kleisauke:
Hmm, perhaps we should also revert commit 598471856aaf8cb694275da12c734ba578b9d543 and 870f9e5ee5f30f4cdb39c98e400150591ccab078 as a follow-up?

jcupitt:
Oh rats, I'd forgotten about that. How frustrating that they have no sane way to mark up tables.

jcupitt:
OK, let's do it! Well done for fixing this.

PR #4152, opened 2024-09-12
Added ability to specify Deflate (zlib) compression level through 'level' parameter

ruven:
There is currently no way to set the deflate compression level (through the TIFFTAG_ZIPQUALITY tag) when saving to TIFF. All deflate-encoded TIFFs, therefore, are compressed using the default level of 6.

This pull requests allows the existing `level` parameter, which is currently only used for zstd to also be usable for deflate.
If this parameter is not set, the compression level defaults to the libtiff default of 6 for Deflate and 9 for Zstd.

### DIFF
--- a/libvips/foreign/vips2tiff.c
+++ b/libvips/foreign/vips2tiff.c
@@ -856,8 +856,8 @@ wtiff_write_header(Wtiff *wtiff, Layer *layer)
 	}
 	if (wtiff->compression == COMPRESSION_ZSTD) {
 		// Set zstd compression level - only accept valid values (1-22)
-		if ((wtiff->level>=1) && (wtiff->level<=22))
-			TIFFSetField(tif, TIFFTAG_ZSTD_LEVEL, wtiff->level);
+		if (wtiff->level)
+			TIFFSetField(tif, TIFFTAG_ZSTD_LEVEL, VIPS_CLIP(1, wtiff->level, 22));
 		if (wtiff->predictor != VIPS_FOREIGN_TIFF_PREDICTOR_NONE)
 			TIFFSetField(tif,
 				TIFFTAG_PREDICTOR, wtiff->predictor);
@@ -865,8 +865,8 @@ wtiff_write_header(Wtiff *wtiff, Layer *layer)
 #endif /*HAVE_TIFF_COMPRESSION_WEBP*/
 
 	// Set zlib compression level - only accept valid values (1-9)
-	if ((wtiff->compression == COMPRESSION_ADOBE_DEFLATE) && (wtiff->level>=1) && (wtiff->level<=9))
-		TIFFSetField(tif, TIFFTAG_ZIPQUALITY, wtiff->level);
+	if ((wtiff->compression == COMPRESSION_ADOBE_DEFLATE) && (wtiff->level))
+		TIFFSetField(tif, TIFFTAG_ZIPQUALITY, VIPS_CLIP(1, wtiff->level, 9));
 
 	if ((wtiff->compression == COMPRESSION_ADOBE_DEFLATE ||
 			wtiff->compression == COMPRESSION_LZW) &&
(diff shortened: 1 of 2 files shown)


libvips/foreign/vips2tiff.c
```diff
@@ -855,13 +855,19 @@ wtiff_write_header(Wtiff *wtiff, Layer *layer)
 		TIFFSetField(tif, TIFFTAG_WEBP_LOSSLESS, wtiff->lossless);
 	}
 	if (wtiff->compression == COMPRESSION_ZSTD) {
-		TIFFSetField(tif, TIFFTAG_ZSTD_LEVEL, wtiff->level);
+		// Set zstd compression level - only accept valid values (1-22)
+		if ((wtiff->level>=1) && (wtiff->level<=22))
+			TIFFSetField(tif, TIFFTAG_ZSTD_LEVEL, wtiff->level);
```
jcupitt, line 860: How about removing the if() and just writing:

```C
TIFFSetField(tif, TIFFTAG_ZSTD_LEVEL, VIPS_CLIP(1, wtiff->level, 22));
```

Probably better than silently ignoring out of range values.

libvips/foreign/vips2tiff.c
```diff
@@ -855,13 +855,19 @@ wtiff_write_header(Wtiff *wtiff, Layer *layer)
...
 		if (wtiff->predictor != VIPS_FOREIGN_TIFF_PREDICTOR_NONE)
 			TIFFSetField(tif,
 				TIFFTAG_PREDICTOR, wtiff->predictor);
 	}
 #endif /*HAVE_TIFF_COMPRESSION_WEBP*/
 
+	// Set zlib compression level - only accept valid values (1-9)
+	if ((wtiff->compression == COMPRESSION_ADOBE_DEFLATE) && (wtiff->level>=1) && (wtiff->level<=9))
+		TIFFSetField(tif, TIFFTAG_ZIPQUALITY, wtiff->level);
```
jcupitt, line 869: We could just clip this as well.

jcupitt (commented):
Hey Ruven, thanks for this. I left a couple of small comments.

ruven:
I put these in an if block because `level` is initialized to zero by default in [vips_foreign_save_tiff_init()](https://github.com/ruven/libvips/blob/3646dc56d3878078ccc92f6e28658a7e3cd6b842/libvips/foreign/tiffsave.c#L425):

> `tiff->level = 0;`

So if the user doesn't explicitly set `level`, the libtiff defaults will be used, which happen to be 6 for zlib and 9 for zstd.

So if we want to keep these defaults, how about keeping the if, but combining it with VIPS_CLIP, so that we only use if the user has explicitly set a value. Something like this?

```
       if (wtiff->level)
            TIFFSetField(tif, TIFFTAG_ZSTD_LEVEL, VIPS_CLIP(1, wtiff->level, 22));
```

Alternatively we could have an identical default for both Deflate and Zstd (maybe not ideal). If we want vips itself to define different defaults for each codec, we would still need if blocks, but one which sets `level` to the default if it is zero. For example:

```
if ((wtiff->compression == COMPRESSION_ADOBE_DEFLATE) {
    if (wtiff->level == 0) wtiff->level = 6
    TIFFSetField(tif, TIFFTAG_ZIPQUALITY, VIPS_CLIP(1, wtiff->level, 9));
}

```
The first option above which leaves the default to libtiff itself to set is maybe the simplest solution.

jcupitt:
I agree, just checking for 0 seems easiest.

ruven:
OK, updated to just check for zero: https://github.com/libvips/libvips/pull/4152/commits/a86afe204128b745cb548791cc2f6fba2fc2ac63

jcupitt:
Great! Thanks for adding this.

jcupitt:
I added a note to the ChangeLog for this, and credited you.

Issue #4154, opened 2024-09-16
Crash in vips_foreign_load_jp2k_pack


mika-fischer:
Certain images from the [openjpeg test data repo](https://github.com/uclouvain/openjpeg-data) cause vips to crash

1. Download e.g. [Bretagne2_4.j2k](https://github.com/uclouvain/openjpeg-data/raw/master/baseline/nonregression/Bretagne2_4.j2k)
2. vips copy Bretagne2_4.j2k foo.jpg
3. Crash

This happens with the windows builds from 8.15.3 down until 8.12.0. 8.11.4 does not crash. Unfortunately these builds do not contain debug info. But with our custom builds we get the following stack traces on Linux and Windows:

```
(gdb) bt
#0  0x00007ffff69c0684 in vips_foreign_load_jp2k_pack () from ../lib/libvips.so.42
#1  0x00007ffff69c2ec5 in vips_foreign_load_jp2k_generate_untiled () from ../lib/libvips.so.42
#2  0x00007ffff6b7198e in vips_region_prepare_to_generate () from ../lib/libvips.so.42
#3  0x00007ffff6b7ece0 in vips_region_prepare_to () from ../lib/libvips.so.42
#4  0x00007ffff6acf552 in vips_tile_cache_gen () from ../lib/libvips.so.42
#5  0x00007ffff6b7044f in vips_region_generate () from ../lib/libvips.so.42
#6  0x00007ffff6b71085 in vips_region_fill () from ../lib/libvips.so.42
#7  0x00007ffff6b7eae9 in vips_region_prepare () from ../lib/libvips.so.42
#8  0x00007ffff6b61909 in vips_image_write_gen () from ../lib/libvips.so.42
#9  0x00007ffff6b7044f in vips_region_generate () from ../lib/libvips.so.42
#10 0x00007ffff6b71085 in vips_region_fill () from ../lib/libvips.so.42
#11 0x00007ffff6b7eae9 in vips_region_prepare () from ../lib/libvips.so.42
#12 0x00007ffff69bd439 in vips_foreign_load_generate () from ../lib/libvips.so.42
#13 0x00007ffff6b7044f in vips_region_generate () from ../lib/libvips.so.42
#14 0x00007ffff6b71085 in vips_region_fill () from ../lib/libvips.so.42
#15 0x00007ffff6b7eae9 in vips_region_prepare () from ../lib/libvips.so.42
[...]

PR #4158, opened 2024-09-18
make jp2kload work with the openjpeg test images

jcupitt:
We were not taking account of x0/y0 correctly.

See https://github.com/libvips/libvips/issues/4154

Thanks @mika-fischer


jcupitt:
Test data here:

https://github.com/uclouvain/openjpeg-data/tree/master/baseline/nonregression

With this PR, we now load `Bretagne2_4.j2k` correctly. 

`Bretagne2_1.j2k` is still crashing argh.

jcupitt:
OK, this now seems to load all the openjpeg test images correctly. 

Thanks again for reporting this, @mika-fischer 

This will be a pain to review :( sorry.

I added two tricky jp2k images from the openjpeg set to our test suite.

kleisauke:
Perhaps this should target the [`8.15`](https://github.com/libvips/libvips/tree/8.15) branch instead? Of course, I can also backport it with a follow-up PR if that's easier. 

FWIW, LLVM does this very neatly with a bot that automates the entire process via `/cherry-pick <commit> <...>`.
https://llvm.org/docs/GitHub.html#backporting-fixes-to-the-release-branches
https://github.com/llvm/llvm-project/blob/main/.github/workflows/issue-release-workflow.yml

I was thinking about adopting a similar approach for libvips, which would avoid the need to retarget PRs and regularly merging release branches back into the master branch.

> This will be a pain to review :( sorry.

Don't worry, reviewing this with the `Hide whitespace changes` option makes it a bit easier.

jcupitt:
It feels like a relatively large change, so I was thinking 8.16 would be a better place for it.

I'll have a look at the LLVM doc, thanks!

> Unfortunately, I still get crashes with the following files

Oh poop. I'll have a look.

I noticed imagemagick gets quite a few of the openjpeg test files wrong too, fwiw.

mika-fischer:
I feel your pain 😬

Just to be clear, I don't really care whether vips can read any of these files correctly. It just should not crash, since we use vips as a library...

jcupitt:
I agree it shouldn't crash.

You could just disable the libvips openjpeg loader, of course, that would probably be a safer fix. If you set the `VIPS_BLOCK_UNTRUSTED` env var, libvips won't run any loaders we haven't fuzzed ourselves.

mika-fischer:
> You could just disable the libvips openjpeg loader, of course, that would probably be a safer fix.

Sure, that's what we did for the moment (falling back to ffmpeg). But of course one of the reasons we use vips in the first place is to be able to read formats like JPEG2000.

> If you set the `VIPS_BLOCK_UNTRUSTED` env var, libvips won't run any loaders we haven't fuzzed ourselves.

I see, but we need JPEG2000, JPEG-XL & PPM, which we would lose that way, so that's not really a good option...

jcupitt:
It's now detecting when decoded pixels don't match the jpeg200 header and flagging an error, which seems to fix that set of crashes.

jcupitt:
> we need JPEG2000, JPEG-XL & PPM,

Ah OK. As long as your users aren't uploading these strange jpeg2000 files, then it should work now.

mika-fischer:
Great, thanks! I can confirm that the crashes are all gone and there are no valgrind warnings for the whole openjpeg folder.

jcupitt:
Great! Thanks for testing @mika-fischer, and thanks for reporting this.

kleisauke (approved):
LGTM!

PR #3663, opened 2023-09-18
v8.14.5 release proposal

kleisauke:
A few bugfixes since 8.14.4:

- fix a crash with alpha plus icc_import and icc_export [jcupitt]
- fix a crash in jxlsave [jcupitt]

PR #4162, opened 2024-09-20
github: add actions to automate backporting fixes

kleisauke:
This adds support for automatically cherry-picking fixes for the release branch using a special 'command' in comments:

```
/cherry-pick <commit1> <commit2> ...
```

This comment will attempt to cherry-pick the given commits to the release branch based on the milestone linked to the to issue or PR. If successful, the changes will be pushed to a new branch on GitHub, from which a pull request will be created.

Context: https://github.com/libvips/libvips/pull/4158#issuecomment-2361150922.

### DIFF
--- a/.github/utils/github-automation.py
+++ b/.github/utils/github-automation.py
@@ -1,0 +1,351 @@
+#!/usr/bin/env python3
+#
+# Originally from:
+# https://github.com/llvm/llvm-project/blob/main/llvm/utils/git/github-automation.py
+#
+# Preserved license:
+# ======- github-automation - LLVM GitHub Automation Routines--*- python -*--==#
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# ==-------------------------------------------------------------------------==#
+
+import argparse
+from git import Repo  # type: ignore
+import github
+import os
+import re
+import sys
+from typing import List, Optional
+
+
+def extract_commit_hash(arg: str):
+    """
+    Extract the commit hash from the argument passed to /action GitHub
+    comment actions. We currently only support passing the commit hash
+    directly or use the GitHub URL, such as
+    https://github.com/libvips/libvips/commit/1a86d4e153536e035d1907652391a26f77cbe1b8
+    """
+    github_prefix = "https://github.com/libvips/libvips/commit/"
+    if arg.startswith(github_prefix):
+        return arg[len(github_prefix):]
+    return arg
+
+
+class BackportWorkflow:
+    CHERRY_PICK_FAILED_LABEL = "cherry-pick-failed"
+
+    """
+    This class implements the sub-commands for the backport-workflow command.
+
+    The execute_command method will automatically choose the correct sub-command
+    based on the text in stdin.
+    """
+
+    def __init__(
+        self,
+        token: str,
+        repo: str,
+        issue_number: int,
+        requested_by: str,
+    ) -> None:
+        self._token = token
+        self._repo_name = repo
+        self._issue_number = issue_number
+        self._requested_by = requested_by
+
+    @property
+    def token(self) -> str:
+        return self._token
+
+    @property
+    def repo_name(self) -> str:
+        return self._repo_name
+
+    @property
+    def repo_owner(self) -> str:
+        return self.repo_name.split("/")[0]
+
+    @property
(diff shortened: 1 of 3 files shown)


kleisauke:
Example:
https://github.com/kleisauke/libvips/issues/5
https://github.com/kleisauke/libvips/pull/6

lovell:
I had a quick look to see if any existing Actions might do something similar, but they all appear to merge/rebase an entire PR rather than allow control over specific commits.

(Introducing the use of milestones would be a good improvement regardless.)

kleisauke:
FWIW, I just found a small issue in this script, which I reported upstream at https://github.com/llvm/llvm-project/issues/109429.

kleisauke:
> We probably also need to treat release branches as distinct branches after this lands (i.e. the 8.x branches do not need to be merged back into the master branch).

Another point worth discussing is the process for updating changelog notes for patch versions. Should this be done gradually (e.g. as part of the backport or as a separate commit)? Or all at once as part of the patch release?

FWIW, GLib does the latter (see e.g. commit https://github.com/GNOME/glib/commit/bd40fbba132fb3d6e5d785ae9d2b05a9e0da8409). However, these changes do _not_ seem to be reflected in the main branch (see e.g. https://github.com/GNOME/glib/blob/main/NEWS).

kleisauke:
This is ready for review now. To summarize:

- Release branches are treated as distinct branches after this lands (i.e. the 8.x branches do not need to be merged back into the master branch).
- All non-backport PRs should target the master branch (this also helps new contributors, see e.g. PR #4242).
- Backport PRs should be merged using GitHub's "Rebase and merge" option (to avoid duplicate references and unnecessary `Co-authored-by:` in commit messages).
- PRs that are likely to be backported (i.e. bug fixes) should _not_ update the ChangeLog directly. The ChangeLog will instead be updated all at once during the patch release (using commit messages as the basis for the entries).

Release branches are still created (i.e "branched") after a minor version is released and before new features are merged into the master branch. We could consider branching earlier (similar to [LLVM's approach](https://discourse.llvm.org/t/llvm-21-x-release-branch-created-main-is-now-22/87374)), but for now, the benefits of doing so appear minimal.

@jcupitt I can help with the releases for the 8.18 cycle (and potentially for any remaining 8.17.x releases) if that would be helpful.

> I just found a small issue in this script, which I reported upstream at https://github.com/llvm/llvm-project/issues/109429.

I opened PR https://github.com/llvm/llvm-project/pull/149653 for this upstream. The script in this PR already contains that change.

> Should this be done gradually (e.g. as part of the backport or as a separate commit)? Or all at once as part of the patch release?

I think the best approach is to update the ChangeLog notes at once in the patch release, since we need to update the date anyway (see e.g. PR #3663). The only caveat is that the ChangeLog on the master branch lists changes only for minor releases, not patch releases (which is probally fine as most users refer to the release notes at https://github.com/libvips/libvips/releases).

jcupitt (approved):

jcupitt:
Sorry for being so slow on this.

The end of an era! libvips enters the age of conventional git usage! I feel sad, but happy too. The modern world awaits.

kleisauke:
Great! I think it makes sense to experiment with this backport workflow during the 8.18.x cycle for commits that exist only on the master branch, but we should probably defer the following points to the 8.19.x cycle:
1. Treat release branches as distinct.
2. All non-backport PRs should target the master branch.
3. Update the ChangeLog notes all at once during a patch release.

If this turns out to be inconvenient, we can always fallback to the previous release workflow.

jcupitt:
I had another thought -- how about a quick cheat-sheet for the new workflow in CONTRIBUTING?

Maybe something like:

```
## New development

All development work goes into `master`, with stable branches forking from that 
roughly every six months. PRs should be squash-merged. 

If a PR adds API, it should be noted in a one-line comment in `ChangeLog`, crediting 
the author. This file is used to help prepare the release notes.

## Bugfixes

Bugfixes also go into `master`. Bugfixes that do not add API (all of them?) do not 
update `CHanegLog`.

If the fix should be backported to a stable branch, that branch should be set as 
the PR milestone, and the special comment `/cherry-pick` in one of the comments on 
the PR used to select the exact commits that need backporting.

    /cherry-pick <commit1> <commit2> ...

A script will automatically create a PR on the stable branch when the bugfix PR 
is merged to master.

PRs on stable branches and must be merged with rebase-and-merge. When a patch 
release is prepared for a stable branch, the release notes will be generated 
from the git log.
``` 

But you should probably write it!

kleisauke:
Good idea! I'll create a follow-up PR for that.

PR #4165, opened 2024-09-21
Update link to OSS-Fuzz issue tracker in README

kleisauke:
The OSS-Fuzz issue tracker has moved to a new location, see: https://github.com/google/oss-fuzz/issues/12508.


jcupitt:
:+1:

PR #4166, opened 2024-09-21
CI: remove redundant step

kleisauke:
This step was consolidated with the next one in commit 52698b8, but re-introduced due to the merge in commit 12dc3b4.


jcupitt:
Ooop!

PR #4164, opened 2024-09-20
call jpeg_set_colorspace for jpeg in tiff

jcupitt:
JPEG in TIFF compression needs the jpeg encoding colourspace set to match the enclosing tiff file.

fixes regression in 8.15.3 from https://github.com/libvips/libvips/pull/3924

see https://github.com/libvips/pyvips/issues/490

reproduce error with

```
vips copy x.jpg x.tif[compression=jpeg,Q=90,pyramid]
```

thanks nahilsobh

PR #4167, opened 2024-09-21
Polish JPEG chroma subsampling handling

kleisauke:
Non-functional change.

Context: https://github.com/libvips/libvips/pull/4164#issuecomment-2365133761.


jcupitt:
Ah so pretty!

PR #4174, opened 2024-09-29
Add jxl chunked save

jcupitt:
This PR adds support for the new libjxl chunked save API. 

We were preparing a whole frame of pixels then writing in one shot, but that's obviously not great for large images. This adds a new save path using JxlEncoderAddChunkedFrame(). With a 30k x 30k sample image I see:

```
$ /usr/bin/time -f %M:%e vips copy st-francis.jpg x.jxl
5719092:586.74
$ /usr/bin/time -f %M:%e vips copy st-francis.jpg x.jxl[chunked]
854360:573.48
```

It's very slightly faster, since it's able to overlap JPEG decode with JXL encode, and needs 7x less memory.

This is version 3 of this PR I think :( and is quite simple. The `NOTES` file outlines earlier (and crazilly complicated) schemes I tried.

TODO:

- this uses the iteration in JxlEncoderAddChunkedFrame() to drive libvips evaluation (and not `vips_sink_*()`), so it's missing things like `preeval`, `eval` etc. -- we should add these

- the old sequential save path can go once testing is over

- we'll need to bump the min libjxl version to at least 0.11

- check chunked load as well


jcupitt:
It seems (from #libjxl on discord) that we'll need to make a threadpool and generate libjxl tiles in parallel. We can't rely on `_data_at()` being executed in parallel.

jcupitt:
We are doing parallel pixel generation by running a libvips pipeline for each 2k x 2k JXL tile. But this means that we emit a `minimise` signal at the end of each crop, and this drops the cache from the `sequential` attached to the jpegload at the end of each tile. 

This PR works, but only by making `sequential()` persist over `minimise`. This will cause huge memory use.

We should probably bite the bullet and implement the incredibly complicated `jxlsave` outlined in `NOTES`.

jcupitt:
Ah, a better fix! Don't emit `minimise` for sub-evaluations, just as we don't emit `preeval`, `eval`, etc.

jcupitt:
Sample JXL HDR image for testing:

https://github.com/immich-app/immich/files/14755028/JPEG.XL.HDR.Test.zip

libvips loads this as a ushort, it looks fine, I guess the HDR element is in the transfer function. There's a synthetic profile attached, but it only supports `perceptual` intent (libvips is `relative` by default). 

```
$ vipsheader -f icc-profile-data JPEG_XL_HDR.jxl | base64 -d > hdr.icc
$ file hdr.icc 
hdr.icc: ColorSync color profile 4.4, type jxl, RGB/Lab-mntr device by jxl, 4344 bytes, 1-12-2019, 0xaa027a64eabe3514 MD5 'RGB_D65_DCI_Per_PeQ'

```
$ vips icc_import JPEG_XL_HDR.jxl x.v --intent perceptual
$ vipsheader x.v
x.v: 6000x4000 float, 3 bands, lab, jxlload
```

It looks nice:

![image](https://github.com/user-attachments/assets/d005fa54-feaf-4a89-a5b3-e99c2de21fd4)

You can get the LAB->sRGB conversion automatically like this:

```
$ vips icc_import JPEG_XL_HDR.jxl x.jpg --intent perceptual
```

Making this image:

![x](https://github.com/user-attachments/assets/4bf0d6d5-4d85-4eaa-8da4-956abda50842)

**Comment by jcupitt** [comment]
We'll should probably make libvips ICC import fall back to perceptual if relative is not supported.

**Comment by jcupitt** [comment]
Useful JXL HDR test images:

https://sneyers.info/hdrtest/

**Comment by jcupitt** [comment]
With libjxl 0.11.1 in release mode and the current version of this PR I see:

```
$ /usr/bin/time -f %M:%e vips copy st-francis.jpg x.jxl
memory: high-water mark 3.99 GB
5475176:59.19
$ /usr/bin/time -f %M:%e vips copy st-francis.jpg x.jxl[chunked]
memory: high-water mark 1.69 GB
2151556:66.88
```

So it's still better, though less than it used to be thanks to libjxl improvements.

I bump the min libjxl version requirement to 0.11, which rules out a lot of current linuxes.

I'll give it another once-over and see if I can can improve it a bit more.

File: libvips/foreign/jxlsave.c [hunk 1]
```diff
@@ -114,65 +122,207 @@ typedef struct _VipsForeignSaveJxl {
...
+{
+	jxl->input_source = (struct JxlChunkedFrameInputSource) {
+		.opaque = jxl,
+		.get_color_channels_pixel_format = vips_foreign_save_jxl_pixel_format,
+		.get_color_channel_data_at = vips_foreign_save_jxl_data_at,
+		.get_extra_channel_pixel_format =
+			vips_foreign_save_jxl_extra_pixel_format,
+		.get_extra_channel_data_at = vips_foreign_save_jxl_extra_data_at,
+		.release_buffer = vips_foreign_save_jxl_input_release_buffer
+	};
```
kleisauke, line 314:
```suggestion
	};

	g_mutex_init(&jxl->tile_lock);
```

libvips/foreign/jxlsave.c
```diff
@@ -302,6 +446,238 @@ vips_foreign_save_jxl_process_output(VipsForeignSaveJxl *jxl)
...
+{
+	VipsForeignSaveJxl *jxl = (VipsForeignSaveJxl *) gobject;
+
+	VIPS_UNREF(jxl->target);
+
+	VIPS_FREEF(JxlThreadParallelRunnerDestroy, jxl->runner);
+	VIPS_FREEF(JxlEncoderDestroy, jxl->encoder);
+
+	VIPS_FREEF(g_hash_table_destroy, jxl->tile_hash);
+
```
kleisauke, line 472:
```suggestion
	g_mutex_clear(&jxl->tile_lock);

```

kleisauke (commented):
`g_mutex_{init,clear}()` still needs to be done for mutexes that are part of dynamically allocated structures.

libvips/foreign/jxlsave.c
jcupitt: I think we can skip the init, since object members are inited to 0. You're right about the _clear though, oops.

kleisauke: Ah, you're right. It looks like GLib uses `g_malloc0()` for these structures.
https://github.com/GNOME/glib/blob/2.84.1/gobject/gtype.c#L407

I had assumed otherwise when I made commit 7e2b786c431c9a0a39ea80ffa6afa42e5df5589d, so we can likely drop many of the `g_mutex_init()` calls now. Nice!

libvips/foreign/jxlsave.c
```diff
@@ -389,93 +754,120 @@ vips_foreign_save_jxl_sink_disc(VipsRegion *region, VipsRect *area, void *a)
...
-				if (JxlEncoderAddBox(jxl->encoder, libjxl_metadata[i].box_type,
-						exif_data, exif_size, JXL_TRUE) != JXL_ENC_SUCCESS) {
-					vips_foreign_save_jxl_error(jxl, "JxlEncoderAddBox");
-					return -1;
-				}
+		if (jxl->delay &&
+			n < jxl->delay_length)
+			header.duration = jxl->delay[n];
+		else
+			header.duration = jxl->gif_delay * 10;
```
kleisauke, line 830: Incorporates the changes from 94502d8928f6838d0af57f2fa63bb54d51a85fc9 and 2672bb658893da7f1cc3a4b1660b154f4308ad84.

```suggestion
		if (!jxl->is_animated)
			header.duration = 0xffffffff;
		else
			header.duration =
				vips_foreign_save_jxl_get_delay(jxl, n);
```

libvips/foreign/jxlsave.c
```diff
@@ -389,93 +754,120 @@ vips_foreign_save_jxl_sink_disc(VipsRegion *region, VipsRect *area, void *a)
...
+	JxlEncoderSetFrameDistance(frame_settings, jxl->distance);
+	JxlEncoderFrameSettingsSetOption(frame_settings,
+		JXL_ENC_FRAME_SETTING_EFFORT, jxl->effort);
+	JxlEncoderSetFrameLossless(frame_settings, jxl->lossless);
 
-				/* The first 4 bytes is offset which is 0 in this case
-				 */
-				memcpy(exif_data + 4, data, length);
+	if (jxl->info.have_animation) {
+		JxlFrameHeader header = { 0 };
```
kleisauke, line 824: Nice! It looks like the `memset()` in `vips_foreign_save_jxl_add_frame()` can also be replaced with zero initialization  now (`= { 0 }`).

kleisauke:
macOS CI failures looks real, but I couldn't reproduce that locally (on Fedora 42 with libjxl 0.11.1).

libvips/foreign/jxlsave.c
jcupitt: ooo yes, nice spot

jcupitt:
> We could also consider keeping it behind a oneshot flag (like jp2kload), but I'm not sure if that's worth it.

Yes, having the old save path still there is useful for benchmarking and testing, but hopefully once this PR is done it'll no longer be useful.

jcupitt:
I've been looking into the macos crashes and it seems to be `g_mutex_clear()`.

For example, running `vips magickload` will make a `magick7` object, then dispose it without running `_build()`. This makes this code run:

```C
    ...
    VIPS_FREEF(magick_destroy_exception, magick7->exception);
    g_mutex_clear(&magick7->lock);
```

And `g_mutex_clear()` will crash inside `pthread_mutex_destroy`:

```
			default enum: none
			allowed enums: none, truncated, error, warning
   revalidate   - Don't use a cached result for this operation, input gboolean
			default: false
operation flags: nocache untrusted 
vips_foreign_load_magick7_dispose: 0x600002318700
Process 59255 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x8)
    frame #0: 0x000000018b08e2d0 libsystem_pthread.dylib`pthread_mutex_destroy + 40
libsystem_pthread.dylib`pthread_mutex_destroy:
->  0x18b08e2d0 <+40>: casa   w8, w9, [x19]
```

`->lock` has the value:

```
  lock = {
    p = 0x0000000000000000
    i = ([0] = 0, [1] = 0)
  }
```

ie. just zero (as you'd expect).

If I move the `g_mutex_init()` from `vips_foreign_load_magick7_build()` to `vips_foreign_load_magick7_init()`, breaking on `_dispose()` shows that lock now has the value:

```
  lock = {
    p = 0x000060000000e980
    i = ([0] = 59776, [1] = 24576)
  }
```

And `g_mutex_clear()` now runs successfully.

The docs say:

> Notice that the GMutex is not initialised to any particular value. Its placement in static storage ensures that it will be initialised to all-zeros, which is appropriate.

Which doesn't seem to be working. I'll try to make a small reproducer and report upstream.

jcupitt:
This seems to reproduce it:

```C
// compile with
//  cc mutex.c `pkg-config glib-2.0 --cflags --libs`

#include <stdio.h>
#include <glib.h>

GMutex lock_static;

int
main(void)
{   
    GMutex lock_stack = { 0 };

    printf("clearing lock in static storage ...\n");
    g_mutex_clear(&lock_static);

    printf("clearing lock on heap ...\n");
    g_mutex_clear(&lock_stack);

    return 0;
}
```

I see:

```
17:25 $ cc mutex.c `pkg-config glib-2.0 --cflags --libs`
✔ ~/try 
17:27 $ ./a.out
clearing lock in static storage ...
Segmentation fault: 11
✘-SEGV ~/try 
```

System:

```
17:27 $ brew info glib
==> glib: stable 2.84.1 (bottled)
17:28 $ uname -a
Darwin cheese-2.local 23.4.0 Darwin Kernel Version 23.4.0: Wed Feb 21 21:45:49 PST 2024; root:xnu-10063.101.15~2/RELEASE_ARM64_T6020 arm64
```

jcupitt:
I made an issue here: https://gitlab.gnome.org/GNOME/glib/-/issues/3682

jcupitt:
I found a bit more out: it happens when glib is NOT configured to use the native glib mutex and instead uses the platform POSIX mutex. It doesn't check for a null pointer before calling `g_mutex_impl_free()`.

It looks like `g_cond()` etc. probably have the same issue.

jcupitt:
The glib devs clarified the docs: if you have a mutex in static storage, it's ok to not init or clear it. It'll leak with some mutex backends, but since it's in static storage, the leak is bounded and doesn't matter.

For dynamically allocated memory, you must init exactly once and you must clear exactly once. Init to zero is not correct.

I think we're going to have to double-check all our GMutex usage :(

kleisauke:
Thanks for clarifying this upstream and fixing this!

> It looks like we must only call `g_mutex_clear()` once, so it should be called from `_finalize()`, not from `_dispose()` :(

Ah, bummer. This reminds me of this suboptimal approach:
https://github.com/libvips/libvips/blob/d0c4d1304fca19b4866d4845169bb1dde131cbdb/libvips/foreign/fits.c#L142-L143

jcupitt:
Yes, I saw that haha. Safe, but a little ugly.

ChangeLog
```diff
@@ -83,8 +85,12 @@ master
 - deprecate "cache" (use tilecache instead)
 - add tests for tokenisation
 - add "unpack_complex" option to vips_getpoint()
+- add zip level compression setting to vipsave [ruven]
 - add deflate compression level setting to tiffsave [ruven]
 - add "smart_deblock" to webpsave [goodusername123]
+- move vips_image_preeval(), vips_image_eval(), vips_image_posteval() into the
+  public API
```
kleisauke, line 92: This changelog note can be moved to 8.17.0 above.

ChangeLog
```diff
@@ -83,8 +85,12 @@ master
...
+- jxlsave uses the new chunked API for low memory use for large images
```
kleisauke, line 93: Can be removed in favor of the one in 8.17.0.

ChangeLog
```diff
@@ -83,8 +85,12 @@ master
 - deprecate "cache" (use tilecache instead)
 - add tests for tokenisation
 - add "unpack_complex" option to vips_getpoint()
+- add zip level compression setting to vipsave [ruven]
```
kleisauke, line 88: Can be removed, it looks like a leftover from rebase.

libvips/foreign/jp2kload.c
```diff
@@ -1068,6 +1068,9 @@ vips_foreign_load_jp2k_generate_tiled(VipsRegion *out,
 		// stop compiler warnings
 		VipsRect hit = { 0 };
 
+		// keep compilers happy
+		hit = (VipsRect) { 0 };
+
```
kleisauke, line 1073: This was already fixes via commit aefa8d11937bad7b06db21239dbcdff1b71d7ea8.

```suggestion
```

libvips/foreign/jxlload.c
```diff
@@ -965,15 +960,37 @@ vips_foreign_load_jxl_header(VipsForeignLoad *load)
...
+				.primaries = JXL_PRIMARIES_SRGB,
+				.transfer_function = JXL_TRANSFER_FUNCTION_SRGB,
+				.rendering_intent = JXL_RENDERING_INTENT_RELATIVE,
+			};
+			if (JxlDecoderSetPreferredColorProfile(jxl->decoder, &enc)) {
+				vips_foreign_load_jxl_error(jxl,
+					"JxlDecoderSetOutputColorProfile");
+				return -1;
+			}
+			if (JxlDecoderSetDesiredIntensityTarget(jxl->decoder, 10000)) {
```
kleisauke, line 988: Should we revert this to its original?

NOTES
kleisauke, line 1: Do we want to keep these notes?

libvips/foreign/jxlload.c
jcupitt: You're right, let's do the HDR stuff in a follow up PR.

kleisauke:
> With libjxl 0.11.1 in release mode and the current version of this PR I see:
> 
> ```
> $ /usr/bin/time -f %M:%e vips copy st-francis.jpg x.jxl
> memory: high-water mark 3.99 GB
> 5475176:59.19
> $ /usr/bin/time -f %M:%e vips copy st-francis.jpg x.jxl[chunked]
> memory: high-water mark 1.69 GB
> 2151556:66.88
> ```

FWIW, I see similar results on my AMD Ryzen 9 7900 workstation.

`master` branch:
```console
$ /usr/bin/time -f %M:%e vips copy ~/st-francis.jpg x.jxl
5533800:50.35
```

This PR:
```console
$ /usr/bin/time -f %M:%e vips copy st-francis.jpg x.jxl
1877692:53.78
```

So, ~3x less memory, similar performance and byte-identical images. Nice!

libvips/foreign/jxlsave.c
```diff
@@ -87,6 +88,10 @@ typedef struct _VipsForeignSaveJxl {
 	gboolean lossless;
 	int Q;
 
+	/* Set on API fail.
+	 */
+	gboolean error;
```
kleisauke, line 93: This seems to be unused, as it's set but never checked.

libvips/foreign/jxlsave.c
```diff
@@ -57,21 +59,20 @@
...
  * - add support encoding images with > 4 bands.
+ *
+ *		see FIXME note in _build()
  */
 
 #define OUTPUT_BUFFER_SIZE (4096)
 
+// max number of scanlines for a jxl tile (ysize arg to _data_at()) ... 2048
+// plus 8 lines of overlap top and bottom
+#define MAX_TILE_HEIGHT (2064)
```
kleisauke, line 74: This definition seems to be unused.

libvips/foreign/jxlsave.c
```diff
@@ -114,65 +120,227 @@ typedef struct _VipsForeignSaveJxl {
...
-
-	/* VipsRegion is not always contiguous, but we need contiguous RGB(A)
-	 * for libjxl. We need to copy each frame to a local buffer.
+	/* Buffer scanlines to make a line of libjxl tiles.
 	 */
-	VipsPel *frame_bytes;
-	size_t frame_size;
+	VipsPel *scanline_buffer;
+	size_t scanline_size;
+	int scanline_y;
```
kleisauke, line 127: These two variables seems to be unused.

libvips/foreign/jxlsave.c
```diff
@@ -114,65 +120,227 @@ typedef struct _VipsForeignSaveJxl {
...
+	 * the mutex.
+	 */
+	GHashTable *tile_hash;
+	GMutex tile_lock;
+
+	/* Current page we are saving.
+	 */
+	VipsImage *page;
+	VipsImage *page_colour;
+	VipsImage *page_extras;
```
kleisauke, line 147: These two images seems to be unused.

libvips/foreign/jxlsave.c
```diff
@@ -114,65 +120,227 @@ typedef struct _VipsForeignSaveJxl {
...
-	int page_count;
-	int page_number;
-
-	/* VipsRegion is not always contiguous, but we need contiguous RGB(A)
-	 * for libjxl. We need to copy each frame to a local buffer.
+	/* Buffer scanlines to make a line of libjxl tiles.
 	 */
-	VipsPel *frame_bytes;
-	size_t frame_size;
+	VipsPel *scanline_buffer;
```
kleisauke, line 125: This seems to be unused, as it's never set (but we try to `VIPS_FREEF()` it).

kleisauke (approved):
LGTM! I noticed a couple of unused things, but that can also be cleaned up in a follow-up PR.

libvips/foreign/jxlsave.c
jcupitt: Nice catch, I'll add some checks.

libvips/foreign/jxlsave.c
jcupitt: Left over from the old save loop I think, I'll remove them.

libvips/foreign/jxlsave.c
jcupitt: Removed!

libvips/foreign/jxlsave.c
jcupitt: I think they were for >4 band support, which we've now dropped for now. I'll remove them.

libvips/foreign/jxlsave.c
jcupitt: Removed.

jcupitt:
Phew! Thanks for the review!

PR #4188, opened 2024-10-08
Update C++ binding

kleisauke:
The now deprecated `VImage::cache()` function is still preserved to ensure ABI compat.

### DIFF
--- a/cplusplus/gen-operators.py
+++ b/cplusplus/gen-operators.py
@@ -7,6 +7,10 @@
 #
 #   pip install --user pyvips
 
+# Rebuild with:
+#   meson compile -Cbuild vips-operators-header
+#   meson compile -Cbuild vips-operators-source
+
 # Sample member declaration:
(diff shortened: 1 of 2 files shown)


cplusplus/gen-operators.py
```diff
@@ -128,6 +132,8 @@ def generate_operation(operation_name, declaration_only=False, indent=''):
             result += f'\n{indent} * @return {details["blurb"]}.'
 
         result += f'\n{indent} */\n'
+        if intro.flags & _OPERATION_DEPRECATED:
+            result += f'{indent}G_DEPRECATED\n'
```
kleisauke, line 136: Note that this uses `G_DEPRECATED` instead of `VIPS_DEPRECATED` since the `extern` specifier cannot be applied to class members.

jcupitt:
Should maybe target the 8.15 branch?

jcupitt:
:+1:

PR #4189, opened 2024-10-08
cplusplus: add `VImage::rawsave_fd` compat function

kleisauke:
Add a C++ compat function for `vips_rawsave_fd()` after the GObject class removal in commit 067b9f5.

https://kleisauke.nl/compat_reports/vips-cpp/8.15.5-rc1_to_rawsave_fd-compat/compat_report.html
https://kleisauke.nl/compat_reports/vips-cpp/8.15.5-rc1_to_master/compat_report.html

FWIW, I'm not sure if we could do anything about `VImage::new_from_memory_steal`, according to https://stackoverflow.com/a/5083976 changing to a const-qualified version would _not_ break ABI/API, at least in C99.

### DIFF
--- a/cplusplus/VImage.cpp
+++ b/cplusplus/VImage.cpp
@@ -1597,7 +1597,7 @@ operator>>=(VImage &a, const std::vector<double> b)
 void
 VImage::rawsave_fd(int fd, VOption *options) const
 {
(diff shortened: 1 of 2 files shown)


jcupitt:
Useful!

PR #4191, opened 2024-10-11
heifsave: expose libheif/aom auto_tiles feature

lovell:
I realise we don't want to expose every single possible underlying encoder feature, however given the relatively poor encoding times of AVIF images this one feels like the most useful addition to libaom I've seen in a while.

The `auto_tiles` property allows the aom encoder to determine the optimum tile count based on the output dimensions and available threads, which in my testing seriously improves multi-core utilisation. Elapsed time is almost halved in some cases. There are claims it reduces memory usage but I've not found much evidence of this yet. The cost of all this is very slightly increased file sizes, mostly from the additional metadata.

The upstream PR for this was https://github.com/strukturag/libheif/pull/1330 (requires libaom v3.10.0+ and the forthcoming libheif v1.18.3+).

### DIFF
--- a/cplusplus/include/vips/VImage8.h
+++ b/cplusplus/include/vips/VImage8.h
@@ -3446,6 +3446,7 @@ public:
 	 *   - **effort** -- CPU effort, int.
 	 *   - **subsample_mode** -- Select chroma subsample operation mode, VipsForeignSubsample.
 	 *   - **encoder** -- Select encoder to use, VipsForeignHeifEncoder.
+	 *   - **auto_tiles** -- Determine optimum tile count, bool.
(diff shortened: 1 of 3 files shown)


jcupitt:
Sure, that sounds like a very useful speedup.

I'll update ChangeLog and the 8.16 release notes.

jcupitt:
Sorry, I was thinking about this overnight. If the filesize increase is very small, how about we just enable this by default?

lovell:
Yes, I like that idea. I guess there's a possibility for artefacts to be introduced along the edges of tiles but I didn't notice this in my testing.

jcupitt:
OK, I made it always on.

Issue #4193, opened 2024-10-11
[8.16] removed symbols


remicollet:
From `abi-compliance-checker`

Lot of symbols have been moved to almostdeprecated.h so are no more available by default (vips_cache...)
Kind of API breakage implying fix to app using libvips

2 symbols are removed from headers
* vips_buf_appendsc
* vips__chomp

But it seems they were previously removed from the shared library, so probably not a real issue.

Can you please clarify ?

remicollet:
Also for the C++ API

```
- VImage::new_from_memory_steal ( void* data, size_t size, int width, int height, int bands, enum VipsBandFormat format ) [static]  1 
- VImage::new_from_memory_steal ( void const* data, size_t size, int width, int height, int bands, enum VipsBandFormat format ) [static]
```

From` abi-compliance-checker`

> The name of the appropriate symbol for this function on binary level has been changed from **_ZN4vips6VImage21new_from_memory_stealEPvmiii14VipsBandFormat** to **_ZN4vips6VImage21new_from_memory_stealEPKvmiii14VipsBandFormat**. This may cause undefined reference linker error in old client applications.

remicollet:
See reports

https://rpms.remirepo.net/compat_reports/libvips/8.15.5-rc1_to_8.16.0-rc1/compat_report.html (with almostdeprecated included)

https://rpms.remirepo.net/compat_reports/libvips-cpp/8.15.5-rc1_to_8.16.0-rc1/compat_report.html (without almostdeprecated)

remicollet:
About deprecation, having to fix vips PHP extension

See https://github.com/libvips/php-vips-ext/pull/55

Notice than including `almostdeprecated.h` generates tons of errors about unknown type, excerpt:

```
/usr/include/vips/almostdeprecated.h:400:9: error: unknown type name 'VipsPlotFn'; did you mean 'VipsStopFn'?
  400 |         VipsPlotFn fn, void *client1, void *client2, void *client3);
      |         ^~~~~~~~~~
      |         VipsStopFn
/usr/include/vips/almostdeprecated.h:403:17: error: unknown type name 'IMAGE'
  403 | int im_plotmask(IMAGE *im, int ix, int iy, PEL *ink, PEL *mask, VipsRect *r);
      |                 ^~~~~
/usr/include/vips/almostdeprecated.h:403:44: error: unknown type name 'PEL'
[...]

PR #4196, opened 2024-10-11
avoid direct include of almostdeprecated.h

remicollet:
In https://github.com/libvips/libvips/issues/4193#issuecomment-2407195135

> almostdeprecated.h cannot be included directly


jcupitt:
Helpful, thank you @remicollet!

PR #4225, opened 2024-10-27
Fix `vips_isprefix()` argument order

kleisauke:
The arguments were reversed, likely due to confusion with `g_str_has_prefix()`. Ensures all metadata is removed when saving with `keep=none`.

See: #4224.

Targets the 8.16 branch.


jcupitt:
I checked all other usage and the rest seem ok. Thanks for the PR!

PR #4231, opened 2024-10-31
add support for multipage jxl

jcupitt:
Multipage JXL images are just animations with the duration field set to 0xffffffff. This PR detects animations with this magic value for all frames and represents them as libvips multipage images (no duration metadata).

A matching change in jxlsave sets duration to 0xffffffff for multipage images.

There's no API change, and this looks pretty safe, so I think it'd be OK for 8.16.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+8.16.1
+
+- support multipage JXL
+
 10/10/24 8.16.0
 
 - allow small offsets for the PDF magic string [project0]
--- a/libvips/foreign/jxlload.c
+++ b/libvips/foreign/jxlload.c
@@ -104,8 +104,12 @@ typedef struct _VipsForeignLoadJxl {
 	uint8_t *xmp_data;
 
 	int frame_count;
-	int *delay;
-	int delay_count;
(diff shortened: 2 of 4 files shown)


jcupitt:
Sample multipage JXL file:

[plot-multipage.zip](https://github.com/user-attachments/files/17589096/plot-multipage.zip)

With this PR, I see:

```
$ vipsheader -a plot-multipage.jxl 
plot-multipage.jxl: 1500x1500 uchar, 3 bands, srgb, jxlload
width: 1500
height: 1500
bands: 3
format: uchar
coding: none
interpretation: srgb
xoffset: 0
yoffset: 0
xres: 1
yres: 1
filename: plot-multipage.jxl
vips-loader: jxlload
n-pages: 5
icc-profile-data: 732 bytes of binary data
orientation: 1
bits-per-sample: 8
```

ie. it looks like a multipage image, not an animated one. This also works correctly:

```
$ vips copy plot-multipage.jxl[n=-1] x.jxl
```

And you can view the multipage image in vipsdisp.

kleisauke (approved):
LGTM!

PR #4237, opened 2024-11-01
Add Magic Kernel support

### LINKED ISSUE #4089: add magic scaler
### Discussed in https://github.com/libvips/libvips/discussions/2155

<sup>Originally posted by **jcupitt** March 20, 2021</sup>
See: 

http://www.johncostella.com/magic

A description of the resize kernel used by facebook and instagram. It should be easy to add to libvips.</div>

akimon658:
Closes #4089

Added support for Magic Kernel Sharp (both 2013 and 2021 version).
Additions in reduceh.cpp are based on [this comment](https://github.com/libvips/libvips/pull/1592#issuecomment-626363031):

> Mask size is always calculated with this pattern: `2 * rint( mask_range * shrink ) + 1`, where the range is 1 for linear, 2 for cubic, 3 for lanczos3.


kleisauke:
Thanks for working on this! According to [`ResampleScope`](https://entropymine.com/resamplescope/), the 2013 variant seems incorrectly(?) implemented, see:
https://github.com/kleisauke/libvips-pr-4237#mks2013

Any thoughts on why? If not, I'd be glad to take a closer look.

libvips/resample/reduceh.cpp
```diff
@@ -125,6 +125,12 @@ vips_reduce_get_points(VipsKernel kernel, double shrink)
 	case VIPS_KERNEL_LANCZOS3:
 		return 2 * rint(3 * shrink) + 1;
 
+	case VIPS_KERNEL_MKS2013:
+		return 2 * rint(5 * shrink) + 1;
```
kleisauke, line 129:
```suggestion
		return 2 * rint(3 * shrink) + 1;
```

libvips/resample/templates.h
```diff
@@ -402,6 +402,48 @@ double inline filter<VIPS_KERNEL_LANCZOS3>(double x)
 	return 0.0;
 }
 
+template <>
+double inline filter<VIPS_KERNEL_MKS2013>(double x)
+{
+	if (x < 0.0)
+		x = -x;
+
+	if (x >= 5.0)
```
kleisauke, line 411:
```suggestion
	if (x >= 2.5)
```

kleisauke (commented):
It looks like the support window of the `VIPS_KERNEL_MKS2013` was wrong. For reference, here are the updated ResampleScope graphs:
https://github.com/kleisauke/libvips-pr-4237/tree/after-review#mks2013

lovell:
Thank you @akimon658 for the PR. Do you think it's worth including both kernels, or would we be better to focus on the more recent 2021 "edition" only? (In addition, I note this kernel is rather useful for edge detection, so a future possible improvement there also.)

libvips/resample/templates.h
akimon658: Sorry, I was misreading the number.

akimon658:
> Do you think it's worth including both kernels, or would we be better to focus on the more recent 2021 "edition" only?

Currently, my team is using the 2013 version. I think adding the 2013 edition would be beneficial as it allows us to migrate to libvips without altering application's behavior.
Additionally, it would not significantly increase the maintenance burden on the libvips team.

jcupitt:
Thanks for doing this work, @akimon658 !

Could you also:

- add it to the set of kernels in the test suite: https://github.com/libvips/libvips/blob/master/test/test-suite/test_resample.py#L77-L101
- add a line to ChangeLog.md, and credit yourself

libvips/resample/templates.h
```diff
@@ -402,6 +402,48 @@ double inline filter<VIPS_KERNEL_LANCZOS3>(double x)
...
+	if (x >= 2.5)
+		return 0.0;
+
+	if (x >= 1.5)
+		return -0.125 * (x - 2.5) * (x - 2.5);
+
+	if (x >= 0.5)
+		return 0.25 * (4.0 * x * x - 11.0 * x + 7.0);
+
+	return 1.0625 - 1.75 * x * x;
```
kleisauke, line 420: Nit: the non-optimized version is probably more readable:
```suggestion
	return 17.0 / 16.0 - 7.0 * x * x / 4.0;
```
(any C++ compiler will optimize it, so the assembly output should be equivalent)

Reference:
mks-2013-formula

akimon658:
In the ChangeLog file, which version should I put my change under?

kleisauke (approved):

jcupitt:
:+1:

j-p-c:
Ah, sorry I missed this—my son Jack just flagged it to me. I would have been happy to provide advice, but it looks like you have it sorted.

j-p-c:
> Hopefully! Thanks for checking @j-p-c

For extreme clarity: I have not checked the correctness of your implementation; I assume that that has been covered off.

PR #4238, opened 2024-11-02
Add Java bindings to the website

lopcode:
Adds the vips-ffm Java bindings to the website, to help folks discover them. I stuck the link at the end of the list, but do let me know if you'd like a different ordering.


jcupitt:
Good idea. Thanks!

PR #4243, opened 2024-11-03
reduce: prefer use of `VIPS_FABS()`

kleisauke:
To ensure we handle negative zero (`-0.0`) correctly, see e.g:
https://stackoverflow.com/a/48609079

### DIFF
--- a/libvips/resample/templates.h
+++ b/libvips/resample/templates.h
@@ -361,8 +361,7 @@ static double inline filter(double x);
 template <>
 double inline filter<VIPS_KERNEL_LINEAR>(double x)
 {
-	if (x < 0.0)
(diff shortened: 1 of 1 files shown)


lovell (approved):

jcupitt:
Should this maybe go into 8.16? It looks like a safe change.

kleisauke:
This also touches the new magic kernel functions, so it won't apply cleanly to 8.16. I'll open a separate backport PR once this lands.

jcupitt:
Ah, true. OK, approved!

kleisauke:
Backport to [`8.16`](https://github.com/libvips/libvips/tree/8.16) available as #4245.

PR #4246, opened 2024-11-05
Revise ChangeLog note

kleisauke:
Context: https://github.com/libvips/ruby-vips/pull/407#issuecomment-2455424610.


jcupitt:
I always get that the wrong way around :(

PR #4249, opened 2024-11-08
Deprecate `GCond `and `GMutex` compat layer

kleisauke:
This PR migrates and deprecates the following functions in favor of their GLib equivalents:
- `vips_g_mutex_new()` is replaced by `g_mutex_init()`
- `vips_g_mutex_free()` is replaced by `g_mutex_clear()`
- `vips_g_cond_new()` is replaced by `g_cond_init()`
- `vips_g_cond_free()` is replaced by `g_cond_clear()`

Additionally, explicit calls to `g_mutex_{init,clear}()` have been removed for mutexes initialized in static storage, as this is only necessary for mutexes that are part of dynamically allocated structures.

Context: https://github.com/libvips/libvips/pull/4134#issuecomment-2323316240.

The abi-compliance-checker results are available at:
https://kleisauke.nl/compat_reports/vips/master_to_deprecate-gcond-gmutex-compat-layer/compat_report.html

Marked as draft to investigate a potential ABI breakage in the `VipsImage`, `VipsSemaphore` and `VipsArea` structures.


kleisauke:
Commit 7f3432eb6e441e3e6b0b9857fb1698d8c287a54b should probably be split into its own PR. This refactor was needed to avoid that `g_mutex_clear()` is called more than once, moving that logic to `->dispose` is probably more appropriate. Note that I tried this first:
```diff
--- a/libvips/foreign/magick2vips.c
+++ b/libvips/foreign/magick2vips.c
@@ -172,6 +172,8 @@ typedef struct _Read {
 static void
 {
+	GMutex *lock = &read->lock;
+
 #ifdef DEBUG
 	printf("magick2vips: read_free: %s\n", read->filename);
 #endif /*DEBUG*/
@@ -181,7 +183,7 @@ read_free(Read *read)
 	VIPS_FREEF(DestroyImageInfo, read->image_info);
 	VIPS_FREE(read->frames);
 	VIPS_FREEF(magick_destroy_exception, read->exception);
-	g_mutex_clear(&read->lock);
+	VIPS_FREEF(g_mutex_clear, lock);
 }
 
 /* Can be called many times.
```

But that didn't work out. The FITS loader has a similar issue (which is also the reason why CI is failing):
https://github.com/libvips/libvips/blob/7f3432eb6e441e3e6b0b9857fb1698d8c287a54b/libvips/foreign/fits.c#L141-L144

Mizokuiam (commented):
Nice work! Here are some thoughts on potential optimizations...

kleisauke:
Commit 24a198af841ebde696a1f1c6347ab29f13ab768a ensures ABI compatibility for the public `VipsSemaphore` structure.  The `abi-compliance-checker` result looks OK now:
https://kleisauke.nl/compat_reports/vips/master_to_deprecate-gcond-gmutex-compat-layer/compat_report.html

This PR is still a draft due to dependence on #4375 (which makes the diff a bit cluttered).

libvips/foreign/fits.c
```diff
@@ -139,7 +139,8 @@ static void
 vips_fits_close(VipsFits *fits)
 {
 	VIPS_FREE(fits->filename);
-	VIPS_FREEF(vips_g_mutex_free, fits->lock);
+	if (fits->line)
+		g_mutex_clear(&fits->lock);
```
kleisauke, line 143: This isn't ideal. An alternative might be merging this with `fitsload.c`, but that is a much larger change.

jcupitt:
This looks OK to me, thank you for doing this horrible job.

I think the size of `VipsSemaphore` changing doesn't matter in practice since we don't document members and never allocate statically or on the stack (the API prevents that anyway), but I can see it's good to keep API checkers happy.

kleisauke:
> I think the size of `VipsSemaphore` changing doesn't matter in practice since we don't document members and never allocate statically or on the stack (the API prevents that anyway), but I can see it's good to keep API checkers happy.

Ah, you're right. Though, we didn't mark these `VipsSemaphore`  members previously as `/*< private >*/`, like we do in `VipsArea`. This PR adds that annotation.

How about a ping-pong strategy? We could land this PR as-is for 8.17 and then revert 2a20e414b4ca091a94967af58b070c1a85ace27d in 8.18. This would hopefully appease the API/ABI checkers.

jcupitt:
Sure, why not. OK to merge now?

kleisauke:
Sure, let's merge! :)

PR #4251, opened 2024-11-12
Fix `--vips-info` CLI flag with GLib >= 2.80

kleisauke:
Use `g_log_writer_default_set_debug_domains()` since resetting `G_MESSAGES_DEBUG` at runtime has no effect for GLib >= 2.80.

Also, remove any checks for the old `G_MESSAGES_DEBUG` env variable and overwrite it directly instead. This means the `--vips-info` CLI flag and the `VIPS_INFO=1` env variable now take precedence over `G_MESSAGES_DEBUG`.

See: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3710.

Targets the 8.16 branch.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,6 +4,7 @@
 - fix PFM byte order on big-endian machines [agoode]
 - morph: fix erode Highway path [kleisauke]
 - morph: fix C-paths with masks containing zero [kleisauke]
+- fix `--vips-info` CLI flag with GLib >= 2.80 [kleisauke]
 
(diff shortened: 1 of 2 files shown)


kleisauke:
FWIW, `VIPS_INFO=1` continued to work with GLib >= 2.80 because `vips_verbose()` was called before any log messages were written. However, this was not the case for the `--vips-info` CLI flag.

jcupitt:
I noticed it wasn't working, but I was too lazy to find out why :( Well done!

PR #4264, opened 2024-11-17
Avoid using `vips_object_argument_isset()` to check for flags

kleisauke:
For example, this ensures that `[no_strip=false]` or `[squash=false]` is properly handled.

Targets the 8.16 branch without a changelog entry, as these arguments are already deprecated.


jcupitt:
:+1:

PR #4266, opened 2024-11-17
Disable `unlimited` flag in fuzzing builds

kleisauke:
The load-specific `unlimited` flag is known to be fuzzing-unfriendly.

Targets the 8.16 branch to prevent CIFuzz from fuzzing this flag as well.

### DIFF
--- a/test/test-suite/test_foreign.py
+++ b/test/test-suite/test_foreign.py
@@ -1426,9 +1426,6 @@ class TestForeign:
             im = pyvips.Image.heifload(AVIF_FILE_HUGE)
             assert im.avg() == 0.0
 
-        im = pyvips.Image.heifload(AVIF_FILE_HUGE, unlimited=True)
-        assert im.avg() == 0.0
-
     @skip_if_no("heifsave")
     def test_avifsave(self):
         self.save_load_buffer("heifsave_buffer", "heifload_buffer",
--- a/libvips/foreign/heifload.c
+++ b/libvips/foreign/heifload.c
@@ -1107,12 +1107,14 @@ vips_foreign_load_heif_class_init(VipsForeignLoadHeifClass *class)
 		G_STRUCT_OFFSET(VipsForeignLoadHeif, autorotate),
 		FALSE);
 
+#ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
 	VIPS_ARG_BOOL(class, "unlimited", 22,
 		_("Unlimited"),
 		_("Remove all denial of service limits"),
 		VIPS_ARGUMENT_OPTIONAL_INPUT,
 		G_STRUCT_OFFSET(VipsForeignLoadHeif, unlimited),
 		FALSE);
+#endif
 }
 
 static gint64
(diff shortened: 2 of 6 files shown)


test/test-suite/test_foreign.py
```diff
@@ -1426,9 +1426,6 @@ def heif_valid(im):
             im = pyvips.Image.heifload(AVIF_FILE_HUGE)
             assert im.avg() == 0.0
 
-        im = pyvips.Image.heifload(AVIF_FILE_HUGE, unlimited=True)
-        assert im.avg() == 0.0
```
kleisauke, line 1430: This regression test was removed to prevent failures in the ASan/UBSan CI job. Additionally, it also seems to fail with libheif >= v1.19.0, see: https://github.com/strukturag/libheif/issues/1389.

lovell: I think we should use your good work in https://github.com/kleisauke/libvips/commit/5bf7cc60b080d397ef62d37840b664a0bafbc76b to deal with this, perhaps as a separate PR targeting the 8.16 branch.

kleisauke: This test was removed since we build libvips with `-DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` in our ASan/UBSan CI job.
https://github.com/libvips/libvips/blob/e38953c5dfa302875c4578ad1222b114ee6eac69/.github/workflows/ci.yml#L82

Work-in-progress commit https://github.com/kleisauke/libvips/commit/5bf7cc60b080d397ef62d37840b664a0bafbc76b increases the default maximum image size to 32768^2, but this appears to be problematic. See for example:
```console
$ vips black x.dz[container=fs,tile_size=256,overlap=0] 32768 32768
$ heif-enc -A -e aom -p speed=0 --tiled-input x_files/15/0_0.jpeg 32768x32768.avif
encoding tiled image, tile size: 256x256 image size: 32768x32768
encoding tile 128 128 (of 128x128)
$ vipsheader 32768x32768.avif[unlimited]
32768x32768.avif: 32768x32768 uchar, 3 bands, srgb, heifload
$ vips copy 32768x32768.avif[unlimited] x.jpg
$ coredumpctl debug
...
Core was generated by `vips copy 32768x32768.avif[unlimited] x.jpg'.
Program terminated with signal SIGSEGV, Segmentation fault.
(gdb) bt
#0  __memcpy_avx512_unaligned_erms () at ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:265
#1  0x00007fb7f090f706 in vips_foreign_load_heif_generate (out_region=0x7fb780092790 [VipsRegion], seq=<optimized out>, a=0x24e0e4a0, b=<optimized out>, stop=<optimized out>) at ../libvips/foreign/heifload.c:999
...
```

I think that `heifload` should be refactored into a tile-based loader before implementing such a change.

lovell: Thanks for the extra info Kleis, I agree.

lovell (approved):

Issue #2983, opened 2022-08-11
10x slowdown when converting to avif in some cases


L3tum:
## Bug report

**Describe the bug**
I'm not sure if the bug is in libvips, libheif or libaom, but you got to start somewhere...

When converting a PNG (specifically PNG, WEBP does not exhibit this for example) image to an AVIF image of the same dimensions, libvips takes 10x longer than when resizing the image as well.

**To Reproduce**
1. Use Image [source](https://user-images.githubusercontent.com/9307432/184093463-2e11d531-7b65-4bf8-8bfb-b7303316ba36.png)
2. `vipsthumbnail.exe .\source.png --size 3840x -o destination.heif[compression=av1,lossless=false,effort=2]`
3. libvips takes roughly 11 seconds on my machine (effort=2 to keep it lower).

**Expected behavior**
Running the above command slightly different results in it only taking ~1 second to be done.
 `vipsthumbnail.exe .\source.png --size 4000x -o destination.heif[compression=av1,lossless=false,effort=2]`

You can also downsize and observe the same behaviour.

 `vipsthumbnail.exe .\source.png --size 3400x -o destination.heif[compression=av1,lossless=false,effort=2]`

Or simply only change one dimension.

 `vipsthumbnail.exe .\source.png --size 3840x1080 -o destination.heif[compression=av1,lossless=false,effort=2]`

**Actual behavior**
Converting to AVIF without resizing the image takes ~10x as long.

**Screenshots**
````powershell
vipsthumbnail.exe --vips-info .\source.png --size 4000x -o destination.heif[compression=av1,effort=2,lossless=false,Q=80,strip=true]
VIPS-INFO: 10:10:54.474: thumbnailing .\artwork_landscape.png
VIPS-INFO: 10:10:54.491: selected loader is VipsForeignLoadPngFile
VIPS-INFO: 10:10:54.497: input size is 3840 x 2160
VIPS-INFO: 10:10:54.504: loading with factor 1 pre-shrink
VIPS-INFO: 10:10:54.519: pre-shrunk size is 3840 x 2160
VIPS-INFO: 10:10:54.524: converting to processing space srgb
VIPS-INFO: 10:10:54.529: residual scale 1,04167 x 1,04167
VIPS-INFO: 10:10:54.535: converting to output space srgb
VIPS-INFO: 10:10:54.540: thumbnailing .\source.png as .\destination.heif[compression=av1,effort=2,lossless=false,Q=80,strip=true]
````

````powershell
[...]

PR #4284, opened 2024-11-24
heifsave: prevent use of AV1 intra block copy feature

lovell:
Helps ensure AVIF encoding time is more predictable and consistent, fixes https://github.com/libvips/libvips/issues/2983

Remains backwards/forwards compatible with or without the upstream change at https://github.com/strukturag/libheif/pull/1408 as it will fail gracefully when the parameter is unavailable.

We might want consider exposing this as an option, either now or later.


kleisauke (approved):

farindk:
Please note that we have changed the name of the parameter to `enable-intrabc` to match the name in AOM.
Either change the name of the parameter or use `heif_encoder_set_parameter_string(..., "aom:enable-intrabc", "0")` to work with any version (even old releases without your PR patch).

Issue #4276, opened 2024-11-21
Libvips 8.16 consume more CPU and Memory than V8.15


euzada:
Hi,

I have a python script to stich a large mosaic (100k x 100k pixels).
I was surprise to notice a degradation in the performance of the latest version V8.16 compared to previous version 8.15.

It look like we lost the multithread feature. Here some statistic using the same files, same PC, same python script. Only difference is the libvips versions.

Same results on Python V3.6 to 3.12

V8.15.2 and V8.15.5
Loading 19 x 27 tiles. Total of 513 tiles.
Saving image of 102372 x 96390 pixels...
Image is now saved.
Process size im memory: 1513.25 MB
Elapsed time: 47s
Note: On windows, CPU usage around 40%, Memory less than 50%. Avegrage Hard drive writing 60MB/s

V8.16 and V8.16 RC2
Loading 19 x 27 tiles. Total of 513 tiles.
Saving image of 102372 x 96390 pixels...
Image is now saved.
Process size im memory: 855.43 MB
Elapsed time: 3m36s
Note: On windows, CPU usage hits quickly 100%, memory increases to almost 98% and takes more than 1min and half to drop an finish the file saving. Average Hard drive writing 10MB/s.

To replicate the results, use the build in the following link: [libvips/build-win64-mxe](https://github.com/libvips/build-win64-mxe/releases). To generate the final results I used tiffsave like the following:
```
im.tiffsave(output_filename, compression="jpeg", bigtiff=True, background=0, tile=True,
                    tile_width=256, tile_height=256, 
                    pyramid=True)
```

Any idea what get changed in V8.16?

jcupitt:
Hi @euzada, 

Could you post the complete python program to help me duplicate this problem? Otherwise I have to write it, and I probably won't match what you have. Also, how big are the tiles you are joining, and what format are they?

kleisauke:
Could you double-check whether this issue is indeed _not_ present in libvips 8.15.5? That would make bisecting this issue much easier.

Within the Windows binaries between version 8.15.2 and 8.15.5, there were a lot of dependency updates. In particular, libtiff was upgraded from 4.6.0 (released Sept. 13, 2023) to 4.7.0 (released Sept. 18, 2024), so one year worth of development.
https://github.com/libvips/build-win64-mxe/compare/v8.15.2..v8.15.5

In contrast, the changes between 8.15.5 and 8.16.0 are significantly smaller, involving only a upgrade of libvips itself.
[...]

PR #4293, opened 2024-11-30
threadpool: improve cooperative downsizing

kleisauke:
Turn the exit flag back into a proper count.

Fixes a regression introduced in commit 27229aa.

Context: #4276.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -8,6 +8,7 @@
 - make `subsample-mode=on` and `lossless=true` mutually exclusive [kleisauke]
 - fix SZI write with openslide4 [goran-hc]
 - heifsave: prevent use of AV1 intra block copy feature [lovell]
+- threadpool: improve cooperative downsizing [kleisauke]
 
 10/10/24 8.16.0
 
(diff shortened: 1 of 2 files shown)


jcupitt:
Ah, nice! I remember I did wonder about the flag (counts are usually much safer), but I couldn't think of a reason it wouldn't work. Threading is hard :(

kleisauke:
It seems the previous implementation was ~14% faster on the NetVips benchmarks. For example, the ~11ms performance gain shown in [this diff](https://github.com/kleisauke/net-vips/commit/7c3b3ef5428e2051ef54674fa5c22ec719d780f8#diff-30dbde008c7d28b9528184b6c4eefff2ff2699b5151eabb220ba934a28d5dffd) appears _only_ with libvips v8.16.0.

Though, I ran these benchmarks on Windows, I haven't (yet) checked if the results are the same on Linux.

kleisauke:
Ah, I forgot to reduce the concurrency in that benchmark, it's _much_ faster with `VIPS_CONCURRENCY=4`, see e.g. commit https://github.com/kleisauke/net-vips/commit/d5e01c66faddfaac1e3d31f71fd3f8fc76dc664d.

Issue #4317, opened 2024-12-12
Regression since version 15.5, "tiff2vips: read error"


CrispinStichartFNSB:
## Bug report

When using version 15.5 or 16.0 to process the attached image, vips crashes with `tiff2vips: read error`. 

Before the error, there are some warnings (the first one about the missing tag is repeated 8 times):

```
(vips.exe:25624): VIPS-WARNING **: 10:48:07.177: ASCII value for tag "Software" contains null byte in value; value incorrectly truncated during reading due to implementation limitations

(vips.exe:25624): VIPS-WARNING **: 10:48:07.285: Premature EOL at line 7212 of strip 0 (got 0, expected 9600)

(vips.exe:25624): VIPS-WARNING **: 10:48:07.287: error in tile 0 x 7200

(vips.exe:25624): VIPS-WARNING **: 10:48:07.288: error in tile 0 x 2150
```

When using version 15.3, it gives some warnings, but _does_ successfully process the image. It gives the above warning about the "Software" tag, as well as a bunch of:

```
(vips.exe:38280): VIPS-WARNING **: 10:50:42.521: Premature EOF at line 7212 of strip 0 (x 0)

(vips.exe:38280): VIPS-WARNING **: 10:50:42.522: Premature EOL at line 7212 of strip 0 (got 0, expected 9600)
```

I was not able to test with 15.4, as the [link to the windows binaries](https://github.com/libvips/build-win64-mxe/releases/tag/v8.15.4) on the release page 404s.

**To Reproduce**

1. Use [the attached image](https://github.com/user-attachments/files/18116811/error_tiff.zip)
2. Run `vips copy error_tiff.tif output.tif`

**Expected behavior**
It copies the image.

**Actual behavior**
It crashes with `tiff2vips: read error`. 

**Environment**

 - OS: Windows 11
 - Vips: 15.5 (16.0 also crashes with the same error)

**Additional context**

`tiffinfo` for the image:

```
> tiffinfo error_tiff.tif
TIFFFetchNormalTag: Warning, ASCII value for tag "Software" contains null byte in value; value incorrectly truncated during reading due to implementation limitations.
=== TIFF directory 0 ===
TIFF Directory at offset 0x8 (8)
  Subfile Type: (0 = 0x0)
  Image Width: 9600 Image Length: 7400
  Resolution: 400, 400 pixels/inch
  Bits/Sample: 1
  Compression Scheme: CCITT Group 4
  Photometric Interpretation: min-is-white
[...]

PR #4319, opened 2024-12-13
Fix libtiff 4.7 compat

jcupitt:
libtiff 4.7 now returns -1 for some warnings
    
Don't bail out if TIFFReadScanline() and TIFFReadEncodedStrip() return -1.
libtiff 4.7.0 and maybe earlier have started returning error values for
truncated strips.
    
Maybe other API points now return error codes for warnings too?
    
see https://github.com/libvips/libvips/issues/4317

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -10,6 +10,7 @@
 - heifsave: prevent use of AV1 intra block copy feature [lovell]
 - threadpool: improve cooperative downsizing [kleisauke]
 - fix alpha shift during colourspace conversions [frederikrosenberg]
+- improve error handling with libtiff 4.7 [CrispinStichartFNSB]
 
 10/10/24 8.16.0
 
--- a/libvips/foreign/tiff2vips.c
+++ b/libvips/foreign/tiff2vips.c
@@ -686,11 +686,10 @@ rtiff_strip_read(Rtiff *rtiff, int strip, tdata_t buf)
 		length = TIFFReadEncodedStrip(rtiff->tiff,
 			strip, buf, (tsize_t) -1);
 
-	if (length == -1) {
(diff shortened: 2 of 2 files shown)


lovell:
I think https://gitlab.com/libtiff/libtiff/-/merge_requests/628 might be the upstream change, which appears to zero out any scan lines that cannot be decoded due to the warning. Is it safe to ignore the return code with older versions of libtiff?

jcupitt:
Ah it's all unclear. Thinking about this again, I think (I think!) libtiff policy is to only return error codes for unrecoverable errors and not for warnings, so this change is probably a libtiff bug. 

I suppose in that case you're right and we should just `#ifdef` this workaround for 4.7.0. I'll ask on the libtiff mailing list.

jcupitt:
Test file from linked issue for reference:

https://github.com/user-attachments/files/18116811/error_tiff.zip

With libtiff 4.7.0, TIFFReadScanline() returns -1 for scanline 7212, while earlier libtiffs appear to keep on going.

jcupitt:
I made an issue on libtiff, let's see what they say.

https://gitlab.com/libtiff/libtiff/-/issues/657

jcupitt:
The result of the libtiff discussion is that we need to revise our error handling in tiff2vips. Read functions like `TIFFReadScanline()` can return non-zero for recoverable errors (though not always), so we should not unconditionally fail, but instead test fail-on.

lovell:
It sounds like this could build quite nicely on top of #4313, so perhaps we should target this fix at 8.17/master?

jcupitt:
I agree, let's junk this and try again on master with a more serious reworking of libtiff error handling. I'll make an issue so this isn't forgotten.

PR #4322, opened 2024-12-19
tiffload: add support for unlimited flag, requires libtiff 4.7.0+

lovell:
Adds a default limit of 20MB for total maximum ("cumulated") memory allocation during tag processing, which should be enough for hundreds if not thousands of tags. I could easily be persuaded to reduce this default limit.

Given libtiff is rather over-zealous in the number of times it reallocates memory, an image containing many thousands of tags can quickly lead to fragmentation. Attempting to process the <1MB sample image at https://issues.oss-fuzz.com/issues/42531230 *with* this memory limitation in place still results in an RSS of ~1GB (the OSS Fuzz issue should auto-close after this PR is merged).

Builds on https://github.com/libvips/libvips/pull/4313 and implements https://github.com/libvips/libvips/discussions/3892 for inclusion in 8.17.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,6 +3,7 @@
 - add `keep_duplicate_frames` option to GIF save [dloebl]
 - add Magic Kernel support [akimon658]
 - tiff: add threadsafe warning/error handlers (requires libtiff 4.5.0+) [lovell]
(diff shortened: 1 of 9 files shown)


jcupitt:
Great! Let's leave the 20mb for now, we can always tune this later.

PR #4329, opened 2024-12-26
Replace carrierwave-vips with CarrierWave in README.md

w-masahiro-ct:
`carrierwave-vips` has not been updated in over 5 years, so its usage is not recommended.
However, [`CarrierWave` now supports a method to directly integrate with `libvips`](https://github.com/carrierwaveuploader/carrierwave#using-vips).
Therefore, I replaced `carrierwave-vips` with `CarrierWave` in README.md.


jcupitt:
Thank you!

PR #4337, opened 2025-01-02
Fix null pointer dereference

peter15914:
The pointer ```class->suffs``` is dereferenced later in the function.
```
for (p = class->suffs; *p; p++)
	if (vips_iscasepostfix(filename, *p))
		return save_class;
```

I added ```return NULL``` to prevent NULL dereference.


lovell (approved):
Thank you for the PR Peter.

The suggestion of a possible dereference here is currently theoretical as all of the savers define `suffs` and this code is unreachable.

However it is definitely a good thing to have a guard to help with future development so that it remains that way.

Issue #4336, opened 2025-01-02
vips_image_get_as_string: output changed for GPSAltitudeRef tag


dloebl:
## Bug report

**Describe the bug**
The output of `vips_image_get_as_string()` changed for the `GPSAltitudeRef` field with v8.16.0. For example:
```
Sea  (Sea , Byte, 1 components, 1 bytes)
```
compared to 
```
Sea level (Sea level, Byte, 1 components, 1 bytes)
```

**To Reproduce**
Use image `test_files/gps_exif.jpeg` from [gps_exif.zip](https://github.com/user-attachments/files/18292260/gps_exif.zip)
```
vipsheader -f exif-ifd3-GPSAltitudeRef test_files/gps_exif.jpeg
```
**Output:** `Sea  (Sea , Byte, 1 components, 1 bytes)`

dloebl:
I've bisected this to 6323ef66630d22f8c78187dee839eb96eb408c36

lovell:
Thanks for reporting and bisecting this problem Daniel, there's a possible fix at https://github.com/libvips/libvips/pull/4339.

(See also https://github.com/libvips/libvips/issues/4002 for a discussion of possible future improvements relating to this.)

dloebl:
Thanks Lovell!
#4339 resolved this, so let's close this issue

PR #4339, opened 2025-01-03
exif: ensure enum entries can be converted to string

lovell:
Enumerated EXIF entries are often 1 byte in length but when represented as a string are up to around 30 chars in length.

This PR ensures support for this fairly common case by, when calculating the maximum length of the string, reducing the multiplier and adding a constant.

Fixes https://github.com/libvips/libvips/issues/4336 and also tested with the original image from https://github.com/libvips/libvips/issues/3857

Before:
```
exif-ifd2-SceneType: Dire (Dire, Undefined, 1 components, 1 bytes)
exif-ifd3-GPSAltitudeRef: Sea  (Sea , Byte, 1 components, 1 bytes)
```
After:
```
exif-ifd2-SceneType: Directly photographed (Directly photographed, Undefined, 1 components, 1 bytes)
exif-ifd3-GPSAltitudeRef: Sea level (Sea level, Byte, 1 components, 1 bytes)
```

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -13,6 +13,7 @@
 - heifsave: set image orientation using irot and imir transformations [lovell]
 - XYZ2Yxy: guard against divide by zero
 - fix MSVC compile error [na-trium-144]
+- exif: ensure enumerated entries can to converted to string values [lovell]
 
 10/10/24 8.16.0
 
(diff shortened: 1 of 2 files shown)


jcupitt:
:+1:!

PR #4344, opened 2025-01-08
gifsave: ensure return code is propagated

lovell:
Whilst investigating https://github.com/lovell/sharp/issues/4257 I discovered that `vips_sink_disc` appears to always return zero during `vips_foreign_save_cgif_build` even when `vips_foreign_save_cgif_sink_disc` returns a non-zero return code. It feels like a bit of a race condition, and I think this is because we need to have the entire input in memory before the CPU-intensive quantisation takes place.

This PR adds a new `aborted` property to track any non-zero return code, which then allows for things like an `eval` callback. The call to `vips_image_eval` occurs for each frame just after the image is remapped to the new palette as this is the most CPU-intensive section. I'd be happy to move the `vips_image_eval` call to a different location if required, or perhaps it might be better to add multiple calls given the CPU-intensive nature of quantisation as a whole.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -14,6 +14,7 @@
 - XYZ2Yxy: guard against divide by zero
 - fix MSVC compile error [na-trium-144]
 - exif: ensure enumerated entries can to converted to string values [lovell]
+- gifsave: add support for eval callback, ensure correct return code [lovell]
 
 10/10/24 8.16.0
 
--- a/libvips/foreign/cgifsave.c
+++ b/libvips/foreign/cgifsave.c
@@ -48,6 +48,7 @@
 #include <string.h>
 
 #include <vips/vips.h>
+#include <vips/internal.h>
 
 #include "pforeign.h"
 #include "quantise.h"
@@ -588,6 +589,12 @@ vips_foreign_save_cgif_write_frame(VipsForeignSaveCgif *cgif)
 	}
 
 	VIPS_FREEF(vips__quantise_image_destroy, image);
+
+	/* Remapping is relatively slow, trigger eval callbacks.
+	 */
+	vips_image_eval(cgif->in, n_pels);
+	if (vips_image_iskilled(cgif->in))
+		return -1;
 
 	/* Set up cgif on first use.
 	 */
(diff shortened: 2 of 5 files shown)


Mizokuiam (commented):
The implementation is clean. Consider adding more error handling...

Mizokuiam:
This is an interesting point. Have you considered...

jcupitt:
I did a little digging. `sink_disc` picks up error returns here:

https://github.com/libvips/libvips/blob/master/libvips/iofuncs/sinkdisc.c#L174

And handles them here:

https://github.com/libvips/libvips/blob/master/libvips/iofuncs/sinkdisc.c#L264

But it does this lazily, since the write is actually in a background thread. It can't detect a write error until the background thread that was doing the write terminates. 

The wait for the exit of the final write happens here:

https://github.com/libvips/libvips/blob/master/libvips/iofuncs/sinkdisc.c#L532

... but we've already set the return code on line 518, so a failure in the final write is never returned.

A simple fix might be to change the end of `vips_sink_disc()` to be:

```C
    /* Just before allocate signalled stop, it set write.buf writing. We
     * need to wait for this write to finish.
     *
     * We can't just free the buffers (which will wait for the bg threads
     * to finish), since the bg thread might see the kill before it gets a
     * chance to write.
     *
     * If the pool exited with an error, write.buf might not have been
     * started (if the allocate failed), and in any case, we don't care if
     * the final write went through or not.
     */
    if (!result)
        vips_semaphore_down(&write.buf->done);

    vips_image_posteval(im);

    /* The final write might have failed, pick up any error code.
     */
    result |= write.buf->write_errno;

    write_free(&write);

    vips_image_minimise_all(im);

    return result;
}
```

ie. OR in any code from the final thread. Though maybe we should also set an error message? Flush does this:

```C
        /* Previous write succeeded?
         */ 
        if (write->buf_back->write_errno) {
            vips_error_system(write->buf_back->write_errno,
                "wbuffer_write", "%s", _("write failed"));
            return -1;
        }   
```

So we should do the same for the final write, or maybe remove the message from flush.

jcupitt:
Nice!

Issue #3475, opened 2023-05-01
Add `default-intent` ICC rendering intent


jcupitt:
Profiles can optionally include a tag saying which intent they recommend. Using (this profile)[http://www.ifra.com/WebSite/ifra.nsf/html/CONT_ISO_DOWNLOADS] I see:

```
$ exiftool ~/pics/ISOnewspaper26v4.icc   
ExifTool Version Number         : 12.56
File Name                       : ISOnewspaper26v4.icc
Directory                       : /home/john/pics
File Size                       : 719 kB
File Modification Date/Time     : 2004:06:30 14:06:30+01:00
File Access Date/Time           : 2023:05:01 13:24:35+01:00
File Inode Change Date/Time     : 2023:05:01 13:21:56+01:00
File Permissions                : -rw-rw-r--
File Type                       : ICC
File Type Extension             : icc
MIME Type                       : application/vnd.iccprofile
Profile CMM Type                : Apple Computer Inc.
Profile Version                 : 2.4.0
Profile Class                   : Output Device Profile
Color Space Data                : CMYK
Profile Connection Space        : Lab
Profile Date Time               : 2004:06:28 20:53:23
Profile File Signature          : acsp
Primary Platform                : Apple Computer Inc.
CMM Flags                       : Not Embedded, Independent
Device Manufacturer             : Unknown (    )
Device Model                    : 
Device Attributes               : Reflective, Glossy, Positive, Color
Rendering Intent                : Perceptual
Connection Space Illuminant     : 0.9642 1 0.82491
Profile Creator                 : Unknown (LOGO)
Profile ID                      : 0
Profile Copyright               : Copyright by LOGO GmbH, Steinfurt
Chromatic Adaptation            : 1 0 0 0 1 0 0 0 1
Media White Point               : 0.6442 0.664 0.50038
Gray Tone Reproduction Curve    : (Binary data 1034 bytes, use -b option to extract)
A To B0                         : (Binary data 95052 bytes, use -b option to extract)
A To B1                         : (Binary data 95052 bytes, use -b option to extract)
A To B2                         : (Binary data 95052 bytes, use -b option to extract)
B To A0                         : (Binary data 132248 bytes, use -b option to extract)
B To A1                         : (Binary data 132248 bytes, use -b option to extract)
B To A2                         : (Binary data 132248 bytes, use -b option to extract)
Gamut                           : (Binary data 35414 bytes, use -b option to extract)
Profile Description             : ISOnewspaper26v4
```

[...]

PR #4347, opened 2025-01-20
colour: use suggested rendering intent as fallback

kleisauke:
When the requested rendering intent is not supported by the profile, fallback to the profile's suggested intent. An error will only be raised if the suggested intent is also unsupported.

Resolves: #3475.

Targets the 8.16 branch.

### DIFF
--- a/libvips/colour/icc_transform.c
+++ b/libvips/colour/icc_transform.c
@@ -159,6 +159,8 @@ typedef struct _VipsIcc {
 	int depth;
 	gboolean black_point_compensation;
 
+	VipsIntent selected_intent;
+
 	VipsBlob *in_blob;
 	cmsHPROFILE in_profile;
 	VipsBlob *out_blob;
@@ -446,7 +448,7 @@ vips_icc_build(VipsObject *object)
 	if (!(icc->trans = cmsCreateTransform(
 			  icc->in_profile, icc->in_icc_format,
 			  icc->out_profile, icc->out_icc_format,
-			  icc->intent, flags)))
+			  icc->selected_intent, flags)))
 		return -1;
 
 	if (VIPS_OBJECT_CLASS(vips_icc_parent_class)->build(object))
@@ -596,8 +598,8 @@ vips_image_is_profile_compatible(VipsImage *image, int profile_bands)
  * Don't set any errors since this is used to test compatibility.
  */
 static cmsHPROFILE
-vips_icc_load_profile_blob(VipsBlob *blob,
-	VipsImage *image, VipsIntent intent, int direction)
+vips_icc_load_profile_blob(VipsIcc *icc, VipsBlob *blob,
+	VipsImage *image, int direction)
 {
 	const void *data;
 	size_t size;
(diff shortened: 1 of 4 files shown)


lovell (approved):
This is really useful, thanks Kleis.

libvips/include/vips/colour.h
```diff
@@ -93,8 +93,12 @@ extern "C" {
 #define VIPS_D3250_Y0 (100.0)
 #define VIPS_D3250_Z0 (45.8501)
 
+/* Note: constants align with those defined in lcms2.h, expect for
```
jcupitt, line 96: Should be "except" I think.

libvips/colour/icc_transform.c
```diff
@@ -617,6 +618,9 @@ vips_icc_load_profile_blob(VipsBlob *blob,
 		return NULL;
 	}
 
+	if (icc->intent == VIPS_INTENT_AUTO)
+		icc->intent = (VipsIntent) cmsGetHeaderRenderingIntent(profile);
```
jcupitt, line 622: Will this break caching? Perhaps have a `VipsIntent selected_intent;` member that's not a constructor parameter.

libvips/colour/icc_transform.c
```diff
@@ -785,7 +789,7 @@ vips_icc_class_init(VipsIccClass *class)
 static void
 vips_icc_init(VipsIcc *icc)
 {
-	icc->intent = VIPS_INTENT_RELATIVE;
+	icc->intent = VIPS_INTENT_AUTO;
```
jcupitt, line 792: Won't changing the default break a lot of old code?

libvips/resample/thumbnail.c
```diff
@@ -1031,7 +1031,7 @@ vips_thumbnail_class_init(VipsThumbnailClass *class)
 		_("Rendering intent"),
 		VIPS_ARGUMENT_OPTIONAL_INPUT,
 		G_STRUCT_OFFSET(VipsThumbnail, intent),
-		VIPS_TYPE_INTENT, VIPS_INTENT_RELATIVE);
+		VIPS_TYPE_INTENT, VIPS_INTENT_AUTO);
```
jcupitt, line 1034: I'm not sure we can change this default.

jcupitt (commented):
I'm not sure we change the default intent. This could change the behaviour of a lot of code.

libvips/include/vips/colour.h
kleisauke: Ah, the infamous typo I always make :sweat_smile:. Thanks!

libvips/colour/icc_transform.c
jcupitt: How about picking the user's requested rendering intent, and if that's not available, falling back to the profile's suggested intent? Then only raising an error if both are missing.

We wouldn't need the AUTO setting, and libvips behaviour with old code would not change.

kleisauke: Good idea! I've incorporated this in the latest changeset and also targeted the [`8.16`](https://github.com/libvips/libvips/tree/8.16) branch instead.

jcupitt (approved):
woo!

PR #4349, opened 2025-01-21
tiffsave: honor disc threshold during pyramid save

kleisauke:
See: https://github.com/kleisauke/net-vips/issues/245.

Test case:
```console
$ curl -LO https://github.com/weserv/images/raw/5.x/test/api/fixtures/PalaisDuLouvre.tif
$ TMPDIR=/doesnotexist vips copy PalaisDuLouvre.tif x.tif[pyramid]
```

Targets the 8.16 branch.


lovell (approved):
:+1:

PR #4350, opened 2025-01-21
fix operation cache

jcupitt:
Quite a few operations change constructor parameters during `_build()`. This changes the operation hash and stops the operation cache working correctly.

For example:

- user sets operation params
- libvips calculates a hash (eg. computes the value 42) and checks the cache
- there are no hits, so the operation is built and the output is available
- the built operation is added to the cache
- but! `_build()` has modified one of the parameters, so the built operation is entered with hash 43
- next time the user uses the same parameters, they will again hash to 42 and you get a cache miss
- the operation is incorrectly built a second time!

The example I found that started this PR was `gifload` modifying the `n` param. I added something to check for the hash changing during build and the test suite now fails in *many* places. We seem to have this error in quite a few operations.

Test C code:

```C
/* compile with
 *
 *      gcc -g -Wall try346.c `pkg-config vips --cflags --libs`
 */

#include <vips/vips.h>

int
main(int argc, char **argv)
{   
    if (VIPS_INIT(argv[0]))
        vips_error_exit(NULL);

    VipsImage *image;
    double avg;

    if (vips_gifload(argv[1], &image, "n", -1, NULL) ||
        vips_avg(image, &avg, NULL))
        vips_error_exit(NULL);
    VIPS_UNREF(image);
    printf("avg = %g\n", avg);

    if (vips_gifload(argv[1], &image, "n", -1, NULL) ||
        vips_avg(image, &avg, NULL))
        vips_error_exit(NULL);
    VIPS_UNREF(image);
    printf("avg = %g\n", avg);

    return 0;
}      
```

Run with eg.:

```
$ VIPS_TRACE=1 ./a.out ~/pics/3198.gif 
vips cache+: gifload filename="/home/john/pics/3198.gif" out=((VipsImage*) 0x5881baafe5f0) n=139 flags=((VipsForeignFlags) VIPS_FOREIGN_SEQUENTIAL) -
vips cache : copy in=((VipsImage*) 0x5881baafe5f0) out=((VipsImage*) 0x5881bab0a0d0) -
vips cache+: linecache in=((VipsImage*) 0x71417000d320) out=((VipsImage*) 0x714170010210) tile-height=1 access=((VipsAccess) VIPS_ACCESS_SEQUENTIAL) -
vips cache+: sequential in=((VipsImage*) 0x71417000d320) out=((VipsImage*) 0x71417000ee20) tile-height=1 -
vips cache+: avg in=((VipsImage*) 0x5881baafe5f0) out=108.677329 -
avg = 108.677
vips cache : copy in=((VipsImage*) 0x5881bab129e0) out=((VipsImage*) 0x5881bab285b0) -
vips cache+: linecache in=((VipsImage*) 0x71413c008d90) out=((VipsImage*) 0x71413c00b640) tile-height=1 access=((VipsAccess) VIPS_ACCESS_SEQUENTIAL) -
vips cache+: sequential in=((VipsImage*) 0x71413c008d90) out=((VipsImage*) 0x71413c00a450) tile-height=1 -
vips cache+: avg in=((VipsImage*) 0x5881bab129e0) out=108.677329 -
avg = 108.677
```

Although it's not showing up, the `gifload` is actually running a second time (you can see there are no cache hits).

After this PR I see:

```
$ VIPS_TRACE=1 ./a.out ~/pics/3198.gif 
vips cache+: gifload filename="/home/john/pics/3198.gif" out=((VipsImage*) 0x5f47f0499600) n=-1 flags=((VipsForeignFlags) VIPS_FOREIGN_SEQUENTIAL) -
vips cache : copy in=((VipsImage*) 0x5f47f0499600) out=((VipsImage*) 0x5f47f04a50e0) -
vips cache+: linecache in=((VipsImage*) 0x74297000d4f0) out=((VipsImage*) 0x7429700103e0) tile-height=1 access=((VipsAccess) VIPS_ACCESS_SEQUENTIAL) -
vips cache+: sequential in=((VipsImage*) 0x74297000d4f0) out=((VipsImage*) 0x74297000eff0) tile-height=1 -
vips cache+: avg in=((VipsImage*) 0x5f47f0499600) out=108.677329 -
avg = 108.677
vips cache*: gifload filename="/home/john/pics/3198.gif" out=((VipsImage*) 0x5f47f0499600) n=-1 flags=((VipsForeignFlags) VIPS_FOREIGN_SEQUENTIAL) -
vips cache*: avg in=((VipsImage*) 0x5f47f0499600) out=108.677329 -
avg = 108.677
```

Now the second `gifload` and the second `avg` are both cache hits and return instantly. The cache is actually working!

### DIFF
--- a/libvips/foreign/dzsave.c
+++ b/libvips/foreign/dzsave.c
@@ -1957,7 +1957,7 @@ vips_foreign_save_dz_build(VipsObject *object)
 	 */
 	if (!vips_object_argument_isset(object, "keep") &&
 		!dz->no_strip)
-		save->keep = VIPS_FOREIGN_KEEP_NONE; // FIXME: Invalidates operation cache
+		save->keep = VIPS_FOREIGN_KEEP_NONE;
 
 	/* Google, zoomify and iiif default to zero overlap, ".jpg".
 	 */
@@ -1966,9 +1966,9 @@ vips_foreign_save_dz_build(VipsObject *object)
 		dz->layout == VIPS_FOREIGN_DZ_LAYOUT_IIIF ||
 		dz->layout == VIPS_FOREIGN_DZ_LAYOUT_IIIF3) {
 		if (!vips_object_argument_isset(object, "overlap"))
-			dz->overlap = 0; // FIXME: Invalidates operation cache
+			dz->overlap = 0;
 		if (!vips_object_argument_isset(object, "suffix"))
-			VIPS_SETSTR(dz->suffix, ".jpg"); // FIXME: Invalidates operation cache
+			VIPS_SETSTR(dz->suffix, ".jpg");
 	}
 
 	/* Google and zoomify default to 256 pixel tiles.
@@ -1976,7 +1976,7 @@ vips_foreign_save_dz_build(VipsObject *object)
 	if (dz->layout == VIPS_FOREIGN_DZ_LAYOUT_ZOOMIFY ||
 		dz->layout == VIPS_FOREIGN_DZ_LAYOUT_GOOGLE) {
 		if (!vips_object_argument_isset(object, "tile_size"))
-			dz->tile_size = 256; // FIXME: Invalidates operation cache
+			dz->tile_size = 256;
 	}
 
 	/* Some iiif writers default to 256, some to 512. We pick 512.
@@ -1984,14 +1984,14 @@ vips_foreign_save_dz_build(VipsObject *object)
 	if (dz->layout == VIPS_FOREIGN_DZ_LAYOUT_IIIF ||
 		dz->layout == VIPS_FOREIGN_DZ_LAYOUT_IIIF3) {
 		if (!vips_object_argument_isset(object, "tile_size"))
-			dz->tile_size = 512; // FIXME: Invalidates operation cache
+			dz->tile_size = 512;
 	}
 
 	/* skip_blanks defaults to 5 in google mode.
 	 */
 	if (dz->layout == VIPS_FOREIGN_DZ_LAYOUT_GOOGLE &&
 		!vips_object_argument_isset(object, "skip_blanks"))
-		dz->skip_blanks = 5; // FIXME: Invalidates operation cache
+		dz->skip_blanks = 5;
 
 	/* Our tile layout.
 	 */
@@ -2014,10 +2014,10 @@ vips_foreign_save_dz_build(VipsObject *object)
 	 */
 	if (dz->layout == VIPS_FOREIGN_DZ_LAYOUT_DZ) {
 		if (!vips_object_argument_isset(object, "depth"))
-			dz->depth = VIPS_FOREIGN_DZ_DEPTH_ONEPIXEL; // FIXME: Invalidates operation cache
+			dz->depth = VIPS_FOREIGN_DZ_DEPTH_ONEPIXEL;
 	}
 	else if (!vips_object_argument_isset(object, "depth"))
-		dz->depth = VIPS_FOREIGN_DZ_DEPTH_ONETILE; // FIXME: Invalidates operation cache
+		dz->depth = VIPS_FOREIGN_DZ_DEPTH_ONETILE;
 
 	/* We don't support onepixel pyramids in gmaps mode. It doesn't make a lot
 	 * of sense anyway, since gmaps tiles are never clipped, so you can't have
@@ -2025,7 +2025,7 @@ vips_foreign_save_dz_build(VipsObject *object)
 	 */
 	if (dz->layout == VIPS_FOREIGN_DZ_LAYOUT_GOOGLE &&
 		dz->depth == VIPS_FOREIGN_DZ_DEPTH_ONEPIXEL)
-		dz->depth = VIPS_FOREIGN_DZ_DEPTH_ONETILE; // FIXME: Invalidates operation cache
+		dz->depth = VIPS_FOREIGN_DZ_DEPTH_ONETILE;
 
 	if (VIPS_OBJECT_CLASS(vips_foreign_save_dz_parent_class)->build(object))
 		return -1;
@@ -2153,7 +2153,7 @@ vips_foreign_save_dz_build(VipsObject *object)
 		if (!vips_object_argument_isset(object, "imagename") &&
 			!vips_object_argument_isset(object, "basename")) {
 			if (filename) {
-				dz->imagename = g_path_get_basename(filename); // FIXME: Invalidates operation cache
+				dz->imagename = g_path_get_basename(filename);
 
 				/* Remove any [options] we may have picked up from the
 				 * filename.
@@ -2162,14 +2162,14 @@ vips_foreign_save_dz_build(VipsObject *object)
 					*p = '\0';
 			}
 			else
-				dz->imagename = g_strdup("untitled"); // FIXME: Invalidates operation cache
+				dz->imagename = g_strdup("untitled");
 		}
 
 		if (!vips_object_argument_isset(object, "dirname")) {
 			if (filename)
-				dz->dirname = g_path_get_dirname(filename); // FIXME: Invalidates operation cache
+				dz->dirname = g_path_get_dirname(filename);
 			else
-				dz->dirname = g_strdup("."); // FIXME: Invalidates operation cache
+				dz->dirname = g_strdup(".");
 		}
 	}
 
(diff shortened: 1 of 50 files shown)


jcupitt:
TODO:

- look through the test suite fails and make similar fixes where appropriate

jcupitt:
(trying to get nip4 to stop and restart gif animations on workspace scroll revealed this issue, ouch)

jcupitt:
It looks like 18 operations probably have problems:

```
FAILED test/test-suite/test_connection.py::TestConnection::test_connection_dz - pyvips.error.Error: unable to call dzsave_target
FAILED test/test-suite/test_conversion.py::TestConversion::test_smartcrop - pyvips.error.Error: unable to call smartcrop
FAILED test/test-suite/test_conversion.py::TestConversion::test_smartcrop_attention - pyvips.error.Error: unable to call smartcrop
FAILED test/test-suite/test_conversion.py::TestConversion::test_smartcrop_rgba - pyvips.error.Error: unable to call smartcrop
FAILED test/test-suite/test_conversion.py::TestConversion::test_smartcrop_rgba_premultiplied - pyvips.error.Error: unable to call smartcrop
FAILED test/test-suite/test_draw.py::TestDraw::test_draw_flood - pyvips.error.Error: unable to call draw_flood
FAILED test/test-suite/test_foreign.py::TestForeign::test_tiff - pyvips.error.Error: unable to call VipsForeignSaveTiffFile
FAILED test/test-suite/test_foreign.py::TestForeign::test_gifload - assert [100, 50, 50, 50, 50] == [0, 50, 50, 50, 50]
FAILED test/test-suite/test_foreign.py::TestForeign::test_dzsave - pyvips.error.Error: unable to call dzsave
FAILED test/test-suite/test_foreign.py::TestForeign::test_heicsave_16_to_12 - pyvips.error.Error: unable to call heifsave_buffer
FAILED test/test-suite/test_foreign.py::TestForeign::test_heicsave_16_to_8 - pyvips.error.Error: unable to call heifsave_buffer
FAILED test/test-suite/test_foreign.py::TestForeign::test_heicsave_8_to_16 - pyvips.error.Error: unable to call heifsave_buffer
FAILED test/test-suite/test_resample.py::TestResample::test_resize - pyvips.error.Error: unable to call resize
FAILED test/test-suite/test_resample.py::TestResample::test_shrink - pyvips.error.Error: unable to call shrink
FAILED test/test-suite/test_resample.py::TestResample::test_thumbnail - pyvips.error.Error: unable to call thumbnail
FAILED test/test-suite/test_resample.py::TestResample::test_similarity - pyvips.error.Error: unable to call similarity
FAILED test/test-suite/test_resample.py::TestResample::test_similarity_scale - pyvips.error.Error: unable to call similarity
FAILED test/test-suite/test_resample.py::TestResample::test_rotate - pyvips.error.Error: unable to call rotate
```

kleisauke:
I agree, we need to address this issue. Some occurrences are harder to spot, such as:
https://github.com/libvips/libvips/blob/aa6db7679d04df777fe39e4823bbf6a432c1fcc3/libvips/colour/icc_transform.c#L997-L1003

Perhaps we should (temporarily?) mark all these constructor parameters as `const` to find these issues at compile-time? I had a quick go with the now-fixed `gifload`, see e.g. commit https://github.com/kleisauke/libvips/commit/5a727119568a03c05941e201d367396c51d058b3.

(I'll open a separate PR for the other commits on that branch later - https://github.com/libvips/libvips/compare/fix-build-cache...kleisauke:fix-build-cache)

jcupitt:
I agree, it'd be good to spot this at compile-time.

The best solution would be to make code like this:

```C
    VIPS_ARG_DOUBLE(reducev_class, "vshrink", 3,
        _("Vshrink"),
        _("Vertical shrink factor"),
        VIPS_ARGUMENT_REQUIRED_INPUT,
        G_STRUCT_OFFSET(VipsReducev, vshrink),
        1.0, 1000000.0, 1.0);
```

Fail to compile if `G_STRUCT_OFFSET()` didn't return a pointer to a const, but I can't think of an easy way to do that.

I think we already spot things like `g_object_set()` on construct parameters after `construct` is set. Maybe we should set `construct` before calling `_build()`, rather than after? Although that might break a lot of things.

jcupitt:
The code in this PR makes finding this error pretty easy -- you see eg.:

```
reducev: arg "vshrink" changed during build, 
   VipsReducev (reducev), shrink an image vertically, 
   reducev in=((VipsImage*) 0x168a3630) out=((VipsImage*) 0x168a4fa0) 
      vshrink=2.000000 kernel=((VipsKernel) VIPS_KERNEL_LANCZOS3) gap=2.000000 -
```

So you just look for uses of `vshrink` in the build method.

This detector is behind a compile flag at the moment. Maybe it should be guarded by something like `VIPS_LEAK` instead? We could then run the test suite with it enabled.

kleisauke:
> The best solution would be to make code like this:
> 
> ```c
>     VIPS_ARG_DOUBLE(reducev_class, "vshrink", 3,
>         _("Vshrink"),
>         _("Vertical shrink factor"),
>         VIPS_ARGUMENT_REQUIRED_INPUT,
>         G_STRUCT_OFFSET(VipsReducev, vshrink),
>         1.0, 1000000.0, 1.0);
> ```
> 
> Fail to compile if `G_STRUCT_OFFSET()` didn't return a pointer to a const, but I can't think of an easy way to do that.

A potential GNU-style approach could be:

```c
#if G_GNUC_CHECK_VERSION(4, 8)
/* For G_STATIC_ASSERT define.
 */
#include <glib/gmacros.h>

#ifdef __cplusplus
#define IS_CONST_MEMBER(STRUCT, ELEM) \
	std::is_const<decltype(STRUCT::ELEM)>::value
#else
#define IS_CONST_MEMBER(STRUCT, ELEM) \
	__builtin_types_compatible_p( \
		glib_typeof((STRUCT) { 0 }.ELEM) *, \
		const glib_typeof((STRUCT) { 0 }.ELEM) *)
#endif

#define VIPS_STRUCT_CONST_OFFSET(STRUCT, ELEM) \
	({ \
		G_STATIC_ASSERT(IS_CONST_MEMBER(STRUCT, ELEM)); \
		G_STRUCT_OFFSET(STRUCT, ELEM); \
	})
#else
#define VIPS_STRUCT_CONST_OFFSET(STRUCT, ELEM) G_STRUCT_OFFSET(STRUCT, ELEM)
#endif /* G_GNUC_CHECK_VERSION (2, 0) */
```

Though, I'm not sure if this would be helpful. For instance, if you modify this:
https://github.com/libvips/libvips/blob/aa6db7679d04df777fe39e4823bbf6a432c1fcc3/libvips/resample/reducev.cpp#L113

to a `const` member, compilation would already fail with:
```
../libvips/resample/reducev.cpp: In function ‘int vips_reducev_build(VipsObject*)’:
../libvips/resample/reducev.cpp:880:42: error: assignment of read-only member ‘_VipsReducev::vshrink’
  880 |                         reducev->vshrink /= int_vshrink;
      |                         ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~
ninja: build stopped: subcommand failed.
```

jcupitt:
Adding something to `ARG_*()` would (hopefully) catch times we forgot to const-ify constructor paramaters.

jcupitt:
I like `__builtin_types_compatible_p()`! That is very cool.

kleisauke:
Actually, I think it's undefined behavior to make these struct members `const`, as `vips_object_{get,set}_property` accesses and modifies them in a non-`const` way. :(

jcupitt:
From memory, `vips_object_{get,set}_property()` will block set after construct, so it's enforcing the behaviour we need, just not at compile time :(

jcupitt:
> Actually, I think it's undefined behavior to make these struct members `const`, as `vips_object_{get,set}_property` accesses and modifies them in a non-`const` way. :(

Ah, you mean writing to a const structure member via a non-const pointer is undefined? Oh dear, that's bad :(

jcupitt:
SO suggests code like:

```C
typedef struct{
    const int *  x;
    const int *  y;
} const_point;

typedef struct{
    int *  x;
    int *  y;
} mutable_point;

typedef union __attribute__((__transparent_union__)) {
    const_point cpoint;
    mutable_point point;
} point;
```

So something like that might work, but it feels terrible.

kleisauke:
> Finding them all will need a lot of const-ifying. Maybe we should delay that for another PR?

I experimented by temporarily marking all input arguments as `const`, checking for compiler errors, noting the affected lines, and then reverting the const-ifying. This resulted in the following commit: https://github.com/kleisauke/libvips/commit/e1e147198b7914eb7c03c7a5e7189baf14b0a23e (though I'm not sure if it caught everything :sweat_smile:).

jcupitt:
Woah! That's nice. Shall I pull those FIXMEs into this PR?

kleisauke:
Feel free to pull that commit into this PR.

jcupitt:
Great! I added them.

I'm inclined to merge this and address the FIXMEs in follow-up PRs. What do you think?

libvips/foreign/dzsave.c
```diff
@@ -2009,38 +2009,23 @@ vips_foreign_save_dz_build(VipsObject *object)
...
-	/* Default to white background. vips_foreign_save_init() defaults to
-	 * black.
-	 */
-	if (!vips_object_argument_isset(object, "background")) {
-		VipsArrayDouble *background;
-
-		/* Using g_object_set() to set an input param in build will
-		 * change the hash and confuse caching, but we don't cache
-		 * savers, so it's fine.
-		 */
```
kleisauke, line 2021: Ah, I just noticed this comment. Perhaps all those FIXMEs in the savers can be removed?

kleisauke (approved):

kleisauke:
Merging this and address the FIXMEs in follow-up PRs sounds good to me!

PR #4354, opened 2025-01-23
Avoid `const`-qualifiers for input string arguments

kleisauke:
These arguments cannot be `const`, see for example:
https://github.com/libvips/libvips/blob/aa6db7679d04df777fe39e4823bbf6a432c1fcc3/libvips/iofuncs/object.c#L1240-L1247

### DIFF
--- a/libvips/colour/profile_load.c
+++ b/libvips/colour/profile_load.c
@@ -49,7 +49,7 @@
 typedef struct _VipsProfileLoad {
 	VipsOperation parent_instance;
 
-	const char *name;
+	char *name;
 	VipsBlob *profile;
 
 } VipsProfileLoad;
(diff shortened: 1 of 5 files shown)


jcupitt:
Hmm do you mean setting a const via a non-const pointer is undefined behaviour? Is this just "not supported by language spec", or is it "will break on some platforms"?

If the danger is only theoretical, I'd be inclined to keep the consts.

kleisauke:
Ah, you're right. I forgot that `const char *` is a mutable pointer to an immutable character/string. This means you can modify the pointer itself, but not the content of the memory it points to.

```c
/* compile with
 *
 *      gcc -g -Wall test-struct.c
 */

#include <stdio.h>

typedef struct _VipsForeignLoadPdf {
	const int password_int;
	const char *password;
} VipsForeignLoadPdf;

int main(void) {
	VipsForeignLoadPdf pdf = { .password_int = 41, .password = "41" };
	//pdf.password_int = 42; // error: assignment of read-only member ‘password_int’
	pdf.password = "42"; // This is fine.

	printf("%s\n", pdf.password);

	return 0;
}
```

I'll abandon this PR.

PR #4355, opened 2025-01-23
Ensure `VIPS_ARG_*` defaults are in-sync with `_init()`

kleisauke:


kleisauke:
Marking as draft to investigate CI failures.

jcupitt:
It'd be nice if there were some way to detect these mismatches automatically :( It's a nasty spot in the GObject API.

Thanks for finding these!

PR #4356, opened 2025-01-23
Simplify `@background` initialization

kleisauke:


jcupitt:
Oh! I wonder why that was using `VipsArea`? Anyway, good spot.

PR #4359, opened 2025-01-24
improve operation cache touch

jcupitt:
On a operation cache hit, we were only touching the hit itself.

As a result, after some operations on a loaded image, the image itself could be forced out by the operations that had been performed on it, even though the underlying image was still being used.

This PR makes cache touches also update the timestamps on all upstream operations, since the upstream operations are also being used (indirectly).


jcupitt:
The libvips operation cache was adapted from the one in nip2 and was supposed to be useful for nip4. 

Now nip4 is (finally!!!) coming to life, of course I'm finding all kinds of issues with the libvips version of the cache. This PR stops animated GIF display forcing load operations out of the cache because of the constant `crop` calls to pull frames out.

kleisauke (approved):

PR #4362, opened 2025-01-27
Prefer use of statically allocated mutexes

kleisauke:
Mutexes in static storage are initialised to all-zeros, which would avoid the need for explicit calls to `g_mutex_init()` and `g_mutex_clear()`.

Split out from #4249.


jcupitt:
Nice cleanup!

PR #4366, opened 2025-01-28
colour: add support for auto-selecting the rendering intent

kleisauke:
Via the `VIPS_INTENT_AUTO` enum value.

Resolves: #3475.

### DIFF
--- a/libvips/include/vips/colour.h
+++ b/libvips/include/vips/colour.h
@@ -93,13 +93,15 @@ extern "C" {
 #define VIPS_D3250_Y0 (100.0)
 #define VIPS_D3250_Z0 (45.8501)
 
-/* Note: constants align with those defined in lcms2.h.
+/* Note: constants align with those defined in lcms2.h, except for
+ * VIPS_INTENT_AUTO, which is libvips-specific.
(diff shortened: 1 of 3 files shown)


libvips/include/vips/colour.h
```diff
@@ -93,13 +93,15 @@ extern "C" {
...
-/* Note: constants align with those defined in lcms2.h.
+/* Note: constants align with those defined in lcms2.h, except for
+ * VIPS_INTENT_AUTO, which is libvips-specific.
  */
 typedef enum {
 	VIPS_INTENT_PERCEPTUAL = 0,
 	VIPS_INTENT_RELATIVE,
 	VIPS_INTENT_SATURATION,
 	VIPS_INTENT_ABSOLUTE,
+	VIPS_INTENT_AUTO,
```
kleisauke, line 104: Note: we cannot define `VIPS_INTENT_AUTO` as `-1` because it would conflict with the return code of `vips_enum_from_nick()`. Therefore, all supported enum values must always be greater than or equal to 0.

jcupitt (approved):

Issue #3210, opened 2022-12-01
should `g_warning` messages be internationalised?


jcupitt:
At the moment, `g_warning()` messages are sometimes internationalised, and sometimes not. We should pick a policy and apply it to all of libvips.

Suggestion: don't bother.

sandstrom:
Yes, I'd also suggest not localization of internal code.

(I know I've suggested this earlier, back in 2020, https://github.com/libvips/libvips/issues/1827)

PR #4369, opened 2025-01-30
Ensure g_warning messages use consistent i18n approach

lovell:
The recommended approach (a.k.a. "policy") is to not provide translations for warning messages.

I've manually removed only a subset of unused entries from the .po files to keep the diff clean. Once this PR is merged I'll do a follow-up to fully refresh the rest of the translation files e.g. remove all other no longer used entries, update line numbers etc.

Fixes https://github.com/libvips/libvips/issues/3210


jcupitt:
Well done for doing this annoying thing!

There are a few clang-format warnings, but they can be fixed in a follow-up.

PR #4370, opened 2025-01-30
i18n: refresh translations

lovell:
Mostly-automated:

- Apply `git clang-format` to commit da5803e
- Remove empty/outdated/non-automated headers from translation files
- Run `meson compile vips8.17-update-po` to automagically update translation files


jcupitt:
:+1:

PR #4378, opened 2025-02-06
Fix typo in ChangeLog

kleisauke:


lovell (approved):
"To be, or not to be, that is the question."

Issue #4393, opened 2025-02-19
improve heifsave fix


jcupitt:
We should revert #4392 and do a better fix in 8.17.

martindisch:
Much appreciated, because we started seeing problems when saving our PNGs as AVIF with 8.16.1👍

martindisch:
Is there a rough timeline for the 8.17 release or is there a lot of work left until it's ready?

jcupitt:
We're hoping to have 8.17 done within a few weeks, we're clearing out PRs and fixing final issues now. Though these things often slip, of course :(

PR #4394, opened 2025-02-19
heifsave: improve alpha channel detection

lovell:
Possible approach for https://github.com/libvips/libvips/issues/4393

### DIFF
--- a/libvips/foreign/heifsave.c
+++ b/libvips/foreign/heifsave.c
@@ -289,8 +289,7 @@ vips_foreign_save_heif_write_page(VipsForeignSaveHeif *heif, int page)
 #endif /*HAVE_HEIF_COLOR_PROFILE*/
 
 	options = heif_encoding_options_alloc();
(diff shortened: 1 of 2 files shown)


libvips/foreign/heifsave.c
```diff
@@ -289,8 +289,7 @@ vips_foreign_save_heif_write_page(VipsForeignSaveHeif *heif, int page)
 #endif /*HAVE_HEIF_COLOR_PROFILE*/
 
 	options = heif_encoding_options_alloc();
-	if (vips_image_hasalpha(save->ready))
-		options->save_alpha_channel = 1;
+	options->save_alpha_channel = save->ready->Bands > 3;
```
lovell, line 292: The default value for `save_alpha_channel` is already `true` so this change should make the intent more explicit.

kleisauke (approved):

lovell:
Updated to remove the merged-in fix from the 8.16 branch.

PR #4398, opened 2025-02-20
heifload: ensure unlimited flag removes all security limits where possible

lovell:
libheif v1.19.0 added new `heif_get_disabled_security_limits` API to remove *all* security limits.

I discovered this whilst investigating https://github.com/lovell/sharp/issues/4335 and removing all limits appears to provide a solution to it.

### DIFF
--- a/libvips/foreign/heifload.c
+++ b/libvips/foreign/heifload.c
@@ -354,6 +354,13 @@ vips_foreign_load_heif_build(VipsObject *object)
 		heif_context_set_maximum_image_size_limit(heif->ctx,
 			heif->unlimited ? USHRT_MAX : 0x4000);
 #endif /* HAVE_HEIF_SET_MAX_IMAGE_SIZE_LIMIT */
+#ifdef HAVE_HEIF_GET_DISABLED_SECURITY_LIMITS
+		if (heif->unlimited) {
+			const struct heif_security_limits *unlimited =
+				heif_get_disabled_security_limits();
+			heif_context_set_security_limits(heif->ctx, unlimited);
+		}
+#endif /* HAVE_HEIF_GET_DISABLED_SECURITY_LIMITS */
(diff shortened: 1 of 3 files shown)


kleisauke:
I think this is still problematic to do, see: https://github.com/libvips/libvips/pull/4266#discussion_r1856394879.

kleisauke:
... it would probably(?) be fine if you relocate this block above the `heif_context_set_maximum_image_size_limit()` call, that way the maximum image size of 16384^2 is still respected (and we hopefully won't segv).

lovell:
Ah, I thought this change felt familiar, I'd forgotten about your previous attempt. I think I'll try to re-work this to only remove the memory allocation limit when `unlimited` is set (rather than remove all security limits).

kleisauke:
PR #4399 should address that issue, so it's no longer a concern.

libvips/foreign/heifload.c
```diff
@@ -354,6 +354,13 @@ vips_foreign_load_heif_build(VipsObject *object)
 		heif_context_set_maximum_image_size_limit(heif->ctx,
 			heif->unlimited ? USHRT_MAX : 0x4000);
 #endif /* HAVE_HEIF_SET_MAX_IMAGE_SIZE_LIMIT */
+#ifdef HAVE_HEIF_GET_DISABLED_SECURITY_LIMITS
+		if (heif->unlimited) {
+			const struct heif_security_limits *unlimited =
+				heif_get_disabled_security_limits();
+			heif_context_set_security_limits(heif->ctx, unlimited);
+		}
```
kleisauke, line 362:
```suggestion
		if (heif->unlimited)
			heif_context_set_security_limits(heif->ctx,
				heif_get_disabled_security_limits());
```

kleisauke (approved):
LGTM once PR #4399 lands.

PR #3513, opened 2023-05-28
heifsave: limit output dimensions

lovell:
Here's a possible fix for https://github.com/libvips/libvips/issues/3509 targetting the 8.14 branch

It takes a similar approach as already used in webpsave:

https://github.com/libvips/libvips/blob/0df1fc51a9cc2a61777d8fba97189f5ceee68511/libvips/foreign/webpsave.c#L672-L675

The arbitrary 32768x32768 limit is to match the libheif default input limit. However, libvips' heifload operation defaults to a limit of 16384x16384, or 65535x65535 when using `unlimited`. Thoughts here welcome, as I can happily be convinced to use alternative arbitrary limits instead.

PR #4399, opened 2025-02-21
heifload: prevent possible int overflow for large images

kleisauke:
i.e. when the `unlimited` flag is set (> 16384x16384).

See: https://github.com/libvips/libvips/pull/4266#discussion_r1856394879.

Targets the 8.16 branch.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -23,6 +23,7 @@
 - matrixload: fix file format detect for some matrix types
 - radload: improve sanity check of colour-related headers [lovell]
 - heifsave: reject multiband images [lovell]
+- heifload: prevent possible int overflow for large images [kleisauke]
 
 10/10/24 8.16.0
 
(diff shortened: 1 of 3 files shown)


jcupitt:
Ah, nice!

I noticed that `heifsave.c` has almost the same issue with `stride`:

```C
    for (y = 0; y < area->height; y++) {
        /* Y in page. 
         */ 
        int page = (area->top + y) / heif->page_height;
        int line = (area->top + y) % heif->page_height;
        VipsPel *p = VIPS_REGION_ADDR(region, 0, area->top + y);
        VipsPel *q = heif->data + line * heif->stride;
```

This feels like an error in `heif_image_get_plane()`, though I suppose it'd be tricky for libheif to change their API.

kleisauke:
Ah, nice spot. `heifsave.c` is limited to a maximum image size of 16384x16384 (after PR #3513), but let me fix that occurrence too.

lovell (approved):
:+1:

Issue #4400, opened 2025-02-24
Libvips fails to process tif image when using streams


frederikrosenberg:
**Describe the bug**
When parsing in the attached tif file using pipes with `cat` vips fails to process the image and fails with an error.

**To Reproduce**
1. Use the [image](https://davidsen.cdn.bizzkit.cloud/5173c523-8575-45cb-8048-9faffa1ff00d/_original/lam-stegt-p%C3%A5-grill.tif)
2. Rename the image to `lam.tif`
3. Run the following command `cat lam.tif | vipsthumbnail stdin -o "lam.webp[keep=none]" --size 1000x`

**Expected behavior**
Expected an image to be created.

**Actual behavior**
Got the following error:
```text
(vips:173553): VIPS-WARNING **: 12:51:02.659: error in tile 0 x 0

(vips:173553): VIPS-WARNING **: 12:51:02.659: error in tile 0 x 14

(vips:173553): VIPS-WARNING **: 12:51:02.660: error in tile 0 x 0

(vips:173553): VIPS-WARNING **: 12:51:02.660: error in tile 0 x 10
TIFFFillStrip: Read error at scanline 4294967295; got 66620 bytes, expected 110657
tiff2vips: read error
```

**Environment**
 - OS: WSL Ubuntu
 - Vips: 8.16.0

**Additional context**
I'm able to process the image without using streaming using `vipsthumbnail lam.tif -o "lam.webp[keep=none]" --size 1000x` with no warnings.

Other images does not observe the same problem.

jcupitt:
Hello @frederikrosenberg,

Curious! It's working fine for me -- I tried 8.16.0 and 8.16.1.

TIFF can't be read with streaming (you need to be able to seek the image), so libvips ought to automatically buffer the whole file in memory before letting the TIFF loader execute. 

I think I remember a bug in an earlier version where this didn't always happen, but 8.16 should be fine. 

Is there any chance you have an earlier libvips installed as well as 8.16? Could the binaries have become mixed up?

frederikrosenberg:
Hello @jcupitt 

Quite interesting. I'm using a locally built libvips from the 8.16 branch and I'm still seeing the error

Version returns the following:
```
> ./vipsthumbnail --version
vips-8.16.1
```

frederikrosenberg:
Hi @jcupitt 

I did some more testing and found that if I use Ubuntu 24.04 LTS in WSL it works. There may be a dependency version causing the problem.

[...]

PR #4403, opened 2025-02-28
tiffload: add missing read loop

kleisauke:
We weren't looping on `vips_source_read()` in tiffload, which could cause failures when reading from pipes.

Resolves: #4400.

Targets the 8.16 branch.

Note that I couldn't find any documentation for this callback, but I noticed that `TIFFOpen*()` behaves similarly, see:
https://gitlab.com/libtiff/libtiff/-/blob/v4.7.0/libtiff/tif_unix.c#L68-95

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -24,6 +24,7 @@
 - radload: improve sanity check of colour-related headers [lovell]
 - heifsave: reject multiband images [lovell]
 - heifload: prevent possible int overflow for large images [kleisauke]
+- tiffload: add missing read loop [kleisauke]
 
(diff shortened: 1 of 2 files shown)


jcupitt:
Well good grief, well done for finding this Kleis. This is a new requirement in libtiff 4.6 and later, is that right?

jcupitt:
I had a look at openout_target_write() and openout_target_read(), but I think they are OK.

jcupitt:
It'll depend on timing as well, so an SSD could work but a slow HDD fail (I think).

PR #4407, opened 2025-03-05
test: reduce severity of `gifsave_timeout`

kleisauke:
On my AMD Ryzen 9 7900 workstation, the `gifsave_timeout` test consistently fails with:
```console
$ meson test -C build --print-errorlogs
...
 8/12 gifsave_timeout          FAIL             2.46s   exit status 1
――― ✀ ―――
expected error return from vips_gifsave_buffer()
――――――――――
```

I suspect the image gets quantized/remapped within 2 seconds, so we can't guarantee that it always returns an error. The best we could do is `g_assert(is_killed)`, which is a no-op in release builds.

Targets the 8.16 branch.

### DIFF
--- a/test/test_timeout_gifsave.c
+++ b/test/test_timeout_gifsave.c
@@ -18,7 +18,6 @@ main(int argc, char **argv)
 	void *buf;
 	size_t len;
 	gboolean is_killed = FALSE;
-	int ret;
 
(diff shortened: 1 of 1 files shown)


lovell (approved):
Alas, further evidence that timeout-based testing is hard :rofl:

PR #4412, opened 2025-03-07
Avoid possible overflow when multiplication result is cast up/down

lovell:
I'm not aware of any specific problems these might have been causing but forcing the relevant type for multiplication should help prevent vulnerabilities in the future.

These were all found via CodeQL.


libvips/colour/LCh2UCS.c
```diff
@@ -165,7 +165,7 @@ vips_col_Ch2hcmc(float C, float h)
 	}
 
 	P = cos(VIPS_RAD(k7 * h + k8));
-	D = k4 + k5 * P * pow(VIPS_FABS(P), k6);
+	D = k4 + k5 * P * (float) pow(VIPS_FABS(P), k6);
```
kleisauke, line 168: As possible follow-up, we could consider using `powf()` and `fabsf()` here, as they are guaranteed to be present in C99.

libvips/colour/LCh2UCS.c
```diff
@@ -165,7 +165,7 @@ vips_col_Ch2hcmc(float C, float h)
...
 	g = C * C * C * C;
 	f = sqrt(g / (g + 1900.0));
```
kleisauke, line 170: ditto, we might use `sqrtf()` here.

libvips/conversion/composite.cpp
```diff
@@ -899,7 +899,7 @@ vips_composite_base_blend3(VipsCompositeSequence *seq,
 			/* You can't sqrt a vector, so we must loop.
 			 */
 			for (int b = 0; b < 3; b++) {
-				double g;
+				float g;
```
kleisauke, line 902: ditto, we could use `g = sqrtf(B[b]);` below.

kleisauke (approved):
Nice!

libvips/colour/LCh2UCS.c
lovell: Good idea, I agree. For consistency, do you think we should define new macros that wrap the `__builtin_*` versions of these, i.e. `VIPS_POWF` and `VIPS_FABSF`?

kleisauke: I wonder if these `__builtin_*` are really needed. IIUC, [`-fno-math-errno`](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-fno-math-errno) would achieve the same effect. That is, speed things up by assuming that these math functions never set `errno`. See for example:
https://ryanburn.com/2020/12/26/why-c-standard-math-functions-are-slow/

kleisauke: According to Godbolt, there is no difference between our current `__builtin_*` functions and the `<math.h>` functions in both GCC and Clang when compiling with `-O2`. `-fno-math-errno` also doesn't appear to make any difference - see:
https://godbolt.org/z/G1vEMP46h

I think commit https://github.com/libvips/libvips/commit/90c1fb9a9444dde41e3db8fd18c41f83c31c7e5f was due to https://sourceware.org/bugzilla/show_bug.cgi?id=17441, which was resolved in glibc 2.23 via commit https://github.com/bminor/glibc/commit/8df4e219e43a4a257d0759b54fef8c488e2f282e.

libvips/conversion/composite.cpp
lovell: My understanding of C++ is that `sqrt` is overloaded for both float and double precision so I think we can get away with it here.

libvips/colour/LCh2UCS.c
lovell: I've incorporated these suggestions as part of this PR as it's a clearer solution for avoiding possible overflow in the `vips_col_Ch2hcmc` function (and there's no loss of precision).

libvips/conversion/composite.cpp
kleisauke: You're right; this isn't an issue in C++. :+1:

libvips/colour/LCh2UCS.c
kleisauke: Great! LGTM.

PR #4414, opened 2025-03-09
Deprecate math macros

kleisauke:
We can simply use the functions from `<math.h>` instead of conditionally relying on GCC and Clang's `__builtin_*` functions.

The likely reason for commit 90c1fb9a9444dde41e3db8fd18c41f83c31c7e5f was due to https://sourceware.org/bugzilla/show_bug.cgi?id=17441, which was resolved in glibc 2.23.

Context: https://github.com/libvips/libvips/pull/4412#discussion_r1986142253.

### DIFF
--- a/libvips/arithmetic/abs.c
+++ b/libvips/arithmetic/abs.c
@@ -116,7 +116,7 @@ vips_abs_build(VipsObject *object)
 		int x; \
 \
 		for (x = 0; x < sz; x++) \
-			q[x] = VIPS_FABS(p[x]); \
+			q[x] = fabs(p[x]); \
 	}
 
 /* Complex abs operation: calculate modulus.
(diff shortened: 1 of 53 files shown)


jcupitt:
Ah funny, `__builtin` used to be a lot faster back in the day. I read this quickly and it seems fine, so :+1: from me.

Yes, conceptually `VIPS_ABS()` is like `VIPS_MAX()` or `VIPS_MIN()`, and not like one of the math functions.

kleisauke:
Thanks! Note that when you merge [`8.16`](https://github.com/libvips/libvips/tree/8.16) back into the master branch, it will likely conflict in `‎libvips/colour/LCh2UCS.c` due to commit 82c7c05cb02a52750251bb4cc69d67f40568cf98. In that case, `fabsf()` should be preferred over `fabs()`.

jcupitt:
Thanks for the tip! Done.

PR #4418, opened 2025-03-12
convi: ensure double sum precision for floats

lovell:
The `CONV_FLOAT` inner loop macro is used for both `float` and `double` images so the type of `sum` should probably match.

I've not had a chance to performance test this change but it has the possibility of making integer convolution of float images a tiny bit faster.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -12,6 +12,7 @@
 - svgload: add support for custom CSS via stylesheet option [lovell]
 - heifload: `unlimited` flag removes all limits (requires libheif 1.19.0+) [lovell]
 - heifsave: improve alpha channel detection [lovell]
+- convi: ensure double sum precision for floats [lovell]
 
 8.16.1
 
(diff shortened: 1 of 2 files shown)


jcupitt:
Hmm, this is likely to be slower, isn't it? Though I've not tried a benchmark ahem. C always uses double for all intermediate values, so:

```C
    float a, b = ... something;
    float d = a * b;
```

Compiles to something like:

```C
    float d = double_to_float(double_multiply(float_to_double(a), float_to_double(b)));
```

If you just use double everywhere you'll save a lot of unnecessary conversions. 

float can useful for specifying storage (eg. `float a[10]` is nice and small), but you should generally use double everywhere else.

jcupitt:
Actually, I say that, but now I can't find a reference to float -> double promotion in the standard. I'll keep looking.

lovell:
Yes, this might be a C99 thing. I think the problem here is that `t` is integer and `p` could be a `float`, which isn't automatically promoted. I've updated the PR to explicitly cast the multiplication up to `double` precision instead.

jcupitt:
I tried with compiler explorer and this:

```C
float fred(float a, float b, float c) { return a * b + c; }
```

Generates single-precision float mul and add instructions on all compilers I tested, so I'm wrong heh. Perhaps I was thinking of the promotion rules around untyped arguments.

PR #4421, opened 2025-03-17
dzsave: ensure IIIF scaleFactors includes all levels

lovell:
Originally reported at https://github.com/lovell/sharp/issues/4346 - the sample image of 39125x34708 split into 1024x1024 tiles correctly generated 7 levels in the IIIF pyramid but it looks like the `scaleFactors` calculation was only considering the first 6.


jcupitt:
LGTM! what a silly error :(

PR #4424, opened 2025-03-19
heifload: improve detection of seek beyond EOF

lovell:
Ensures we check both the `old_position` and `result` for a possible error when seeking.

Fixes https://issues.oss-fuzz.com/issues/404288018


jcupitt:
Oh nice. Does this stop the annoying bad seeks errors we were getting?

lovell:
Not sure, perhaps we should check the return value of `old_position` sooner so we don't even perform the `SEEK_SET`?

PR #4429, opened 2025-03-21
Prevent possible use-after-free during vips_shutdown when also using deprecated vips7 plugins

lovell:
The impact of this is negligible as it's in deprecated shutdown code that almost no-one will (or should) be using, but sensible to fix all the same. At worst I guess a libvips process could have crashed during process shutdown.


jcupitt:
I think this code was actually OK --- after the `g_free()`, the thing plug points to is invalid, but the value of the pointer (as a key in the plugin list) is fine. But I agree it looks tidier to do the free later.

ChangeLog
```diff
@@ -14,6 +14,7 @@
 - heifsave: improve alpha channel detection [lovell]
 - convi: ensure double sum precision for floats [lovell]
 - improve guard against corrupt ICC profiles with older lcms2 versions [kleisauke]
+- prevent use-after-free during `vips_shutdown` with deprecated plugins [lovell]
```
jcupitt, line 17: Maybe remove this line, since it's not a use-after-free bug?

lovell: Yes, I agree, I didn't want this to sound alarmist and yet, reading it back, it definitely sounds alarmist. I've updated the commit message too.

jcupitt:
:+1:

Issue #4205, opened 2024-10-15
`jp2kload: Invalid tile part index` error on some JPEG 2000 files


mbklein:
## Bug report

**Describe the bug**
When loading certain JP2 files (using the `vips` CLI or the `sharp` library for NodeJS), it fails with an error similar to the following:
```
(vips:17855): VIPS-WARNING **: 16:12:11.207: error in tile 1024 x 0
jp2kload: Invalid tile part index for tile number 1. Got 7, expected 0

jp2kload: Fail to read the current marker segment (0xff90)

jp2kload: Failed to decode the codestream in the JP2 file
```

**To Reproduce**

The file in question came from a downstream user of my project, and I haven't yet gotten clearance to attach it to a public GitHub issue. If there's another way I can send it (e.g., via email), I would be happy to do so. I will attach it to a subsequent comment if and when I get permission.

Vips CLI:

```shell
$ vipsheader y1980-25.jp2
y1980-25.jp2: 1772x2000 uchar, 3 bands, srgb, jp2kload
$ vips tiffsave y1980-25.jp2 y1980-25.tif

(vips:17855): VIPS-WARNING **: 16:12:11.207: error in tile 1024 x 0
jp2kload: Invalid tile part index for tile number 1. Got 7, expected 0

jp2kload: Fail to read the current marker segment (0xff90)

jp2kload: Failed to decode the codestream in the JP2 file
```

sharp:
```shell
$ node
Welcome to Node.js v22.6.0.
Type ".help" for more information.
> const sharp = require('sharp');
> sharp.versions
{ vips: '8.15.3', sharp: '0.33.5' }
> const img = sharp('y1980-25.jp2');
undefined
> img.toFile('y1980-25.tif')
Promise {
  <pending>,
  [Symbol(async_id_symbol)]: 272,
  [Symbol(trigger_async_id_symbol)]: 6
}
> Uncaught Error: jp2kload: Invalid tile part index for tile number 1. Got 7, expected 0

jp2kload: Fail to read the current marker segment (0xff90)

jp2kload: Failed to decode the codestream in the JP2 file
    at Sharp.toFile (./node_modules/sharp/lib/output.js:90:19)
```

**Expected behavior**

The image is loaded and processed properly.

**Actual behavior**

See errors above.

**Environment**

 - OS: MacOS 15.0.1, Fedora Linux 39
 - Vips: 8.15.3, 8.15.4
 - Vips compiled with openjpeg2 v2.5.2

**Additional context**

[...]

PR #4432, opened 2025-03-22
Add jp2kload oneshot

jcupitt:
Some versions of openjpeg can fail to decode some tiled jpeg2000 images. 

This PR adds a `oneshot` flag that flips `jp2kload` into "one shot" mode, ie. entire frames are loaded in one call, rather than a tile at a time. 

This will use a lot more memory, and is usually slower, but it will make it possible to load files that otherwise could not be loaded.

closes https://github.com/libvips/libvips/issues/4205


jcupitt:
Any thoughts on the name? `oneshot` might not be obvious.

lovell:
> Any thoughts on the name? oneshot might not be obvious.

Using https://xkcd.com/2254/ as inspiration how about "long haul"? :stuck_out_tongue_winking_eye:

jcupitt:
hehe I think "oneshot" is a bit clearer

PR #4433, opened 2025-03-23
Fix vips7 compatibility layer for *magick

kleisauke:
Tested using:
```console
$ vips im_magick2vips test/test-suite/images/MARBLES.BMP x.png
```

Context: https://github.com/libvips/libvips/pull/4375#issuecomment-2729738142.


jcupitt:
:+1:

PR #4436, opened 2025-03-23
threadset: prevent possible thread oversubscription on Wasm

kleisauke:
Track the number of threads that haven't reached their entry point to prevent thread oversubscription, which could occur on Wasm.

This issue likely did not affect other environments, where thread spawning is usually instant.

See: https://github.com/kleisauke/wasm-vips/issues/92.

### DIFF
--- a/libvips/iofuncs/threadset.c
+++ b/libvips/iofuncs/threadset.c
@@ -311,7 +311,9 @@ vips_threadset_run(VipsThreadset *set,
 
 	g_async_queue_lock(set->queue);
 
-	/* Create a new thread if there are no waiting threads in the queue.
+	/* Create or reuse an idle thread if there are at least as many tasks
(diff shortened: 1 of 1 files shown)


kleisauke:
In that case, we still need two counts: the number of items in the queue and the number of waiting threads. Calling [`g_async_queue_length_unlocked()`](https://docs.gtk.org/glib/method.AsyncQueue.length_unlocked.html) is a bit simpler since it provides exactly that.

The comment "Create a new thread if there are no waiting threads in the queue." might be a misleading. It's more like "Create (or reuse an idle thread) if there are at least as many tasks in the queue as waiting threads."

> Mark as draft as this is a hot code path. Will perform thorough regression testing first.

I can confirm that this doesn't cause regressions under heavy server load. This is ready for review now.

jcupitt:
OK, let's do it!

PR #4439, opened 2025-03-30
CI: fix macOS build

kleisauke:


.github/workflows/ci.yml
```diff
@@ -71,6 +71,8 @@ jobs:
         if: runner.os == 'macOS'
         run: |
           echo "PKG_CONFIG_PATH=$(brew --prefix mozjpeg)/lib/pkgconfig:$(brew --prefix libarchive)/lib/pkgconfig:$PKG_CONFIG_PATH" >> $GITHUB_ENV
+          # Workaround for https://github.com/Homebrew/homebrew-core/issues/212112
+          sed -i'.bak' -e '/^libdir/s/=.*/=${prefix}\/lib/' -e '/^includedir/s/=.*/=${prefix}\/include/' $(brew --prefix cfitsio)/lib/pkgconfig/cfitsio.pc
```
jcupitt, line 75: Could this break again when homebrew fix the issue? Maybe it's better to just wait for them to update?

jcupitt: Ah, I think homebrew have fixed this now, I see:

```
✔ ~ 
09:54 $ brew update && brew upgrade
... snip
✔ ~ 
09:54 $ pkg-config cfitsio --cflags
-I/opt/homebrew/Cellar/cfitsio/4.6.2/include
✔ ~ 
09:54 $ 
```

jcupitt:
\o/

PR #4440, opened 2025-03-31
Add CITATION.cff

jcupitt:


kleisauke:
Great! Is there an alternative link that we could use? Or is https://www.imaging.org/ still in the process of releasing the online papers?

jcupitt:
I think they've still not put up the paper links. You can get an abstract here, but I don't think there's any more available yet.

https://www.imaging.org/IST/IST/Conferences/EI/EI2025/Conference/C_HPCI.aspx#HPCI-178

PR #4441, opened 2025-03-31
Add C declaration for fitsload_source

kleisauke:
Also, fix the remaining issues in `function-list.md`.

### DIFF
--- a/doc/function-list.md
+++ b/doc/function-list.md
@@ -25,10 +25,10 @@ libvips has a set of C wrapper functions for calling operators, in this
 case [method@Image.gamma]:
 
 ```c
-  VipsImage *fred = ...;
-  VipsImage *jim;
+VipsImage *fred = ...;
+VipsImage *jim;
 
-  if (vips_gamma(fred, &jim, NULL))
+if (vips_gamma(fred, &jim, NULL))
     ...error;
 ```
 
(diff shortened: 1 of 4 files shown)


doc/function-list.md
```diff
@@ -194,7 +194,7 @@ API docs each function links to for more details.
 | `jpegsave_buffer` | Save image to jpeg buffer | [method@Image.jpegsave_buffer] |
 | `jpegsave_mime` | Save image to jpeg mime | [method@Image.jpegsave_mime] |
 | `jpegsave_target` | Save image to jpeg target | [method@Image.jpegsave_target] |
-| `jxlload` | Load jpeg-xl image | [method@Image.jxlload] |
+| `jxlload` | Load jpeg-xl image | [func@jxlload] |
```
jcupitt, line 197: I'm puzzled, you'd think `method` would work for all these.

kleisauke: Hmm, I think it should be `ctor`, just like `vips_image_new_from_file()` and friends.
https://kleisauke.nl/vips-gi-docgen/ctor.Image.new_from_file.html

jcupitt: Yes, I had that too :( Looking at https://gnome.pages.gitlab.gnome.org/gi-docgen/linking.html it seems to say that `method` can find class methods, but I think it'll only do it for C functions which take the class struct as a first parameter.

I think we'll have to use `func` for all the `*load` operations, unfortunately :(

PR #4443, opened 2025-03-31
revise docs in iofuncs for gi-docgen

jcupitt:
I added `::: seealso` and it looks quite nice!

image


jcupitt:
I can't get `VIPS_FORMAT_DOUBLE` to link. I tried:

```
[enum@Format.DOUBLE]
[enum@BandFormat.DOUBLE]
[enum@BandFormat.FORMAT_DOUBLE]
```

but no luck :( I left them just in backticks.

Perhaps this is because the enum is not named consistently? Maybe renaming as `VipsFormat` would fix this.

jcupitt:
>  it seems they need to be linked fully qualified 

:face_with_head_bandage: I would not have guess that. Thanks!

kleisauke:
Those admonitions look great! Indeed, "optional arguments" can also benefit from that. :+1:

jcupitt:
OK, I'll try and do the other optional args as well. I'm just using a vim macro, there's probably a better way.

jcupitt:
What a horrible PR, but I think it's done.

jcupitt:
.... I don't think this is worth reviewing, it's just a lot of mostly mechanical changes to the doc comments. Let's just commit, and if there are any errors, we can fix them as we find them.

PR #4444, opened 2025-03-31
doc: make the load functions part of the image ctors

kleisauke:
Old: https://kleisauke.nl/vips-gi-docgen-old/func.jpegload.html
New: https://kleisauke.nl/vips-gi-docgen/ctor.Image.jpegload.html

Context: https://github.com/libvips/libvips/pull/4441#discussion_r2020935684.

### DIFF
--- a/doc/fixup-load-as-ctor.py
+++ b/doc/fixup-load-as-ctor.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python3
+
+import argparse
+import xml.etree.ElementTree as ET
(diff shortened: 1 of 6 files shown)


kleisauke:
As a possible follow-up, I think we could do the same for all other `func@` operations listed in  [`doc/function-list.md`](https://github.com/libvips/libvips/blob/master/doc/function-list.md?plain=1).

jcupitt:
That's great! I didn't realise you could write fixups like that.

PR #4448, opened 2025-04-02
share and reuse openslide connections

jcupitt:
openslide_open() can take many seconds for some slide formats. This PR adds a cache to let these connections be reused between openslideload calls, and across threads.


jcupitt:
Openslide handles (not just tile read) can now be shared between threads. It doesn't try to lock calls to other openslide entry points (eg. `openslide_get_error()` etc.). The openslide docs say:

> **An openslide_t object can be used concurrently from multiple threads
without locking.**

So I hope this is OK.

jcupitt:
Benchmarks:

- `vipsheader` on a huge DICOM WSI takes 5s before and after this patch
- `nip4 xxx` on the same huge DICOM takes 5s after this patch, but took 40s before, since it had to repeatedly reopen the slide
- zooming after this patch is much faster, since changing a pyramid level no longer requires opening a new page

jcupitt:
It doesn't drop the cache on `vips_shutdown()`, perhaps it should.

libvips/foreign/openslideconnection.c
```diff
@@ -0,0 +1,252 @@
...
+
+	g_mutex_unlock(&vips_openslideconnection_lock);
+
+	g_mutex_lock(&connection->lock);
+
+	/* We do the open outside the main lock (just in the connection lock)
+	 * since it can take many seconds for some slides.
+	 */
+	if (!connection->osr)
+		connection->osr = openslide_open(connection->filename);
```
bgilbert, line 226: `openslide_open` can validly return NULL, in which case this will rerun on every open attempt.  Probably fine, since the NULL case is an early exit which is pretty cheap (equivalent to `openslide_detect_vendor`).

libvips/foreign/openslideconnection.c
```diff
@@ -0,0 +1,252 @@
...
+
+	oldest = NULL;
+	g_hash_table_foreach(vips_openslideconnection_cache,
+		vips_openslideconnection_oldest_cb, &oldest);
+
+	return oldest;
+}
+
+static void
+vips_openslideconnection_trim(void)
```
bgilbert, line 132: You could possibly simplify this path by replacing the `time` mechanism with a global LRU `GQueue` of unreferenced connections.

libvips/foreign/openslideconnection.c
```diff
@@ -0,0 +1,252 @@
...
+	g_mutex_lock(&vips_openslideconnection_lock);
+
+	VipsOpenslideConnection *connection;
+	connection = g_hash_table_lookup(vips_openslideconnection_cache, filename);
+
+	/* We unref in the main lock since openslide_close() is always relatively
+	 * quick.
+	 */
+	if (connection)
+		vips_openslideconnection_unref(connection);
```
bgilbert, line 247: If we're called on an `openslide_t` in error state, we should detect that and stop handing that `openslide_t` out to new requestors, so errors don't become unhelpfully sticky.  Unfortunately the fact that we currently only take a `filename` argument makes this difficult.  If it were me, I'd have this function take a `VipsOpenslideConnection`, have `vips__openslideconnection_open` return that in an out-arg, and remove the `VipsOpenslideConnection` from the hash table if the `openslide_t` is closed from an error state.

libvips/foreign/openslideconnection.c
bgilbert, line 226: While you're here, you might want to check for the existence of `openslide_cache_create()` from `meson setup`, and share an OpenSlide cache among all the open handles.  Otherwise you're giving up as much as 32 MiB of RAM per handle.

bgilbert (commented):
Thanks for this.  I was never comfortable with VIPS reopening the slide for each level, so I'll be happy to see this go in.

All OpenSlide methods are indeed thread-safe (except `openslide_close` of course).  `openslide_get_error` is safe because error states are terminal and cause all future operations on the handle to fail.

> It will also break `revalidate`, unfortunately.

Is that for handling changes to the underlying slide file?  OpenSide isn't robust to that case.

libvips/foreign/openslideconnection.c
jcupitt: You're right, I thought about using a queue rather than a timestamp, but we're only going to be keeping a couple of connections around (they are relatively "heavy", as you note), so I didn't think the extra complexity was worth it.

libvips/foreign/openslideconnection.c
jcupitt: That's a good point. I think dropping the error state connection from cache on ref-count == 0 (to force a reopen) is probably the best solution.

libvips/foreign/openslideconnection.c
jcupitt: Ah good idea, I'll do this too.

jcupitt: Ah it'll ref again on the second failure. I'll fix this too.

jcupitt:
Thank you again @bgilbert !

libvips/foreign/openslideconnection.c
bgilbert: Well, we're only keeping a couple idle connections, but potentially unlimited active connections depending on what libvips' caller does.

bgilbert: I ended up implementing this in https://github.com/libvips/libvips/pull/4455.

PR #4449, opened 2025-04-02
doc: move sections to separate files (batch 1/2)

kleisauke:
The task is simple: find a `SECTION:` and move it to a separate file, and fix up the gtk-doc links and syntax to use gi-docgen instead. This is batch 1 of 2; the second one will be handled in a follow-up PR.

Rendered output: https://kleisauke.nl/vips-gi-docgen-sections.


doc/developer-checklist.md
```diff
@@ -21,7 +21,8 @@ nina.jpg: 6048x4032 uchar, 3 bands, srgb, jpegload
 
 I see:
 
-```bash/usr/bin/time -f %M:%e vips resize nina.jpg x.jpg 0.1
+```bash
```
jcupitt, line 24: Ooops!

doc/libvips-arithmetic.md
```diff
@@ -65,7 +65,7 @@ generally the principle is that the output type should be large enough to
 represent the whole range of possible values, except that int never becomes
 float.
 
-## Arithmetic functions
+## Functions
 
 * [method@Image.add]
```
jcupitt, line 70: We are setting ourselves up for a lot of work if these lists are going to be maintained by hand :(

jcupitt:
Well done! This is a horrible maintenance job.

doc/developer-checklist.md
kleisauke: It seems that my initial regex that places `bash` for these CLI commands was too greedy. 😅

doc/libvips-arithmetic.md
kleisauke: I'll check if we can automate this. Perhaps we could inspect the resulting GIR file and extract only the functions from the `arithmetic/` directory.

kleisauke: Commit 27016bb23cb4a1acf7e0b5404e399d57b8c24d41 adds a sanity check script to verify that all symbols are listed on the relevant pages. It caught a few overlooked symbols, which were fixed separately in commit fa378cd0bc98291904433404cd20275b49823258. Linux CI was modified to cover this, see commit 6122324dbbec6e79db30063bbba08b3f76e54fc5 and e.g. these now fixed failures:
https://github.com/libvips/libvips/actions/runs/14282791823/job/40034288723#step:10:48

jcupitt: Ah very nice!

PR #4453, opened 2025-04-06
try renaming some sections

jcupitt:
maybe we can make the extra docs section a little easier to read

with this PR, it looks like:

image

### DIFF
--- a/doc/binding.md
+++ b/doc/binding.md
@@ -1,7 +1,5 @@
 Title: Writing bindings for libvips
 
-# Writing bindings for libvips
-
 There are full libvips bindings for quite a few environments now, including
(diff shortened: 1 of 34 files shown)


jcupitt:
Since it seems we can't do any grouping to things in the extra-docs section.

kleisauke (approved):
LGTM!

> Since it seems we can't do any grouping to things in the extra-docs section.

This is being tracked upstream at https://gitlab.gnome.org/GNOME/gi-docgen/-/issues/168.

jcupitt:
> This is being tracked upstream at https://gitlab.gnome.org/GNOME/gi-docgen/-/issues/168

Great! Though the priority seems low hehe.

PR #4455, opened 2025-04-06
openslideconnection: fixes and some cleanups

bgilbert:
The first three commits are genuine fixes, the last two IMO are improvements but feel free to do with them as you'd like.

I've smoke-tested this but haven't exercised the cache code in a meaningful way.


jcupitt:
Thanks! I'll do some extra testing in another PR.

PR #4456, opened 2025-04-07
openslideconnection: fix link failure with modules enabled

kleisauke:
Alternatively, we could make `vips__openslideconnection_leak()` always available, just like:
https://github.com/libvips/libvips/blob/dbc3b86054208b2d9915e5ebdf7a0c3d022013e0/libvips/include/vips/internal.h#L102-L105

But I'm not sure if that's worth it, since this logic is only for `VIPS_LEAK`.


jcupitt:
Oooop! Sorry, what an idiot.

PR #4460, opened 2025-04-10
redo annotations for "colour"

jcupitt:
I noticed `vips_image_get_typeof()` is wrapped as `method@Image.image_get_typeof`. It should probably just be `method@Image.get_typeof`. I suppose we need to add another thing to `fixup-gir-for-doc.py`.


kleisauke:
> I noticed `vips_image_get_typeof()` is wrapped as `method@Image.image_get_typeof`. It should probably just be `method@Image.get_typeof`.

I just opened PR #4462 for this.

doc/rename.sed
```diff
@@ -199,6 +200,7 @@ s/vips_\(icc_export\)()/[method@Image.\1]/g
 s/vips_\(icc_import\)()/[method@Image.\1]/g
 s/vips_\(icc_transform\)()/[method@Image.\1]/g
 s/vips_\(ifthenelse\)()/[method@Image.\1]/g
+s/vips_\(image_get_typeof\)()/[method@Image.\1]/g
```
kleisauke, line 203:
```suggestion
s/vips_image_\(get_typeof\)()/[method@Image.\1]/g
```

kleisauke (approved):

Issue #4343, opened 2025-01-06
Black image after linear thumbnail attempt


chcunningham:
## Bug report

**Describe the bug**
Thanks for the wonderful library!

vipsthumnail produces a mostly black image for a particular input. I'm unfortunately not able to share the input publicly but I will email the file to jcupitt at gmail.com.

The vipsthumbnail command is executed with the linear flag. This fails for the original image due to a problematic ICC profile* that doesn't support relative output intent. Following the [recommendation here](https://github.com/libvips/libvips/issues/2329#issuecomment-871544408), the ICC profile issue is mitigated using icc_transform with `--intent perceptual`. The transformed result looks normal (just a tad darker). Running the thumbnail command on this transformed result produces a mostly black (actually: very very dark) output. 

*Aside: the problematic ICC profile in the original image is the "Apple Wide Color Sharing Profile", which [seems to be output by Apple's Photos app](https://community.adobe.com/t5/camera-raw-discussions/p-apple-generated-jpeg-with-icc-profile-quot-apple-wide-color-sharing-profile-quot-won-t-import/td-p/14355907) when exporting photos as "Most compatible". LMK if icc_transform --intent perceptual is still the best path for such images. Maybe it would be better for me to drop the --linear flag in these cases, with resize accuracy issues being the lesser evil?

**To Reproduce**

```
$ vipsthumbnail original_image.png --size="2560>" --linear -o thumbnail_attempt1.png`

(vipsthumbnail:78858): VIPS-WARNING **: 10:26:51.972: profile does not support relative output intent
vipsthumbnail: unable to thumbnail original_image.png
icc_export: no output profile

$ vips icc_transform original_image.png icc_transformed_image.png srgb --intent perceptual
$ vipsthumbnail icc_transformed_image.png --size="2560>" --linear -o thumbnail_attempt2.png
```

**Expected behavior**
thumbnail_attempt2.png colors should match those from icc_transformed_image.png

**Actual behavior**
thumbnail_attempt2.png is almost entirely black.

**Screenshots**
Please see email.

**Environment**
(please complete the following information)
 - OS: Mac OS Sonoma 14.7.1
 - Vips: vips-8.16.0 (also previous version, 8.15.3)

**Additional context**
For other images, using vipsthumbnail for icc_transformed images seems to be generally working fine, including for images with the Apple Wide Color Sharing Profile. This single problematic  image is the only one I've seen fail so far.

jcupitt:
[...]

PR #4467, opened 2025-04-15
fix linear resize of 16-bit rgba

jcupitt:
We were changing the interpretation in icc_import and icc_export, but we were not rescaling the alpha. As a result, 16-bit RGBA images had a alpha of 65535 in PCS, not 255.

This PR moves alpha channel scaling into the base VipsColour class, so we get correct scaling automatically, and moves XYZ <-> scRGB conversion back on top of VipsColourCode, partially reverting #3627

see https://github.com/libvips/libvips/issues/4343

### DIFF
--- a/libvips/colour/colour.c
+++ b/libvips/colour/colour.c
@@ -165,8 +165,6 @@ vips_colour_build(VipsObject *object)
 	VipsImage **extra_bands;
 	VipsImage *out;
 
-	int i;
-
 #ifdef DEBUG
 	printf("vips_colour_build: ");
 	vips_object_print_name(object);
@@ -177,11 +175,10 @@ vips_colour_build(VipsObject *object)
 		return -1;
 
 	if (colour->n > MAX_INPUT_IMAGES) {
-		vips_error(class->nickname,
-			"%s", _("too many input images"));
+		vips_error(class->nickname, "%s", _("too many input images"));
 		return -1;
 	}
-	for (i = 0; i < colour->n; i++)
+	for (int i = 0; i < colour->n; i++)
 		if (vips_image_pio_input(colour->in[i]))
 			return -1;
 
@@ -201,7 +198,7 @@ vips_colour_build(VipsObject *object)
 		VipsImage **new_in = (VipsImage **)
 			vips_object_local_array(object, colour->n);
 
-		for (i = 0; i < colour->n; i++) {
+		for (int i = 0; i < colour->n; i++) {
 			if (vips_check_bands_atleast(class->nickname,
 					in[i], colour->input_bands))
 				return -1;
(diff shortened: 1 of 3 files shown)


jcupitt:
We could potentially move scRGB2BW, scRGB2sRGB and sRGB2scRGB on top of VipsColourCode now and remove the alpha handling they have. It'd be a bit slower, but a lot simpler.

kleisauke:
> We could potentially move scRGB2BW, scRGB2sRGB and sRGB2scRGB on top of VipsColourCode now and remove the alpha handling they have. It'd be a bit slower, but a lot simpler.

I had a quick look, but these classes also have a special paths for 16-bit pixels (see e.g. the `vips_scRGB2BW16()` and `vips_scRGB2RGB16()` callers). Perhaps we could consider splitting those 16-bit paths into their own class as well?

libvips/colour/colour.c
```diff
@@ -245,40 +242,43 @@ vips_colour_build(VipsObject *object)
 
 	if (vips_image_generate(out,
 			vips_start_many, vips_colour_gen, vips_stop_many,
-			in, colour)) {
-		g_object_unref(out);
```
kleisauke, line 249: Was the removal of this unref intentional? I also noticed a missing unref in the error condition around line 240, i.e. if the `vips__profile_set()` call fails.

libvips/colour/colour.c
```diff
@@ -245,40 +242,43 @@ vips_colour_build(VipsObject *object)
...
 		if (extra_bands[i]) {
-			VipsImage *t1, *t2;
+			VipsImage **t = (VipsImage **) vips_object_local_array(object, 5);
+
+			double max_alpha_before =
+				vips_interpretation_max_alpha(extra_bands[i]->Type);
+			double max_alpha_after =
+				vips_interpretation_max_alpha(out->Type);
+
+			VipsImage *in;
```
kleisauke, line 260: nit: `in` was already declared outside the loop, perhaps we should use a different variable name here? This would also appease any `-Wshadow` warnings in this file.
```
../libvips/colour/colour.c: In function ‘vips_colour_build’:
../libvips/colour/colour.c:260:36: warning: declaration of ‘in’ shadows a previous local [-Wshadow]
  260 |                         VipsImage *in;
      |                                    ^~
../libvips/colour/colour.c:164:21: note: shadowed declaration is here
  164 |         VipsImage **in;
      |             
```

libvips/colour/colour.c
```diff
@@ -245,40 +242,43 @@ vips_colour_build(VipsObject *object)
...
-	}
 
 	/* Reattach higher bands, if necessary. If we have more than one input
 	 * image, just use the first extra bands.
 	 */
-	for (i = 0; i < colour->n; i++)
+	for (int i = 0; i < colour->n; i++)
 		if (extra_bands[i]) {
-			VipsImage *t1, *t2;
+			VipsImage **t = (VipsImage **) vips_object_local_array(object, 5);
```
kleisauke, line 253: nit:
```suggestion
			VipsImage **t = (VipsImage **) vips_object_local_array(object, 3);
```
(although I understand that oversizing this array can be useful for future adjustments, feel free to disregard this)

libvips/colour/colour.c
jcupitt: Oop! No, a copy-paste error from an earlier version of this PR, I'll fix it, thanks!

libvips/colour/colour.c
jcupitt: Nice catch, I'll rename it `alpha`.

kleisauke (approved):
Nice!

jcupitt:
> I had a quick look, but these classes also have a special paths for 16-bit pixels (see e.g. the vips_scRGB2BW16() and vips_scRGB2RGB16() callers). Perhaps we could consider splitting those 16-bit paths into their own class as well?

I think I can see how to do it. I'll try a follow-up PR.

jcupitt:
(and thanks for the review!)

PR #4468, opened 2025-04-16
move scRGB2BW on top of ColourCode

jcupitt:
saves some alpha messing

todo: 

- [x] some tests
- [x] also scRGB2sRGB and sRGB2scRGB
- [x] move vips_image_decode() call from colourcode to colour class?


kleisauke:
Nice!

PR #4472, opened 2025-04-17
update conversion docs

jcupitt:
update conversion docs to new format, revise arithmetic and colour

I added "optional arg" annotations as tips (not notes), and moved then to the end of the comment block:

image

- using tips makes them look different from seealso
- the first para of a doc comment is supposed to be a one-line summary, so we can't put optional args there, unfortunately

The Image class page now looks like:

image

So ifthenelse now has a sensible description, though icc_transform does not.


jcupitt:
... on second thoughts, I don't think this is worth reviewing, it's just comment changes. If we notice anything, we can fix it later.

Issue #4476, opened 2025-04-23
stray file created by dzsave


jcupitt:
Try:

```
$ ls
$ vips icc_transform ../k2.jpg x.dz[container=fs] srgb 
$ ls -al
total 84
drwxrwxr-x   3 john john  4096 Apr 23 04:01 .
drwxr-xr-x 150 john john 69632 Apr 23 03:56 ..
-rw-r--r--   1 john john     0 Apr 23 04:01 x.dz
-rw-rw-r--   1 john john   206 Apr 23 04:01 x.dzi
drwxrwxr-x  14 john john  4096 Apr 23 04:01 x_files
$ 
```

A stray length 0 file called `x.dz` is created.

jcupitt:
Oooof and you can't run it twice sigh.

```
$ vips copy ../k2.jpg x.dz[container=fs]
$ vips copy ../k2.jpg x.dz[container=fs]
VipsForeignSave: "x.dzi" is not a known file format
```

jcupitt:
We have a PR, I'll close.

PR #4482, opened 2025-04-28
lazy create of file targets

jcupitt:
`vips_target_new_to_file()` used to immediately create the target file. If the filename was not actually used (eg. by `dzsave`), a stray file would be left behind.

This PR moves the file create to the first `_write()` operation.

See https://github.com/libvips/libvips/issues/4476

### DIFF
--- a/libvips/iofuncs/target.c
+++ b/libvips/iofuncs/target.c
@@ -115,25 +115,10 @@ vips_target_build(VipsObject *object)
 		return -1;
 	}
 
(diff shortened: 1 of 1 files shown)


kleisauke (approved):
LGTM! 

I was going to suggest moving the `connection->filename && connection->tracked_descriptor == -1` check outside the function (as `vips_target_write_real()` can be called many times), but it's probably inlined anyway.

PR #4488, opened 2025-04-30
rename thumbnail's profile arguments

jcupitt:
We use `input-profile` and `output-profile` everywhere else, but `thumbnail` calls them `import-profile` and `export-profile`, confusingly. LittleCMS seems to use `input-profile` and `output-profile` too.

### DIFF
--- a/tools/vipsthumbnail.c
+++ b/tools/vipsthumbnail.c
@@ -105,6 +105,8 @@
  * 2/10/20
  * 	- support "stdin" as a magic input filename for thumbnail_source
  * 	- support ".suffix" as a magic output format for stdout write
+ * 30/4/25
+ *  - rename import/export profile args as input/oputput
  */
 
 #ifdef HAVE_CONFIG_H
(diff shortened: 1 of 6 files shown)


tools/vipsthumbnail.c
```diff
@@ -159,12 +161,12 @@ static GOptionEntry options[] = {
 		G_OPTION_ARG_STRING, &output_format,
 		N_("output to FORMAT"),
 		N_("FORMAT") },
-	{ "export-profile", 'e', 0,
-		G_OPTION_ARG_FILENAME, &export_profile,
+	{ "output-profile", 0, 0,
```
kleisauke, line 164: Perhaps the short name (i.e. `-e` and `-i`) should be kept on the non-deprecated options? This would ensure that `vipsthumbnail --help` would print these.

master branch:
```console
$ vipsthumbnail --help | grep profile=
  -e, --export-profile=PROFILE     export with PROFILE
  -i, --import-profile=PROFILE     import untagged images with PROFILE
```

This PR:
```console
$ vipsthumbnail --help | grep profile=
  --output-profile=PROFILE      export with PROFILE
  --input-profile=PROFILE       import untagged images with PROFILE
```

jcupitt: Maybe, though `-i` and `-e` are not obvious shortcuts.

kleisauke (approved):

jcupitt:
Thanks for the review!

PR #4495, opened 2025-05-01
test and fix SAVEABLE

jcupitt:
`vips__foreign_convert_saveable()` doesn't seem to work for all image types, start testing it more carefully.

`test_matrix` tests matrixsave, which uses `SAVEABLE_MONO` ... test conversion of all image types.


jcupitt:
Sample bug in master from these problems -- RGB to mono works OK:

```
$ vips copy k2.jpg x.csv
$ vipsheader x.csv
x.csv: 1450x2048 double, 1 band, b-w, csvload
```

So the `SAVEABLE_MONO` tag on `csvsave` is successfully converting RGB to mono before the save, but:

```
$ vipsheader lion.png 
lion.png: 425x425 uchar, 4 bands, srgb, pngload
$ vips copy lion.png x.csv
csvsave_target: image must one band
```

It fails for RGBA. The saveable system needs simplifying and testing more carefully.

jcupitt:
This replaces the old `VipsSaveable` enum with a set of `VipsForeignSaveable` flags. Savers set flags for the image types they support, eg. `MONO | RGB | ALPHA` etc. There's an implementation of `VipsSaveable` in `almostdeprecated.h` for compat, though libvips itself no longer uses it. 

The conversion to a saveable image type is supposed to be simpler and easier to understand and verify. There are a lot more tests!

This fixes quite a few obscure bugs, as noted in the top comment.

kleisauke:
CIFuzz caught an interesting fuzzer bug in this PR. You can download the reproducer here:
https://github.com/libvips/libvips/actions/runs/14878870815?pr=4495
```console
$ vipsheader crash-c8c0ef1d5d8a06465e3a728c0cd8d37b59239cbe
crash-c8c0ef1d5d8a06465e3a728c0cd8d37b59239cbe: 1x1 dpcomplex, 1 band, error
$ vips copy crash-c8c0ef1d5d8a06465e3a728c0cd8d37b59239cbe x.jpg
Aborted (core dumped)
```

ChangeLog
```diff
@@ -26,6 +26,7 @@ master
 - rename `thumbnail`'s export/import profile options as input/output for
   consistency
 - restore vips_remosaic(), it was not being linked
+- deprecate VipsSaveable, add VipsSaveableFlags
```
kleisauke, line 29:
```suggestion
- deprecate VipsSaveable, add VipsForeignSaveable
```

libvips/include/vips/foreign.h
```diff
@@ -308,24 +308,42 @@ void vips_foreign_load_invalidate(VipsImage *image);
...
+	VIPS_SAVEABLE_RGBA =
+		VIPS_SAVEABLE_RGB | VIPS_FOREIGN_SAVEABLE_ALPHA,
+	VIPS_SAVEABLE_RGBA_ONLY =
+		VIPS_FOREIGN_SAVEABLE_RGB | VIPS_FOREIGN_SAVEABLE_ALPHA,
+	VIPS_SAVEABLE_RGB_CMYK =
+		VIPS_SAVEABLE_RGB | VIPS_FOREIGN_SAVEABLE_CMYK,
+	VIPS_SAVEABLE_ANY =
+		VIPS_FOREIGN_SAVEABLE_ANY,
+	VIPS_SAVEABLE_LAST = 99,
 } VipsSaveable;
```
kleisauke, line 347: We could still move this to `almostdeprecated.h` and alias `vips_saveable_get_type()` to `vips_foreign_saveable_get_type()` for compat, see e.g. commit https://github.com/kleisauke/libvips/commit/865d67c2416eb6d48dfb3d9d218c326601479b91. Not sure if that's the right approach, though.

jcupitt:
> CIFuzz caught an interesting fuzzer bug in this PR. 

It was a nasty bug in `vips_image_guess_format()`. Good to fix that.

libvips/include/vips/foreign.h
jcupitt: Ah maybe that is better. It'll be ages before we can get rid of the use of `saveable_get_type` in pyvips.

kleisauke (approved):
Great!

PR #4499, opened 2025-05-07
doc: port remainder

kleisauke:


jcupitt:
That's great! I suppose we could do `%gboolean`, `%gint`, etc. as well.

kleisauke:
Good idea. The last revision of this PR also ports those GLib types.

PR #4500, opened 2025-05-07
doc: avoid gi-docgen linking syntax for non-doc comments

kleisauke:
I missed these in PR #4498.


kleisauke:
Ah, hold on, I'll also check `method@` this time.

kleisauke:
That's all, I think.

jcupitt:
Hurrah!

PR #4501, opened 2025-05-07
Ignore VipsSaveable type when generating introspection data

kleisauke:
Otherwise, it will end up including `vips_saveable_get_type()` in the docs.


jcupitt:
Good idea!

PR #4504, opened 2025-05-08
doc: port remainder to gi-docgen linking syntax

kleisauke:
I missed these in PR #4499 (`grep`-ing in `Vips-8.0.gir` is much easier than though all the files in `libvips/*`).

While we're here, change the enum members to uppercase in the Markdown files for consistency.

### DIFF
--- a/libvips/iofuncs/buf.c
+++ b/libvips/iofuncs/buf.c
@@ -494,10 +494,10 @@ vips_buf_appendd(VipsBuf *buf, int d)
 /**
  * vips_buf_appendgv:
  * @buf: the buffer
- * @value: #GValue to format and append
+ * @value: [struct@GObject.Value] to format and append
  *
- * Format and append a #GValue as a printable thing. We display text line "3144
- * bytes of binary data" for BLOBs like icc-profile-data.
+ * Format and append a [struct@GObject.Value] as a printable thing.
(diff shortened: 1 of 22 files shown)


jcupitt:
> grep-ing in Vips-8.0.gir is much easier than though all the files in libvips/*

Damn, why didn't I think of that! Duh.

libvips/iofuncs/buf.c
```diff
@@ -494,10 +494,10 @@ vips_buf_appendd(VipsBuf *buf, int d)
 /**
  * vips_buf_appendgv:
  * @buf: the buffer
- * @value: #GValue to format and append
+ * @value: [struct@GObject.Value] to format and append
  *
- * Format and append a #GValue as a printable thing. We display text line "3144
- * bytes of binary data" for BLOBs like icc-profile-data.
+ * Format and append a [struct@GObject.Value] as a printable thing.
+ * We display text line "3144 bytes of binary data" for BLOBs like icc-profile-data.
```
jcupitt, line 500: I think this is a typo and should be "like"

kleisauke: Nice catch. 50571ba0e4825394110d623b558b726c1ea482ff should fix that.

Issue #4509, opened 2025-05-11
`VipsForeignSaveable` regression with RGBA image


kleisauke:
Performing a recomb operation on an RGBA image produce an incorrect result since commit 27ab9b058dffd290a35a012c822004e216a21b84.

**Reproducer**
```console
$ cat <<EOT > sepia-4-band.mat
4 4
0.3588 0.7044 0.1368 0 
0.2990 0.5870 0.1140 0
0.2392 0.4696 0.0912 0
0      0      0      1
EOT
$ curl -LO https://github.com/weserv/images/raw/5.x/test/api/fixtures/alpha-layer-1-fill.png
$ vips recomb alpha-layer-1-fill.png sepia.png sepia-4-band.mat
```

**Expected image**
![Expected image](https://github.com/user-attachments/assets/b0d221c8-b239-4068-98f6-dcee61bd751e)

**Actual image**
![Actual image](https://github.com/user-attachments/assets/3b91fee1-6869-41d6-bf00-95175eba022f)

**Additional context**
This issue was originally discovered in [`weserv/images`](https://github.com/weserv/images)' test suite while preparing for the upcoming libvips 8.17 release.

PR #4510, opened 2025-05-11
Avoid cast shifting during saving

kleisauke:
Resolves: #4509.
Possibly also fixes https://issues.oss-fuzz.com/issues/416442575.

Marked as draft due to missing test.


libvips/foreign/foreign.c
```diff
@@ -1548,6 +1548,7 @@ vips__foreign_convert_saveable(VipsImage *in, VipsImage **ready,
 	 */
 	if (in->Coding == VIPS_CODING_NONE &&
 		(saveable & VIPS_FOREIGN_SAVEABLE_ANY) &&
+		format &&
```
kleisauke, line 1551: This is a drive-by fix, but it might be redundant since `->format_table` already defaults to this:
https://github.com/libvips/libvips/blob/70b5eef9918110143f8fba8624311368d81e97ad/libvips/foreign/foreign.c#L1863-L1865

If `format` is guaranteed to never be `NULL`, we could simplify remove the `&& format` check on line 1643 instead.

jcupitt: We also call `vips__foreign_convert_saveable()` directly from tiffsave.c and dzsave.c, though format is non-NULL there too, you're right. We could maybe add a `g_assert(format)` at the start of the function?

kleisauke: Added an assert for both `format` and `coding` to be non-NULL and removed the redundant `&& format` check.

jcupitt:
Thanks for fixing more dumbness!

PR #4511, opened 2025-05-12
Increase minimum version of libheif dependency to 1.11.0

lovell:
Aligns with the libheif version provided by Debian 11, which seems to be the oldest version provided by any actively supported flavour of Linux (Ubuntu 20.04 is EOL this month).


jcupitt:
Nice cleanup.

PR #4514, opened 2025-05-12
heifload: limit per-image memory usage to 2GB by default

lovell:
The upstream default is 4GB, which is too big for the fuzz test environment and IMHO is too big for a sensible default. Halving the default limit to 2GB means all the current "Out-of-memory" fuzz test failures will go away, and seems to be sufficient to process all the HEIF images I've tested with. It is however an(other) arbitrary value and I'd be happy to modify it based on further feedback.

The existing `unlimited` flag can be used to remove this.

Before:
```
$ vips heifload clusterfuzz-testcase-minimized-rawsave_buffer_fuzzer-6038980984635392
clusterfuzz-testcase-minimized-rawsave_buffer_fuzzer-6038980984635392: bad seek to 1024
...
heif: Invalid input: Unexpected end of file (2.100)
```
After:
```
$ vips heifload clusterfuzz-testcase-minimized-rawsave_buffer_fuzzer-6038980984635392
clusterfuzz-testcase-minimized-rawsave_buffer_fuzzer-6038980984635392: bad seek to 1024
...
heif: Memory allocation error: Security limit exceeded: Memory usage of 3774939392 bytes for the 'saio' table exceeds the security limit of 2147483648 bytes of total memory usage (6.1000)
```
After with `--unlimited` flag:
```
$ vips heifload clusterfuzz-testcase-minimized-rawsave_buffer_fuzzer-6038980984635392 --unlimited
clusterfuzz-testcase-minimized-rawsave_buffer_fuzzer-6038980984635392: bad seek to 1024
...
heif: Invalid input: Unexpected end of file (2.100)
```

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -32,6 +32,7 @@ master
 - jxlsave writes in chunks, for lower memory use on large images (but we now
   need libjxl 0.11 at minimum)
 - increase minimum version of libheif dependency to 1.11.0
+- heifload: limit per-image memory usage to 2GB (requires libheif 1.20.0+).
 
 8.16.1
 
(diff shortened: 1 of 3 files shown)


jcupitt (approved):
2GB sounds reasonable to me.

PR #4515, opened 2025-05-12
doc: minor gi-docgen improvements

kleisauke:
See the individual commits for more details. As a possible follow-up, we could remove the `doc/rename.{sed,sh}` scripts now.


jcupitt:
Wow you found found a lot of small errors. I agree, those rename scripts can go now.

Issue #4512, opened 2025-05-12
pngsave with bitdepth 16 setting does not seem to work


jcupitt:
Try:

```
$ vips pngsave k2.jpg x.png --bitdepth 16
pngsave: invalid buffer size
wbuffer_write: write failed
system error: Unknown error -1
wbuffer_write: write failed
system error: Unknown error -1
memory: high-water mark 12.49 MB
error buffer: pngsave: invalid buffer size
wbuffer_write: write failed
system error: Unknown error -1
wbuffer_write: write failed
system error: Unknown error -1
```

Fails with any RGB JPEG image.

kleisauke:
It looks like the libpng saver (i.e. when built with `-Dspng=disabled`) segfaults instead. However, both savers seem to work when you do:
```console
$ vips colourspace zebra.jpg x.png[bitdepth=16] rgb16
$ vipsheader x.png 
x.png: 4120x2747 ushort, 3 bands, rgb16, pngload
```
(though, the `bitdepth` setting isn't required in this case)

jcupitt:
It seems to be just that the PNG saver never actually changes the format of the image (!!). Adding this to `vips_foreign_save_spng_build()` seems to fix it:

```C

    /* Deprecated "colours" arg just sets bitdepth large enough to hold
     * that many colours.
     */
    if (vips_object_argument_isset(object, "colours"))
        spng->bitdepth = ceil(log2(spng->colours));

   // new code:

    /* The bitdepth param can change the interpretation.
     */
    VipsInterpretation interpretation;
    if (in->Bands > 2) {
       if (spng->bitdepth > 8)
           interpretation = VIPS_INTERPRETATION_RGB16;
       else
           interpretation = VIPS_INTERPRETATION_sRGB;
    }
    else {
       if (spng->bitdepth > 8)
           interpretation = VIPS_INTERPRETATION_GREY16;
       else
           interpretation = VIPS_INTERPRETATION_B_W;
    }
    if (in->Type != interpretation) {
        VipsImage *x;

        if (vips_colourspace(in, &x, interpretation, NULL)) {
            g_object_unref(in);
            return -1;
        }
        g_object_unref(in);
        in = x;
    }

    // end of new code

    /* If this is a RGB or RGBA image and a low bit depth has been
     * requested, enable palettisation.
     */
    if (in->Bands > 2 &&
        spng->bitdepth < 8)
        spng->palette = TRUE;
```

[...]

PR #4516, opened 2025-05-12
fix the pngsave bitdepth param

jcupitt:
Saving an 8-bit image as 16-but PNG (and vice versa) didn't work correctly.

See https://github.com/libvips/libvips/issues/4512

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -31,6 +31,7 @@ master
   public API
 - jxlsave writes in chunks, for lower memory use on large images (but we now
   need libjxl 0.11 at minimum)
+- increase minimum version of libheif dependency to 1.11.0
 
 8.16.1
 
--- a/libvips/arithmetic/boolean.c
+++ b/libvips/arithmetic/boolean.c
@@ -622,7 +622,7 @@ vips_boolean_constv(VipsImage *in, VipsImage **out,
  * Perform various boolean operations on an image against an array of
  * constants.
  *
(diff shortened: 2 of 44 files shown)


jcupitt:
This is a bit more complicated than you'd think because `colourspace("srgb")` on a float image can give a float result, rather than forcing uint8.

For example:

```
$ vips colourspace k2.jpg x.v scrgb
$ vips colourspace x.v y.v srgb
$ vipsheader y.v
y.v: 1450x2048 uchar, 3 bands, srgb, jpegload
```

The float scrgb image becomes uint8 srgb, so precision is lost. But:

```
$ vips cast k2.jpg x.v double
$ vips colourspace x.v y.v srgb
$ vipsheader y.v
y.v: 1450x2048 double, 3 bands, srgb, jpegload
```

Now it stays float.

This seems unpredictable and is probably a bad idea. What do you think? Should colourspace conversion to srgb always output uint8, and conversion to rgb16 always output ushort? 

There's a potential loss of precision, but if you want high precision srgb you should probably be using scrgb.

jcupitt:
1. `colourspace` now always outputs the expected format for each colourspace, so sRGB is always uint3, XYZ is always float, and so on.
2. The PNG savers let you use the `bitdepth` arg to select rgb16 output, though I'm not sure how useful this is since it's just an 8-bit image shifted up.

kleisauke (approved):
Nice!

PR #4521, opened 2025-05-17
Revise `VipsForeignSaveable` enum members

kleisauke:


jcupitt:
We have an `_ALL` member for all flag enums, and they can be useful. Though this is an internal thing, so maybe it doesn't matter.

kleisauke:
I removed this because I assumed the `_ANY` member had the same meaning as `_ALL`.

jcupitt:
Ah OK! 

`ANY` means "don't touch the format or interpretation, this saver supports anything". It's used for eg. TIFF or `.v`, where almost everything can be represented, eg. seven band float. It's not really a flag bit since, if this is set, none of the other flag bits mean anything.

`ALL` means a saver supports all of the image types that convert saveable can generate, so BW | RGB | CMYK, with a single optional alpha.

kleisauke:
Ah, in that case `ALL = 31` should probably be `= 15`, i.e. `VIPS_FOREIGN_SAVEABLE_MONO | VIPS_FOREIGN_SAVEABLE_RGB | VIPS_FOREIGN_SAVEABLE_CMYK | VIPS_FOREIGN_SAVEABLE_ALPHA`.

jcupitt:
You're right, that'd be clearer.

kleisauke:
OK, I also changed `_ANY` to `= 0`, but let me know if you'd prefer it the other way.

jcupitt:
You're right, 0 would work for `_ANY`, neat.

Issue #4494, opened 2025-05-01
vips_hough_line_vote ignoring votes for extreme values of rho and theta


ecbypi:
## Bug report

Please let me know if you need anything else.  

**Environment**
 - OS: macOS Sequoia 15.4.1
 - Vips: 8.16.0 (installed via macports)
 - Ruby-vips: 2.2.3
 - Ruby: 3.4.2

**Describe the bug**
The calculations for `r`, `ri`, `xd`, and `yd` in [`vips_hough_line_vote`](https://github.com/libvips/libvips/blob/2672bb658893da7f1cc3a4b1660b154f4308ad84/libvips/arithmetic/hough_line.c#L106) appear to be incorrect, causing curves in the Hough image to be cut off for either low angles at high values of `r` or high angles at low values of `r`.

In a test script (below), I was able to address this by:
1. Scaling and shifting `ri` to accommodate negative values
2. Calculating`xd` and `yd` by dividing `x` and `y` with the diagonal of the image instead of their respective image dimensions.  This ensures `r` will stay within 1 and the adjusted `ri` within the bounds of `height`.

**To Reproduce**
I discovered this behavior with the following Ruby script.  I was using it to clarify my understanding of how the `width` and `height` arguments to `hough_line` should be translated back to values of theta and rho.  

What the script does:
1. Create a 512 by 512 pixel black image with two pixels set to white.  
3. Create a Vips image from the nested array of pixel values; the image is output to a file for debugging
4. Calculate the Hough space with `Vips::Image#hough_line`
5. Calculate the Hough space manually using the changes mentioned under **Describe the bug**
7. Output the Hough images and the images highlighting intersections in Hough images to files (examples attached below)
8. Print intersections to stderr

```ruby
#!/usr/bin/env ruby

# This script creates a 512x512 image with individual pixels set to white
#
# The `points` array is a list of points I was testing to see if they intersect
# in the output `hough_line`

require "ruby-vips"
require "pp"

# Check points at different orientations to each other
#
# Only rendering two points at at time for simplicity
points = [
  # Corner pixels to test boundaries
  #[1, 1],
  #[1, 512],
  #[512, 1],
  #[512, 512],

  # horizontal line
  #[128, 256],
  #[384, 256],

  # vertical line
  #[256, 128],
  #[256, 384],

  # Points where no intersection is found `Vips::Image#hough_line`
  # points that would produce theta > 135; no intersection is found
  #[118, 128],
  #[274, 384],

[...]

PR #4525, opened 2025-05-20
improve scaling of hough_line feature space

jcupitt:
we were not scaling against the diagonal size, nor correctly considering negative angles

thanks ecbypi

see https://github.com/libvips/libvips/issues/4494

### DIFF
--- a/libvips/arithmetic/hough_line.c
+++ b/libvips/arithmetic/hough_line.c
@@ -70,14 +70,12 @@ vips_hough_line_build(VipsObject *object)
 	VipsHoughLine *hough_line = (VipsHoughLine *) object;
 	int width = hough_line->width;
 
-	int i;
-
 	if (!(hough_line->sin = VIPS_ARRAY(object, 2 * width, double)))
 		return -1;
 
 	/* Map width to 180 degrees, width * 2 to 360.
 	 */
-	for (i = 0; i < 2 * width; i++)
+	for (int i = 0; i < 2 * width; i++)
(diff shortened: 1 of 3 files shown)


jcupitt:
With this change, the sample ruby hough transform in the linked issue and libvips produce (almost) the same result.

libvips/arithmetic/hough_line.c
```diff
@@ -107,22 +105,26 @@ vips_hough_line_vote(VipsHough *hough, VipsImage *accumulator, int x, int y)
...
 	guint *data = (guint *) accumulator->data;
 
-	int i;
-
-	for (i = 0; i < width; i++) {
+	for (int i = 0; i < width; i++) {
 		int i90 = i + width / 2;
 		double r = xd * hough_line->sin[i90] + yd * hough_line->sin[i];
-		int ri = height * r;
+		int ri = r * (height / 2) + height / 2;
```
kleisauke, line 124: This can be simplified to:
```suggestion
		int ri = (r + 1) * (height / 2.0);
```

Though, I think the compiler already does this.

kleisauke (approved):

libvips/arithmetic/hough_line.c
jcupitt: You're right twice, the compiler will do that, and it's clearer.

jcupitt:
I updated the workspace, here's nip4 running it:

Image

http://www.rollthepotato.net/~john/hough4.ws

It seems to work fine for non-square images and non-square hough space.

PR #4526, opened 2025-05-20
Minor formatting improvements

kleisauke:
Non-functional change.

### DIFF
--- a/libvips/foreign/svgload.c
+++ b/libvips/foreign/svgload.c
@@ -29,7 +29,7 @@
  * 5/6/22
  * 	- allow random access
  * 20/5/25
- *  - support high bit depth rendering (128-bit)
+ * 	- support high bitdepth rendering (128-bit)
  */
 
 /*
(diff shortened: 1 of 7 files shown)


libvips/foreign/svgload.c
```diff
@@ -1115,7 +1116,7 @@ vips_foreign_load_svg_buffer_init(VipsForeignLoadSvgBuffer *buffer)
  * During the CSS cascade, the specified stylesheet will be applied with a
  * User Origin. This feature requires librsvg 2.48.0 or later.
  *
- * Setting @high_bitdepth TRUE enables 128-bit scRGB output."
+ * Set @high_bitdepth to enable 128-bit scRGB output (32-bit per channel).
```
jcupitt, line 1119: Maybe we should say "request"? The flag doesn't guarantee float output and you'll get a warning and uchar if it's not supported.

Perhaps `svgload` should fail if float rendering is not available? I can't decide :( 

OTOH, pretty soon all librsvgs will support this, so failure won't be possible and it won't matter.

jcupitt: ... let's leave this alone.

jcupitt:
\o/

PR #4528, opened 2025-05-21
doc: remove redundant rename scripts

kleisauke:
As mentioned in PR #4515.


jcupitt:
Oooop! Good catch, I forgot about them.

PR #4532, opened 2025-05-22
svgload: replaced vips_gamma with linearization through LUT usage

kstanikviacbs:
As it was pointed out in discussion here: https://github.com/libvips/libvips/pull/4529 - performing additional `vips_gamma` operation does not guarantee enough similarity between `high_biptdepth=true|false` outputs. 

As suggested - it turns out that `librsvg` - instead of simple pow operation applies sRGB curve as shown here:
https://github.com/GNOME/librsvg/blob/2.60.0/rsvg/build.rs#L28

This PR introduces port of mentioned logic and uses it as part of `svgload` pipeline. 

Please let me know if there's already an existing equivalent function available I should use instead and if there's a better place to apply it to pixels.

### DIFF
--- a/libvips/foreign/cairo.c
+++ b/libvips/foreign/cairo.c
@@ -127,88 +127,35 @@ vips__bgra2rgba(guint32 *restrict p, int n)
 	}
 }
 
-/**
- * @brief Converts a normal sRGB color value to a linear sRGB color value.
- *
- * This function removes the gamma correction from a non-linear sRGB component,
- * transforming it into a linear light intensity value. It implements the
- * inverse sRGB Electro-Optical Transfer Function (EOTF).
- *
- * @param c A single non-linear sRGB color component, expected in the [0, 1] range.
- * @return The corresponding linear sRGB color component, also in the [0, 1] range.
- *
- * @note This function is the C port of the `linearize` function found in librsvg.
- * For reference, the Rust source code for this function is available at:
- * https://github.com/GNOME/librsvg/blob/2.60.0/rsvg/build.rs#L28
- */
-static inline float
-linearize(float c)
-{
-	// The threshold (0.04045f) is the point where the linear segment of the
-	// sRGB transfer function (used for encoding) meets the power-law segment.
-	// This value is derived directly from the sRGB specification: 12.92 * 0.0031308.
-	if (c <= 0.04045f) {
-		return c / 12.92f;
-	}
-	else {
-		// Apply the inverse power-law function for sRGB linearization.
-		// All numerical literals use the 'f' suffix to explicitly ensure
-		// single-precision float-point arithmetic throughout the calculation.
-		return powf(((c + 0.055f) / 1.055f), 2.4f);
-	}
-}
+extern float vips_v2Y_16[65536];
+void vips_col_make_tables_RGB_16(void);
 
-/**
- * @brief Converts a row of Cairo-style premultiplied RGBA128F pixels to linear RGBA.
- *
- * This function processes 'n' pixels in the 'p' buffer. For each pixel, it performs
- * two main operations:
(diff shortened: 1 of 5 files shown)


jcupitt:
libvips already has this code (more or less!) here:

https://github.com/libvips/libvips/blob/master/libvips/colour/LabQ2sRGB.c#L151-L158

It'd be good to only have this in one place, though it's a pretty simple thing, perhaps it doesn't matter.

We could also use the 16 bit libvips LUT and do piecewise linear interpolation, it might be quicker.

It's a shame librsvg doesn't have a flag to enable linear output.

kstanikviacbs:
Yeah, I haven't found any option to control the output in terms of gamma in librsvg :( 

Thanks for you snippet, works on my end, but I'm not quite sure if 16-bit would be enough (it's all about rendering high quality gradients with SVG (4K) and thus quality here is quite crucial. I will give it a try and test few scenarios and will keep you posted.

jcupitt:
You could add piecewise linear interpolation with eg.:

```C
extern int vips_v2Y_16[];
static inline float
linearize_lut(float c)
{
    float cf = VIPS_CLIP(0, c * 65535, 65535);
    int ci = (int) cf;
    return vips_v2Y_16[ci] +
        (vips_v2Y_16[ci + 1] - vips_v2Y_16[ci]) * (cf - ci);
}
```

It's still a bit faster than the float path.

kstanikviacbs:
Thanks for your support with the snippets and being involved in discussion @jcupitt . I've checked outputs with my requirements and the first approach with 16-bit LUT (no interpolation) was already good enough so I updated the PR with it.

I also did a few comparison tests for `high-bitdepth` outputs and it looked very similar this time.

libvips/foreign/cairo.c
```diff
@@ -127,21 +127,31 @@ vips__bgra2rgba(guint32 *restrict p, int n)
 	}
 }
 
+extern float vips_v2Y_16[65536];
```
jcupitt, line 130: This is a bit ugly. Let's move this declaration to `include/vips/internal.h`, and remove it from `colour/pcolour.h`.

Same for `vips_col_make_tables_RGB_16()`.

libvips/foreign/cairo.c
```diff
@@ -127,21 +127,31 @@ vips__bgra2rgba(guint32 *restrict p, int n)
...
-		float b = p[2];
+		float r = VIPS_CLIP(0, p[0], 1);
+		float g = VIPS_CLIP(0, p[1], 1);
+		float b = VIPS_CLIP(0, p[2], 1);
 		float a = p[3];
 
+		// linearize the values with LUT
+		r = vips_v2Y_16[(int) (65535 * r)];
+		g = vips_v2Y_16[(int) (65535 * g)];
+		b = vips_v2Y_16[(int) (65535 * b)];
```
kleisauke, line 153: nit: I was going to suggest doing `FCLIP` here (which is slower), but then I noticed we just want an int result. How about doing this instead?:
```suggestion
		// CLIP is much faster than FCLIP, and we want an int result
		int ri = VIPS_CLIP(0, (int) (p[0] * 65535), 65535);
		int gi = VIPS_CLIP(0, (int) (p[1] * 65535), 65535);
		int bi = VIPS_CLIP(0, (int) (p[2] * 65535), 65535);

		// linearize the values with LUT
		float r = vips_v2Y_16[ri];
		float g = vips_v2Y_16[gi];
		float b = vips_v2Y_16[bi];
		float a = p[3];
```

libvips/foreign/cairo.c
kstanikviacbs: Agreed, I wasn't sure what would be the best way to deal with it, thanks for catching that. Applied.

jcupitt:
\o/

PR #4533, opened 2025-05-22
Apply i18n policy to all g_warning messages

kleisauke:
In line with PR #4369.

This change was mostly automated with ast-grep:
```yaml
id: remove_i18n_from_g_warning
language: c
# Define rewriters
rewriters:
  - id: i18n-remove
    rule:
      kind: call_expression
      pattern: $ARG
    constraints:
      ARG: { regex: ^_ } # Only match if it starts with '_'
    transform:
      NEW_ARG:
        substring:
          source: $ARG
          startChar: 2
          endChar: -1
    fix: $NEW_ARG
# Find the target node
rule:
  pattern:
    context: g_warning($$$ARGS);
    selector: call_expression
# Apply rewriters to sub node
transform:
  REMOVE_I18N_ARGS:
    rewrite:
      rewriters: [i18n-remove]
      source: $$$ARGS
fix: g_warning($REMOVE_I18N_ARGS)
```
```console
$ ast-grep scan -r remove-i18n-from-g_warning.yml -U libvips/
Applied 106 changes
```

### DIFF
--- a/libvips/arithmetic/measure.c
+++ b/libvips/arithmetic/measure.c
@@ -161,10 +161,10 @@ vips_measure_build(VipsObject *object)
 				 */
 				if (dev * 5 > fabs(avg) &&
 					fabs(avg) > 3)
-					g_warning(_("%s: "
-								"patch %d x %d, "
-								"band %d: "
-								"avg = %g, sdev = %g"),
+					g_warning("%s: "
+							  "patch %d x %d, "
+							  "band %d: "
+							  "avg = %g, sdev = %g",
 						class->nickname,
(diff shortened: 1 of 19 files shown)


kleisauke:
Good point, I've just removed the no longer used entries from the .po files. We should probably re-run `meson compile vips8.17-update-po` as a follow-up, as there are many line numbers that have also changed.

lovell (approved):

jcupitt:
I'd not come across ast-grep, that's really useful! What a shame it uses YAML though haha

PR #4537, opened 2025-05-26
test: remove compat for old libvips and pyvips versions

kleisauke:
Assume the test suite is run with the latest versions of libvips and pyvips.

### DIFF
--- a/test/test-suite/test_connection.py
+++ b/test/test-suite/test_connection.py
@@ -102,8 +102,6 @@ class TestConnection:
 
         assert (im - self.mono).abs().max() == 0
 
-    @skip_if_no("csvload_source")
-    @skip_if_no("csvsave_target")
     def test_connection_csv(self):
         x = pyvips.Target.new_to_memory()
         self.mono.csvsave_target(x)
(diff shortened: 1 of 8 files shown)


test/test-suite/test_connection.py
```diff
@@ -102,8 +102,6 @@ def test_connection_matrix(self):
 
         assert (im - self.mono).abs().max() == 0
 
-    @skip_if_no("csvload_source")
-    @skip_if_no("csvsave_target")
```
jcupitt, line 106: We can disable csv load and save with a meson option, so maybe this test should stay?

test/test-suite/test_iofuncs.py
```diff
@@ -71,18 +71,18 @@ def test_revalidate(self):
...
-        load1 = pyvips.Image.new_from_file(filename);
+        load1 = pyvips.Image.new_from_file(filename)
         assert load1.width == im1.width
 
         im2 = pyvips.Image.black(20, 20)
         im2.write_to_file(filename)
 
         # this will use the old, cached load
-        load2 = pyvips.Image.new_from_file(filename);
+        load2 = pyvips.Image.new_from_file(filename)
```
jcupitt, line 81: hehe oops

test/test-suite/test_connection.py
kleisauke: There are no Meson options to disable CSV load/save, so we can assume it's always available. This is likely because the CSV loader must be explicitly called (since `->is_a_source` always returns `FALSE`), meaning it doesn't contribute to the potential attack surface.

PR #4543, opened 2025-05-29
use g_realloc, not realloc

jcupitt:
I noticed some (probably very seldom used!) code that was calling realloc/free. 

Calling realloc in a loop like this is probably incredibly slow, especially on windows, though I doubt if it matters much, I think we won't hit this path often. It would be better to grow the allocation by successive doubling, we'd get O(logN), not O(N).

I fixed a few bad line breaks too.


kleisauke (approved):

jcupitt:
:+1:

PR #4548, opened 2025-06-02
foreign: inline parent build chain calls where applicable

kleisauke:
Non-functional change. Also, ensure buffer loaders consistently reference the correct parent class (some were incorrectly referencing the file loader's parent class).


jcupitt:
Nice! Yes, it was a bit messy before.

PR #4550, opened 2025-06-03
Lower libjxl version requirement to 0.9.x

kleisauke:
RHEL 10 (and its derivatives) only includes libjxl 0.10.3 in EPEL. EPEL 8 and 9 provide only libjxl 0.7.0, and even that upgrade was somewhat controversial due to ABI/API concerns (see e.g. https://phpc.social/@remi/109381841149751828).

@remicollet FYI. This would ensure that the `vips-jxl` package remains available for RHEL 10 with the upcoming libvips 8.17 release in your repository (it can't be lowered any further, unfortunately).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -30,7 +30,7 @@ master
 - move vips_image_preeval(), vips_image_eval(), vips_image_posteval() into the
   public API
 - jxlsave writes in chunks, for lower memory use on large images (but we now
-  need libjxl 0.11 at minimum)
+  need libjxl 0.9 at minimum)
 - increase minimum version of libheif dependency to 1.11.0
 - improve scaling of hough_line feature space [ecbypi]
 - heifload: limit per-image memory usage to 2GB (requires libheif 1.20.0+)
--- a/meson.build
+++ b/meson.build
@@ -506,9 +506,9 @@ if libheif_dep.found()
     cfg_var.set('HAVE_HEIF_MAX_TOTAL_MEMORY', cpp.has_member('struct heif_security_limits', 'max_total_memory', prefix: '#include <libheif/heif.h>', dependencies: libheif_dep))
 endif
 
-# need v0.11+ for chunked write
-libjxl_dep = dependency('libjxl', version: '>=0.11', required: get_option('jpeg-xl'))
-libjxl_threads_dep = dependency('libjxl_threads', version: '>=0.11', required: get_option('jpeg-xl'))
+# need v0.9+ for JxlEncoderAddChunkedFrame()
+libjxl_dep = dependency('libjxl', version: '>=0.9', required: get_option('jpeg-xl'))
+libjxl_threads_dep = dependency('libjxl_threads', version: '>=0.9', required: get_option('jpeg-xl'))
 libjxl_found = libjxl_dep.found() and libjxl_threads_dep.found()
 libjxl_module = false
 if libjxl_found


jcupitt:
I think I made it 0.11 (from memory!) because some aspects of chunked write only started working properly then. Though now of course I can't remember which bits didn't work properly :(

We might need to put the old write path back and use ifdefs, or simply revert chunked write.

jcupitt:
I reread the PR for chunked jxl save and maybe the 0.11 requirement came from HDR support, which (in the end) we never implemented?

https://github.com/libvips/libvips/pull/4174

kleisauke:
I asked the JPEG XL devs on Discord (in the `#libjxl` channel) for clarification. It looks like `JPEGXL_SO_MINOR_VERSION` is always incremented, even when there are no ABI/API changes.
https://github.com/libjxl/libjxl/blob/v0.11.1/lib/CMakeLists.txt#L12-L24

remicollet:
> @remicollet FYI. This would ensure that the `vips-jxl` package remains available for RHEL 10 with the upcoming libvips 8.17 release in your repository (it can't be lowered any further, unfortunately).

As EPEL-8 and EPEL-9 still have 0.7, this means the vips-jxl feature will be **removed**.
This sounds terrible (especially as EL-10 is very young, clones are not even released, most users are still using 9)

Of course, 0.11 is more problematic (Only Fedora >= 42), while 0.10 is available on Fedora >= 41 and EPEL-10

Perhaps I should consider providing the new vips version only on EL-10 (and Fedora)

remicollet:
It would be nice if the new feature (requiring a bump requirement) were implemented conditionally...

jcupitt:
I did some testing and you're right, 0.9 seems OK. Phew! Maybe it was needed for the HDR stuff I was looking at and that got removed before the merge.

@remicollet supporting 0.7 will mean having loads of redundant code hidden behind ifdefs :( It's a difficult choice -- rapidly changing libraries like libjxl are always tricky to support. I expect the Rust rewrite they are doing now will force more breakage when it lands. 

I'll see if we can put the libjxl 0.7 save path back.

kleisauke:
> It's a difficult choice -- rapidly changing libraries like libjxl are always tricky to support.

Indeed. Fortunately, libjxl is now considering keeping `JPEGXL_SO_MINOR_VERSION` at `11`, as no ABI/API changes are expected in the upcoming v0.12.x release.

> I'll see if we can put the libjxl 0.7 save path back.

I just opened PR #4551 for this. :tada:

kleisauke:
Note to self: libheif may need to be lowered to 1.7.0 (RPM Fusion provides version 1.7.0 on EL8, and version 1.16.1 on EL9 via the `libheif-freeworld` package).

jcupitt:
> libheif may need to be lowered to 1.7.0

Oooof, that's true. @remicollet please let us know what else you run into with 8.17.

remicollet:
> Note to self: libheif may need to be lowered to 1.7.0 (RPM Fusion provides version 1.7.0 on EL8, and version 1.16.1 on EL9 via the libheif-freeworld package).

More exactly:
- EL-8 have libheif 1.7.0 from RPMfusion
- EL-9 have libheif 1.16.1 from EPEL-9 (libheif-freeworld is optional, only for the optional encoder)
- EL-10 have libheif 1.17.2 from EPEL-10 
- Fedora have libheif 1.19.8 from official repo

Yes, requiring 1.11.0 is indeed a problem.

jcupitt:
Kleis has skilfully lowered it to 1.7.0: https://github.com/libvips/libvips/pull/4552

PR #4552, opened 2025-06-04
Lower libheif version requirement to 1.7.0

kleisauke:
For compat with RHEL 8 (and its derivatives).

Context: #4550.


jcupitt:
Lovely.

Issue #3380, opened 2023-03-12
vips_magickload behaves differently from vips_magickload_buffer when dealing with NEF files


lkempf:
## Bug report

**Describe the bug**
When loading a NEF file vips_magickload produces a image of the correct dimension while vips_magickload_buffer produces a image of dimension 160x120 (probably some embedded thumbnail). This small image is also produced when loading the image as a TIFF (which is what libvips identifies the image as).

**To Reproduce**
1. Use Image https://www.mannyphoto.com/D700D3/_DSC4175.NEF
2. Execute the following in python
   ```
import pyvips
img_bytes = open('./_DSC4175.NEF', 'rb').read()
img1 = pyvips.Image.magickload('./_DSC4175.NEF')
img2 = pyvips.Image.magickload_buffer(img_bytes)
print(img1.width, img2.width)
   ```
3. Output is `4284 160` not the expected `4284 4284`.

**Expected behavior**
Both methods should produce the same image.

**Actual behavior**
Loading from buffer produces a much smaller image.

**Environment**
 - OS: Arch Linux
 - Vips: vips-8.14.1

kleisauke:
ImageMagick doesn't have a buffer recognizer for Nikon's NEF file format. Since this file format uses the standard TIFF header, it tries to load the buffer through its TIFF loader. libvips has a special path to workaround this limitation of *magick, but that does not include support for NEF. 

I had a quick attempt to add a NEF sniffer for this, see e.g. commit https://github.com/kleisauke/libvips/commit/ab5d0e356f4f6f2ca4cbca2a07c0aff7b3a8ae34, but somehow(?) this would cause a regression on the file-based loader, as it outputs `160 4284` on the above reproducer.

lkempf:
That's not the output of the file based loaders, that's the width for both loaders. Sorry for the confusing output format.

kleisauke:
> this would cause a regression on the file-based loader, as it outputs `160 4284` on the above reproducer.

I just reported this upstream at https://github.com/ImageMagick/ImageMagick/issues/6242.

kleisauke:
I just opened PR #3460 for this.

uhthomas:
I can see that https://github.com/libvips/libvips/pull/3460 did not actually include a fix for this @kleisauke. Do you have any ideas on how we can reliably differentiate between NEF and TIFF files?

kleisauke:
[...]

PR #4562, opened 2025-06-09
Add the basic support for loading RAW format files with libRAW

lxsameer:
This PR adds support for basic libraw supports. I tried to do the bare minimum so I can get early feedback and build upon that.

Looking forward to your feedback.

### DIFF
--- a/libvips/foreign/dcrawload.c
+++ b/libvips/foreign/dcrawload.c
@@ -1,0 +1,342 @@
+/* load RAW camera files with libraw
+ */
+
+/*
+
+	This file is part of VIPS.
+
+	VIPS is free software; you can redistribute it and/or modify
+	it under the terms of the GNU Lesser General Public License as published by
+	the Free Software Foundation; either version 2 of the License, or
+	(at your option) any later version.
+
+	This program is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU Lesser General Public License for more details.
+
+	You should have received a copy of the GNU Lesser General Public License
+	along with this program; if not, write to the Free Software
+	Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+	02110-1301  USA
+
+ */
+
+/*
+
+	These files are distributed with VIPS - http://www.vips.ecs.soton.ac.uk
+
+ */
+
+/*
+#define DEBUG
+ */
+
+#ifdef HAVE_CONFIG_H
+#include <config.h>
+#endif /*HAVE_CONFIG_H*/
+#include <glib/gi18n-lib.h>
+
+#include <stdio.h>
+#include <stdlib.h>
+
+#include <vips/vips.h>
+
+#ifdef HAVE_LIBRAW
+
+#include <libraw/libraw.h>
+
+#define VIPS_TYPE_FOREIGN_LOAD_DCRAW (vips_foreign_load_dcraw_get_type())
+#define VIPS_FOREIGN_LOAD_DCRAW(obj) \
+	(G_TYPE_CHECK_INSTANCE_CAST((obj), \
+		VIPS_TYPE_FOREIGN_LOAD_DCRAW, VipsForeignLoadDcRaw))
+
+static const char *vips_foreign_dcraw_suffs[] = {
+	".arw", ".cr2", ".cr3", ".crw", ".dng", ".nef", ".nrw",
+	".orf", ".pef", ".raf", ".raw", ".rw2", ".srw", ".x3f",
+	".erf", ".kdc", ".mdc", ".mos", ".pxn", ".srf",
+	".3fr", ".ari", ".cap", ".cin", ".dcr", ".fff", ".iiq", ".k25",
+	".mrw", ".ori", ".rwl", ".sr2", NULL
+};
+
+typedef struct _VipsForeignLoadDcRaw {
+	VipsForeignLoad parent_object;
+
+	/* LibRaw processor.
+	 */
+	libraw_data_t *raw_processor;
+
+	/* Internal Libraw processed image
+	 */
+	libraw_processed_image_t *processed;
+
+	/* Load from this source (set by subclasses).
+	 */
+	VipsSource *source;
+
+} VipsForeignLoadDcRaw;
+
+typedef VipsForeignLoadClass VipsForeignLoadDcRawClass;
+
+G_DEFINE_ABSTRACT_TYPE(VipsForeignLoadDcRaw, vips_foreign_load_dcraw,
+	VIPS_TYPE_FOREIGN_LOAD);
+
+static void
+vips_foreign_load_dcraw_dispose(GObject *gobject)
+{
+	VipsForeignLoadDcRaw *raw = (VipsForeignLoadDcRaw *) gobject;
+
+	VIPS_FREEF(libraw_dcraw_clear_mem, raw->processed);
+	VIPS_FREEF(libraw_close, raw->raw_processor);
+	VIPS_UNREF(raw->source);
+
+	G_OBJECT_CLASS(vips_foreign_load_dcraw_parent_class)->dispose(gobject);
+}
+
+static int
+vips_foreign_load_dcraw_build(VipsObject *object)
+{
+	VipsForeignLoadDcRaw *raw = VIPS_FOREIGN_LOAD_DCRAW(object);
+	VipsObjectClass *class = VIPS_OBJECT_GET_CLASS(raw);
+	raw->raw_processor = libraw_init(0);
+
+	if (!raw->raw_processor) {
+		vips_error(class->nickname, _("unable to initialize libraw"));
+		return -1;
+	}
+
+	return VIPS_OBJECT_CLASS(vips_foreign_load_dcraw_parent_class)
+		->build(object);
+}
+
+static VipsForeignFlags
+vips_foreign_load_dcraw_get_flags_filename(const char *filename)
+{
+	return 0;
+}
+
+static VipsForeignFlags
+vips_foreign_load_dcraw_get_flags(VipsForeignLoad *load)
+{
+	return 0;
+}
+
+static int
+vips_foreign_load_dcraw_header(VipsForeignLoad *load)
+{
+	VipsForeignLoadDcRaw *raw = (VipsForeignLoadDcRaw *) load;
+	VipsObjectClass *class = VIPS_OBJECT_GET_CLASS(raw);
+
+	const void *data;
+	size_t length;
+	VipsImage *image;
+	VipsImage *x;
+	GDateTime *dt;
+
+	/* Map the whole source into memory.
+	 */
+	if (!(data = vips_source_map(raw->source, &length)))
+		return -1;
+
+	/* Enforce VIPS_FORMAT_USHORT output.
+	 */
+	raw->raw_processor->params.output_bps = 16;
+
+	/* Apply camera white balance.
+	 */
+	raw->raw_processor->params.use_camera_wb = 1;
+
+	int result = libraw_open_buffer(raw->raw_processor, data, length);
+	if (result != LIBRAW_SUCCESS) {
+		vips_error(class->nickname, "%s : %s",
+			_("unable to read the source"), libraw_strerror(result));
+		return -1;
+	}
+
+	result = libraw_unpack(raw->raw_processor);
+	if (result != LIBRAW_SUCCESS) {
+		vips_error(class->nickname, "%s : %s",
+			_("unable to unpack the source"), libraw_strerror(result));
+		return -1;
+	}
+
+	/* Process the image (demosaicing, white balance, etc.)
+	 */
+	result = libraw_dcraw_process(raw->raw_processor);
+
+	if (result != LIBRAW_SUCCESS) {
+		vips_error(class->nickname,
+			"error processing RAW data: %s\n", libraw_strerror(result));
+		return -1;
+	}
+
+	if (!(raw->processed =
+				libraw_dcraw_make_mem_image(raw->raw_processor, &result))) {
+		vips_error(class->nickname,
+			"error creating processed image: %s\n", libraw_strerror(result));
+		return -1;
+	}
+
+	if (!(image = vips_image_new_from_memory(
+			  raw->processed->data, raw->processed->data_size,
+			  raw->processed->width, raw->processed->height,
+			  raw->processed->colors, VIPS_FORMAT_USHORT)))
+		return -1;
+
+	VIPS_SETSTR(image->filename,
+		vips_connection_filename(VIPS_CONNECTION(raw->source)));
+
+	/* Set custom metadata.
+	 */
+	vips_image_set_string(image, "raw-make",
+		raw->raw_processor->idata.make);
+	vips_image_set_string(image, "raw-model",
+		raw->raw_processor->idata.model);
+	vips_image_set_double(image, "raw-iso",
+		raw->raw_processor->other.iso_speed);
+	vips_image_set_double(image, "raw-shutter",
+		raw->raw_processor->other.shutter);
+	vips_image_set_double(image, "raw-aperture",
+		raw->raw_processor->other.aperture);
+	vips_image_set_double(image, "raw-focal-length",
+		raw->raw_processor->other.focal_len);
+
+	if (raw->raw_processor->other.timestamp &&
+		(dt = g_date_time_new_from_unix_utc(
+			 raw->raw_processor->other.timestamp))) {
+		vips_image_set_string(image, "raw-timestamp",
+			g_date_time_format_iso8601(dt));
+		g_date_time_unref(dt);
+	}
+
+	/* What a hack. Remove the @out that's there now and replace it with
+	 * our image.
+	 */
+	g_object_get(load, "out", &x, NULL);
+	g_object_unref(x);
+	g_object_unref(x);
+
+	g_object_set(load, "out", image, NULL);
+
+	return 0;
+}
+
+static void
+vips_foreign_load_dcraw_class_init(VipsForeignLoadDcRawClass *class)
+{
+	GObjectClass *gobject_class = G_OBJECT_CLASS(class);
+	VipsObjectClass *object_class = (VipsObjectClass *) class;
+	VipsOperationClass *operation_class = VIPS_OPERATION_CLASS(class);
+	VipsForeignLoadClass *load_class = (VipsForeignLoadClass *) class;
+
+	gobject_class->dispose = vips_foreign_load_dcraw_dispose;
+
+	object_class->nickname = "dcrawload_base";
+	object_class->description = _("load RAW camera files");
+	object_class->build = vips_foreign_load_dcraw_build;
+
+	/* LibRaw is fuzzed, but not by us.
+	 */
+	operation_class->flags |= VIPS_OPERATION_UNTRUSTED;
+
+	load_class->get_flags = vips_foreign_load_dcraw_get_flags;
+	load_class->header = vips_foreign_load_dcraw_header;
+	load_class->load = NULL;
+}
+
+static void
+vips_foreign_load_dcraw_init(VipsForeignLoadDcRaw *raw)
+{
+}
+
+typedef struct _VipsForeignLoadDcRawFile {
+	VipsForeignLoadDcRaw parent_object;
+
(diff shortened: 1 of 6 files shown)


libvips/foreign/librawload.c
```diff
@@ -0,0 +1,368 @@
...
+}
+
+static int
+vips_foreign_load_libraw_header(VipsForeignLoad *load)
+{
+	VipsForeignLoadLibRaw *raw = (VipsForeignLoadLibRaw *) load;
+	int result;
+
+	if (raw->source == NULL) {
+		vips_error("librawload", "source property is nulll!\n");
```
jcupitt: You can get this from the class with this at the top of the function:

```C
    VipsObjectClass *class = VIPS_OBJECT_GET_CLASS(raw);
```

Then call `vips_error()` with eg.:

```C
vips_error(class->nickname, "source property is null!");
```

jcupitt: ... although for this exact `if()`, I'd use `g_assert(raw->source);`, since a NULL source pointer will be an internal libvips error, not a user error.

jcupitt: There's no trailing `\n` on vips error messages.

libvips/foreign/librawload.c
```diff
@@ -0,0 +1,368 @@
...
+
+	gobject_class->dispose = vips_foreign_load_libraw_dispose;
+	gobject_class->set_property = vips_object_set_property;
+	gobject_class->get_property = vips_object_get_property;
+
+	object_class->nickname = "librawload_base";
+	object_class->description = _("load raw camera files");
+	object_class->build = vips_foreign_load_libraw_build;
+
+	foreign_class->priority = 100;
```
jcupitt: That's quite high priority, are you are it needs to be this high? Slower sniffers need to be near the back of the queue or they'll delay things like JPEG detection.

libvips/foreign/librawload.c
```diff
@@ -0,0 +1,368 @@
...
+
+	foreign_class->suffs = vips_foreign_libraw_suffs;
+
+	load_class->get_flags_filename = vips_foreign_load_libraw_get_flags_filename;
+	load_class->get_flags = vips_foreign_load_libraw_get_flags;
+	load_class->header = vips_foreign_load_libraw_header;
+	load_class->load = vips_foreign_load_libraw_load;
+
+	VIPS_ARG_STRING(class, "filename", 1,
```
jcupitt: The filename arg should be on the file loader, not the base class.

libvips/foreign/librawload.c
```diff
@@ -0,0 +1,368 @@
...
+vips_foreign_load_libraw_file_class_init(VipsForeignLoadLibRawFileClass *class)
+{
+	GObjectClass *gobject_class = G_OBJECT_CLASS(class);
+	VipsObjectClass *object_class = (VipsObjectClass *) class;
+	VipsForeignClass *foreign_class = (VipsForeignClass *) class;
+
+	gobject_class->set_property = vips_object_set_property;
+	gobject_class->get_property = vips_object_get_property;
+
+	object_class->nickname = "libraw_load";
```
jcupitt: This should be `librawload` for consistency with other loaders.

libvips/foreign/librawload.c
```diff
@@ -0,0 +1,368 @@
...
+vips_foreign_load_libraw_class_init(VipsForeignLoadLibRawClass *class)
+{
+	GObjectClass *gobject_class = G_OBJECT_CLASS(class);
+	VipsObjectClass *object_class = (VipsObjectClass *) class;
+	VipsForeignClass *foreign_class = (VipsForeignClass *) class;
+	VipsForeignLoadClass *load_class = (VipsForeignLoadClass *) class;
+
+	gobject_class->dispose = vips_foreign_load_libraw_dispose;
+	gobject_class->set_property = vips_object_set_property;
+	gobject_class->get_property = vips_object_get_property;
```
jcupitt: You only need `get_property` and `set_property` for classes with arguments.

libvips/foreign/librawload.c
```diff
@@ -0,0 +1,368 @@
...
+}
+
+static int
+vips_foreign_load_libraw_file_build(VipsObject *object)
+{
+	VipsForeignLoadLibRaw *raw = VIPS_FOREIGN_LOAD_LIBRAW(object);
+
+	if (raw->filename)
+		if (!(raw->source = vips_source_new_from_file(raw->filename))) {
+			vips_error("libraw_load", _("Can not read the input file into a source"));
```
jcupitt: You only need to set an error message for errors you discover directly. Here you can just write:

```C
if (raw->filename && 
	!(raw->source = vips_source_new_from_file(raw->filename)))
	return -1;
```

libvips/foreign/librawload.c
```diff
@@ -0,0 +1,368 @@
...
+	".erf", ".kdc", ".mdc", ".mos", ".pxn", ".srf",
+	NULL
+};
+
+typedef struct _VipsForeignLoadLibRaw {
+	VipsForeignLoad parent_object;
+
+	/* Filename for load.
+	 */
+	char *filename;
```
jcupitt: This should be in the `VipsForeignLoadLibRawFile` subclass.

libvips/foreign/librawload.c
```diff
@@ -0,0 +1,368 @@
...
+
+G_DEFINE_TYPE(VipsForeignLoadLibRawFile, vips_foreign_load_libraw_file,
+	vips_foreign_load_libraw_get_type());
+
+static void
+vips_foreign_load_libraw_dispose(GObject *gobject)
+{
+	VipsForeignLoadLibRaw *raw = (VipsForeignLoadLibRaw *) gobject;
+
+	if (raw->processed) {
```
jcupitt: There's a macro for this, you can write:

```C
VIPS_FREEF(libraw_dcraw_clear_mem, raw->processed);
```

FREEF meaning **FREE** with **F**unction.

libvips/foreign/librawload.c
```diff
@@ -0,0 +1,368 @@
...
+}
+
+static VipsForeignFlags
+vips_foreign_load_libraw_get_flags(VipsForeignLoad *load)
+{
+	return 0;
+}
+
+static int
+vips_foreign_load_libraw_header(VipsForeignLoad *load)
```
jcupitt: libvips loaders are split into something to just get the metadata (the `_header()` method), and something to decode the image (the `_load()` method).

libraw doesn't support header read, so I'd put everything into `_load()` and just not define `_header()`.

libvips/foreign/meson.build
```diff
@@ -123,6 +123,14 @@ if not openslide_module
     foreign_sources += openslide_module_sources
 endif
 
+libraw_sources = files(
```
jcupitt: You only need special rules here if we want to make the raw loader a loadable module. For now, let's just add `librawload.c` to `foreign_sources` at the top of `meson.build`.

meson_options.txt
```diff
@@ -236,6 +236,11 @@ option('nsgif',
   value: true,
   description: 'Build with nsgif')
 
+option('libraw',
+  type: 'boolean',
+  value: true,
```
jcupitt: We want this to work like tiff, so:

```meson
option('libraw',
  type: 'feature',
  value: 'auto',
  description: 'Build with libraw')
```

Would probably be better.

libvips/foreign/librawload.c
```diff
@@ -0,0 +1,368 @@
...
+
+	if (raw->source == NULL) {
+		vips_error("librawload", "source property is nulll!\n");
+		return -1;
+	}
+
+	/* Use the source object as a buffer foor libraw.
+	 */
+	result = libraw_open_buffer(raw->raw_processor,
+		vips_source_sniff(raw->source, vips_source_length(raw->source)),
```
jcupitt: This is a thing to just get the first few bytes of the file for quick file format detection. 

Although it will work for getting the whole file, it's better to use `vips_source_map()`, perhaps:

```C
    size_t length;
	result = libraw_open_buffer(raw->raw_processor,
		vips_source_map(raw->source, &length), length); 
```

jcupitt: ... though you should really detect a failure from `source_map()`, so I'd write:

```C
void *data;
size_t length;
if (!(data = vips_source_map(raw->source, &length))
    return -1;
result = libraw_open_buffer(raw->raw_processor, data, length);
if (result != LIBRAW_SUCCESS) {
    vips_error(class->nickname, "%s : %s", _("read error"), libraw_strerror(result));
    return -1;
}
```

libvips/foreign/librawload.c
```diff
@@ -0,0 +1,368 @@
...
+		vips_source_length(raw->source));
+
+	if (result != LIBRAW_SUCCESS) {
+		vips_error("librawload", "%s : %s", _("unable to read the source"), libraw_strerror(result));
+		return -1;
+	}
+
+	libraw_unpack(raw->raw_processor);
```
jcupitt: I think this can fail, from a quick look at the API docs.

libvips/foreign/librawload.c
```diff
@@ -0,0 +1,368 @@
...
+	gobject_class->set_property = vips_object_set_property;
+	gobject_class->get_property = vips_object_get_property;
+
+	object_class->nickname = "librawload_base";
+	object_class->description = _("load raw camera files");
+	object_class->build = vips_foreign_load_libraw_build;
+
+	foreign_class->priority = 100;
+
+	foreign_class->suffs = vips_foreign_libraw_suffs;
```
jcupitt: You only need to define this on the subclass that handles filename load.

jcupitt:
This is great @lxsameer, thank you for doing this work!

It seems to work OK for me, here's nip4 loading a cr2 sample:

image

I made some (mostly) general comments.

I think the big one is probably to only define `_load()` and not `_header()`. Unfortunately, libraw seems not to have a thing to detect supported RAW files quickly, nor something to just get the metadata. This means this will have to be a low priority loader, like `magickload` (which has similar problems).

libvips/foreign/librawload.c
kleisauke: I suspect this is due to the issue mentioned in https://github.com/libvips/libvips/issues/3380#issuecomment-1676364634. As an alternative, we could lower the priority of `tiffload`, guarded by `HAVE_LIBRAW`.

mertalev:
Would you be able to set `output_bps=16` and `use_camera_wb=1`  by default like [ImageMagick](https://github.com/ImageMagick/ImageMagick/blob/a024f0560a87ba405ad0d86e9ecc754048fad608/coders/dng.c#L293)?

libvips/foreign/librawload.c
kleisauke: ... thinking about this further, we also could prevent TIFF-based RAW images from being loaded via libtiff. I previously mentioned this approach in the linked comment above and just cleaned up the commit (now guarded behind `HAVE_LIBRAW`) in https://github.com/kleisauke/libvips/commit/955c46776e12060475ae0e8e3075736ba0128e2e.

While this solution isn't very future-proof (and is essentially a wall of shame), it might help stop vendors from using standard TIFF signatures for their own custom RAW formats. :sweat_smile:

lxsameer:
Hey folks, thanks for the review, I'll improve the code and ping you again for a review

lxsameer:
@jcupitt just a quick question, do you have a certain priority in mind?

jcupitt:
Now I think again, I think you can just not set priority. 

You define a set of suffixes and no `_is_a()` method, so your loader will be picked based JUST on the filename suffix, there will be no file type testing and the performance won't matter.

jcupitt:
I had another thought last night -- the `librawload` name is a bit awkward, we don't use the lib prefix anywhere else. How about `dcrawload`, since libraw is a wrapper for `dcraw`?

lxsameer:
@jcupitt absolutely, I'll change the name

libvips/foreign/librawload.c
lxsameer: I removed the ` _header()` and merged it into `_load()`. Now I'm in trouble, do I still need to set `load->out`?

lxsameer:
Updated the commit. But unfortunately, in my test program, when I call `vips_image_write_to_file` on the VipsImage that I get from calling `vips_dcrawload`, I get the following error:

```
VipsRegion: images do not match in pixel size
```
I don't know what might be the issue, as far as I can see, I set the same size for both `load->out` and `load->real`. Any idea?

libvips/foreign/librawload.c
kleisauke: Looking at:
https://github.com/libvips/libvips/blob/594187812907bca9499205b8d2e0e35d724f5e3f/libvips/foreign/foreign.c#L1149-L1151

I think it's vice-versa, so put everything into `_header()` and set `->load` to `NULL`, just like `vipsload`.

kleisauke:
It looks like everything needs to be moved into `_header()`, and `->load` should be set to `NULL`, similar to how it's done in `vipsload`. I did a quick refactor with commit https://github.com/kleisauke/libvips/commit/624accf8dca95ae6add403ea459d7d57f93505c8, which seems to work on my end. Feel free to cherry-pick.

BTW, seeing `#include <vips/intl.h>` made me wonder, is this code already used somewhere? Or just based on an old loader by accident?

libvips/foreign/librawload.c
jcupitt: Ooop! My bad, sorry.

libvips/foreign/dcrawload.c
```diff
@@ -0,0 +1,373 @@
...
+
+	/* Use the source object as a buffer for libraw.
+	 */
+
+	data = vips_source_map(raw->source, &length);
+	if (!data)
+		return -1;
+
+	raw->raw_processor->params.output_bps = 16;
+	raw->raw_processor->params.use_auto_wb = 1;
```
mertalev: The param I mentioned is `use_camera_wb`, not `use_auto_wb` (though it seems ImageMagick sets both?). See [this](https://github.com/immich-app/immich/pull/3806) for more context.

libvips/foreign/dcrawload.c
```diff
@@ -0,0 +1,373 @@
...
+#define VIPS_FOREIGN_LOAD_DCRAW(obj) \
+	(G_TYPE_CHECK_INSTANCE_CAST((obj), \
+		VIPS_TYPE_FOREIGN_LOAD_DCRAW, VipsForeignLoadDcRaw))
+
+static const char *vips_foreign_dcraw_suffs[] = {
+	".arw", ".cr2", ".cr3", ".crw", ".dng", ".nef", ".nrw",
+	".orf", ".pef", ".raf", ".raw", ".rw2", ".srw", ".x3f",
+	".erf", ".kdc", ".mdc", ".mos", ".pxn", ".srf",
+	NULL
+};
```
mertalev: This is missing some formats: `".3fr", ".ari", ".cap", ".cin", ".dcr", ".fff", ".iiq", ".k25", ".mrw", ".ori", ".rwl", ".sr2"` (comparing with [this](https://github.com/immich-app/immich/blob/77a362f0c0fa6f7757519ab664c68c1ba0253cf2/server/src/utils/mime-types.ts#L4)).

lxsameer:
@kleisauke Thank you, used your commit in my and added you as the co-author

libvips/foreign/dcrawload.c
```diff
@@ -0,0 +1,343 @@
...
+#ifdef HAVE_LIBRAW
+
+#include <libraw/libraw.h>
+
+#define VIPS_TYPE_FOREIGN_LOAD_DCRAW (vips_foreign_load_dcraw_get_type())
+#define VIPS_FOREIGN_LOAD_DCRAW(obj) \
+	(G_TYPE_CHECK_INSTANCE_CAST((obj), \
+		VIPS_TYPE_FOREIGN_LOAD_DCRAW, VipsForeignLoadDcRaw))
+
+static const char *vips_foreign_dcraw_suffs[] = {
```
mertalev, line 54: Question: is the suffix converted to lower case before checking this array? I assume it is

mertalev (commented):
I can't comment much on the overall implementation, but the extensions and libraw params look good!

libvips/foreign/dcrawload.c
jcupitt: Yes, libvips does case folding on that test.

lxsameer: i hope so

meson.build
```diff
@@ -429,6 +429,12 @@ if openexr_dep.found()
     cfg_var.set('HAVE_OPENEXR', true)
 endif
 
+libraw_dep = dependency('libraw', required: get_option('raw'))
```
jcupitt: Looking at the docs:

https://www.libraw.org/docs/API-notes.html#threads

I think we probably want `libraw_r`, ie. the thread-safe version. Plain `libraw` uses a global state and you can't run two conversions at the same time in the same process.

I see:

```
$ pkg-config libraw_r --libs
-lraw_r -lstdc++ -fopenmp -llcms2 -llcms2_fast_float -llcms2_threaded -pthread -lm 
```

So I think `dependency('libraw_r', required: get_option('raw'))` would be enough.

Alternatively, we could have a single `libraw_init()` and put a lock around it, but that sounds annoying, and probably slower.

libvips/foreign/dcrawload.c
```diff
@@ -0,0 +1,343 @@
...
+
+	if (!(raw->processed =
+				libraw_dcraw_make_mem_image(raw->raw_processor, &result))) {
+		vips_error(class->nickname,
+			"error creating processed image: %s\n", libraw_strerror(result));
+		return -1;
+	}
+
+	if (!(image = vips_image_new_from_memory(
+
```
jcupitt: typo?

libvips/foreign/dcrawload.c
```diff
@@ -0,0 +1,343 @@
...
+	 */
+	libraw_data_t *raw_processor;
+
+	/* Internal Libraw processed image
+	 */
+	libraw_processed_image_t *processed;
+
+	/* Load from this source (set by subclasses).
+	 */
+	VipsSource *source;
```
jcupitt, line 75: Looking at the libraw docs, I think it only really supports file input. Many RAW formats have separate files for metadata, so for example it needs to be given a filename (perhaps `"xxx.rw2"`), then be able to open `"xxx.jpg"` to get any EXIF data.

I think right now you'll work for any RAW file from any source, but you'll miss some metadata for some formats. We could change it to only work for filename sources, but then we'd not be able to load from memory or pipes.

What do you think?

jcupitt (commented):
It looks great! I think I found a small bug in threading. and I have a question about the sort of interface we should support.

libvips/foreign/dcrawload.c
jcupitt: One solution would be to use the libraw filename interface if possible. So instead of:

```C
	if (!(data = vips_source_map(raw->source, &length)))
		return -1;
	int result = libraw_open_buffer(raw->raw_processor, data, length);
```

We could have something like:

```C
if (vips_source_is_file(raw->source)) {
    const char *filename = vips_connection_filename(VIPS_CONNECTION(raw->source));
    result = libraw_open_file(raw->raw_processor, filename);
}
else {
    if (!(data = vips_source_map(raw->source, &length)))
	return -1;
    result = libraw_open_buffer(raw->raw_processor, data, length);
}
```

Which might get us the best of both.

lxsameer: IMHO, let's restrict this patch to only formats that are just one file, to keep things simple, then we can add those more complicated suffixes later on.

jcupitt: OK, then let's merge. We can have follow-up PRs to add or fix relatively minor things.

jcupitt (approved):

jcupitt:
I think this looks great. Thank you again for doing all this work @lxsameer.

I suppose we should have some tests, but they can come in a followup PR. This is an API change, so we need a line in the changelog too (and we should credit you!), but again that can be a followup.

I've made a stable 8.17 branch for the current release, so I think we are ready to merge to master.

lxsameer:
Awesome, thank you.
Looking forward to this PR being merged and starting new PRs

jcupitt:
OK, let's merge and fix anything else later. Nice job!

jcupitt:
I bumped the version in master to 8.18 ready for the next dev cycle, and credited you in the changelog @lxsameer, I hope that's OK.

https://github.com/libvips/libvips/commit/8c5c65fddc782dc5914caf52ea4c17de61f9cb73

jcupitt:
I made https://github.com/libvips/libvips/pull/4577 for any follow-up work.

PR #4572, opened 2025-06-14
tiff: use correct log domain in threadsafe warning handlers

kleisauke:
This ensures that warnings emitted from these handles can still be suppressed via the `VIPS_WARNING=0` env variable.


lovell (approved):
This was my fault, sorry, thanks for fixing.

Issue #4568, opened 2025-06-12
Regression with v8.17: PNG to JPEG conversion with DisplayP3 profile


dloebl:
While updating from libvips v8.16 to v8.17 one of our tests found a regression for a PNG (DisplayP3) to JPEG conversion.

**To Reproduce**
```
vips copy display_p3.png 817.jpeg
``` 
[display_p3.png](https://github.com/user-attachments/assets/76fc107f-593a-44f3-ac43-18756d184581)

**v8.16.1:**
<img src="https://github.com/user-attachments/assets/dab15e03-7c42-4273-84cb-95347ffebf0a" width="200"  />

**v8.17:**
<img src="https://github.com/user-attachments/assets/1addae27-4be6-4651-9da2-a107fe6bad80" width="200"  />

kleisauke:
Bisection reports that 899ca59bda8cc45c3a8fd3f43167a1e096781380 is the culprit. Which is somewhat unexpected since that commit resolves a regression noted in #4509.

Here's the normalized `--vips-cache-trace` / `VIPS_TRACE=1` diff:
```diff
@@ -1,12 +1,12 @@
 vips cache : pngload filename="/home/kleisauke/Downloads/display_p3.png" out=((VipsImage*) 0x952c3e0) flags=((VipsForeignFlags) VIPS_FOREIGN_SEQUENTIAL) access=((VipsAccess) VIPS_ACCESS_SEQUENTIAL) -
 vips cache : copy in=((VipsImage*) 0x952c3e0) out=((VipsImage*) 0x953dd70) -
 vips cache+: colourspace in=((VipsImage*) 0x953dd70) out=((VipsImage*) 0x9541df0) space=((VipsInterpretation) VIPS_INTERPRETATION_RGB16) -
 vips cache : copy in=((VipsImage*) 0x9541df0) out=((VipsImage*) 0x9544890) -
 vips cache+: flatten in=((VipsImage*) 0x9541df0) background=0  out=((VipsImage*) 0x95438e0) -
 vips cache : copy in=((VipsImage*) 0x95438e0) out=((VipsImage*) 0x9549670) -
-vips cache+: cast in=((VipsImage*) 0x95438e0) out=((VipsImage*) 0x9548570) format=((VipsBandFormat) VIPS_FORMAT_UCHAR) -
[...]

PR #4574, opened 2025-06-16
Shift 16-bit output down to 8-bit when unsupported by saver

kleisauke:
Resolves: #4568.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,7 @@ master
 - improve function checks in meson [kleisauke]
 - tiff: use correct log domain in threadsafe warning handlers [kleisauke]
 - shift 16-bit output down to 8-bit when unsupported by saver [kleisauke]
+- text: prevent use of rgba subpixel anti-aliasing [lovell]
 
 5/6/25 8.17.0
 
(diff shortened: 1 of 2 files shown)


libvips/foreign/foreign.c
```diff
@@ -1643,7 +1643,17 @@ vips__foreign_convert_saveable(VipsImage *in, VipsImage **ready,
...
 	 */
 	if (in->Coding == VIPS_CODING_NONE) {
-		if (vips_cast(in, &out, format[in->BandFmt], NULL)) {
+		/* If the saver does not support 16-bit output, automatically
+		 * shift it down. This is the behaviour we want for saving an
+		 * RGB16 image as JPEG, for example.
+		 */
+		gboolean needs_shift =
+			!vips_band_format_is8bit(in->BandFmt) &&
+			vips_band_format_is8bit(format[in->BandFmt]);
```
kleisauke, line 1652: Note: to handle 8-bit to 16-bit promotion automatically, we could perhaps adjust this to:
```suggestion
			vips_band_format_is8bit(in->BandFmt) !=
			vips_band_format_is8bit(format[in->BandFmt]);
```

But I'm not sure if that is wanted.

Issue #4338, opened 2025-01-03
Text generated with RGBA on Ubuntu 24.10 is badly rendered


shun-recall:
## Bug report

**Describe the bug**

Using the following vips command over Ubuntu 24.10 is generating yellow text colored RGBA with bad colors/aliasing:

```sh
vips text hello_world.png "<span foreground='yellow'>Hello World</span>" --width 500 --height 100 --dpi 300 --font "Arial 48" --rgba
```

**To Reproduce**

Use Docker `ubuntu:24.10` image and libvips `8.16.0` and run the following command:

```sh
vips text hello_world.png "<span foreground='yellow'>Hello World</span>" --width 500 --height 100 --dpi 300 --font "Arial 48" --rgba
```

The following image will be generated with wrong green color around the text:

![hello_world_bad](https://github.com/user-attachments/assets/5a55d536-d80d-4b74-b669-fbcc87ec34e3)

**Expected behavior**

Doing exactly the same over `ubuntu:22.04` Docker image and libvips `8.16.0` is resulting in a correct output:

```sh
vips text hello_world.png "<span foreground='yellow'>Hello World</span>" --width 500 --height 100 --dpi 300 --font "Arial 48" --rgba
```

![hello_world](https://github.com/user-attachments/assets/a998bdc3-2ebc-4157-a84d-3ec6b0c23412)

kleisauke:
Sounds like a issue in Pango (or any of its sub-dependencies) which libvips uses for text rendering. I couldn't reproduce this in Debian 12 (Bookworm) or Fedora Rawhide.
```bash
$ docker run -e DEBIAN_FRONTEND=noninteractive -v $(pwd):/src -w /src -it debian:bookworm sh -c "\
    apt-get update -y && \
    apt-get install -y --no-install-recommends libvips-tools && \
    vips text hello_world.png \"<span foreground='yellow'>Hello World</span>\" --width 500 --height 100 --dpi 300 --font \"Arial 48\" --rgba"
```

```bash
$ docker run -v $(pwd):/src -w /src -it fedora:rawhide sh -c "\
    dnf install -y vips-tools && \
    vips text hello_world.png \"<span foreground='yellow'>Hello World</span>\" --width 500 --height 100 --dpi 300 --font \"Arial 48\" --rgba"
```

jcupitt:
I agree, this is a pango bug in ubuntu 23.10, not a libvips bug. I'll close.

lovell:
[...]

PR #4575, opened 2025-06-17
text: prevent use of rgba subpixel anti-aliasing

lovell:
It looks like newer versions of pango enable subpixel anti-aliasing by default, which can lead to rather colourful but unwanted artefacts around the edge of rendered text.

This previously came up in https://github.com/libvips/libvips/issues/4338 but was considered at the time to be a pango bug. A new issue came up in sharp at https://github.com/lovell/sharp/issues/4417 so I took a quick look and found `cairo_font_options_set_antialias` to enforce the previous default anti-aliasing approach.

Before:

After:


libvips/create/text.c
```diff
@@ -502,6 +502,15 @@ vips_text_build(VipsObject *object)
 		return -1;
 	}
 
+	if (text->rgba) {
+		/* Prevent use of subpixel anti-aliasing to avoid artefacts.
+		 */
+		cairo_font_options_t *opts = cairo_font_options_create();
+		cairo_font_options_set_antialias(opts, CAIRO_ANTIALIAS_GRAY);
+		pango_cairo_context_set_font_options(text->context, opts);
```
jcupitt: This is part of context setup, maybe it would be more logical just after:

```C
text->context = pango_font_map_create_context(vips_text_fontmap);
```

lovell: Moved, thanks. (I think I put it there as I was initially experimenting with the cairo context rather than the pango context.)

jcupitt:
Nice!

PR #4577, opened 2025-06-23
more work on dcrawload

jcupitt:
libraw has a buffer and a file interface. If you use the file interface, it's able to read secondary files and extract more metadata.

TODO

- [x] add `dcrawload_source` and `_buffer`
- [ ] tests and a sample file
- [x] support more metadata
- [x] add options for bitdepth and whitebalance control, others?
- [x] thumbnail support

### DIFF
--- a/libvips/foreign/dcrawload.c
+++ b/libvips/foreign/dcrawload.c
@@ -98,8 +98,8 @@ vips_foreign_load_dcraw_build(VipsObject *object)
 {
 	VipsForeignLoadDcRaw *raw = VIPS_FOREIGN_LOAD_DCRAW(object);
 	VipsObjectClass *class = VIPS_OBJECT_GET_CLASS(raw);
-	raw->raw_processor = libraw_init(0);
 
+	raw->raw_processor = libraw_init(0);
 	if (!raw->raw_processor) {
 		vips_error(class->nickname, _("unable to initialize libraw"));
 		return -1;
@@ -110,33 +110,24 @@ vips_foreign_load_dcraw_build(VipsObject *object)
 }
 
 static VipsForeignFlags
-vips_foreign_load_dcraw_get_flags_filename(const char *filename)
+vips_foreign_load_dcraw_get_flags(VipsForeignLoad *load)
 {
 	return 0;
 }
 
-static VipsForeignFlags
-vips_foreign_load_dcraw_get_flags(VipsForeignLoad *load)
+static void
+vips_foreign_load_dcraw_error(VipsForeignLoadDcRaw *raw,
+	const char *message, int code)
 {
-	return 0;
+	VipsObjectClass *class = VIPS_OBJECT_GET_CLASS(raw);
+
+	vips_error(class->nickname, "%s: %s", message, libraw_strerror(code));
 }
 
 static int
 vips_foreign_load_dcraw_header(VipsForeignLoad *load)
 {
 	VipsForeignLoadDcRaw *raw = (VipsForeignLoadDcRaw *) load;
-	VipsObjectClass *class = VIPS_OBJECT_GET_CLASS(raw);
-
-	const void *data;
-	size_t length;
-	VipsImage *image;
-	VipsImage *x;
-	GDateTime *dt;
-
-	/* Map the whole source into memory.
-	 */
-	if (!(data = vips_source_map(raw->source, &length)))
-		return -1;
 
 	/* Enforce VIPS_FORMAT_USHORT output.
 	 */
@@ -146,37 +137,51 @@ vips_foreign_load_dcraw_header(VipsForeignLoad *load)
 	 */
 	raw->raw_processor->params.use_camera_wb = 1;
 
-	int result = libraw_open_buffer(raw->raw_processor, data, length);
+	/* We can use the libraw file interface for filename sources. This
+	 * interface can often read more metadata, since it can open secondary
+	 * files.
+	 */
+	int result;
+	if (vips_source_is_file(raw->source)) {
+		const char *filename =
+			vips_connection_filename(VIPS_CONNECTION(raw->source));
+
+		result = libraw_open_file(raw->raw_processor, filename);
+	}
+	else {
+		size_t length;
+		const void *data;
+
+		if (!(data = vips_source_map(raw->source, &length)))
+			return -1;
+		result = libraw_open_buffer(raw->raw_processor, data, length);
+	}
 	if (result != LIBRAW_SUCCESS) {
-		vips_error(class->nickname, "%s : %s",
-			_("unable to read the source"), libraw_strerror(result));
+		vips_foreign_load_dcraw_error(raw, _("unable to read"), result);
 		return -1;
 	}
 
 	result = libraw_unpack(raw->raw_processor);
 	if (result != LIBRAW_SUCCESS) {
-		vips_error(class->nickname, "%s : %s",
-			_("unable to unpack the source"), libraw_strerror(result));
+		vips_foreign_load_dcraw_error(raw, _("unable to unpack"), result);
 		return -1;
 	}
 
-	/* Process the image (demosaicing, white balance, etc.)
+	/* Process the image (demosaicing, white balance, etc.).
 	 */
 	result = libraw_dcraw_process(raw->raw_processor);
-
 	if (result != LIBRAW_SUCCESS) {
-		vips_error(class->nickname,
-			"error processing RAW data: %s\n", libraw_strerror(result));
+		vips_foreign_load_dcraw_error(raw, _("unable to process"), result);
 		return -1;
 	}
 
 	if (!(raw->processed =
-				libraw_dcraw_make_mem_image(raw->raw_processor, &result))) {
-		vips_error(class->nickname,
-			"error creating processed image: %s\n", libraw_strerror(result));
+			libraw_dcraw_make_mem_image(raw->raw_processor, &result))) {
+		vips_foreign_load_dcraw_error(raw, _("unable to build image"), result);
 		return -1;
 	}
 
+	VipsImage *image;
 	if (!(image = vips_image_new_from_memory(
 			  raw->processed->data, raw->processed->data_size,
 			  raw->processed->width, raw->processed->height,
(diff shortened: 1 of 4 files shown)


mertalev:
Ooh, does this mean it can read edits and metadata from XMP sidecar files? Or is this a different feature?

jcupitt:
Yes, XMP sidecar support should be possible I think, though it won't work right now.

jcupitt:
... would you have a test file with an XMP sidecar? It'd be useful for dev.

mertalev:
Awesome! I don't have one at the moment, but I asked the Immich Discord server for samples 🤞

jcupitt:
It seems to support XMP and ICC profiles now. I see:

```
$ time vipsheader -a sample1.cr2
sample1.cr2: 8736x5856 uchar, 3 bands, srgb, dcrawload
width: 8736
height: 5856
bands: 3
format: uchar
coding: none
interpretation: srgb
xoffset: 0
yoffset: 0
xres: 1
yres: 1
filename: sample1.cr2
vips-loader: dcrawload
raw-make: Canon
raw-model: EOS 5DS
raw-software: 
raw-iso: 125
raw-shutter: 0.0166667
raw-aperture: 7.1
raw-focal-length: 60
raw-timestamp: 2015-07-23T15:52:30Z
xmp-data: 8193 bytes of binary data
raw-lens: EF24-105mm f/4L IS USM

real	0m0.041s
user	0m0.023s
sys	0m0.018s
```

It's nice and quick too, since header read is split from image reconstruction.

I made it default to 8 bit output, you use `--bitdepth=16` to get 16 bit output.

There's a lot more metadata we could support, but I don't know how useful they are. Maybe they should be added as needed.

I added load from a memory buffer and load from a generic source, so you can now do:

```
cat sample1.cr2 | vips dcrawload_source "[descriptor=0]" x.v
```

Though I'm not sure how useful that is.

We should add thumbnail load too, perhaps:

```
vips dcrawload sample1.cr2[thumbnail] x.jpg
```

jcupitt:
It's a little quicker than the old imagemagick path, since it defaults to 8 bit read now. On this PC I see:

```
$ time vips copy sample1.cr2 x.jpg

real	0m2.659s
user	0m5.834s
sys	0m1.511s
```

With the imagemagick loader it's:

```
$ time vips magickload sample1.cr2 x.jpg

real	0m3.035s
user	0m7.518s
sys	0m2.172s
```

That's to make a 8736 x 5856 pixel jpeg from a 70mb .cr2 raw:

image

Memory use is down a bit too:

```
john@banana ~/pics $ /usr/bin/time -f %M:%e vips magickload sample1.cr2 x.jpg
1140616:3.00
john@banana ~/pics $ /usr/bin/time -f %M:%e vips copy sample1.cr2 x.jpg
734736:2.64
```

Down from 1.1gb to 730mb. The images look identical.

mertalev:
Wow, that's a big difference! Really excited to start using this loader.

If it helps, a user shared [these](https://drive.google.com/drive/folders/1TVqy7RkQnvjajTgPgXaYjVXbt0MYrEkN?usp=sharing) files with a .pp3 file apparently used by RawTherapee.

I'm also curious how the API would work for these secondary files. Would this be exposed as a file path in sharp etc., or would it be handled internally?

jcupitt:
> I'm also curious how the API would work for these secondary files.

libraw does all that -- if you open `/some/dir/sample1.cr2` it'll look for extra metadata in `/some/dir/sample1.xml` (just example extensions, I don't think cr2 uses side files).

jcupitt:
I added thumbnail support too. It looks like this:

```
$ vipsheader -a sample1.cr2
sample1.cr2: 8736x5856 uchar, 3 bands, srgb, dcrawload
width: 8736
height: 5856
bands: 3
format: uchar
coding: none
interpretation: srgb
xoffset: 0
yoffset: 0
xres: 1
yres: 1
filename: sample1.cr2
vips-loader: dcrawload
raw-make: Canon
raw-model: EOS 5DS
raw-software: 
raw-iso: 125
raw-shutter: 0.0166667
raw-aperture: 7.1
raw-focal-length: 60
raw-timestamp: 2015-07-23T15:52:30Z
xmp-data: 8193 bytes of binary data
raw-lens: EF24-105mm f/4L IS USM
raw-thumbnail-data: 17472 bytes of binary data
```

So there's a metadata item containing a JPEG thumbnail.

You can read it out like this:

```
$ vipsheader -f raw-thumbnail-data sample1.cr2 | base64 -d > x.jpg
```

To make:

x

It also seems to work well for your users sample images.

`vipsthumbnail` doesn't yet know about using libraw thumbnails, so thumbnailing raw images is still slow.

```
$ time vipsthumbnail sample1.cr2
memory: high-water mark 46.14 MB

real	0m3.944s
user	0m6.967s
sys	0m1.343s
```

We should fix that in a follow-up PR.

kleisauke:
TIL that `dcraw` is named after its author, Dave Coffin. It doesn't stand for "digital camera RAW".

libvips/foreign/dcrawload.c
```diff
@@ -310,6 +517,62 @@ vips_foreign_load_dcraw_file_init(VipsForeignLoadDcRawFile *file)
...
+vips_foreign_load_dcraw_buffer_class_init(
+	VipsForeignLoadDcRawBufferClass *class)
+{
+	GObjectClass *gobject_class = G_OBJECT_CLASS(class);
+	VipsObjectClass *object_class = (VipsObjectClass *) class;
+
+	gobject_class->set_property = vips_object_set_property;
+	gobject_class->get_property = vips_object_get_property;
+
+	object_class->nickname = "rawload_buffer";
```
kleisauke:
```suggestion
	object_class->nickname = "dcrawload_buffer";
```

libvips/include/vips/foreign.h
```diff
@@ -878,6 +878,12 @@ int vips_gifsave_target(VipsImage *in, VipsTarget *target, ...)
 VIPS_API
 int vips_dcrawload(const char *filename, VipsImage **out, ...)
 	G_GNUC_NULL_TERMINATED;
+VIPS_API
+int vips_dcrawload_source(VipsSource *source, VipsImage **out, ...)
+	G_GNUC_NULL_TERMINATED;
+VIPS_API
+int vips_dcrawload_buffer(void *buf, size_t len, VipsImage **out, ...)
+	G_GNUC_NULL_TERMINATED;
```
kleisauke: nit: this is usually sorted.
```suggestion
VIPS_API
int vips_dcrawload_buffer(void *buf, size_t len, VipsImage **out, ...)
	G_GNUC_NULL_TERMINATED;
VIPS_API
int vips_dcrawload_source(VipsSource *source, VipsImage **out, ...)
	G_GNUC_NULL_TERMINATED;
```

libvips/foreign/dcrawload.c
```diff
@@ -227,25 +356,98 @@ vips_foreign_load_dcraw_class_init(VipsForeignLoadDcRawClass *class)
...
+	GObjectClass *gobject_class = G_OBJECT_CLASS(class);
+	VipsObjectClass *object_class = (VipsObjectClass *) class;
+	VipsOperationClass *operation_class = VIPS_OPERATION_CLASS(class);
+
+	gobject_class->set_property = vips_object_set_property;
+	gobject_class->get_property = vips_object_get_property;
+
+	object_class->nickname = "dcrawload_source";
+	object_class->build = vips_foreign_load_dcraw_source_build;
+
```
kleisauke, line 445: Perhaps we should also a `->is_a_source` that always returns `FALSE`? Similar to `csvload_source`:
https://github.com/libvips/libvips/blob/8c5c65fddc782dc5914caf52ea4c17de61f9cb73/libvips/foreign/csvload.c#L625-L633

Otherwise, `vips_foreign_find_load_source()` would trigger this warning:
https://github.com/libvips/libvips/blob/8c5c65fddc782dc5914caf52ea4c17de61f9cb73/libvips/foreign/foreign.c#L751-L752

libvips/foreign/dcrawload.c
```diff
@@ -310,6 +517,62 @@ vips_foreign_load_dcraw_file_init(VipsForeignLoadDcRawFile *file)
...
+{
+	GObjectClass *gobject_class = G_OBJECT_CLASS(class);
+	VipsObjectClass *object_class = (VipsObjectClass *) class;
+
+	gobject_class->set_property = vips_object_set_property;
+	gobject_class->get_property = vips_object_get_property;
+
+	object_class->nickname = "rawload_buffer";
+	object_class->build = vips_foreign_load_dcraw_buffer_build;
+
```
kleisauke, line 578: Ditto, perhaps we should a `->is_a_buffer` here that always returns `FALSE`. For example:
```c
/* Compile with:
 */

#include <vips/vips.h>

int
main(int argc, char *argv[])
{
	if (VIPS_INIT(argv[0]))
		vips_error_exit(NULL);

	VipsSource *source;
		return 0;

	vips_foreign_find_load_buffer(NULL, 0);

	return 0;
}
```

```console
$ ./a.out 

(process:33687): VIPS-WARNING **: 12:39:48.840: loader dcrawload_source has no is_a_source method

(process:33687): VIPS-WARNING **: 12:39:48.841: loader dcrawload_buffer has no is_a_buffer method
```

jcupitt:
Yes, he's an infamous bit-twiddling hacker. `dcraw` is a great program used by everyone, but he's uninterested in making or maintaining a library, he just puts out ever more comprehensive versions of a windows-only command-line program for conversion.

There have been many attempts to make a portable library from it with a stable API, and libraw is the best so far. They have to keep revising themselves whenever he changes `dcraw` again. But who knows how long it'll be updated for.

libvips/foreign/dcrawload.c
jcupitt: Oh, good point!

jcupitt:
We can't add `is_a` tests for sources, since `libraw_open()` needs to have the whole object, it can't sniff from the first few bytes.

We possibly could for file and buffer since `libraw_open()` is reasonably quick (a few ms). However, libraw has to come before JPEG and TIFF, since those formats are commonly used as containers for cameras raw files, and that means we'd be adding a few ms to all file detection.

I think only detecting from the filename suffix is probably the best compromise.

jcupitt:
Thanks for the review!

jcupitt:
@mertalev any testing would be very helpful! This will be in libvips 8.18, due in the autumn.

kleisauke:
I think Immich uses sharp, so it likely requires a change similar to commit https://github.com/lovell/sharp/commit/67ff930535a04378689d3f3940a90ab5d2965794 (but only for the file loader).

mertalev:
Very cool! I'll give it a spin when I get a chance

lovell:
I've added (future-proof) support for this feature to sharp via commit https://github.com/lovell/sharp/commit/df5454e7dc582256fb0c29a7d080054d0efedc99

Issue #4576, opened 2025-06-18
IIIF Issues with 8.17


cmahnke:
The release notes for 8.17 don't mention #4421  so it took me a bit longer to narrow it down ;)

## Bug report

The referenced change adds a new `scaleFactor` within the tiles array of a IIIF `info.json` file. This can confuse some IIIF Viewers. I've tested against OpenLayers and OpenSeadragon.
While it's not really noticeble using OpenLayers, since it transparently falls back to the next possible tile set and only logs the problem to the console, OpenLayers isn't as forgiving by default...

You can test OpenLayers here:
https://openlayers.org/en/latest/examples/iiif.html

For the next few days one can use this example. But it will be replaced with a file regenerated by Vips 8.16.1 very soon:
https://static.projektemacher.org/backsteinexpressionismus/post/messegelaende-koeln-3/front/info.json

When opening with the OpenLayers demonstrator you can see a request to 
https://static.projektemacher.org/backsteinexpressionismus/post/messegelaende-koeln-3/front/full/439,/0/default.jpg
Which hasn't been generated by VIPS. For reference, this is the generated `info.json`:

```
{
  "@context": "http://iiif.io/api/image/2/context.json",
  "@id": "https://static.projektemacher.org/backsteinexpressionismus/post/messegelaende-koeln-3/front",
  "profile": [
    "http://iiif.io/api/image/2/level0.json",
    {
      "formats": [
        "jpg"
      ],
      "qualities": [
        "default"
      ]
    }
  ],
  "protocol": "http://iiif.io/api/image",
  "tiles": [
    {
      "scaleFactors": [
        1,
        2,
        4,
        8
      ],
      "width": 512
    }
  ],
  "width": 3509,
  "height": 2506
}
```
The `scaleFactor` of `8` causes the viewer to  fetch an non-exiting image.

lovell:
Could this be a problem with OpenLayers or its configuration expecting to find a (missing) `full` region directory instead of the pixel-based `0,0,3509,2506` directory?

The image https://static.projektemacher.org/backsteinexpressionismus/post/messegelaende-koeln-3/front/0,0,3509,2506/439,/0/default.jpg does appear to exist.

There was a discussion related to this a few years ago at https://github.com/libvips/libvips/discussions/2036 (it looks like you commented at the time).

cmahnke:
Thanks, for having a look, actually I did't do that much research on my own yet - I completely forgot about #2036...

[...]

PR #4578, opened 2025-06-23
dzsave: IIIF: use named region of 'full' when no crop takes place

lovell:
Will improve compliance with IIIF v2 and v3, the specs of which both mandate support for the "full" region.

- https://iiif.io/api/image/2.0/compliance/#region
- https://iiif.io/api/image/3.0/compliance/#31-region

I've checked OpenLayers and OpenSeadragon and they both contain the same logic so it should address most of https://github.com/libvips/libvips/discussions/2036

- https://github.com/openlayers/openlayers/blob/7cf17b80aeefe7e6de867703f4239b362b19496b/src/ol/source/IIIF.js#L263-L270
- https://github.com/openseadragon/openseadragon/blob/a02c2456d972dd23f9c8d106ffbd2b29efcf56d1/src/iiiftilesource.js#L509-L513

This PR only handles the "region" part of the URL and doesn't touch the "size" as (I think) as long as the tile size is smaller than the original image size in either (or both) dimensions then a "max" tile will not be requested. If the tile size is larger than the original image in both dimensions then there's only one tile/layer, so you don't really need dzsave.

I've also added a couple of initial, basic IIIF layout tests as there were none. I used the sample dimensions from #4576 for these as it should be fixed too.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -9,6 +9,7 @@ master
 - text: prevent use of rgba subpixel anti-aliasing [lovell]
 - tiffsave: always apply resolution unit conversion [kleisauke]
 - cache: suppress invalidation errors in release builds [kleisauke]
+- dzsave: IIIF: use named region of 'full' when no crop takes place [lovell]
 
 5/6/25 8.17.0
 
--- a/libvips/foreign/dzsave.c
+++ b/libvips/foreign/dzsave.c
@@ -1212,6 +1212,9 @@ tile_name(Level *level, int x, int y)
 			save->ready->Xsize - left);
 		int height = VIPS_MIN(dz->tile_size * level->sub,
 			save->ready->Ysize - top);
+		gboolean is_region_full = left == 0 && top == 0 &&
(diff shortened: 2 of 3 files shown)


jcupitt:
Hi @lovell, I'm so confused :( I thought `full` meant the full size image, not the first shrink to fit inside a single tile. 

If it just means a one-tile version of the image (a tile that contains the `full` image I guess?) then I agree, this is a very simple fix. Phew!

lovell:
Yes, I think the confusion stems from IIIFv2 using the concept of "full" to represent both "region" and "size" but with different meanings. IIIFv3 fixes the _naming things is hard_ problem for "size" by replacing "full" with "max".

The discussion about this will be celebrating its 10th birthday later this year - see https://github.com/IIIF/api/issues/620

jcupitt:
> Yes, I think the confusion stems from IIIFv2 using the concept of "full" to represent both "region" and "size" but with different meanings.

*Yare yare*, as jojo says :roll_eyes:

PR #4590, opened 2025-06-27
semaphore: simplify `GCond` initialization

kleisauke:
This member is marked as `/*< private >*/` and is never allocated statically or on the stack, something the API disallows anyway. Therefore, we can simply use `GCond` directly here.

https://kleisauke.nl/compat_reports/vips/master_to_semaphore-simplify-gcond-init/compat_report.html

Context: https://github.com/libvips/libvips/pull/4249#issuecomment-2740143439.


jcupitt:
:+1:

Issue #4530, opened 2025-05-22
Add magickload_source() operation


HFO4:
## Bug report

Trying to generate thumbnails for psd/html file that are supported via ImageMagick. 

* If source is loaded from local file via `vipsthumbnail`, everything works fine;
* If source is piped via stdin or loaded from file via `thumbnail_source`, it will fail with `VipsForeignLoad: source is not in a known format`

[test.psd.zip](https://github.com/user-attachments/files/20381634/test.psd.zip)

**To Reproduce**

1. unzip [test.psd.zip](https://github.com/user-attachments/files/20381634/test.psd.zip), you will get `test.psd`;
2. execute `vipsthumbnail test.psd`, thumbnail can be generated;
3. execute `cat test.psd | vips thumbnail_source '[descriptor=0]'  .png 100`, got error `VipsForeignLoad: source is not in a known format`;
4. execute `vips thumbnail_source '[filename=test.psd]'  .png 100`, got error `VipsForeignLoad: source is not in a known format`;

**Expected behavior**
Thumbnail can be generated no matter how to specify the source.

**Actual behavior**
Format cannot be detected if loading ImageMagick supported source from `thumbnail_source`.

**Screenshots**

**Environment**
macOS 15.5

**Additional context**

jcupitt:
Hello @HFO4,

The `magickload` operation in libvips only works for files and memory buffers, unfortunately. Someone needs to implement load from source.

Let's tag this as an enhancement.

jcupitt:
We have a PR, I'll close.

PR #4592, opened 2025-06-28
add magickload_source

jcupitt:
test with eg.:

```
$ cat MARBLES.BMP | vips magickload_source "[descriptor=0]" x.v
$ cat MARBLES.BMP | vipsthumbnail stdin -o x.jpg
```

ImageMagick doesn't support generic IO, so it just maps the source into memory and uses the buffer interface.

magick6 and magick7.

See https://github.com/libvips/libvips/issues/4530

### DIFF
--- a/libvips/foreign/magick6load.c
+++ b/libvips/foreign/magick6load.c
@@ -1120,15 +1120,15 @@ vips_foreign_load_magick_source_header(VipsForeignLoad *load)
 		const char *filename =
 			vips_connection_filename(VIPS_CONNECTION(magick_source->source));
 
-		g_strlcpy(magick7->image_info->filename, filename, MagickPathExtent);
-		magick_sniff_file(magick7->image_info, filename);
-		magick7->image = ReadImage(magick7->image_info, magick7->exception);
+		g_strlcpy(magick->image_info->filename, filename, MagickPathExtent);
+		magick_sniff_file(magick->image_info, filename);
+		magick->image = ReadImage(magick->image_info, magick->exception);
 	}
 	else {
 		size_t length;
 		const void *data;
 
-		if (!(data = vips_source_map(raw->source, &length)))
+		if (!(data = vips_source_map(magick_source->source, &length)))
 			return -1;
 		magick_sniff_bytes(magick->image_info, data, length);
 		magick->image = BlobToImage(magick->image_info,
--- a/libvips/foreign/magick6load.c
+++ b/libvips/foreign/magick6load.c
@@ -1074,6 +1074,102 @@ vips_foreign_load_magick_buffer_init(VipsForeignLoadMagickBuffer *buffer)
 {
 }
 
+typedef struct _VipsForeignLoadMagickSource {
+	VipsForeignLoadMagick parent_object;
+
+	VipsSource *source;
+
+} VipsForeignLoadMagickSource;
+
+typedef VipsForeignLoadMagickClass VipsForeignLoadMagickSourceClass;
+
+G_DEFINE_TYPE(VipsForeignLoadMagickSource, vips_foreign_load_magick_source,
+	vips_foreign_load_magick_get_type());
+
+static gboolean
+vips_foreign_load_magick_source_is_a_source(VipsSource *source)
+{
+	const unsigned char *p;
+
+	// just use the first 100 byes, we don't want to force too much into
+	// memory
+	return (p = vips_source_sniff(source, 100)) &&
+		magick_ismagick(p, 100);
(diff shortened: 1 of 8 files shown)


libvips/foreign/magick6load.c
```diff
@@ -1074,6 +1074,102 @@ vips_foreign_load_magick_buffer_init(VipsForeignLoadMagickBuffer *buffer)
...
+
+G_DEFINE_TYPE(VipsForeignLoadMagickSource, vips_foreign_load_magick_source,
+	vips_foreign_load_magick_get_type());
+
+static gboolean
+vips_foreign_load_magick_source_is_a_source(VipsSource *source)
+{
+	const unsigned char *p;
+
+	// just use the first 100 byes, we don't want to force too much into
```
kleisauke:
```suggestion
	// just use the first 100 bytes, we don't want to force too much into
```

libvips/foreign/magick7load.c
```diff
@@ -939,6 +939,100 @@ vips_foreign_load_magick7_buffer_init(VipsForeignLoadMagick7Buffer *buffer)
...
+
+G_DEFINE_TYPE(VipsForeignLoadMagick7Source, vips_foreign_load_magick7_source,
+	vips_foreign_load_magick7_get_type());
+
+static gboolean
+vips_foreign_load_magick7_source_is_a_source(VipsSource *source)
+{
+	const unsigned char *p;
+
+	// just use the first 100 byes, we don't want to force too much into
```
kleisauke:
```suggestion
	// just use the first 100 bytes, we don't want to force too much into
```

kleisauke:
It looks like IM7 provides `SetCustomStreamReader()` and related functions. However, based on my understanding of `CustomStreamToImage()`, it simply maps the stream into memory or to a temporary file on disk (if the coder lacks the `CoderBlobSupportFlag` flag).

jcupitt:
Haha so just the same as us then :)

At some point in the future I suppose we can update this thing to use a proper stream interface, if they add one.

jcupitt:
> CoderBlobSupportFlag

I guess we could also check this and fall back to the filename interface.

kleisauke (commented):
Great! I assume the C declaration (i.e. `vips_magickload_source()`) was intentionally omitted?

jcupitt:
> intentionally omitted?

Ahem, I added the decl, good catch.

I made it use the filename interface if the source has an associated filename, it'll broaden compatibility a little.

kleisauke:
> Ahem, I added the decl, good catch.

Ah, I think we also need to add the C implementation of this to `magickload.c`, since `magick{6,7}load.c` only contains the GObject-specific parts.

> I made it use the filename interface if the source has an associated filename, it'll broaden compatibility a little.

Great!

libvips/foreign/magick6load.c
```diff
@@ -1074,6 +1074,113 @@ vips_foreign_load_magick_buffer_init(VipsForeignLoadMagickBuffer *buffer)
...
+
+#ifdef DEBUG
+	printf("vips_foreign_load_magick_source_header: %p\n", load);
+#endif /*DEBUG*/
+
+	if (vips_source_is_file(magick_source->source)) {
+		const char *filename =
+			vips_connection_filename(VIPS_CONNECTION(magick_source->source));
+
+		g_strlcpy(magick7->image_info->filename, filename, MagickPathExtent);
```
kleisauke:
```suggestion
		g_strlcpy(magick->image_info->filename, filename, MagickPathExtent);
```

libvips/foreign/magick6load.c
```diff
@@ -1074,6 +1074,113 @@ vips_foreign_load_magick_buffer_init(VipsForeignLoadMagickBuffer *buffer)
...
+		magick_sniff_file(magick7->image_info, filename);
+		magick7->image = ReadImage(magick7->image_info, magick7->exception);
+	}
+	else {
+		size_t length;
+		const void *data;
+
+		if (!(data = vips_source_map(raw->source, &length)))
```
kleisauke:
```suggestion
		if (!(data = vips_source_map(magick_source->source, &length)))
```

kleisauke (commented):
These suggestions were caught by CodeQL:
https://github.com/libvips/libvips/actions/runs/15945176481/job/44978306353?pr=4592

jcupitt:
I'm rushing and missing stuff :( I've now actually tested against im6 and it seems fine, phew.

jcupitt:
Great! Thank you for the help!

PR #4594, opened 2025-06-29
Simplify using `g_type_class_peek()`

kleisauke:
Only possible for classes that have already been referenced before.

### DIFF
--- a/libvips/iofuncs/operation.c
+++ b/libvips/iofuncs/operation.c
@@ -292,14 +292,9 @@ vips_operation_class_usage_classify(VipsArgumentClass *argument_class)
 /* Display a set of flags as "a:b:c"
  */
 static void
-vips__flags_to_str(VipsBuf *buf, GType type, guint value)
+vips__flags_to_str(VipsBuf *buf, GFlagsClass *flags, guint value)
 {
-	GTypeClass *class = g_type_class_ref(type);
-	GFlagsClass *flags = G_FLAGS_CLASS(class);
-
-	gboolean first;
-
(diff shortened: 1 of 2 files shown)


libvips/iofuncs/operation.c
```diff
@@ -321,19 +316,18 @@ vips_operation_pspec_usage(VipsBuf *buf, GParamSpec *pspec)
...
-		GTypeClass *class = g_type_class_ref(type);
+		/* GParamSpecEnum holds a ref on the class so we just peek.
+		 */
+		GEnumClass *genum = g_type_class_peek(type);
 		GParamSpecEnum *pspec_enum = (GParamSpecEnum *) pspec;
 
-		GEnumClass *genum;
 		int i;
 
 		/* Should be impossible, no need to warn.
```
lovell: Should we also update these "impossible" comments as `g_type_class_peek` can return null?

kleisauke: The `GParamSpec` holds a ref to the class, so `g_type_class_peek()` would never return `NULL` in this case.

I've updated these to use `g_assert()`, which is probably more appropriate in this case.

kleisauke:
For reference, this PR was motivated by a similar simplification made in GLib via commit https://gitlab.gnome.org/GNOME/glib/-/commit/ab1111b8bbb7a6057248658013eb23d80f4721cf.

jcupitt:
In practice, glib can never unload classes, so the extra refs don't matter. But I agree that peek looks neater.

Issue #4595, opened 2025-07-02
smartcrop: arg "width" changed during build error with InterestingAll + SizeBoth combination


isuh88:
### Description
Environment:
libvips 8.17.0
govips (Go binding)
Production server environment

### Issue:
When using `vips.LoadThumbnailFromBuffer()` with `vips.InterestingAll` and `vips.SizeBoth` parameters, the following error occurs consistently in production:
```text
smartcrop: arg "width" changed during build, VipsSmartcrop (smartcrop), extract an area from an image, smartcrop input=((VipsImage*) 0x1a6794c0) attention-x=0 out=((VipsImage*) 0x1a1d6e10) attention-y=0 width=600 height=400 interesting=((VipsInteresting) VIPS_INTERESTING_ALL)
```

### Observations:
Only occurs with InterestingAll + SizeBoth combination
Other all combinations work fine
Cannot reproduce in local test environment
Occurs consistently in production server environment

### Question:
What specific aspect of the smartcrop implementation should we investigate when this "arg width(or height) changed during build" error occurs? Is this related to memory management, thread safety, or parameter validation in the smartcrop operation?

isuh88:
In addition) I found that the error does not occur when the input width and height parameters exactly match the aspect ratio of the original image.

When the input width and height does not match the original image aspect & generate thumbnail with VIPS_INTERESTING_ALL`, (not with `SIZE_FORCE`) the error occurs.

lovell:
If you'll excuse the pun, what was it that interested you in the use of `VIPS_INTERESTING_ALL`? It's probably the least useful value, with `VIPS_INTERESTING_ENTROPY` and `VIPS_INTERESTING_ATTENTION` usually producing better results.

The reason for the error is that as of 8.17.0, the use of `VIPS_INTERESTING_ALL` with libvips operation cache will generate this error due to:

https://github.com/libvips/libvips/blob/a5a0d0a857b63852d177d19e4bf54963e10557d1/libvips/conversion/smartcrop.c#L389-L390

You might be able to workaround this by switching off the cache but I'm unsure of the govips API for this.

kleisauke:
These errors are only issued when libvips' internal leak checker is enabled, either when the `VIPS_LEAK` env variable is set or when `vips_leak_set(TRUE)` is explicitly called. In govips, you can likely disable this by setting the `ReportLeaks` configuration option to `false`.
[...]

PR #4596, opened 2025-07-02
cache: suppress invalidation errors in release builds

kleisauke:
These diagnostics are only useful during development and debugging.

See: #4595.
Targets the 8.17 branch.


jcupitt:
I agree, this should not be on in production.

PR #4599, opened 2025-07-04
Fix pdfiumload.c vips_foreign_load_pdf_generate()

zhifengzhuang:
- Fix pdfiumload.c vips_foreign_load_pdf_generate(). Inconsistent PDF page dimensions (varying width/height) may generate empty Bitmaps, leading to crashes in PDFium versions prior to 6633.

- Draw a background color on outRegion to maintain consistency with Poppler's rendering behavior.

### DIFF
--- a/libvips/foreign/pdfiumload.c
+++ b/libvips/foreign/pdfiumload.c
@@ -582,6 +582,11 @@ vips_foreign_load_pdf_generate(VipsRegion *out_region,
 	 */
 	vips_region_black(out_region);
 
+    /*
+     * Same with popperload
+     */
+    vips_region_paint_pel(out_region, r, pdf->ink);
+
 	top = r->top;
 	while (top < VIPS_RECT_BOTTOM(r)) {
 		VipsRect rect;
@@ -592,39 +597,43 @@ vips_foreign_load_pdf_generate(VipsRegion *out_region,
 		if (vips_foreign_load_pdf_get_page(pdf, pdf->page_no + i))
 			return -1;
 
-		vips__worker_lock(&vips_pdfium_mutex);
+        if (rect.width > 0 && rect.height > 0) {
+            // Ensure rect not empty
 
-		/* 4 means RGBA.
-		 */
-		bitmap = FPDFBitmap_CreateEx(rect.width, rect.height, 4,
-			VIPS_REGION_ADDR(out_region, rect.left, rect.top),
-			VIPS_REGION_LSKIP(out_region));
+            vips__worker_lock(&vips_pdfium_mutex);
 
-		/* Only paint the background if there's no transparency.
-		 */
-		if (!FPDFPage_HasTransparency(pdf->page)) {
-			FPDF_DWORD ink = *((guint32 *) pdf->ink);
+            /* 4 means RGBA.
+             */
+            bitmap = FPDFBitmap_CreateEx(rect.width, rect.height, 4,
+                VIPS_REGION_ADDR(out_region, rect.left, rect.top),
+                VIPS_REGION_LSKIP(out_region));
 
-			FPDFBitmap_FillRect(bitmap,
-				0, 0, rect.width, rect.height, ink);
-		}
+            /* Only paint the background if there's no transparency.
+             */
+            if (!FPDFPage_HasTransparency(pdf->page)) {
+                FPDF_DWORD ink = *((guint32 *) pdf->ink);
 
-		// pdfium writes bgra by default, we need rgba
-		FPDF_RenderPageBitmap(bitmap, pdf->page,
-			pdf->pages[i].left - rect.left,
-			pdf->pages[i].top - rect.top,
-			pdf->pages[i].width, pdf->pages[i].height,
-			0, FPDF_ANNOT | FPDF_REVERSE_BYTE_ORDER);
+                FPDFBitmap_FillRect(bitmap,
+                    0, 0, rect.width, rect.height, ink);
+            }
 
-		FPDF_FFLDraw(pdf->form, bitmap, pdf->page,
-			pdf->pages[i].left - rect.left,
-			pdf->pages[i].top - rect.top,
-			pdf->pages[i].width, pdf->pages[i].height,
-			0, FPDF_ANNOT | FPDF_REVERSE_BYTE_ORDER);
+            // pdfium writes bgra by default, we need rgba
+            FPDF_RenderPageBitmap(bitmap, pdf->page,
+                pdf->pages[i].left - rect.left,
+                pdf->pages[i].top - rect.top,
+                pdf->pages[i].width, pdf->pages[i].height,
+                0, FPDF_ANNOT | FPDF_REVERSE_BYTE_ORDER);
 
-		FPDFBitmap_Destroy(bitmap);
+            FPDF_FFLDraw(pdf->form, bitmap, pdf->page,
+                pdf->pages[i].left - rect.left,
+                pdf->pages[i].top - rect.top,
+                pdf->pages[i].width, pdf->pages[i].height,
+                0, FPDF_ANNOT | FPDF_REVERSE_BYTE_ORDER);
 
-		g_mutex_unlock(&vips_pdfium_mutex);
+            FPDFBitmap_Destroy(bitmap);
+
+            g_mutex_unlock(&vips_pdfium_mutex);
+        }
 
 		top += rect.height;
 		i += 1;
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,6 +3,7 @@ master
 - add dcrawload, dcrawload_source, dcrawload_buffer: load raw camera files
   using libraw [lxsameer]
 - add magickload_source: load from a source with imagemagick
+- fix pdfiumload.c vips_foreign_load_pdf_generate(): crash risk when pdfium < 6633 [daniel]
 
 8.17.1
 


kleisauke:
> * Fix pdfiumload.c vips_foreign_load_pdf_generate(). Inconsistent PDF page dimensions (varying width/height) may generate empty Bitmaps, leading to crashes in PDFium versions prior to 6633.

I wonder if commit 2fdab9ba70f5ebf22588943e951339ceb19b184e is also related here.

> * Draw a background color on outRegion to maintain consistency with Poppler's rendering behavior.

This effectively reverts commit 985fb1d1d0b2d74a769bc0e191771db017a22ebd, I'm not sure if that's wanted.

zhifengzhuang:
Commit https://github.com/libvips/libvips/commit/2fdab9ba70f5ebf22588943e951339ceb19b184e seems not to related here. Older pdfium(<6633) will crash when an empty bitmap pass to FPDF_FFLDraw. We should check the rect width&height before we call FPDF_FFLDraw?
Commit https://github.com/libvips/libvips/commit/985fb1d1d0b2d74a769bc0e191771db017a22ebd seems to related here. When there is no overlapping area between the VipsRegion and the PDF, actively draw the background color to ensure the PDFium output matches that of Poppler.

jcupitt:
Hi @zhifengzhuang, thank you for this PR!

> I wonder if commit https://github.com/libvips/libvips/commit/2fdab9ba70f5ebf22588943e951339ceb19b184e is also related here.

I think that's unrelated -- this PR fixes the case where you have a narrow page and the tilecache asks for a rect outside the page boundary. 

poppler deals with this gracefully, but (as @zhifengzhuang says) older pdfiums will crash. 

> This effectively reverts commit https://github.com/libvips/libvips/commit/985fb1d1d0b2d74a769bc0e191771db017a22ebd, I'm not sure if that's wanted.

I agree, I'm not sure about the change to background. I think this is a case where PDFium has better transparency support than poppler, so we don't want to match the behaviour exactly.

libvips/foreign/pdfiumload.c
```diff
@@ -582,6 +582,11 @@ vips_foreign_load_pdf_generate(VipsRegion *out_region,
 	 */
 	vips_region_black(out_region);
 
+    /*
+     * Same with popperload
+     */
+    vips_region_paint_pel(out_region, r, pdf->ink);
```
jcupitt: I don't think we should do this. We set the background when required with `FPDFBitmap_FillRect()` below.

libvips/foreign/pdfiumload.c
```diff
@@ -592,39 +597,43 @@ vips_foreign_load_pdf_generate(VipsRegion *out_region,
 		if (vips_foreign_load_pdf_get_page(pdf, pdf->page_no + i))
```
jcupitt: This could move inside the rect size test `if()`, eg.:

```C
    top = r->top;
    while (top < VIPS_RECT_BOTTOM(r)) {
        FPDF_BITMAP bitmap; 

        /* Is the rect within this page? It might not be if the output is more
         * than one tile wide and this page is narrower.
         */
        VipsRect rect;
        vips_rect_intersectrect(r, &pdf->pages[i], &rect);
        if (rect.width > 0 && 
            rect.height > 0) {

            if (vips_foreign_load_pdf_get_page(pdf, pdf->page_no + i))
                return -1;

            vips__worker_lock(&vips_pdfium_mutex);

            /* 4 means RGBA.
             */ 
            bitmap = FPDFBitmap_CreateEx(rect.width, rect.height, 4,
                VIPS_REGION_ADDR(out_region, rect.left, rect.top),
                VIPS_REGION_LSKIP(out_region));
... etc
```

kleisauke:
I can confirm that this issue reproduces with [PDFium 6611](https://github.com/bblanchon/pdfium-binaries/releases/tag/chromium%2F6611) but does not occur with [PDFium 6666](https://github.com/bblanchon/pdfium-binaries/releases/tag/chromium%2F6666). Tested using:

<details>
  <summary>Details</summary>

```bash
# export PDFIUM_VERSION=6666
export PDFIUM_VERSION=6611
mkdir -p $HOME/pdfium-linux-x64/lib/pkgconfig
curl -Ls https://github.com/bblanchon/pdfium-binaries/releases/download/chromium%2F$PDFIUM_VERSION/pdfium-linux-x64.tgz | tar xzC $HOME/pdfium-linux-x64
cat > $HOME/pdfium-linux-x64/lib/pkgconfig/pdfium.pc << EOF
prefix=$HOME/pdfium-linux-x64
exec_prefix=\${prefix}
libdir=\${exec_prefix}/lib
includedir=\${prefix}/include

Name: pdfium
Description: PDFium
Version: $PDFIUM_VERSION
Libs: -L\${libdir} -lpdfium
Cflags: -I\${includedir}
EOF

export PKG_CONFIG_PATH=$HOME/pdfium-linux-x64/lib/pkgconfig${PKG_CONFIG_PATH:+:$PKG_CONFIG_PATH}
export LD_LIBRARY_PATH=$HOME/pdfium-linux-x64/lib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}
pkg-config --exists --print-errors "pdfium >= 4200" && echo OK

meson setup build --prefix=/usr -Ddeprecated=false -Dexamples=false -Dintrospection=disabled
meson compile -Cbuild
sudo meson install -Cbuild
```
</details>

It looks like this was fixed with commit https://pdfium.googlesource.com/pdfium/+/f8e68068d9301f24993afd94c270860dc0244674, i.e. this diff:
```diff
@@ -696,8 +727,13 @@ FPDF_EXPORT void FPDF_CALLCONV FPDF_FFLDraw(FPDF_FORMHANDLE hHandle,
                                             int size_y,
                                             int rotate,
                                             int flags) {
-  FFLCommon(hHandle, bitmap, nullptr, page, start_x, start_y, size_x, size_y,
-            rotate, flags);
+  CFX_DIBitmap* cbitmap = CFXDIBitmapFromFPDFBitmap(bitmap);
+  if (!cbitmap) {
+    return;
+  }
+
+  FFLCommon(hHandle, page, cbitmap, start_x, start_y, size_x, size_y, rotate,
+            flags);
 }
 
 #if defined(PDF_USE_SKIA)
```

Which ensures `NULL` bitmaps are handled gracefully in `FPDF_FFLDraw()`.

zhifengzhuang:
Thanks for replying. I remove vips_region_paint_pel and update test if().

libvips/foreign/pdfiumload.c
zhifengzhuang: > I don't think we should do this. We set the background when required with `FPDFBitmap_FillRect()` below.

So maybe we could paint background color to out_region when rect is empty?

zhifengzhuang:
> I can confirm that this issue reproduces with [PDFium 6611](https://github.com/bblanchon/pdfium-binaries/releases/tag/chromium%2F6611) but does not occur with [PDFium 6666](https://github.com/bblanchon/pdfium-binaries/releases/tag/chromium%2F6666). Tested using:
> 
> Details
> It looks like this was fixed with commit https://pdfium.googlesource.com/pdfium/+/f8e68068d9301f24993afd94c270860dc0244674, i.e. this diff:
> 
> ```diff
> @@ -696,8 +727,13 @@ FPDF_EXPORT void FPDF_CALLCONV FPDF_FFLDraw(FPDF_FORMHANDLE hHandle,
>                                              int size_y,
>                                              int rotate,
>                                              int flags) {
> -  FFLCommon(hHandle, bitmap, nullptr, page, start_x, start_y, size_x, size_y,
> -            rotate, flags);
> +  CFX_DIBitmap* cbitmap = CFXDIBitmapFromFPDFBitmap(bitmap);
> +  if (!cbitmap) {
> +    return;
> +  }
> +
> +  FFLCommon(hHandle, page, cbitmap, start_x, start_y, size_x, size_y, rotate,
> +            flags);
>  }
>  
>  #if defined(PDF_USE_SKIA)
> ```
> 
> Which ensures `NULL` bitmaps are handled gracefully in `FPDF_FFLDraw()`.

Yes，it was fixed https://pdfium.googlesource.com/pdfium/+/f8e68068d9301f24993afd94c270860dc0244674 (Just between 6632 & 6633)

kleisauke:
I've updated this PR to target the `8.17` branch and revised the commit message and changelog note accordingly.

libvips/foreign/pdfiumload.c
```diff
@@ -584,47 +584,56 @@ vips_foreign_load_pdf_generate(VipsRegion *out_region,
...
-		 */
-		bitmap = FPDFBitmap_CreateEx(rect.width, rect.height, 4,
-			VIPS_REGION_ADDR(out_region, rect.left, rect.top),
-			VIPS_REGION_LSKIP(out_region));
+			/* 4 means RGBA.
+			 */
+			if (!(bitmap = FPDFBitmap_CreateEx(rect.width, rect.height, 4,
+					VIPS_REGION_ADDR(out_region, rect.left, rect.top),
+					VIPS_REGION_LSKIP(out_region)))) {
+				g_mutex_unlock(&vips_pdfium_mutex);
```
jcupitt: Can this really fail? If it can, we should set an error message, perhaps:

```suggestion
                vips_pdfium_error();
                g_mutex_unlock(&vips_pdfium_mutex);
```

kleisauke: Ah, looking at the docs, I think this could only fail (i.e. OOM) when not using an external buffer. Let's remove this check.

libvips/foreign/pdfiumload.c
```diff
@@ -584,47 +584,56 @@ vips_foreign_load_pdf_generate(VipsRegion *out_region,
...
+				return -1;
+			}
```
kleisauke:
```suggestion
			bitmap = FPDFBitmap_CreateEx(rect.width, rect.height, 4,
				VIPS_REGION_ADDR(out_region, rect.left, rect.top),
				VIPS_REGION_LSKIP(out_region));
```

jcupitt:
Great! Let's merge!

Thank you for doing this work, @zhifengzhuang, this is a good thing to fix. This change will be in libvips 8.17.1.

PR #4600, opened 2025-07-05
Update CITATION.cff

kleisauke:


jcupitt:
I'd totally forgotten :( Nice!

kleisauke:
I think it was only published very recently (probably less than 2 days ago), so no worries.

PR #4604, opened 2025-07-06
use much larger mmap windows

jcupitt:
At the moment, formats like JPEG will convert to uncompressed `.v` images in tmp and then be mapped into memory for processing.  In order to limit VMEM usage, this mapping is windowed, so only a few 100 scanlines are mapped at once.

However, this can cause severe performance issues with applications needing random pixel access, for example, image viewers. A viewer which is updating the screen in parallel will make scattered reads to the image, and each read will force the mmap window to scroll. With a large image, almost all time will be spent blocked on ioctl().

This PR makes the mmap window *much* larger, up to about 20,000 scanlines. Modern machines have much more VMEM available, so I think this will be OK.

Performance of interactive applications can improve by several 100 times.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,7 +3,8 @@ master
 - add dcrawload, dcrawload_source, dcrawload_buffer: load raw camera files
   using libraw [lxsameer]
 - add magickload_source: load from a source with imagemagick
-- improve random access mode for many file formats by adding fully mapped images
+- larger mmap windows on 64-bit machines improve random access mode for many
+  file formats
 
 8.17.1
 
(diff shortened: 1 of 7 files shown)


jcupitt:
Here's git master viewing a large file:

https://youtu.be/79MzqJyB6Nc

And here's this PR:

https://youtu.be/UjRMomizz_s

git master takes about 30s to update the screen, with this PR it's under 0.5s.

jcupitt:
We could change the mmap windows size with the access hint, so sequential images keep the small window we have now, random images get the entire file mapped. Maybe that would be a better compromise?

jcupitt:
It now only mmaps the complete image when you open a format like JPG in random access mode and it gets converted via a temporary disc file.

jcupitt:
I had yet another rethink. It now uses large (up to 10gb) windows on 64-bit machines, and small windows (up to 10mb) on 32-bit machines. This is nice and simple, and I think it should perform well everywhere.

Issue #4597, opened 2025-07-02
Support for PDF boxes


jebbench:
How feasable would it be to add an option to pdfload to specify which of the PDF boxes to use when reading the PDF?

Our usecase is creating thumbnails from PDFs with printer marks on, we want to use the trim box so the thumbnail is what will be produced off the back of a guillotine (rather than the current default which shows what will be printed including the printers marks).

This comment (https://github.com/libvips/libvips/issues/379#issuecomment-181357031) explains the PDF boxes well.

lovell:
libvips can read PDF documents via either poppler or pdfium. The former seems to support only bounding and crop boxes, and the latter supports those plus trim, art and bleed boxes.

- https://poppler.freedesktop.org/api/glib/poppler-Poppler-Page.html#poppler-page-get-bounding-box
- https://pdfium.googlesource.com/pdfium/+/refs/heads/main/public/fpdf_transformpage.h#130

These pages provide a bit more info about how the boxes relate to each other, which might help someone who is implementing this feature.

- https://wiki.scribus.net/canvas/PDF_Boxes_:_mediabox,_cropbox,_bleedbox,_trimbox,_artbox
- https://www.prepressure.com/pdf/basics/page-boxes

I've labelled this issue as a future possible enhancement.

PR #4605, opened 2025-07-10
pdfload: control region to be rendered via page_box (pdfium only)

lovell:
Each page in a PDF document can specify up to 5 different, optional page boxes, namely _media_, _crop_, _trim_, _bleed_ and _art_. A PDF renderer should, by default, use the _crop_ box if it exists otherwise use the page dimensions, and this behaviour remains the same.

This PR adds a new `page_box` parameter to the pdfium-powered `pdfload` to allow control over which page box should define the region to render. If the requested page box is not present, it will fallback to using the crop box (which by default is the page dimensions).

As pdfium uses two different coordinate systems, this change ensures the page box calculations occur using the _page_ coordinate system before we start to request dimensions in the device-specific _client_ coordinate system. (When I started work on this I added some logic to libvips to convert between the two systems, but the final approach here ends up being much simpler.)

From what I can tell, the pdfium API calls to read/write page box info are not thread safe so these are wrapped in the mutex.

I've updated the pdfium-based operation descriptions to include a "pdfium" suffix to expose (in a rather brittle manner) which PDF library is being used as this is required by the tests. I'd be very happy to improve this if there's a better way.

The existing PDF tests are expanded to cover this new logic, including a new test fixture.

The fuzz test environment already includes pdfium and therefore includes coverage for pdfiumload so I've not added anything further here.

I investigated how we might implement this via Poppler, however its (glib-based) C API does not expose any page box information. To implement page box rendering with Poppler we would have to either switch to using its C++ API or modify Poppler itself, both of which are beyond the scope of this feature.

media

crop

bleed

trim

art

Closes https://github.com/libvips/libvips/issues/4597


jcupitt:
How about adding this param to the poppler loader as well, but making it do nothing? Or maybe issue a warning? We could possibly wire it up to something in the future if poppler expose this feature or we switch to the poppler C++ API. 

I think this is what we do for other operations which can change their behaviour with compile time flags.

I suppose we could have a abstract base PDF class with the args and then two implementation sub classes, but that sounds like a PITA.

jcupitt:
LGTM!

Issue #4609, opened 2025-07-13
vips_rank() fails with large windows


jcupitt:
The histogram path for the rank filter causes errors with windows of 10x10 and up. 

Reproducer:

```python
#!/usr/bin/env python3

import pyvips

pyvips \ 
    .Image.svgload_buffer(b"""
        <svg viewBox="0 0 2000 2000">
        <rect width="100%" height="100%" fill="black"/>
        <circle r="500" cx="1000" cy="1000" fill="white"/>
        </svg>
     """) \
    .rank(10, 10, 0) \
    .write_to_file("x.png")
```

Makes:

<img width="1128" height="1029" alt="Image" src="https://github.com/user-attachments/assets/27bf1e51-64de-4196-a7e6-d2f3e720f839" />

Referring discussion:

https://github.com/libvips/libvips/discussions/4606#discussioncomment-13745393

jcupitt:
We have a PR, I'll close.

PR #4610, opened 2025-07-16
fix an off by one error in rank

jcupitt:
only there for three years sigh

see https://github.com/libvips/libvips/issues/4609

### DIFF
--- a/libvips/morphology/rank.c
+++ b/libvips/morphology/rank.c
@@ -517,7 +517,7 @@ vips_rank_build(VipsObject *object)
 	 * too many recalculations on overlaps.
 	 */
 	if (vips_image_pipelinev(rank->out,
-			VIPS_DEMAND_STYLE_SMALLTILE, in, NULL))
+			VIPS_DEMAND_STYLE_FATSTRIP, in, NULL))
 		return -1;
 	rank->out->Xsize -= rank->width - 1;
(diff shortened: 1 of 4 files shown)


libvips/morphology/rank.c
```diff
@@ -175,62 +168,58 @@ vips_rank_generate_uchar(VipsRegion *out_region,
...
-			q[b] = i;
+			q[b] = value;
 
-			/* Adapt histogram -- remove the pels from
-			 * the left hand column, add in pels for a
-			 * new right-hand column.
+			/* Adapt histogram -- remove the pels from the left hand column,
+			 * add in pels for a new right-hand column.
 			 */
+			const int next = bands * rank->width;
```
jcupitt, line 218: this is the fix, it was `bands * (rank->width - 1)`

jcupitt:
I reformatted to our more modern style, which make the fix hard to spot, but see the comment above.

The error only really shows when running large rank filters on binary images, so maybe the three year discovery time is not so bad.

I checked `vips_hist_local()` (also has a rolling histogram) and it's fine.

kleisauke:
Was `8.17` intentionally used as the base branch? It looks like we should drop commit cc61cb24d2a6091d60a1c136920c3e147d7ba5ff and merge this directly into `master` via the CLI instead.

Issue #4612, opened 2025-07-19
XWD support


noodlefan:
Hello,

It'd be nice if we had support for the X Window Dump file format in libvips to allow taking screenshots via the xwd(1) utility that comes with the X11 distribution by default.

Example command with xwd(1) and convert(1) from ImageMagick to take a PNG screenshot:
`$ xwd -root | convert xwd:- test.png`

Possible usage examples and actual outputs:
`$ xwd -root | vips stdin test.png`
`VipsImage: unable to load source`
`$ xwd -root -out test.xwd && vips copy test.xwd test.png`
`VipsForeignLoad: "test.xwd" is not a known file format`

Counter arguments:
- Keep around an ImageMagick installation for taking screenshots: duplicates efforts
- Use a full-featered screenshot-taking utility like scrot: more packages to manage

Thank you.

jcupitt:
Hello, you're in luck! This is in 8.18, due in the autumn.

https://github.com/libvips/libvips/pull/4592

ghost:
> Hello, you're in luck! This is in 8.18, due in the autumn.
> 
> [#4592](https://github.com/libvips/libvips/pull/4592)

Amazing! Thanks a lot and goodluck with the release :D

kleisauke:
> ```console
> $ xwd -root -out test.xwd && vips copy test.xwd test.png
> ```

Note that you could already make this work with:
```console
$ xwd -root -out test.xwd && vips magickload xwd:test.xwd test.png
```
(assuming you have build libvips against ImageMagick)

We could perhaps add a dedicated sniffer for this format using:
```diff
--- a/libvips/foreign/magick.c
+++ b/libvips/foreign/magick.c
@@ -138,6 +138,15 @@ magick_sniff(const unsigned char *bytes, size_t length)
 		(bytes[4] != 0 || bytes[5] != 0))
 		return "ICO";
 
+	if (length >= 8 &&
+		bytes[1] == 0 &&
+		bytes[2] == 0 &&
+		bytes[5] == 0 &&
+		bytes[6] == 0 &&
+		((bytes[4] == 7 && bytes[7] == 0) ||
+			(bytes[4] == 0 && bytes[7] == 7)))
+		return "XWD";
+
 	if (length >= 18 &&
 		(bytes[1] == 0 ||
 			bytes[1] == 1) &&
```

Though, I'm not sure how common this image format is. For example, the xwd utility doesn't seems to be supported on my Fedora 42 workstation.

[...]

PR #4613, opened 2025-07-19
add a sniffer for the XWD format

jcupitt:
otherwise we mistake it for TGA

see https://github.com/libvips/libvips/issues/4612


kleisauke (approved):
:+1:

PR #4620, opened 2025-08-03
validate poppler page size

jcupitt:
Poppler can return pages less than 1 pixel wide or high. Check for this and flag an error.

Thanks Yang Luo

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,7 @@
 8.17.2
 
 - rank: fix an off-by-one error [larsmaxfield]
+- popplerload: check page size [Yang Luo]
 
 7/7/25 8.17.1
 
--- a/libvips/foreign/popplerload.c
+++ b/libvips/foreign/popplerload.c
@@ -312,7 +312,6 @@ vips_foreign_load_pdf_header(VipsForeignLoad *load)
 	VipsForeignLoadPdf *pdf = VIPS_FOREIGN_LOAD_PDF(load);
 
 	int top;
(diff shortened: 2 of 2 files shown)


jcupitt:
We don't always validate all metadata returned by load libraries (we mostly assume they do this).

Perhaps we should add more protection against this type of error? For example, we could add:

```C
if (!vips_object_sanity(VIPS_OBJECT(image)))
    return -1;
```

Near the start of major functions such as `vips_image_generate()`, or at strategic points in the base class of `VipsForeign`.

`vips_image_sanity()` can be quite slow (it walks the entire pipeline testing inter-image links). We could maybe only do the walk when `--vips-leak` is enabled.

jcupitt:
I added some extra sanity checks as a backstop. They wouldn't have stopped this error, but they should catch other missed validations.

Is this too large a change for stable? We could delay until 8.18.

lovell:
Should the target branch be updated to `8.17`?

jcupitt:
Oh drat, I always forget to set the target, sorry.

lovell:
I'm unsure about adding these new `vips_object_sanity` calls given it uses `printf`, which is more of a CLI approach than libvips-as-a-library behaviour.

jcupitt:
Oh, you're right, `vips_object_sanity()` sends stuff to stdout.

It'd be a simple change to make sanity log an error, or perhaps that should wait until 8.18?

lovell (approved):

PR #4629, opened 2025-08-13
pdfiumload: allow both dpi and scale to be set

kleisauke:
In line with commit 0c7b65e1569f28ceee74c9b5e348edb3a18aff68.

Targets the 8.17 branch.


jcupitt:
That was a silly one to drop! Nice find.

PR #4632, opened 2025-08-15
pdfiumload: allow both dpi and scale to be set (#4629)

jcupitt:
In line with commit 0c7b65e1569f28ceee74c9b5e348edb3a18aff68.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,7 @@
 8.17.2
 
 - rank: fix an off-by-one error [larsmaxfield]
+- pdfiumload: allow both dpi and scale to be set [kleisauke]
 
 7/7/25 8.17.1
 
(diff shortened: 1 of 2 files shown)


kleisauke:
Was this opened by mistake? We usually merge releases branches back to the master branch via the CLI.

jcupitt:
I just thought it might be a little easier, sorry for the noise.

kleisauke:
Ah, this could also be done via a PR, we'd just need to ensure it's merged with a rebase merge rather than a squash merge (just like the backport PRs). 

The "You're all set — the [8.17](https://github.com/libvips/libvips/tree/8.17) branch can be safely deleted." below this message is also a bit tricky, we should never actually click that button. :sweat_smile:

jcupitt:
Hehe yes that would be bad

PR #4637, opened 2025-08-18
Add new metadata fields 'tile-width' and 'tile-height'.

jbaiter:
We have a use case where we need to know the tile size in the encoded image to give hints to clients which regions to request for optimal performance.

For this, we add two new integer metadata fields, `tile-width` and `tile-height`, which are set if tiles are recognized during image parsing.

This PR implements this for the following image formats:
- TIFF
- JPEG2000
- EXR
- OpenSlide

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,8 @@ master
 - larger mmap windows on 64-bit machines improve random access mode for many
   file formats
 - pdfload: control region to be rendered via `page_box` [lovell]
+- add vips_image_get_tile_width(), vips_image_get_tile_height(): get tile
+  cache geometry hints [jbaiter]
 
 8.17.2
 


jcupitt:
Hi @jbaiter, this looks like a good PR, nice job!

I though of a possible simplification: instead of adding the code to set `VIPS_META_TILE_WIDTH` etc. to each loader, how about doing this in the `tilecache` operator? I think all of these loaders set the tile cache tile size to match the image file tile size, so it ought to work.

Something like (untested):

```C
    if (vips_image_pipelinev(conversion->out,
            VIPS_DEMAND_STYLE_SMALLTILE, block_cache->in, NULL))
        return -1; 
    
    // hint cache geometry to downstream operators
    vips_image_set_int(conversion->out,
        VIPS_META_TILE_WIDTH, block_cache->tile_width);
    vips_image_set_int(conversion->out, 
        VIPS_META_TILE_HEIGHT, block_cache->tile_height);

    if (vips_image_generate(conversion->out,
            vips_start_one, vips_tile_cache_gen, vips_stop_one,
            block_cache->in, cache))
        return -1;
```

in `vips_tile_cache_build()`.

I wonder if getting the libvips sinks to check these hints could be a performance win in some cases? Perhaps that's getting too far off piste for this PR.

jbaiter:
Thank you for the suggestion, I now implemented it that way, much cleaner! :-)

As for the potential performance improvements by using these hints, can you go into some more details? I'm only starting with the libvips code, but if the tile cache already knows about the "optimal" tile size for reading the images, where would the performance wins in the sinks come from? Less copying by making the output tile size match the input tile size?

jcupitt:
Yes, the sinks allocate a buffer of $n lines, then loop filling it with tiles from the pipeline, usually 128x128 pixels. You can see what it's doing with the `--vips-progress` flag.

```
$ vips invert k2.jpg x.jpg --vips-progress
vips temp-6: 1450 x 2048 pixels, 32 threads, 1450 x 16 tiles, 640 lines in buffer
vips temp-6: done in 0.0237s          
```

With a JPEG source, it knows to demand in full-width lines (since libjpeg supplies decoded pixels in lines) and (by chance) is asking for 16 at a time. It is using up to 32 threads on this PC, so it needs 16 * 32 == 512 lines, plus a bit. It uses double buffer and write-behind, so it's really allocating two 640 line buffers. When the first one fills, it swaps the buffers and a background thread will write those 640 lines to the output JPEG in parallel with the computation of the next 640 lines. 

You can see the rules it uses for calculating the buffer height here:

https://github.com/libvips/libvips/blob/master/libvips/iofuncs/thread.c#L252-L332

The win might come with eg. a tiled TIFF source:

```
$ vips copy k2.jpg x.tif[tile,compression=jpeg,tile-width=192,tile-height=16]
$ vips invert k2.tif x.jpg --vips-progress
vips temp-6: 1450 x 2048 pixels, 32 threads, 128 x 128 tiles, 640 lines in buffer
vips temp-6: done in 0.048s          
```

Now it knows the TIFF is tiled, so it's fetching in square tiles, but the crazy tile size I chose for the TIFF will mean each 128x128 tile the tilecache makes will need to be assembled from many TIFF tiles with lots of copies. If the sink used 192x16 tiles that could all be skipped.

I don't know how common this is! But I think it could probably help in some cases.

Thinking again, there's another case too: large vertical shrinks subdivide the fetch to avoid memory spikes. Those could also perhaps also aim to match the input image tile size.

All of this goes out of the window if there are eg. rotates or blurs, of course :( since they'll always mess up tile alignment.

jcupitt:
... I meant to ask about your use case. Could you explain what you are doing that sees a useful performance boost from tile alignment? How large a boost do you get?

jbaiter:
Thank you for the explanation, sounds like it could indeed give a decent boost in some situations! :-)

As for our use case, we're working on a new [IIIF Image API](https://iiif.io/api/image/3.0/) server implementation, and the spec has the option to indicate to clients which tile sizes to request. The main access pattern for IIIF Image Servers is multiple subsequent requests for tiles from a larger image, and using tiles that are aligned with what's in the input images can reduce load on the server. We haven't quantified it, though.

jcupitt:
Ooop, sorry, I forgot one thing: this is an API addition, so we need a line in the ChangeLog and a credit. Perhaps:

```
- add vips_image_get_tile_width(), vips_image_get_tile_height(): get tile cache geometry hints [jbaiter]
```

jcupitt:
Great! 

Thank you for doing this work, @jbaiter

PR #4640, opened 2025-08-20
enable caching of vips_system()

jcupitt:
We were not caching vips_system() calls, since they could potentially have side effects. However, almost all the time we are doing things like running imagemagick commands, so caching is very helpful.

This patch enables caching of vips_system(), and adds better hashing of compound argument types.


jcupitt:
This PR makes nip4 menu items like "Magick / Annotate" 9 times faster (!!!).

It could be bad if we want to force execution. Maybe we should move the `"revalidate"` flag into VipsOperation (it's on VipsForeignLoad right now).

kleisauke:
> It could be bad if we want to force execution. Maybe we should move the `"revalidate"` flag into VipsOperation (it's on VipsForeignLoad right now).

I'm not sure that's the best approach, e.g. the `"revalidate"` flag doesn't make sense for operations that are already "nocache". 

FWIW, it reminds me of commit https://github.com/kleisauke/net-vips/commit/fa03ac416b95a528467cb7e7e6a9e86f5f888e31, which removes those `"revalidate"` flags in NetVips from the source loaders (already uncached) and the buffer loaders (always wrapped in a new `VipsBlob`, so uncacheable).

jcupitt:
Hmm true.

Maybe a `nocache` or `revalidate` argument just for system? I think it could define a VipsOperation::get_flags method that flagged `NOCACHE`.

jcupitt:
I added a `nocache` optional arg to `system` to disable caching of a call (eg. if you know that you always want this call to execute).

Though perhaps the sense should be flipped and it should be called `cache`? We'd have full back compat then.

jcupitt:
... I flipped it to `cache`, so we should have compatibility now.

kleisauke (approved):
Great!

PR #4641, opened 2025-08-21
openslideload_source: flag operation as "nocache"

kleisauke:
See commit cb58d7d9600282802d2d04a04254f6d9dfefe203 for rationale.

History:
Commit 1214f942f9397db25cfec19d5712b88480965a0f moved this flag to the base class, and commit 53db48d49cea96a699fb74c6713fe810e0ed805b later removed it, likely under the assumption that the `_source` class still had this flag set.

Targets the 8.17 branch.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -5,6 +5,7 @@
 - pdfiumload: allow both dpi and scale to be set [kleisauke]
 - svgload: honor DPI when scaling elements with non-pixel units [kleisauke]
 - disable redundant Highway AVX512 targets [kleisauke]
+- openslideload_source: flag operation as "nocache" [kleisauke]
 
 7/7/25 8.17.1
 
--- a/libvips/foreign/openslideload.c
+++ b/libvips/foreign/openslideload.c
@@ -1155,6 +1155,7 @@ vips_foreign_load_openslide_source_class_init(
 {
 	GObjectClass *gobject_class = G_OBJECT_CLASS(class);
 	VipsObjectClass *object_class = (VipsObjectClass *) class;
+	VipsOperationClass *operation_class = VIPS_OPERATION_CLASS(class);
 	VipsForeignLoadClass *load_class = (VipsForeignLoadClass *) class;
 
 	gobject_class->set_property = vips_object_set_property;
@@ -1163,6 +1164,8 @@ vips_foreign_load_openslide_source_class_init(
 	object_class->nickname = "openslideload_source";
 	object_class->description = _("load source with OpenSlide");
 	object_class->build = vips_foreign_load_openslide_source_build;
+
+	operation_class->flags |= VIPS_OPERATION_NOCACHE;
 
 	load_class->is_a_source =
 		vips_foreign_load_openslide_source_is_a_source;


kleisauke:
Found by this (not yet committed) change in NetVips:
```diff
--- a/samples/NetVips.Samples/Samples/GenerateImageClass.cs
+++ b/samples/NetVips.Samples/Samples/GenerateImageClass.cs
@@ -156,7 +156,8 @@ private string GenerateFunction(string operationName, string indent = "
         bool mutable = false, IReadOnlyList<Introspect.Argument> outParameters = null)
     {
         using var op = Operation.NewFromName(operationName);
-        if ((op.GetFlags() & Enums.OperationFlags.DEPRECATED) != 0)
+        var flags = op.GetFlags();
+        if ((flags & Enums.OperationFlags.DEPRECATED) != 0)
         {
             throw new ArgumentException($"No such operator. Operator \"{operationName}\" is deprecated");
         }
@@ -172,8 +173,8 @@ private string GenerateFunction(string operationName, string indent = "
         // we are only interested in non-deprecated args
         var optionalInput = intro.OptionalInput
             .Where(arg => (arg.Value.Flags & Enums.ArgumentFlags.DEPRECATED) == 0)
-            // Drop "revalidate" option from source and buffer loaders as they are already uncached
-            .Where(arg => arg.Key != "revalidate" || (firstArgType != GValue.SourceType && firstArgType != GValue.BlobType))
+            // Drop "revalidate" option from buffer loaders and operations marked "nocache" as they are already uncached
+            .Where(arg => arg.Key != "revalidate" || ((flags & Enums.OperationFlags.NOCACHE) == 0 && firstArgType != GValue.BlobType))
             .Select(x => x.Value)
             .ToArray();
         var optionalOutput = intro.OptionalOutput
```

jcupitt:
That's annoying. How about adding this to operation.c:

```C
static void
vips_operation_summary_class(VipsObjectClass *object_class, VipsBuf *buf)
{
    VipsOperationClass *operation_class = VIPS_OPERATION_CLASS(object_class);

    VIPS_OBJECT_CLASS(vips_operation_parent_class)
        ->summary_class(object_class, buf);
 
    GType gtype = G_OBJECT_CLASS_TYPE(operation_class);
    if (!G_TYPE_IS_ABSTRACT(gtype) &&
        operation_class->flags & VIPS_OPERATION_NOCACHE)
        vips_buf_appendf(buf, ", nocache");
}
```

Now `vips -l` displays eg.:

```
$ vips -l | grep _source
          VipsForeignLoadCsvSource (csvload_source), load csv, nocache, priority=0, untrusted, is_a_source, get_flags, header, load
...
```

And we can test that every source loader is tagged correctly with:

```
$ vips -l | grep _source | grep -v nocache
          VipsForeignLoadOpenslideSource (openslideload_source), load source with OpenSlide, priority=100, untrusted, is_a_source, get_flags, get_flags_filename, header, load
$ 
```

(before this PR, of course hehe)

kleisauke:
Good idea! Let's do that as a follow-up.

jcupitt:
What should we do about your netvips commit? Do you think `revalidate` should move out of the base class?

kleisauke:
I'll apply that NetVips commit when regenerating the methods for libvips 8.18. 

I'm not sure there's much benefit to moving the `revalidate` flag to `VipsOperation`, it would likely create unnecessary noise in downstream bindings. The only use case I can see is benchmarking libvips, but in that case it's probably better to disable the cache globally rather than individually for each operation.

jcupitt:
OK, let's merge then!

PR #4648, opened 2025-08-26
operation: include "nocache" flag in class summary

kleisauke:
Context: https://github.com/libvips/libvips/pull/4641#issuecomment-3210822736.


jcupitt:
Well remembered!

PR #4651, opened 2025-08-28
linear: avoid single-band image when all constants are equal

kleisauke:
Fixes a regression introduced in commit 9cc54a0.

### DIFF
--- a/libvips/arithmetic/linear.c
+++ b/libvips/arithmetic/linear.c
@@ -103,9 +103,12 @@ typedef struct _VipsLinear {
 	 */
 	gboolean uchar;
 
+	/* TRUE if a and b each contain only one unique value.
+	 */
+	gboolean single_element;
+
 	/* Our constants expanded to match arith->ready in size.
 	 */
-	int n;
 	double *a_ready;
 	double *b_ready;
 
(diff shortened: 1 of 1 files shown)


kleisauke:
FWIW, we previously assigned directly to `linear->a->n` and `linear->b->n`, so this may also need to be updated:
https://github.com/libvips/libvips/blob/42cf3136be20c945368c344490b74994671696cb/libvips/arithmetic/linear.c#L230
https://github.com/libvips/libvips/blob/42cf3136be20c945368c344490b74994671696cb/libvips/arithmetic/linear.c#L289

(perhaps we could hoist the `a_n` and `b_n` locals into the struct and adjust the corresponding checks accordingly?)

jcupitt:
You're right, I broke it a bit, sorry.

`linear->n` is the size of the constants. We should probably have that as 1 for all-elements-equal, and test that value in the macro.

kleisauke:
Commit c7a53c809706625f73558bca1a2308f79745ea73 should fix that.

kleisauke:
Is it OK to land this?

jcupitt:
Sorry, I keep meaning to test this and forgetting :( I'll try again tonight.

jcupitt (approved):
LGTM!

jcupitt:
Sorry for being so incredibly slow on this one :( I agree, it looks fine!

kleisauke:
No worries! It was only a short-lived regression and had no impact on any stable releases. :relieved:

Issue #4649, opened 2025-08-27
Loading some compressed TIFF files does not work with 8.17 (but were ok with 8.16)


harri-314:
## Bug report

**Describe the bug**

Attached is a deflate compressed TIFF (created with RawTherapee) that does not load correctly in 8.17.0 nor in 8.17.1. The loaded image is all zeroes. 

The file is loaded correctly with 8.16.0 and 8.16.1. 

Tested all four versions on windows-11 and also 8.17.1 on OpenSuse Tumbleweed with the same problem.

**To Reproduce**
1. Use the attached image `fail.tif`
2. For instance from the command line run `vips max fail.tif`
3. Correct output should be 65535, but 8.17 version prints 0.

**Expected behavior**
Output should be 65535.000, confirmed with versions 8.16.x

**Actual behavior**
Output is 0.000 produced by versions 8.17.x

**Environment**
(please complete the following information)
 - OS: Windows-11 and OpenSuse
 - Vips: 8.17.0 and 8.17.1

[fail.tif](https://github.com/user-attachments/files/22006907/fail.tif)

jcupitt:
Hi @harri-314,

How odd, it all seems to be working for me on ubuntu 25.04. I tried with 8.17 and 8.18 (head of master). It works with vipdisp too:

<img width="1128" height="1029" alt="Image" src="https://github.com/user-attachments/assets/8b4acdc6-12c1-4db3-b084-cb1f88b1dd0c" />

I'll try on win10.

jcupitt:
You're right, this fails on win10 with our official binary:

```
jcupi@DESKTOP-HGI6HBR MINGW64 /f/vips-dev-w64-all-8.17.1/vips-dev-8.17/bin
$ ./vips.exe max z:\fail.tif
0.000000
```

I suppose it's been built with a wonky version of zlib. We switched to `zlib-ng`, perhaps that fails with 16-bit images?

kleisauke:
I was able to reproduce this on my Fedora 42 workstation. Opening the file with the `unlimited` flag seems to fix this.
```console
$ vips max ~/Downloads/fail.tif 
0.000000
$ vips max ~/Downloads/fail.tif[unlimited]
65535.000000
```

This makes commit dee0d89ff385407b65f8526657a5cdb3b1e82bb9 relevant here.

jcupitt:
My ubuntu is using a different zlib, which presumably is working:

```
$ dpkg -S libz.so.1
[...]

PR #4652, opened 2025-08-29
tiffload: improve handling of memory-related errors

lovell:
Ensures TIFF processing halts:
- for known-fatal errors relating to memory allocation
- for all other errors when `fail_on` is set to the `error` level or higher

Also, in a separate commit, increases the memory limit from 20MB to 50MB.

Fixes https://github.com/libvips/libvips/issues/4649

### DIFF
--- a/libvips/foreign/tiff2vips.c
+++ b/libvips/foreign/tiff2vips.c
@@ -612,6 +612,19 @@ static int
 rtiff_handler_error(TIFF *tiff, void* user_data,
 	const char *module, const char *fmt, va_list ap)
 {
+	if (user_data) {
+		Rtiff *rtiff = (Rtiff*) user_data;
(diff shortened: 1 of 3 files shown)


libvips/foreign/tiff2vips.c
```diff
@@ -612,6 +612,19 @@ static int
...
 	const char *module, const char *fmt, va_list ap)
 {
+	if (user_data) {
+		Rtiff *rtiff = (Rtiff*) user_data;
+		if (rtiff->fail_on >= VIPS_FAIL_ON_ERROR) {
+			rtiff->failed = TRUE;
+		} else if (fmt) {
+			/* Inspect message for always-fatal errors.
+			*/
+			if (strncmp(fmt, "Cumulated memory", 16) == 0) {
```
jcupitt, line 622: I checked libtiff and this message doesn't seem to be i18n-ed, phew!

Issue #4654, opened 2025-09-03
v8.17.x compile error: The type name `GLib.InputStreamClass' could not be found


stydxm:
## Affected

since [1741857](https://github.com/libvips/libvips/commit/1741857876e54f90c23c0fcc45f411d4522ce341)

## Command

```bash
meson setup -Dexamples=false -Ddocs=true -Dvapi=true -Dnifti-prefix-dir=/usr -Dpdfium=disabled -Dquantizr=disabled -Dspng=disabled -Dcfitsio=enabled -Dcgif=enabled -Dexif=enabled -Dfftw=enabled -Dfontconfig=enabled -Darchive=enabled -Dheif=enabled -Dimagequant=enabled -Dheif=enabled -Djpeg=enabled -Djpeg-xl=enabled -Dlcms=enabled -Dmagick=enabled -Dmatio=enabled -Dnifti=enabled -Dopenexr=enabled -Dopenjpeg=enabled -Dopenslide=enabled -Dhighway=enabled -Dorc=enabled -Dpangocairo=enabled -Dpng=enabled -Dpoppler=enabled -Drsvg=enabled -Dtiff=enabled -Dwebp=enabled -Dzlib=enabled build
cd build
meson compile
```

## Error

```log
[544/549] Generating libvips/vips.vapi with a custom command
FAILED: [code=1] libvips/vips.vapi 
/usr/bin/vapigen --quiet --library=vips --directory=/home/stydxm/code/libvips/build/libvips --pkg=glib-2.0 --pkg=gio-2.0 --pkg=gobject-2.0 --metadatadir=/home/stydxm/code/libvips/libvips /home/stydxm/code/libvips/build/libvips/Vips-8.0.gir
Vips-8.0.gir:22558.7-22558.62: warning: Signal `Vips.Image.posteval' conflicts with method of the same name
Vips-8.0.gir:22807.7-22807.58: warning: Signal `Vips.Image.preeval' conflicts with method of the same name
Vips-8.0.gir:29354.7-29354.54: warning: Creation method of abstract class cannot be public.
Vips-8.0.gir:30580.7-30581.73: warning: Creation method of abstract class cannot be public.
Vips-8.0.gir:32046.7-32046.50: warning: Creation method of abstract class cannot be public.
Vips-8.0.gir:35559.7-35559.56: warning: Field `Vips.Source.decode' conflicts with method of the same name
[...]

PR #4655, opened 2025-09-03
Fix vapi build

jcupitt:
It looks like Gio.gir is not wrapping GInputStreamClass, so we can't include it in stuff we wrap. Marking the whole of the class private seems to fix this.

I've not tested the .vapi file we make.

See https://github.com/libvips/libvips/issues/4654

Thanks to @stydxm for the clear report.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -9,7 +9,7 @@
 - remove `vipsprofile.1` man page from install [kleisauke]
 - tiffload: ensure processing halts for memory-related errors [lovell]
 - tiffload: increase memory limit from 20MB to 50MB [lovell]
-- fix vapi build [jcupitt]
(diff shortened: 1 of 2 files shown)


jcupitt:
It looks like `Gio.gir` is not wrapping `GInputStreamClass`, so we can't include it in stuff we wrap. Marking the whole of the class private seems to fix this.

jcupitt:
Perhaps we should include `-Dvapi=true` in our test builds? It might help catch things like this sooner. Any thoughts?

kleisauke:
I found the underlying issue, see PR #4656. Adding `-Dvapi=true` to our CI builds sounds good to me.

jcupitt:
Well done! Your fix is a lot better.

Issue #4664, opened 2025-09-12
[unlimited] flag needed for multi-strip TIFFs despite small tag overhead - clarification needed


datatango:
I'm converting TIFF files to pyramidal TIFFs using vips-8.17.1 with:

`vips tiffsave <input> <output> --tile --tile-width=256 --tile-height=256 --pyramid --compression none`

Issue: Numerous conversions produce black output TIFFs but adding [unlimited] resolves it:

`vips tiffsave <input>[unlimited] <output> --tile --tile-width=256 --tile-height=256 --pyramid --compression none`

<img width="3020" height="1708" alt="Image" src="https://github.com/user-attachments/assets/326c3458-6fda-4a95-b62f-fae204a0cc5d" />

Question: [Documentation](https://www.libvips.org/API/current/ctor.Image.tiffload.html) states the [unlimited] flag addresses libtiff 4.7.0+'s 20MB tag processing limit. However, I'm seeing unexpected behavior:

Files requiring [unlimited] flag (black output without it):

- Multi-strip TIFFs with 'large' strip data but small tag overhead

- Example: https://ark.lib.uchicago.edu/ark:61001/b2002q31hv55/00000001/file.tif
    - Tags: ~314 bytes total (well under 20MB)
    - 12 strips with 'large' strip data (~23 MB)

Files working without [unlimited] flag:

- Single-strip TIFFs with >300MB image data

- Example: https://ark.lib.uchicago.edu/ark:61001/b2008kf2f543/00000001/file.tif

My understanding: The 20MB limit should apply to tag metadata processing, not strip/image data. Yet files with minimal tag overhead but multiple strips seem to trigger this limit, while single-strip files with much larger data don't.

- Does the memory limit affect strip data loading beyond just tag processing?

- Is VIPS applying additional memory constraints when reading multi-strip TIFFs?

Any clarification on when [unlimited] is actually needed would be helpful for processing large TIFF collections efficiently. Thank you.

lovell:
This is likely to be the same as https://github.com/libvips/libvips/issues/4649

jcupitt:
Yes, I think this is fixed in 8.17.2. We should probably make another release.

> while single-strip files with much larger data don't.

My understanding is that this depends on the compression. Uncompressed strips can be read directly with no buffering, but most compression types will require a read of the whole strip to a buffer, then decompression to the output. Allocating this buffer will add to the per-image limit.

jcupitt:
[...]

PR #4665, opened 2025-09-13
tiffload: ensure docs match memory limit behaviour

lovell:
See commit dcb1df8 (discovered via https://github.com/libvips/libvips/issues/4664).


jcupitt:
Maybe we could remove the reference to tag processing? The limit seems to be for all memory allocations on each file handle.

Maybe:

```
When using libtiff 4.7.0+, the TIFF loader will limit memory allocation
for each tiff file to 50mb to prevent denial of service attacks. Set 
@unlimited to remove this limit.
```

PR #4669, opened 2025-09-16
fix tiffsave with non-default resolution-unit

jcupitt:
A typo was stopping resolution-unit inch metadata applying correctly.

Before this PR, with this test image:

https://openslide.cs.cmu.edu/download/openslide-testdata/Aperio/CMU-1.svs

I see:

```
$ vipsheader -f xres ~/pics/openslide/CMU-1.svs
2004.0080160320642
```

so 2004.01 pixels per mm in the original, but:

```
$ vips crop CMU-1.svs x.tif 30000 10000 100 100
$ tiffinfo x.tif
=== TIFF directory 0 ===
TIFF Directory at offset 0x9c48 (40008)
...
  Resolution: 20040.1, 20040.1 pixels/inch
```

The resolution in the saved TIFF has not been converted to pixels per inch correctly, it's used cm instead.

See https://github.com/libvips/pyvips/issues/552

Thanks to @machur for finding this.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,5 +1,5 @@
 project('vips', 'c', 'cpp',
-    version: '8.17.2',
+    version: '8.17.3',
     meson_version: '>=0.55',
     default_options: [
         # this is what glib uses (one of our required deps), so we use it too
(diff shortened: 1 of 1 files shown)


machur:
@jcupitt Just a quick note: in my tests, the output resolutions for different slides were scaled in multiples of 10 (either up or down, depending on the input slide). I’m not entirely sure how this fix addresses that, especially since it involves inches - my knowledge here is very limited though...

kleisauke (approved):
Oops, my bad. Thanks for fixing this!

PR #4680, opened 2025-09-24
Improve kill handling

jcupitt:
Some small fixes found while improving load cancel behaviour in nip4:

- libvips was not setting `O_TMPFILE` on temp files on linux (the dir check in `vips__open()` was breaking it)
- in some cases `vips_foreign_load_start()` could fail without setting an error message, causing confusion in applications
- various tiny improvements


kleisauke (approved):
Nice!

PR #4686, opened 2025-09-28
add a fast uhdr detector

jcupitt:
Add a quick ultrahdr detector (just looks for an MPF block in APP2) and use it to select `uhdrload`.

With this PR I see:

```
$ vipsheader -a ultra-hdr.jpg 
width: 3840
height: 2160
bands: 3
format: uchar
coding: none
interpretation: srgb
xoffset: 0
yoffset: 0
xres: 1
yres: 1
filename: ultra-hdr.jpg
vips-loader: uhdrload
gainmap-gamma: 1 1 1 
gainmap-offset-sdr: 0 0 0 
gainmap-offset-hdr: 0 0 0 
gainmap-hdr-capacity-min: 1
gainmap-hdr-capacity-max: 100
gainmap-use-base-cg: 1

$ vipsheader -a ../k2.jpg 
../k2.jpg: 1450x2048 uchar, 3 bands, srgb, jpegload
width: 1450
height: 2048
bands: 3
format: uchar
coding: none
interpretation: srgb
xoffset: 0
yoffset: 0
xres: 2.835
yres: 2.835
filename: ../k2.jpg
vips-loader: jpegload
jpeg-multiscan: 0
jpeg-chroma-subsample: 4:2:0
exif-data: 186 bytes of binary data
resolution-unit: in
exif-ifd0-Orientation: 1 (Top-left, Short, 1 components, 2 bytes)
exif-ifd0-XResolution: 72009/1000 (72.009, Rational, 1 components, 8 bytes)
exif-ifd0-YResolution: 72009/1000 (72.009, Rational, 1 components, 8 bytes)
exif-ifd0-ResolutionUnit: 2 (Inch, Short, 1 components, 2 bytes)
exif-ifd0-YCbCrPositioning: 1 (Centred, Short, 1 components, 2 bytes)
exif-ifd2-ExifVersion: Exif Version 2.1 (Exif Version 2.1, Undefined, 4 components, 4 bytes)
exif-ifd2-ComponentsConfiguration: Y Cb Cr - (Y Cb Cr -, Undefined, 4 components, 4 bytes)
exif-ifd2-FlashpixVersion: FlashPix Version 1.0 (FlashPix Version 1.0, Undefined, 4 components, 4 bytes)
exif-ifd2-ColorSpace: 65535 (Uncalibrated, Short, 1 components, 2 bytes)
exif-ifd2-PixelXDimension: 1450 (1450, Long, 1 components, 4 bytes)
exif-ifd2-PixelYDimension: 2048 (2048, Long, 1 components, 4 bytes)
orientation: 1
```

So it successfully detects a uhdr JPG and route the load the the specialist uhdr loader. I'm not sure why the uhdr image is missing EXIF, I suppose that needs checking.

I also tried doing the routing to `uhdrload` during the `jpegload` header parse, but having a fast select function was a lot cleaner.

We'll need another PR to automatically route uhdr images to `uhdrsave` based on metadata.

### DIFF
--- a/libvips/foreign/uhdrload.c
+++ b/libvips/foreign/uhdrload.c
@@ -51,10 +51,7 @@
 #ifdef HAVE_UHDR
 
 #include "pforeign.h"
-
-const char *vips__uhdr_suffs[] = {
-	NULL
-};
+#include "jpeg.h"
 
 #include <ultrahdr_api.h>
 
@@ -101,6 +98,64 @@ typedef struct _VipsForeignLoadUhdrClass {
 
 G_DEFINE_ABSTRACT_TYPE(VipsForeignLoadUhdr, vips_foreign_load_uhdr,
 	VIPS_TYPE_FOREIGN_LOAD);
+
+/* We need a fast uhdr detector that only looks at the header. We don't want
+ * to pull the whole image in!
+ *
+ * Use vanilla libjpeg only, and just check for an MPF block in an APP2
+ * marker.
+ */
+static int
+vips_foreign_load_uhdr_is_a(VipsSource *source)
+{
+	VipsImage *context = vips_image_new();
+
+	ReadJpeg *jpeg;
+	if (!(jpeg = vips__readjpeg_new(source, context, 1, VIPS_FAIL_ON_NONE,
+			  FALSE, FALSE))) {
+		VIPS_UNREF(context);
+		return 0;
+	}
+
+	/* Here for longjmp() from vips__new_error_exit() during
+	 * cinfo->mem->alloc_small() or jpeg_read_header().
+	 */
+	if (setjmp(jpeg->eman.jmp)) {
+		VIPS_UNREF(context);
+		return 0;
+	}
+
+	if (vips__readjpeg_open_input(jpeg)) {
+		VIPS_UNREF(context);
+		return 0;
+	}
+
+	struct jpeg_decompress_struct *cinfo = &jpeg->cinfo;
+
+	/* Read JPEG header.
+	 */
+	jpeg_save_markers(cinfo, JPEG_APP0 + 2, 0xffff);
+	jpeg_read_header(cinfo, TRUE);
+
+	gboolean found;
+
+	found = FALSE;
+	for (jpeg_saved_marker_ptr p = cinfo->marker_list; p; p = p->next)
+		switch (p->marker) {
+		case JPEG_APP0 + 2:
+			if (p->data_length > 4 &&
+				vips_isprefix("MPF", (char *) p->data))
+				found = TRUE;
+			break;
+
+		default:
+			break;
+		}
+
+	VIPS_UNREF(context);
+
+	return found;
+}
 
 const char *
 vips__uhdr_error_str(uhdr_codec_err_t err)
@@ -659,6 +714,7 @@ vips_foreign_load_uhdr_class_init(VipsForeignLoadUhdrClass *class)
 {
 	GObjectClass *gobject_class = G_OBJECT_CLASS(class);
 	VipsObjectClass *object_class = (VipsObjectClass *) class;
+	VipsForeignClass *foreign_class = (VipsForeignClass *) class;
 	VipsForeignLoadClass *load_class = (VipsForeignLoadClass *) class;
 
 	gobject_class->dispose = vips_foreign_load_uhdr_dispose;
@@ -668,6 +724,10 @@ vips_foreign_load_uhdr_class_init(VipsForeignLoadUhdrClass *class)
 	object_class->nickname = "uhdrload_base";
 	object_class->description = _("load a UHDR image");
 	object_class->build = vips_foreign_load_uhdr_build;
+
+	/* We need to be higher poriority than jpegload.
+	 */
+	foreign_class->priority = 100;
 
 	load_class->get_flags = vips_foreign_load_uhdr_get_flags;
 	load_class->header = vips_foreign_load_uhdr_header;
@@ -723,12 +783,26 @@ vips_foreign_load_uhdr_file_build(VipsObject *object)
 		->build(object);
 }
 
+static gboolean
+vips_foreign_load_uhdr_file_is_a(const char *filename)
+{
+	VipsSource *source;
+
+	if (!(source = vips_source_new_from_file(filename)))
+		return FALSE;
+	gboolean is_a = vips_foreign_load_uhdr_is_a(source);
+	VIPS_UNREF(source);
+
+	return is_a;
+}
+
 static void
 vips_foreign_load_uhdr_file_class_init(VipsForeignLoadUhdrFileClass *class)
 {
 	GObjectClass *gobject_class = G_OBJECT_CLASS(class);
 	VipsObjectClass *object_class = (VipsObjectClass *) class;
 	VipsForeignClass *foreign_class = (VipsForeignClass *) class;
+	VipsForeignLoadClass *load_class = (VipsForeignLoadClass *) class;
 
 	gobject_class->set_property = vips_object_set_property;
 	gobject_class->get_property = vips_object_get_property;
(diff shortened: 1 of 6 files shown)


gregbenz:
Great to see an automated method to route gain maps through ulthdr to regain the gain map. The ultimate goal is to be able to handle simple transcodes (crop, compress, resize) and not have to do anything unique based on whether the image has a gain map or not, it should just work by default. 


Routing any image using MPF (multi-picture format) should catch any JPG using a gain map. 

However, it raises a couple questions. 

(1) it might catch a non-gain map image using MPF, perhaps an image with a depth map. I can't think of many examples, but the format predates gain maps and was intended for other use, so just a watch out. 

(2) AVIF / HEIF is not far off and have a different mechanism for encoding auxiliary images (and likely the binary metadata, though I haven't reviewed the details closely yet).

Libultrahdr has a probe utility meant to test if an image is a valid gain map, and I'm wondering if that might be a way to avoid false positives as well as help prepare for future AVIF / HEIF support (both are already in the experimental branch for libultrahdr).

jcupitt:
We can't use the libultrahdr tester since it scans the whole image -- we need a quick tester that just looks at the header. I thought about parsing the MPF attributes and looking for uhdr hints, but for now I think just looking for a MPF block is probably enough.

> The ultimate goal is to be able to handle simple transcodes (crop, compress, resize) and not have to do anything unique based on whether the image has a gain map or not, it should just work by default.

I don't think libvips itself will ever do automatic gain map changes. If they are required (by crop, for example), they will need to be handled explicitly by higher level systems, eg. vipsthumbnail or sharp. 

Copy and resize should just work since no gain map change is necessary.

> (2) AVIF / HEIF is not far off

Yes, hopefully the new libultrahdr will handle these!

gregbenz:
@jcupitt Agreed, certainly it needs to be performant. It wasn't my impression that the libultrahdr probe scans the whole image, but I haven't tested it. If it doesn't meet expectations for efficiency, it may be ideal to add a request there or submit changes in that repo - this same need will be very common for other downstream users of libultrahdr (similar questions came up with the imagemagick integration to help route images through libultrahdr only if the source has a gain map).

libultra contains functions to help crop, resize, compress, and flip/mirror the image. Any reason not to leverage those? The alteration of the map is fairly simple (and resize is already altering the map - so the key missing piece here would be cropping). The gain map metadata does not need to change (and generally should not - though you could try to optimize gamma/range of the encoding to balance quality/compression during transcoding, this is optional and probably not trivial to do well as it would be easy to create problems).

jcupitt:
We could add something to the libvips `crop` operation to also crop any attached gainmap, but then one operation would have this extra bit of hidden magic, which feels like a design disaster. We could add gainmap manipulation to all operators, but that would be a huge amount of work, and would probably be very inefficient. 

It's **much** better to leave any necessary gainmap manipulation to higher level libraries, such as sharp or vipsthumbnail. libvips can provide basic mechanisms, but other levels will need to implement whatever gainmap processing they need.

> libultra contains functions to help crop, resize, compress, and flip/mirror the image. Any reason not to leverage those?

Those implementations will not be useful for a streaming library like libvips. They will be slow and will need lots of memory.

jcupitt:
> It wasn't my impression that the libultrahdr probe scans the whole image, but I haven't tested it.

libvips needs a tester than can run very early in image detection -- well before any pixel data has appeared. Imagine an image being read from a pipe, for example, you need to spot uhdr after only a few 100 bytes.

libultrahdr has `is_uhdr_image()`, but it gets and tests all the gainmap metadata, so it will need to read in almost the entire image.

gregbenz:
@jcupitt You're the expert on how this library is used by downstream developers and I trust your judgement. 

I would just say that the scenario which sounds ideal to me is that if someone asks libvips to crop or resize, it handles the map automatically in a consistent way. In nearly every case, the correct handling of the map would be predetermined and not something a developer should alter. The only exception to that would be if someone wished to set the map to a lower scaling factor (ie go from 1:1 to 1:2 or 1:4 to trade off quality for size). Other than that, the resizing should always use the same ratio as the resizing applied to the base, and the crop should always be aligned. It would seem to me that libvips should automatically handle the map in a consistent way, unless there is a deliberate request to do otherwise for those rare cases which it might have value. There aren't many degrees of freedom here which are useful for basic transcoding of an existing gain map, but many places you could cause trouble if you need to manage all the various inputs.

The end goal being that one could update to the latest libvips and with no change in how you request it to resize/crop/compress, it should handle gain maps automatically and with highest fidelity for the requested quality factors. If that is not practical given real use of libvips, I understand. But I believe it would be ideal if the benefits of gain map support in libvips would automatically flow through to downstream users without any development work (especially so as to avoid a learning curve, as there are many ways you might make mistakes - such as not cropping the map consistent with the base, etc).

libvips/foreign/uhdrload.c
```diff
@@ -51,10 +51,7 @@
 #ifdef HAVE_UHDR
 
 #include "pforeign.h"
-
-const char *vips__uhdr_suffs[] = {
-	NULL
-};
+#include "jpeg.h"
```
lovell, line 54: Should we make `HAVE_UHDR` dependent on `HAVE_JPEG` in `meson.build`?

jcupitt:
libvips is like a lego set -- a box of small bits you can assemble to make things. It does not see the bigger picture, so it doesn't know what the source is, what the destination is, or what the purpose is. This means it can't really do good gainmap processing.

libvips is used by higher level things, like sharp and vipsthumbnail. Here, the source, destination and purpose are known very well, so they have enough information to be able to decide how to process gainmaps. That's the level at which things like gainmap crop will need to be implemented.

To give a silly example, libvips has `wrap`:

https://www.libvips.org/API/current/method.Image.wrap.html

This moves the image origin -- you use it to transform Fourier transforms so the optical axis is in the centre of the image. Should this operation make the matching change to the gainmap?

`wrap` is used during Fourier space filtering, so you'd do forward FFT, then wrap, then multiply in frequency space to filter, then unwrap, then inverse FFT. You'd end up wrapping and unwrapping the gain map for no reason :(

libvips/foreign/uhdrload.c
jcupitt: uhdrload and save already assume they have libjpeg, but you're right, we should probably make it a formal requirement.

gregbenz:
@jcupitt Thanks for the perspective/education, helpful for a newcomer. I know gain maps very well, but not libvips. Very excited to see this critical tool add support to enable those downstream tools to manage gain maps properly.

libvips/foreign/uhdrload.c
jcupitt: I added a thing to `meson.build` for this. Thanks!

libvips/foreign/uhdrload.c
```diff
@@ -102,6 +99,64 @@ typedef struct _VipsForeignLoadUhdrClass {
...
+	if (vips__readjpeg_open_input(jpeg)) {
+		VIPS_UNREF(context);
+		return 0;
+	}
+
+	struct jpeg_decompress_struct *cinfo = &jpeg->cinfo;
+
+	/* Read JPEG header.
+	 */
+	jpeg_save_markers(cinfo, JPEG_APP0 + 2, 0xffff);
```
kleisauke, line 137: Note to self: investigate use of `jpeg_set_marker_processor()` as a possible follow-up, see:
https://github.com/lovell/sharp/issues/3984#issuecomment-2317449630

kleisauke (approved):
LGTM!

libvips/foreign/uhdrload.c
jcupitt: Good idea!

jcupitt: ... lets switch to `jpeg_set_marker_processor()` everywhere in a follow-up PR.

PR #4687, opened 2025-09-30
automatically select uhdr save

jcupitt:
If jpegsave detects an image that can be saved as uhdr (either a gainmap, or an scRGB image), automatically send the save on to uhdrsave.

With this PR, UltraHDR images just work in simple cases:

```
$ vips copy ultra-hdr.jpg x.jpg
$ vipsheader x.jpg
x.jpg: 3840x2160 uchar, 3 bands, srgb, uhdrload
```

You can see the output has been detected as an uhdr image.

Thumbnail works too:

```
$ vipsthumbnail ultra-hdr.jpg
$ vipsheader tn_ultra-hdr.jpg 
tn_ultra-hdr.jpg: 128x72 uchar, 3 bands, srgb, uhdrload
```

Though the gainmap on the thumbnail has not been downsized, so it's not very efficient. It also won't crop the gainmap if you select crop, so that needs fixing.

### DIFF
--- a/libvips/foreign/jpegsave.c
+++ b/libvips/foreign/jpegsave.c
@@ -59,6 +59,9 @@
 typedef struct _VipsForeignSaveJpeg {
 	VipsForeignSave parent_object;
 
+	// set by subclasses
+	VipsTarget *target;
+
 	/* Quality factor.
 	 */
 	int Q;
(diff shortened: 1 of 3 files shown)


jcupitt:
It saves as uhdr if the image has a gainmap, or if it's scRGB (linear float sRGB 0-1, with out of range values allowed).

This means you can do (for example):

```
$ vips flip ultra-hdr.jpg[hdr] x.jpg vertical
```

For an up-down flip. With the `hdr` input flag, this will process in scRGB and regenerate the gainmap on save.

We could also select ultraHDR save for things like XYZ and LAB images, I'm not sure if that's useful.

Issue #4693, opened 2025-10-10
Specific file causes vipsthumbnail to go into an infinite loop


vansante:
## Bug report

**Describe the bug**
I run a thumbnailer service for a filestorage platform, and because a file was mis-identifiied as an image, it got fed into libvips to be thumbnailed, which causes it to go into an infinite loop, effectively becoming a DoS on the thumbnailer.

**To Reproduce**
1. Use file `loopfile` (zipped in attachment)
2. Run `vipsthumbnail topfile`
3. Observe how the process never finishes

**Expected behavior**
The process reports an invalid file format.

**Actual behavior**
An infinite loop with lots of cpu usage occurs.

**Environment**
(please complete the following information)
 - OS: `Ubuntu 24.04.3 LTS`
 - Vips: `vips-8.15.1` (my local machine, I noticed it on version `8.16.1` on the production service)

**Additional context**
I managed to cut the file down to less than 200 bytes, with the issue still occurring, but the original file was like 12k.

[loopfile.zip](https://github.com/user-attachments/files/22842919/loopfile.zip)

lovell:
Thanks for reporting, this should be fixed by https://github.com/libvips/libvips/pull/4694 and will be part of the future v8.17.3 release.

jcupitt:
8.17.3 is out now with this fix, so I'll close. Thanks for the report @vansante!

PR #4694, opened 2025-10-10
svgload: ignore short/malformed compressed input

lovell:
Suggested fix for #4693, targets the 8.17 branch.


jcupitt:
Great! We could maybe add the bad file from #4693 to the tests, it's pretty tiny.

lovell:
Thanks, good idea, test added.

kleisauke:
Looking at https://github.com/zlib-ng/zlib-ng/blob/2.2.5/test/test_small_buffers.cc#L58-L62, I think there was only a check missing for `err == Z_STREAM_END` (i.e. the input stream fully decompresses). How about applying commit https://github.com/kleisauke/libvips/commit/70d7b6ea1190dba2c5d5b4275d6b47a05d1c5925?

lovell:
Much simpler, updated, thank you. Having checked the zlib source for other possible statii this will also detect `Z_NEED_DICT`.

PR #4696, opened 2025-10-13
tile geometry metadata is not being set

jcupitt:
We added some metadata to expose tile geometry, but this is set in _load() when we add the tilecache, not in _header(), so it's not visible downstream because of lazy loading.

Add a test, fix the test in the next commit.

See https://github.com/libvips/libvips/pull/4637


jcupitt:
\o/

Issue #4702, opened 2025-10-17
dzsave with Zoomify may write wrong number of NUMTILES to ImageProperties.xml


hifi55:
Creating a Zoomify tile set, the NUMTILES property in ImageProperties.xml may report a slightly less value.

Setup
------
Running vips 8.15.1 on Windows 7 x64 from the command line, but a test with vips 8.17.2 on Windows 10 also shows this bug. CPU is a Xeon quad core.

Use a large image like this landscape (15000x5052 px):
Image: https://i.pinimg.com/originals/a7/e0/30/a7e030dd44130f6edbddb5ae725d782c.jpg

Create a custom Zoomify layout with a rather sparse tile set:
vips dzsave landscape.jpg landscapezf --layout zoomify --background 0,0,0 --skip-blanks 100 --tile-size 256 --overlap 2

The generated tile set always has **1553** tiles so ImageProperties.xml should read

<IMAGE_PROPERTIES WIDTH="15000" HEIGHT="5052" NUMTILES="**1553**" NUMIMAGES="1" VERSION="1.8" TILESIZE="256" />

however it might randomly generate slightly less NUMTILES like

<IMAGE_PROPERTIES WIDTH="15000" HEIGHT="5052" NUMTILES="**_1552_**" NUMIMAGES="1" VERSION="1.8" TILESIZE="256" />

--------------------

The source uses tile_count to track the NUMTILES:

		/* And count tiles so far in this level.
		 */
		n += y * level->tiles_across + x;

		vips_snprintf(subdir, VIPS_PATH_MAX, "TileGroup%d", n / 256);
		vips_snprintf(name, VIPS_PATH_MAX,
			"%d-%d-%d%s", level->n, x, y, dz->file_suffix);

		/* Used at the end in ImageProperties.xml
		 */
		dz->tile_count += 1;

and writes the output with
	vips_dbuf_writef(&dbuf, "<IMAGE_PROPERTIES "
							"WIDTH=\"%d\" HEIGHT=\"%d\" NUMTILES=\"%d\" "
							"NUMIMAGES=\"1\" VERSION=\"1.8\" "
							"TILESIZE=\"%d\" />\n",
		dz->level->width,
		dz->level->height,
		dz->tile_count,
		dz->tile_size);

That seems quite straightforward, so perhaps there is some race condition in the multicore processing that causes this glitch.

kleisauke:
[...]

PR #4704, opened 2025-10-20
dzsave: Zoomify: ensure correct tile count in ImageProperties.xml

kleisauke:
Resolves: #4702.

Targets the 8.17 branch.

### DIFF
--- a/libvips/foreign/dzsave.c
+++ b/libvips/foreign/dzsave.c
@@ -303,7 +303,7 @@ struct _VipsForeignSaveDz {
 
 	/* Count zoomify tiles we write.
 	 */
-	int tile_count;
+	int tile_count; // (atomic)
 
 	/* Where we write ... can be the filesystem, or a zip.
 	 */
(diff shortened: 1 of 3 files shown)


kleisauke:
It seems the CI Codespell job failed as a result of the AWS outage, see: https://health.aws.amazon.com/health/status.
> ```
> ERROR: failed to authorize: failed to fetch oauth token: unexpected status from POST request to https://auth.docker.io/token: 503 Service Unavailable: <html><body><h1>503 Service Unavailable</h1>
> ```

libvips/foreign/dzsave.c
```diff
@@ -303,7 +303,7 @@ struct _VipsForeignSaveDz {
 
 	/* Count zoomify tiles we write.
 	 */
-	int tile_count;
+	int tile_count; // (atomic)
```
lovell, line 306: Does this need to be marked as `volatile`? The advice at https://docs.gtk.org/glib/func.atomic_int_get.html is a little ambiguous.

kleisauke: I don't think it needs to be marked as `volatile`. For reference, see:
https://gitlab.gnome.org/GNOME/glib/-/commit/fab561f8d05794329184cd81f9ab9d9d77dcc22a (and [the useful site](https://isvolatileusefulwiththreads.in/c/) it links :smile:)
https://gitlab.gnome.org/GNOME/glib/-/commit/83e48d8ac1fee98059e2305d8909dca26190bddc

lovell: Ha ha, thank you, the history of `volatile` in C has been, er, volatile.

PR #4707, opened 2025-10-21
fix VipsSaveable with c++

jcupitt:
C++ is much stricter than C about enum assignment, so building a c++ program that assigned VIPS_SAVEABLE_RGB to a VipsForeignSaveable would fail with an error.

Reimplement VipsSaveable just using macros.

Thanks @scossu

See https://github.com/harvard-lts/kakadu-vips/issues/23


kleisauke (approved):
LGTM!

PR #4708, opened 2025-10-25
add "gainmap" to VipsKeep

jcupitt:
And revise the logic in vips_foreign_save_remove_metadata().

### DIFF
--- a/libvips/foreign/foreign.c
+++ b/libvips/foreign/foreign.c
@@ -1712,36 +1712,60 @@ vips__foreign_convert_saveable(VipsImage *in, VipsImage **ready,
 	return 0;
 }
 
+/* What VipsKeep does a metadata item belong to?
+ *
+ * Return 0 (KEEP_NONE) for never remove, ie. an unrecognised field, or
+ * one which is used to control saver behaviour, like "delay" or
+ * "orientation".
+ */
+static VipsForeignKeep
+vips_foreign_get_keep(const char *field)
+{
+	if (g_str_equal(field, VIPS_META_EXIF_NAME) ||
+		// derived from exif
+		vips_isprefix("exif-", field) ||
+		// also comes from exif
+		vips_isprefix("jpeg-thumbnail-data", field))
+		return VIPS_FOREIGN_KEEP_EXIF;
+
+	if (g_str_equal(field, VIPS_META_XMP_NAME))
+		return VIPS_FOREIGN_KEEP_XMP;
+
+	if (g_str_equal(field, VIPS_META_IPTC_NAME))
+		return VIPS_FOREIGN_KEEP_IPTC;
+
+	if (g_str_equal(field, VIPS_META_ICC_NAME))
+		return VIPS_FOREIGN_KEEP_ICC;
+
+	if (vips_isprefix("gainmap", field))
+		return VIPS_FOREIGN_KEEP_GAINMAP;
+
+	/* OTHER is a metadata item that:
+	 *
+	 *	- one or more savers will write to a format
+	 *	- is optional
+	 *	- is not used to control saver behaviour
+	 */
+	if (vips_isprefix("png-comment-", field) ||
+		g_str_equal(field, VIPS_META_IMAGEDESCRIPTION) ||
+		g_str_equal(field, VIPS_META_PHOTOSHOP_NAME) ||
+		vips_isprefix("magickprofile-", field) ||
+		vips_isprefix("magick-", field))
+		return VIPS_FOREIGN_KEEP_OTHER;
+
+	return 0;
+}
+
 static void *
 vips_foreign_save_remove_metadata(VipsImage *image,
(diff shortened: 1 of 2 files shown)


jcupitt:
Test with eg.:

```
$ vips copy ultra-hdr.jpg[hdr] x.v[keep=gainmap]
$ vipsheader -a x.v
x.v: 3840x2160 float, 3 bands, scrgb
width: 3840
height: 2160
bands: 3
format: float
coding: none
interpretation: scrgb
xoffset: 0
yoffset: 0
xres: 1
yres: 1
filename: x.v
gainmap-gamma: 1 1 1 
gainmap-offset-sdr: 0 0 0 
gainmap-offset-hdr: 0 0 0 
gainmap-hdr-capacity-min: 1
gainmap-hdr-capacity-max: 100
gainmap-use-base-cg: 1
$
```

So EXIF has gone, but the gainmap metadata is still there.

I also revised the logic for OTHER, the old code was very confusing to me and I'm not sure if it worked. It ought to be safer now. I hope I've not become confused.

I removed the test on the result of `vips_image_remove()` since the FALSE return is not an error, just a note that there was no metadata of that name removed. For example, it'll return FALSE for `"width"` and not set an error.

libvips/foreign/foreign.c
```diff
@@ -1718,26 +1718,30 @@ vips_foreign_save_remove_metadata(VipsImage *image,
...
-		(strcmp(field, VIPS_META_ICC_NAME) == 0 &&
+		(g_str_equal(field, VIPS_META_ICC_NAME) &&
 			(keep & VIPS_FOREIGN_KEEP_ICC)) ||
-		(keep & VIPS_FOREIGN_KEEP_OTHER))
+		(vips_isprefix("gainmap", field) &&
+			(keep & VIPS_FOREIGN_KEEP_GAINMAP)))
+		return NULL;
+
+	/* OTHER is everything that doesn't match one of the above tests.
+	 */
```
kleisauke: I think(?) the original intent of that old check was to define an allow list of which metadata should be classified as "other".

For example, when saving with `keep=none`, this would now also remove `"delay"`, `"loop"`, `"resolution-unit"`, and others (grep for `vips_image_get*` in the savers), which affect saver behavior rather than representing metadata within the image.

libvips/foreign/foreign.c
```diff
@@ -1718,26 +1718,30 @@ vips_foreign_save_remove_metadata(VipsImage *image,
...
+		(g_str_equal(field, VIPS_META_XMP_NAME) &&
 			(keep & VIPS_FOREIGN_KEEP_XMP)) ||
-		(strcmp(field, VIPS_META_IPTC_NAME) == 0 &&
+		(g_str_equal(field, VIPS_META_IPTC_NAME) &&
 			(keep & VIPS_FOREIGN_KEEP_IPTC)) ||
-		(strcmp(field, VIPS_META_ICC_NAME) == 0 &&
+		(g_str_equal(field, VIPS_META_ICC_NAME) &&
 			(keep & VIPS_FOREIGN_KEEP_ICC)) ||
-		(keep & VIPS_FOREIGN_KEEP_OTHER))
+		(vips_isprefix("gainmap", field) &&
```
kleisauke: How about?:
```suggestion
		(g_str_equal(field, "gainmap") &&
```
(I'm not sure if multiple gain maps are possible)

jcupitt: There's quite a bit of gainmap metadata, eg.:

```
$ vipsheader -a ultra-hdr.jpg 
width: 3840
height: 2160
bands: 3
format: uchar
coding: none
interpretation: srgb
xoffset: 0
yoffset: 0
xres: 1
yres: 1
filename: ultra-hdr.jpg
vips-loader: uhdrload
gainmap-gamma: 1 1 1 
gainmap-offset-sdr: 0 0 0 
gainmap-offset-hdr: 0 0 0 
gainmap-hdr-capacity-min: 1
gainmap-hdr-capacity-max: 100
gainmap-use-base-cg: 1
$ 
```

So the `isprefix` gets them all.

libvips/foreign/foreign.c
jcupitt: Ah you're right, I'd not thought of that. Perhaps we need a table of libvips metadata names with a `VipsKeep` flag for each one? It's a bit ad-hoc right now.

libvips/foreign/foreign.c
kleisauke: Ah, indeed. Thanks for the clarification.

libvips/foreign/foreign.c
kleisauke: Good idea! Using a table would be neater.

jcupitt: I'll have a go tomorrow.

jcupitt:
I fiddled with it a bit and a function to classify metadata fields seemed neater. I also looked though the savers and found a few more items we should keep or remove.

PR #4713, opened 2025-10-30
OSS-Fuzz: add workaround for llvm/llvm-project#163369

kleisauke:
In line with commit 023a0f48bf187dc8903cb76c02f64af6012b0edf.

Context: https://github.com/google/oss-fuzz/issues/5272#issuecomment-3456966272.

### DIFF
--- a/fuzz/oss_fuzz_build.sh
+++ b/fuzz/oss_fuzz_build.sh
@@ -102,6 +102,10 @@ popd
 
 # libspng
 pushd $SRC/libspng
+# Workaround for https://github.com/llvm/llvm-project/issues/163369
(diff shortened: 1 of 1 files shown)


kleisauke:
Tested with:
```console
$ python infra/helper.py build_fuzzers --sanitizer thread libvips > build_log.txt
$ mkdir -p tmp
$ python infra/helper.py run_fuzzer --corpus-dir=$PWD/tmp/ libvips generic_buffer_with_args_fuzzer -- -max_total_time=1800
```

The idea behind this is to enable TSan in CIFuzz as well, as most dependencies are built from source there (which ensures they also include TSan instrumentation).

jcupitt:
Nice!

Issue #4673, opened 2025-09-18
How can I output files relative to cwd instead of file location using vipsthumbnail?


zjturner:
I want something that is kind of a mix of the absolute and relative output modes.  I'm using a build system that does not expose absolute paths to output artifacts, so it's impossible for me to specify an absolute path.  At the same time, it also does not allow writing to the source tree, it's treated as read only.  So I also can't output my processed images relative to the input file.

It sandboxes the output folder and operates in terms of relative paths from the build process's working directory.  So it might have the process cwd set to something like `D:\build-output-sandbox` and it wants to write to `out\target\foo.png`.  It's able to pass the build tool a path such as that, and it expects the output to appear at `D:\build-output-sandbox\out\target\foo.png`.  But there appears to be no way to do this.

I'm willing to make a PR for this, but I don't know how likely something like this is to be accepted or what kind of design you'd accept.  For example I could imagine an `--relative-output-mode=[input_relative|cwd_relative]` argument, but that seems kind of clunky.  Thoughts?

zjturner:
Actually, an even better solution perhaps is to allow the command line parser to interpolate `%d` instead of just `%s` like it does now.    `%d` would be replaced with the current process working directory before the program looks to see whether an absolute or relative path is used.  This way, I can simply modify my command line to `--output %d/%s.png` and it will behave exactly as I need it to behave, and the interface meshes well with what's already there.

Would you accept this patch?

jcupitt:
Hi @zjturner, 

We could also allow more complex relative paths, would that work?

Right now, `vipsthumbnail` has a (pretty crude) thing to detect output to `.suffix` (used for sending files to stdout). It gets triggered too easily and stops outputs like `../a/%s.jpg` from working.

If we change it to be:

```diff
diff --git a/tools/vipsthumbnail.c b/tools/vipsthumbnail.c
index 852cde250..b759a5638 100644
--- a/tools/vipsthumbnail.c
+++ b/tools/vipsthumbnail.c
@@ -359,10 +359,11 @@ thumbnail_process(VipsObject *process, const char *name)
        /* If the output format is something like ".jpg", we write to stdout
         * instead.
         *
-        * (but allow "./%s.jpg" as a output format)
+        * (but allow "./%s.jpg", "../a/%s.jpg", etc. as a output format)
         */
-       if (vips_isprefix(".", output_format) &&
-               !vips_isprefix("./", output_format)) {
+       if (output_format[0] == '.' &&
[...]

PR #4714, opened 2025-10-31
add --path option to vipsthumbnail

jcupitt:
This PR deprecates the old "output" option with its strange absolute / relative pathname behaviour and adds a new "path" option.

This new option allows three substitution characters: basename without suffix, dirname, and cwd. For example:

```
vipsthumbnail a/b/c.jpg --path %c/%d/%s.tif
```

- `%c` is whatever `cwd` is, so an absolute path, perhaps `/d/e/f`
- `%d` would be `a/b`
- `%s` would be `c`

In this case, the thumbnail would be written to `/d/e/f/a/b/c.tif`.

See https://github.com/libvips/libvips/issues/4673

Thanks to @zjturner for the suggestion.

### DIFF
--- a/tools/vipsthumbnail.c
+++ b/tools/vipsthumbnail.c
@@ -314,8 +314,6 @@ thumbnail_output_format(void)
 
 /* Given (eg.) "/poop/somefile.png", write @im to the thumbnail name,
  * (eg.) "/poop/tn_somefile.jpg".
- *
- * If
  */
 static int
 thumbnail_write_file(VipsObject *process, VipsImage *im, const char *filename)


libvips/include/vips/internal.h
```diff
@@ -311,7 +311,7 @@ guint64 vips__parse_size(const char *size_string);
 /* TODO(kleisauke): VIPS_API is required by vipsthumbnail.
  */
 VIPS_API
-int vips__substitute(char *buf, size_t len, char *sub);
+int vips__substitute(char *buf, size_t len, char c, char *sub);
```
kleisauke: Despite being a private API (indicated by the double underscore), I think `abi-compliance-checker` will still complain about this change (because the `VIPS_API` macro is applied to this function).

tools/vipsthumbnail.c
```diff
@@ -235,54 +240,95 @@ static GOptionEntry options[] = {
...
-	g_free(file);
+	return NULL;
+}
 
-	if (vips_image_write_to_file(im, output_name, NULL)) {
-		g_free(output_name);
+/* Given (eg.) "/poop/somefile.png", write @im to the thumbnail name,
+ * (eg.) "/poop/tn_somefile.jpg".
+ *
+ * If
```
kleisauke: nit: there's a stray comment here (seems to be pre-existing).
```suggestion
```

libvips/include/vips/internal.h
jcupitt: Ah good point. Fixed!

kleisauke (approved):
LGTM!

jcupitt:
Thanks for reviewing!

Issue #4692, opened 2025-10-09
vipsthumbnail: 8.17 size at least


esiqveland:
vipsthumbnail: 8.17 size

There seems to be some weird behaviour with sizing of this image.

`VIPS_INFO=1 vipsthumbnail --size='1920>' -o balloon.jpg balloon.jp2`

I would expect one of the dimensions to be roughly 1920 in size, but the result is:

```
$ vipsheader balloon.jpg
balloon.jpg: 85x116 uchar, 3 bands, srgb, jpegload
```

Steps to reproduce:

```
wget https://github.com/bitsgalore/jp2kMagic/raw/refs/heads/master/sampleImages/balloon.jp2

VIPS_INFO=1 vipsthumbnail --size '1920>' -o balloon.jpg balloon.jp2
VIPS-INFO: 11:17:34.467: found /opt/homebrew/Cellar/vips/8.17.2/lib/vips-modules-8.17
VIPS-INFO: 11:17:34.472: using configure-time prefix
VIPS-INFO: 11:17:34.472: VIPS_PREFIX = /opt/homebrew/Cellar/vips/8.17.2
VIPS-INFO: 11:17:34.472: VIPS_LIBDIR = /opt/homebrew/Cellar/vips/8.17.2/lib
VIPS-INFO: 11:17:34.472: prefix = /opt/homebrew/Cellar/vips/8.17.2
VIPS-INFO: 11:17:34.472: libdir = /opt/homebrew/Cellar/vips/8.17.2/lib
VIPS-INFO: 11:17:34.472: searching "/opt/homebrew/Cellar/vips/8.17.2/lib/vips-modules-8.17"
VIPS-INFO: 11:17:34.472: loading "/opt/homebrew/Cellar/vips/8.17.2/lib/vips-modules-8.17/vips-heif.dylib"
VIPS-INFO: 11:17:34.482: loading "/opt/homebrew/Cellar/vips/8.17.2/lib/vips-modules-8.17/vips-magick.dylib"
VIPS-INFO: 11:17:34.485: loading "/opt/homebrew/Cellar/vips/8.17.2/lib/vips-modules-8.17/vips-poppler.dylib"
VIPS-INFO: 11:17:34.489: loading "/opt/homebrew/Cellar/vips/8.17.2/lib/vips-modules-8.17/vips-openslide.dylib"
[...]

PR #4715, opened 2025-10-31
fix vipsthumbnail pyr layer selection

jcupitt:
A >= rather than > meant we were always picking the smallest pyramid layer, not the best.

Test with:

```
vips copy any-image.jpg x.tif[tile,pyramid,compression=jpeg]
vipsthumbnail x.tif --size=1920 -o x.jpg --vips-info
```

Before you'll see eg.:

```
VIPS-INFO: 16:19:40.970: loading with factor 6 pre-shrink
VIPS-INFO: 16:19:40.970: pre-shrunk size is 68 x 45
```

After perhaps:

```
VIPS-INFO: 16:38:23.060: input size is 4368 x 2912
VIPS-INFO: 16:38:23.076: loading pyramid page 1
VIPS-INFO: 16:38:23.076: pre-shrunk size is 2184 x 1456
```

ie. it's now selecting the correct pyr level.

See https://github.com/libvips/libvips/issues/4692

Thanks @esiqveland

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+8.17.4
+
+- vipsthumbnail: fix pyramidal image layer selection [esiqveland]
+
 30/10/25 8.17.3
 
 - tiffsave: fix saved resolution with a non-default resolution-unit metadata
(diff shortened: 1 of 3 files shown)


kleisauke (approved):
LGTM! This looks similar to the fix in commit 6f0548dc5c272f02481f0a201bbf086cc1152b15.

(I've fixed a small typo in the PR title and description, hope that's OK).

jcupitt:
> LGTM! This looks similar to the fix in commit https://github.com/libvips/libvips/commit/6f0548dc5c272f02481f0a201bbf086cc1152b15.

Exactly! I noticed that too. We probably should have spotted this bug at the same time :( 

Thanks for the review + patches Kleis.

PR #4717, opened 2025-11-03
call terminateprocess to exit

jcupitt:
clang on win10 can generate exes which occasionally deadlock on process exit, see https://github.com/llvm/llvm-project/issues/110954

on windows, call TerminateProcess instead to bypass orderly shutdown


jcupitt:
I added an internal `VIPS_API void vips__win32_terminate(int ret);` to get all this craziness in one place.

Issue #3912, opened 2024-03-27
HDR AVIF input images are outputted as flat and desaturated


mertalev:
## Bug report

 HDR AVIF input images look very desaturated in the output image. This happens for both `srgb` and `rgb16` pipeline colorspaces and both `srgb` and `p3` target ICC profiles. It also occurs when outputting to any of JPEG, WebP, AVIF or PNG (these are the outputs formats I tested).

**To Reproduce**
1. Use the raw0016-positive.avif image from [this](https://github.com/immich-app/immich/files/14773511/avif.zip) zip
2. Convert to any of JPEG, WebP or PNG
3. Note the desaturated colors

**Expected behavior**

When outputting to JPEG or WebP, the colors should be somewhat representative of the input image (minus the HDR, of course). When outputting to 10-bit AVIF, it should have roughly the same colors including HDR.

**Actual behavior**

The output image is unnaturally desaturated.

**Screenshots**

Apple Preview:

![raw0016-positive-preview](https://github.com/libvips/libvips/assets/101130780/2923a843-9844-4e85-9940-acfee2514c43)

libvips:

![immich-thumbnail](https://github.com/libvips/libvips/assets/101130780/ba2efb95-8784-42aa-a491-7dfffc14ec52)

**Environment**
(please complete the following information)
 - OS: Debian Bookworm
 - Vips: 8.15.2
 - sharp 0.33.3
 - libheif 1.15.1 (also tried a build with 1.17.6 with same results)

**Additional context**

libheif has a convert_hdr_to_8bit setting that's referenced [here](https://github.com/libvips/libvips/blob/89928220eb2e745cd941fc56343474212babc9f2/meson.build#L518), but doesn't seem to be used anywhere.

jcupitt:
Hi @mertalev,

You two images above ^^ look identical on firefox on linux, but different in firefox on macos, interestingly.

libvips uses ICC profiles for colour management, but your image has no ICC profile. I would guess it is using nclx for encoding HDR, and that is (as far as I know) only supported by heif. I tried in imagemagick6 and 7, and they both gave the same result as libvips.

We could add nclx support, but that would be a huge amount of work. Maybe you could save your image with an ICC profile? It might at least be less work to support.

jcupitt:
... you're right, we could also add an `sdr` (for example) option to `heifload` which enabled `convert_hdr_to_8bit`. That might be a simple fix.

mertalev:
It does apparently use nclx:
```
[...]

PR #4723, opened 2025-11-10
[WIP] Create synthetic ICC profile from nclx tag in libheif

Starbix:
This is a WIP PR trying to fix https://github.com/libvips/libvips/issues/3912.

HDR AVIF (and most likely HEIF) files that lack an ICC profile signal their interpretation via CICP values (as defined in H.273) in the nclx field. libvips ignores this information, leading to very desaturated colors.

This PR attempts to create an ICC profile from the CICP. Creating a simple parametric curve ICC profile does not work for PQ/HLG as the curves defined by lcms2 aren't able to cover PQ or HLG. Thus in this PR I simply applied ITU-R BT.2020-2 curves.
image

This doesn't solve the wrong brightness. Defined in ICC v4.4 there is a CICP tag which can signal the exact transfer function.  This is interpreted by Chromium and macOS Preview (I didn't test other clients). Unfortunately, this tag is not supported by lcms2: (https://github.com/mm2/Little-CMS/issues/439). Meaning in case of an `icc_transform` that HDR information is lost again. But at least it makes it already easier for downstream clients (such as Immich) to handle HDR from libheif more easily (e.g. by not doing an icc_transform)
In a second step we should add a tonemapping curve instead so clients (such as libvips itself) that don't support the cicp tag in the ICC profile can still produce a decent looking image. relevant: https://issues.chromium.org/issues/40239687, https://github.com/discord/lilliput/pull/218

I only tested this with a P3 PQ HDR file created by Adobe Lightroom.
I have attached three files: the original HDR AVIF, the output JPG by upstream libvips, and the JPG with an HDR ICC profile.
[examples.zip](https://github.com/user-attachments/files/23455197/examples.zip)

### DIFF
--- a/libvips/foreign/heifload.c
+++ b/libvips/foreign/heifload.c
@@ -28,6 +28,8 @@
  * 	- add @unlimited
  * 13/03/23 MathemanFlo
  * 	- add bits per sample metadata
+ * 10/11/25 Starbix
+ * - read CICP tag and create ICC profile
  */
 
 /*
@@ -61,7 +63,7 @@
 #define DEBUG_VERBOSE
 #define VIPS_DEBUG
 #define DEBUG
- */
+*/
 
 #ifdef HAVE_CONFIG_H
 #include <config.h>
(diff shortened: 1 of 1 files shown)


jcupitt:
Hello @Starbix, I thought I'd repeat something said in chat so we have a record.

Another possible approach (which I think is maybe what you were proposing in chat) could be to change `heifload` to tag the image with the NCLX metadata, somewhat in the way `uhdrload` does.

For example, with git master libvips you get:

```
$ vipsheader -a ultra-hdr.jpg
width: 3840
height: 2160
bands: 3
format: uchar
coding: none
interpretation: srgb
xoffset: 0
yoffset: 0
xres: 1
yres: 1
filename: ultra-hdr.jpg
vips-loader: uhdrload
gainmap-data: 31738 bytes of binary data
gainmap-gamma: 1 1 1 
gainmap-offset-sdr: 0 0 0 
gainmap-offset-hdr: 0 0 0 
gainmap-hdr-capacity-min: 1
gainmap-hdr-capacity-max: 100
gainmap-use-base-cg: 1
```

ie. there are a set of extra metadata tags holding the transform. On `uhdrsave`, these are used to set the metadata in the output file.

We could do something similar for heifload (and other loaders and savers which support nclx, like jxl), and add a new operation (maybe called `nclx2scRGB`?) which applied the nclx metadata to the image, computing scRGB (3 x float bands, 0-1 linear for black to white, out of range values for HDR). This would let us feed these images into an ICC-profile-managed workflow.

We could also add `VIPS_INTERPRETATION_NCLX` and treat it like any other colourspace. That would let things like:

```
$ vips uhdrsave x.avif y.jpg
```

work too, which might be useful (the gainmap would be generated automatically).

What do you think? It might be simpler than trying to get lcms to accept an HDR profile reliably.

Starbix:
Yes that sounds like a very sensible approach that doesn't abuse ICC. I'd need to further understand the inner workings of lcms2 to try some approach (and also investigate how to map to scRGB). But the code should probably live somewhere else than heifload.c.

jcupitt:
Great! Let's try this approach.

jcupitt:
We need to get libvips 8.18 out of the door pretty soon, let's move this to the 8.19 cycle in the first half of nest year.

Starbix:
Ah yeah definitely, I don't have that much time for the rest of the semester. I'll try to work on it once I have more capacity.

jcupitt:
Great! I'll have some time after June next year.

Starbix:
Superseded by #4818

PR #4728, opened 2025-11-15
uhdrload: return early if source is not JPEG

kleisauke:
Test case:
```c
/* Compile with:
 */

#include <vips/vips.h>

int
main(int argc, char *argv[])
{
	if (VIPS_INIT(argv[0]))
		vips_error_exit(NULL);

	VipsSource *source;
		return 0;

	return 0;
}
```

Before:
```console
$ gcc -g -Wall test.c `pkg-config vips --cflags --libs`
$ ./a.out

(process:301493): VIPS-WARNING **: 11:39:52.417: read gave 1 warnings

(process:301493): VIPS-WARNING **: 11:39:52.417: vips_mapfile: unable to mmap
system error: Invalid argument
VipsJpeg: premature end of JPEG image
VipsJpeg: Not a JPEG file: starts with 0xff 0xd9
```

After:
```console
$ gcc -g -Wall test.c `pkg-config vips --cflags --libs`
$ ./a.out
<EMPTY OUTPUT />
```

### DIFF
--- a/libvips/foreign/uhdrload.c
+++ b/libvips/foreign/uhdrload.c
@@ -108,6 +108,9 @@ G_DEFINE_ABSTRACT_TYPE(VipsForeignLoadUhdr, vips_foreign_load_uhdr,
 static int
 vips_foreign_load_uhdr_is_a(VipsSource *source)
 {
+	if (!vips__isjpeg_source(source))
+		return 0;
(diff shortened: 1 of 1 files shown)


jcupitt:
Ah, nice find.

PR #4733, opened 2025-11-19
expose the exact parameter for webp saving

yoshiyoshyosh:
I sometimes work with "images-as-data" using vips, such as for texture masking or metallic maps in game engines. webp is the format that allows the best file-size compression while being supported by all the programs I use

when alpha is part of the data, though, fully-transparent pixels will be messed with for better compression as part of libwebp's default settings. this, of course, messes up the data.

libwebp has an "exact" parameter to keep the color data from transparent pixels. I think it'd be a good fit to expose in vips.

### DIFF
--- a/libvips/foreign/webpsave.c
+++ b/libvips/foreign/webpsave.c
@@ -112,6 +112,10 @@ typedef struct _VipsForeignSaveWebp {
 	 */
 	gboolean lossless;
 
+	/* Preserve color values from transparent pixels.
+	 */
+	gboolean exact;
+
 	/* Lossy compression preset.
 	 */
 	VipsForeignWebpPreset preset;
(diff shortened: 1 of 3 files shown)


libvips/foreign/webpsave.c
```diff
@@ -839,6 +844,13 @@ vips_foreign_save_webp_class_init(VipsForeignSaveWebpClass *class)
 		G_STRUCT_OFFSET(VipsForeignSaveWebp, lossless),
 		FALSE);
 
+	VIPS_ARG_BOOL(class, "exact", 12,
```
jcupitt, line 847: The numbers are supposed to be unique -- could you make this 24, or (if you can be bothered) renumber the args?

jcupitt (approved):

jcupitt:
Hi @yoshiyoshyosh, sure, this sounds like a good idea. I saw one tiny thing.

libvips/foreign/webpsave.c
yoshiyoshyosh: sounds good. there was another non-unique number in there (smart_deblock vs passes), so I reordered them based on how they show up in the CLI already if that's fine.

jcupitt:
Thanks for adding this feature, @yoshiyoshyosh!

PR #4744, opened 2025-11-24
ppmsave: cast to target format when colourspace is unsupported

kleisauke:
After commit 64b145578fb1cb1c67b1187b89e5327ca603901d, the `vips_colourspace()` operation is only applied when conversion to the target colourspace is actually possible. If the colourspace is unsupported, we still need to perform a cast. For example, saving a `MULTIBAND` or `MATRIX` float image as PPM still requires casting to uchar.

This partially reverts commit c9086c2f4ebefddfa086c26fc54727a5fd7c0f08. 

Targets the 8.17 branch.

### DIFF
--- a/libvips/foreign/ppmsave.c
+++ b/libvips/foreign/ppmsave.c
@@ -223,6 +223,7 @@ vips_foreign_save_ppm_build(VipsObject *object)
 	VipsImage *image;
 	char *magic;
 	char *date;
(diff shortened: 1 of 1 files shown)


kleisauke:
A variant for `pngsave` is available in #4753.

jcupitt (approved):

PR #4745, opened 2025-11-24
Add dzsave gainmap

jcupitt:
TODO:

- [x] add some tests

### DIFF
--- a/libvips/foreign/dzsave.c
+++ b/libvips/foreign/dzsave.c
@@ -2135,7 +2135,7 @@ vips_foreign_save_dz_build(VipsObject *object)
 		 */
 		dz->skip_blanks = -1;
 
-		/* Direct mode soes not support gainmaps.
+		/* Direct mode does not support gainmaps.
 		 */
 		dz->direct = FALSE;
 	}


jcupitt:
Test with:

```console
$ dzsave ultra-hdr.jpg x --keep gainmap
$ vipsheader -f gainmap-data x_files/10/1_0.jpeg | base64 -d > map-tile.jpg
```

And verify that `map-tile.jpg` is the expected size and crop position. `test_uhdr_dzsave` does this test.

You need the `--keep gainmap` since dzsave strips all tile metadata by default, but of course you need to retain it here.

libvips/foreign/dzsave.c
```diff
@@ -2092,6 +2122,33 @@ vips_foreign_save_dz_build(VipsObject *object)
...
+		g_object_ref(dz->gainmap);
+
+		dz->gainmap_hscale = (double) dz->gainmap->Xsize / save->ready->Xsize;
+		dz->gainmap_vscale = (double) dz->gainmap->Ysize / save->ready->Ysize;
+
+		/* Don't check for blanks, too annoying with a gainmap as well.
+		 */
+		dz->skip_blanks = -1;
+
+		/* Direct mode soes not support gainmaps.
```
kleisauke:
```suggestion
		/* Direct mode does not support gainmaps.
```

PR #4748, opened 2025-11-27
add uhdr2scRGB

jcupitt:
Transform uhdr (8-bit RGB with a gainmap) into scRGB.

TODO 

- [x] add tests

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -176,12 +176,21 @@ if meson.can_run_host_binaries()
 else
     have_target_clones = cc.links(target_clones_check, args: '-Werror', name: 'Has target_clones attribute')
 endif
-# ThreadSanitizer is incompatible with the target_clones attribute
-# See: https://github.com/llvm/llvm-project/issues/163369
-if have_target_clones and get_option('b_sanitize') == 'thread'
-    have_target_clones = false
+# `target_clones` attribute is incompatible with:
+# - ThreadSanitizer on Clang and GCC
+#   https://github.com/llvm/llvm-project/issues/163369
+#   https://gcc.gnu.org/PR122894
+# - Profile-guided optimization on GCC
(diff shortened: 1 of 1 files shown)


jcupitt:
It seems to work, but the result is about 30% too bright, I'm not sure why. I'll keep digging ...

Test with:

```
$ vips uhdr2scRGB ultra-hdr.jpg x.v
$ vips copy ultra-hdr.jpg[hdr] x2.v
```

And hopefully they should match (but they don't, right now).

libvips/include/vips/colour.h
```diff
@@ -203,6 +203,9 @@ VIPS_API
 int vips_XYZ2CMYK(VipsImage *in, VipsImage **out, ...)
 	G_GNUC_NULL_TERMINATED;
 
+VIPS_API
+int vips_uhdr2scRGB(VipsImage *in, VipsImage **out, ...);
+
```
kleisauke, line 209:
```suggestion
	G_GNUC_NULL_TERMINATED

```

jcupitt:
Ah, found it, we just need to not gamma-correct the gainmap pixels.

libvips is quite a bit quicker than libultrahdr, interestingly. These are exactly equivalent, but the `uhdr2scRGB` case does the gainmap processing in libvips:

```
$ time vips copy ultra-hdr.jpg[hdr] x.v
real	0m0.402s
user	0m0.527s
sys	0m0.129s

$ time vips uhdr2scRGB ultra-hdr.jpg x2.v
real	0m0.276s
user	0m0.366s
sys	0m0.103s
```

The libultrahdr code is here:

https://github.com/google/libultrahdr/blob/main/lib/src/jpegr.cpp#L1595-L1680

They've not done much tuning.

PR #4750, opened 2025-11-29
Regenerate POTFILES.in, C++ binding and function list

kleisauke:

PR #4751, opened 2025-11-29
add uhdr docs

jcupitt:
Add a chapter to the docs introducing uhdr.

Draft for now since it raises a few questions:

- add scRGB2uhdr .. inverse of uhdr2scRGB? only works if there's a gainmap on the scRGB image
- don't use libuhdr linear import, just call uhdr2scRGB on result
- call uhdr2scRGB automatically from sRGB2scRGB
- uhdr should be a supported colourspace? not clear how this would interact with nclx, best to leave it
- verify that uhdrsave will only use `gainmap-data` if `gainmap` is missing
- saving scRGB as uhdr always recomputes the gainmap, is this the best behaviour?

### DIFF
--- a/doc/uhdr.md
+++ b/doc/uhdr.md
@@ -63,7 +63,7 @@ write with the [method@Image.uhdrsave] operation, embedding the gainmap and the
 associated metadata in the output image.
 
 Intermediate operations which change the image geometry will also need to
-update the `"gainmap-data" metadata item, the mechanisms for doing this are
+update the `"gainmap-data"` metadata item, the mechanisms for doing this are
 described below. The other gainmap fields should probably not be changed
 unless the intention is to alter the image appearance.
 


jcupitt:
Sample formatted docs: http://www.rollthepotato.net/~john/vips/uhdr.html

doc/function-list.md
kleisauke, line 1: This file will be regenerated via PR #4750.

doc/uhdr.md
```diff
@@ -0,0 +1,208 @@
...
+gainmap-use-base-cg: 1
+```
+
+The gainmap metadata is copied unmodified through any processing operations.
+If you save an image with gainmap metadata to a JPEG file, libvips will do the
+write with the [method@Image.uhdrsave] operation, embedding the gainmap and the
+associated metadata in the output image.
+
+Intermediate operations which change the image geometry will need to also
+update the `"gainmap0-data" metadata item, the mechanisms for doing this are
```
kleisauke:
```suggestion
update the `"gainmap-data" metadata item, the mechanisms for doing this are
```

doc/function-list.md
jcupitt: Ah good point, I'll remove this.

jcupitt:
I updated the docs for the removal of the `hdr` argument:

http://www.rollthepotato.net/~john/vips/uhdr.html

Any thoughts? I think it seems a bit less crazy now.

doc/uhdr.md
```diff
@@ -0,0 +1,158 @@
...
+gainmap-use-base-cg: 1
+```
+
+The gainmap metadata is copied unmodified through any processing operations.
+If you save an image with gainmap metadata to a JPEG file, libvips will do the
+write with the [method@Image.uhdrsave] operation, embedding the gainmap and the
+associated metadata in the output image.
+
+Intermediate operations which change the image geometry will also need to
+update the `"gainmap-data" metadata item, the mechanisms for doing this are
```
kleisauke:
```suggestion
update the `"gainmap-data"` metadata item, the mechanisms for doing this are
```

kleisauke:
The documentation looks great. Should we also support `jpegload`'s `autorotate` argument in `uhdrload`?

As an aside, I looked into how Android's Bitmap API handles gainmaps. It appears that automatic gainmap handling is applied only to basic operations, while for more advanced operations, the gainmap is preserved unchanged. See: https://developer.android.com/media/grow/ultra-hdr/edit#advanced-edits.

jcupitt:
> Should we also support jpegload's autorotate argument in uhdrload?

I'm not sure -- it's not that useful, and we don't need to support the whole of the `jpegload` / `jpegsave` API (there's a lot of it, ouch), just the bits we need for `thumbnail` etc.

>  As an aside, I looked into how Android's Bitmap API handles gainmaps. It appears that automatic gainmap handling is applied only to basic operations, while for more advanced operations, the gainmap is preserved unchanged. See: https://developer.android.com/media/grow/ultra-hdr/edit#advanced-edits.

Ah not very different from us then, well done for checking!

PR #4752, opened 2025-11-30
uhdrload: ensure ICC is at least 15 bytes before inspection

kleisauke:
Resolves: https://issues.oss-fuzz.com/issues/464597460


kleisauke:
For reference, see PR #3109 for a similar fix and  this section of `jpeg2vips.c`:
https://github.com/libvips/libvips/blob/0a0e4987af39325e519cc0663deee42aa2e2de11/libvips/foreign/jpeg2vips.c#L688-L692

jcupitt (approved):

PR #4753, opened 2025-11-30
pngsave: cast to target format when colourspace is unsupported

kleisauke:
Similar to #4744, but for `pngsave`. Test case:
```python
#!/usr/bin/env python3
import pyvips

im = pyvips.Image.new_from_array([1, 2, 3, 4])
buf = im.pngsave_buffer()
im2 = pyvips.Image.new_from_buffer(buf, "")

assert im.avg() == im2.avg()
```

Targets the 8.17 branch.


gcsideal:
Thanks, lua-vips builds correctly again.

PR #4754, opened 2025-11-30
uhdrload: remove "hdr" argument

jcupitt:
Since uhdr2scRGB() is much faster, and we save 200+ lines of code.

I'll update the docs in the other PR.


kleisauke (approved):
Great!

jcupitt:
Yes, `uhdrload("file.jpg").uhdr2scRGB()` is twice the speed of `uhdrload("filejpg", hdr=True)`, and lower mem use too.

PR #4756, opened 2025-11-30
add gainmap() and set(VImage) to cpp API

jcupitt:
Since they'll be needed for anyone processing UHDR images.

We'll need similar additions for pyvips, ruby-vips etc.

### DIFF
--- a/cplusplus/include/vips/VImage8.h
+++ b/cplusplus/include/vips/VImage8.h
@@ -551,7 +551,7 @@ public:
 	/**
 	 * The associated gainmap image, if any.
 	 */
-	const VImage
+	VImage
 	gainmap() const
 	{
 		return VImage(vips_image_get_gainmap(get_image()));


kleisauke:
Perhaps `get_gainmap()` could be renamed to `gainmap()` for consistency with the other getters (e.g. `filename()`, `width()`, etc.)?

jcupitt:
You're right, that would be better. Thanks!

There are a lot of other getters we could wrap, but maybe leave them until someone asks.

cplusplus/include/vips/VImage8.h
```diff
@@ -549,10 +549,10 @@ class VImage : public VObject {
 	}
 
 	/**
-	 * Return the gainmap image.
+	 * The associated gainmap image, if any.
 	 */
-	VImage
-	get_gainmap() const
+	const VImage
```
kleisauke: clang-tidy says: "Return type 'const vips::VImage' is 'const'-qualified at the top level, which may reduce code readability without improving const correctness".
```suggestion
	VImage
```

See: https://clang.llvm.org/extra/clang-tidy/checks/readability/const-return-type.html

cplusplus/include/vips/VImage8.h
```diff
@@ -670,6 +679,15 @@ class VImage : public VObject {
 			free_fn, data, length);
 	}
 
+	/**
+	 * Set the value of an image metadata item on an image.
+	 */
+	void
+	set(const char *field, const VImage value)
```
kleisauke, line 686: clang-tidy says: "The const qualified parameter 'value' is copied for each invocation; consider making it a reference"
```suggestion
	set(const char *field, const VImage &value)
```
See: https://clang.llvm.org/extra/clang-tidy/checks/performance/unnecessary-value-param.html

kleisauke: Some other C++ functions have the same issue, but I'm not sure whether fixing them would break ABI compat.

jcupitt: This one is OK -- `VImage` is just a pointer, so using a reference is actually slower.

jcupitt:
Thanks for the review!

PR #4760, opened 2025-12-01
cplusplus: apply a few clang-tidy suggested fixes

kleisauke:

https://kleisauke.nl/compat_reports/vips-cpp/master_to_clang-tidy/compat_report.html

### DIFF
--- a/cplusplus/include/vips/VError8.h
+++ b/cplusplus/include/vips/VError8.h
@@ -54,6 +54,8 @@ public:
 
 	/**
 	 * Get a reference to the underlying C string.
+	 * Note: this override must be preserved for ABI, removing it
+	 * would also eliminate the `_ZNK4vips6VError4whatEv` symbol.
 	 */
 	const char *
 	what() const noexcept override
@@ -65,6 +67,13 @@ public:
 	 * Print the error message to a stream.
 	 */
 	void ostream_print(std::ostream &) const;
+
+private:
+	/**
+	 * ABI padding to preserve original VError size.
+	 */
+	char _abi_padding[sizeof(std::exception) + sizeof(std::string) -
+		sizeof(std::runtime_error)] = {};
 };
 
 VIPS_NAMESPACE_END


cplusplus/include/vips/VError8.h
```diff
@@ -42,30 +42,23 @@ VIPS_NAMESPACE_START
...
 	 * Get a reference to the underlying C string.
 	 */
-	virtual const char *
-	what() const throw()
+	const char *
+	what() const noexcept override
 	{
-		return _what.c_str();
+		return std::runtime_error::what();
 	}
```
kleisauke, line 64: I considered removing `what()`, since it adds no functional value beyond what the base class (`std::runtime_error`) already provides. However, doing so appears to break ABI compat, see:
https://kleisauke.nl/compat_reports/vips-cpp/master_to_clang-tidy-rev2/compat_report.html
https://github.com/kleisauke/libvips/compare/cpp-clang-tidy..cpp-clang-tidy-rev2

jcupitt:
Wow I'm amazed you can change the exception base class without breaking ABI compatibility. The C++ ABI rules are very mysterious.

Is it possible it could break with some C++ compilers and not others?

cplusplus/include/vips/VImage8.h
```diff
@@ -293,7 +293,7 @@ class VOption {
 	 * A copy is taken of the object.
 	 */
 	VOption *
-	set(const char *name, const VObject value);
+	set(const char *name, VObject value);
```
jcupitt: These `const` decls might not mean anything to the compiler, but they do signal something to a human reading the header or the docs, don't they? I wonder if clang-tidy is being too strict with this rule.

kleisauke: You're right, I dropped this commit.

kleisauke:
My understanding is that this is safe because `std::runtime_error` also stores a `std::string` internally. However, I need to verify this, since the change in class size could(?) break code that catches `VError` by value (`catch (VError err)`) rather than by reference (`catch (const VError &err)`).

kleisauke:
... let's be on the safe side, commit 943d1fc5ab14870c9223d59f59f1e806c4671b2e ensures ABI compatibility, I now see:
https://kleisauke.nl/compat_reports/vips-cpp/master_to_clang-tidy-abi/compat_report.html

lovell:
I've been investigating error message corruption in v8.18.0 on Windows and I suspect this change might have exposed us to https://github.com/microsoft/STL/issues/2114. The call to `vips_error_buffer` now takes place sooner at construction time rather than deferring to a later `what()` call. Perhaps we should use `vips_error_buffer_copy` and manage the lifetime within the exception?

kleisauke:
@lovell According to PR https://github.com/nodejs/node-addon-api/pull/1059, this define appears to be the root cause:
https://github.com/nodejs/node/blob/v25.4.0/common.gypi#L485-L486

I'm not sure if we should support the `_HAS_EXCEPTIONS=0` case. This changeset appears to resolve the issue in sharp:
https://github.com/lovell/sharp/compare/0.35...kleisauke:sharp:temp-err-2
(i.e. the define also needs to be propagated to consumers of `VError8.h`)

lovell:
@kleisauke Thanks, that explains it, I was so close! It looks like STL will be altering its behaviour to copy the message so we can leave this alone.

Issue #4761, opened 2025-12-01
Possible ref leak when thumbnailing Ultra HDR images


kleisauke:
To reproduce:
```console
$ vipsthumbnail test/test-suite/images/ultra-hdr.jpg --vips-leak
vips_threadset_free: peak of 24 threads
3 objects alive:
0) VipsSource (0x15becfe0), count=2
VipsSource (source), input source, 
1) VipsForeignLoadJpegBuffer (0x15759760), count=1
VipsForeignLoadJpegBuffer (jpegload_buffer), load jpeg from buffer, priority=50, is_a_buffer, get_flags, get_flags_filename, header, load, jpegload_buffer buffer=VIPS_TYPE_BLOB, data = 0x15be49a0, length = 31738 out=((VipsImage*) 0x15becd10) shrink=8 flags=((VipsForeignFlags) VIPS_FOREIGN_SEQUENTIAL) -
2) VipsImage (0x15becd10) 1056 bytes, count=1
VipsImage (image), image class, 120x67 uchar, 1 band, b-w, jpegload_buffer
5 VipsArea alive
	0x15bfb8f0 count = 1, bytes = 518
	0x15bfb840 count = 1, bytes = 5
	0x15bed3a0 count = 1, bytes = 15
	0x15759870 count = 1, bytes = 31738
	0x157581d0 count = 1, bytes = 31738
memory: high-water mark 2.80 MB
```

jcupitt:
Ah I'm sure this worked, I must have broken it again :( I'll have a look.

jcupitt:
Ugh, it was the "clever" use of `&&` in `get_gainmap`, sigh. It's fixed again now.

Issue #4737, opened 2025-11-23
SIGSEGV while vipsthumbnail -a uhdr.jpg


ri-char:
## Bug report

**Describe the bug**
SIGSEGV

**To Reproduce**
1. Download this [image.jpg](https://github.com/user-attachments/assets/cd59fa85-1bc8-4731-bda4-ded002f67d88)
2. Run `vipsthumbnail -a image.jpg`
3. See error

**Expected behavior**
Generate thumbnail normally.

**Actual behavior**
Segmentation fault

**Backtrace**
```
#0  0x00007ffff7c620c8 in vips_image_get_typeof (image=0x0, name=0x7ffff7cc5c19 "gainmap-data")
    at ../libvips/iofuncs/header.c:1463
#1  0x00007ffff7a77175 in vips_foreign_save_jpeg_build (object=0x555556c7bd70) at ../libvips/foreign/jpegsave.c:151
#2  0x00007ffff7a783cc in vips_foreign_save_jpeg_buffer_build (object=0x555556c7bd70) at ../libvips/foreign/jpegsave.c:423
#3  0x00007ffff7c465eb in vips_object_build (object=0x555556c7bd70) at ../libvips/iofuncs/object.c:372
#4  0x00007ffff7c5b89d in vips_cache_operation_buildp (operation=0x7fffffff9310) at ../libvips/iofuncs/cache.c:1062
#5  0x00007ffff7c66445 in vips_call_required_optional (operation=0x7fffffff9310, required=0x7fffffff9350, 
    optional=0x7fffffff9460) at ../libvips/iofuncs/operation.c:974
#6  0x00007ffff7c665b2 in vips_call_by_name (operation_name=0x7ffff7cc5fbb "jpegsave_buffer", option_string=0x0, 
    required=0x7fffffff9350, optional=0x7fffffff9460) at ../libvips/iofuncs/operation.c:1014
#7  0x00007ffff7c66cee in vips_call_split (operation_name=0x7ffff7cc5fbb "jpegsave_buffer", optional=0x7fffffff9460)
    at ../libvips/iofuncs/operation.c:1121
#8  0x00007ffff7a78b26 in vips_jpegsave_buffer (in=0x0, buf=0x7fffffff9678, len=0x7fffffff9670)
    at ../libvips/foreign/jpegsave.c:703
[...]

PR #4762, opened 2025-12-02
Fix get gainmap (again)

jcupitt:
get_gainmap could refleak ... rework in a simpler way to make it obviously correct

See https://github.com/libvips/libvips/issues/4761


kleisauke:
This appears to regress issue #4737 and is causing test failures in the pytest suite. Sorry for not noticing it earlier.

jcupitt:
:groan: it should be `||` not `&&`, my fault, I'll make (another) PR

Issue #4616, opened 2025-07-26
Enhancement: support effort level 10 for jxlsave


cheebusjeebus:
`cjxl --help -v -v -v -v` describe of using `--allow_expert_options` for enablement of effort level 11 (`-e 11`).

i attempt with `vips jxlsave "image.png" "output.jxl" --lossless --effort 11` but he said:
```
(vips.exe:18728): GLib-GObject-CRITICAL **: 10:59:33.015: value "11" of type 'gint' is invalid or out of range for property 'effort' of type 'gint'
```

upon `vips -v` he said `vips-8.17.1`. i am utilizing him from here: https://github.com/libvips/build-win64-mxe/releases/download/v8.17.1/vips-dev-w64-all-8.17.1.zip

FWIW there is also effort level 10 of `cjxl` and `vips` also said same error (but he said 10 not 11 inside message)

please consider support for this efforts levels. thank you for good software entirely :)

lovell:
The upstream logic is here:

https://github.com/libjxl/libjxl/blob/73beeb5409cd805cf8e957b2f33467be74dfbfe2/lib/jxl/encode.cc#L1613-L1624

It looks like a value of 10 was added via https://github.com/libjxl/libjxl/pull/3248 in libjxl v0.10.0.

A value of 11 is limited to "expert" mode, which is not exposed, and I suspect libvips is the wrong tool at that point.

jcupitt:
Hi @cheebusjeebus, I made a quick PR to allow effort=10. Let's move any discussion there.

PR #4763, opened 2025-12-02
allow effort=10 for jxlsave

jcupitt:
libjxl 0.10 allows effort=10, and 11 in expert mode

allow effort 10, but don't try to implement expert mode encoding

See https://github.com/libvips/libvips/issues/4616


kleisauke:
I reported the CIFuzz failure here: https://github.com/google/highway/pull/2805#issuecomment-3602660593.

jcupitt:
Thank you!

Issue #4607, opened 2025-07-11
Add support for tune parameter of libaom


adityapatadia:
Libheif has a `tune` parameter which can be used to control how the quality is tuned for output of AVIF. Can we expose that parameter in libvips (and also sharp)?

https://github.com/strukturag/libheif/blob/c164ed719bd8fafd6cdc2181c6fc08afb7401b93/libheif/plugins/encoder_aom.cc#L294

jcupitt:
Sure, if this would be useful. Let's tag this as an enhancement issue.

jcupitt:
I made a PR for this, let's move any discussion there.

adityapatadia:
@lovell just a nudge, it would be good to have this in sharp as well.

PR #4764, opened 2025-12-02
heifsave: add "tune"

jcupitt:
Thanks @adityapatadia

See https://github.com/libvips/libvips/issues/4607

### DIFF
--- a/libvips/foreign/heifsave.c
+++ b/libvips/foreign/heifsave.c
@@ -673,7 +673,7 @@ vips_foreign_save_heif_build(VipsObject *object)
 		return -1;
 	}
 
-	if (vips_object_argument_isset(object, "tune")) {
+	if (heif->tune) {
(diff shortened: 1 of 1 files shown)


libvips/foreign/heifsave.c
```diff
@@ -668,6 +673,16 @@ vips_foreign_save_heif_build(VipsObject *object)
 		return -1;
 	}
 
+	if (vips_object_argument_isset(object, "tune")) {
```
kleisauke: nit: since this string defaults to `NULL`, we usually just do a simple check:
```suggestion
	if (heif->tune) {
```
See for example:
https://github.com/libvips/libvips/blob/5900859e7e5bf9fcfb8b06378b07547b8d0f77b2/libvips/create/text.c#L435

(but I don't mind either way)

jcupitt:
Thanks for the review!

Issue #4262, opened 2024-11-16
add support for oklab colour space


jcupitt:
oklab is a new-ish perceptual colour space:

https://en.wikipedia.org/wiki/Oklab_color_space

It's very like CIELAB, but more accurate. more stable, and no more complex. It should be easy to add support.

cc: @nrobidoux

mceachen:
oklch would be splendid as well.

jcupitt:
We have a PR for this feature, let's move any discussion there. Thanks for suggesting this!

PR #4766, opened 2025-12-02
Add oklab

jcupitt:
Add support for the oklab colour space, see:

https://en.wikipedia.org/wiki/Oklab_color_space

Referring issue: https://github.com/libvips/libvips/issues/4262

TODO 

- [x] add oklch as well
- [x] conversion to oklab is slow because of the `cbrtf()`, we should probably use a LUT (eg. `vips_XYZ2Lab()` does this). though HDR values would be tricky
- [x] it's possible to get L over 1.0 from sRGB which seems odd, perhaps we are not normalising correctly? 
- [x] update main docs

### DIFF
--- a/libvips/colour/Oklab2XYZ.c
+++ b/libvips/colour/Oklab2XYZ.c
@@ -76,9 +76,13 @@ vips_Oklab2XYZ_line(VipsColour *colour, VipsPel *out, VipsPel **in, int width)
 		const float s = sp * sp * sp;
 
 		// M1 inv to get D65 normalised XYZ
-		q[0] = l *  1.22701385 + m * -0.55779998 + s *  0.28125615;
-		q[1] = l * -0.04058018 + m *  1.11225687 + s * -0.07167668;
-		q[2] = l * -0.07638128 + m * -0.42148198 + s *  1.58616322;
+		float X = l *  1.22701385 + m * -0.55779998 + s *  0.28125615;
+		float Y = l * -0.04058018 + m *  1.11225687 + s * -0.07167668;
+		float Z = l * -0.07638128 + m * -0.42148198 + s *  1.58616322;
+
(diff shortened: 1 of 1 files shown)


jcupitt:
It seems to work! 

Running:

```
$ vips colourspace theo.jpg x.v oklab
```

Makes:

image

Pixels are 0-1 for L, roughly +0.5 to -0.5 for ab. It round-trips correctly.

jcupitt:
I made a tiny tester:

```python
#!/usr/bin/env python3

import sys
import pyvips

if len(sys.argv) != 6:
    print(f"usage: {sys.argv[0]} source-space dest-space v1 v2 v3")
    print(f"  eg.: {sys.argv[0]} oklab srgb 0.5 0 0")
    sys.exit(1)

src = sys.argv[1]
dst = sys.argv[2]
value = [float(arg) for arg in sys.argv[3:6]]
pixel = (pyvips.Image.black(1, 1) + value).copy(interpretation=src)
pixel = pixel.colourspace(dst)

print(pixel(0, 0))
```

To convert values between any pair of colourspaces, for example:

```
$ ./cvt.py oklab xyz 1.0 0 0
[95.04700469970703, 100.0, 108.83000183105469]
```

Which is D65-X0, Y0 and Z0, as expected. Or:

```
$ ./cvt.py srgb oklab 255 255 255
[1.0000016689300537, 2.418591975583695e-06, -0.00011364112287992612]
$ ./cvt.py srgb oklab 128 128 128
[0.5998718738555908, 1.422204150003381e-06, -6.817534449510276e-05]
$ ./cvt.py srgb oklch 255 0 0
[0.6279258728027344, 0.2576846182346344, 29.22317886352539]
```

Which all look OK.

kleisauke (approved):
LGTM!

PR #4767, opened 2025-12-02
Conditionally use C++14 for Highway sources

kleisauke:
Highway requires at least C++14 starting from version 1.4.0.

See: https://github.com/google/highway/pull/2805

### DIFF
--- a/fuzz/oss_fuzz_build.sh
+++ b/fuzz/oss_fuzz_build.sh
@@ -168,6 +168,8 @@ EOF
 
 # highway
 pushd $SRC/highway
+# FIXME: Remove once Highway 1.4.0 is released.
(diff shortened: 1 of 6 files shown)


jcupitt:
Oooof what a pain. Maybe we should just jump to C++14 as well? Would it make supporting older RHEL difficult?

kleisauke:
Good point, it looks like RHEL 8 (our minimum supported baseline) provides out-of-the-box support for C++14.
https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html-single/developing_c_and_cpp_applications_in_rhel_8/index#changes-in-gcc-in-rhel-8_changes-in-toolchain-since-rhel-7

Closing in favor of #4769.

PR #4768, opened 2025-12-03
cplusplus: explicitly take ownership of gainmap

kleisauke:
Test case included.

### DIFF
--- a/cplusplus/include/vips/VImage8.h
+++ b/cplusplus/include/vips/VImage8.h
@@ -554,7 +554,7 @@ public:
 	VImage
 	gainmap() const
 	{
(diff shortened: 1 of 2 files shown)


cplusplus/include/vips/VImage8.h
```diff
@@ -554,7 +554,7 @@ class VImage : public VObject {
 	VImage
 	gainmap() const
 	{
-		return VImage(vips_image_get_gainmap(get_image()));
+		return VImage(vips_image_get_gainmap(get_image()), NOSTEAL);
```
jcupitt, line 557: Ooop, nice catch!

PR #4771, opened 2025-12-04
make cpp's get_gainmap return a true ref

jcupitt:
This PR makes get_gainmap return a VImage which holds a true reference to the gainmap image. This makes it safer to use, while being no more complex for the caller.

### DIFF
--- a/cplusplus/include/vips/VImage8.h
+++ b/cplusplus/include/vips/VImage8.h
@@ -549,7 +549,7 @@ public:
 	}
 
 	/**
-	 * The associated gainmap image. nullptr for no gainmap.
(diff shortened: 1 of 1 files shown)


kleisauke:
I think `NOSTEAL` works the same way as this, see for example:
https://github.com/libvips/libvips/blob/11e74eea51510fc9bbb2d66f18c0e7541732f5fe/cplusplus/include/vips/VImage8.h#L96-L101

(it's a bit confusing)

jcupitt:
Oh haha, I thought it was the opposite, sigh. I'll delete this PR.

PR #4773, opened 2025-12-05
libvips 8.18 release notes

jcupitt:


jcupitt:
This links to some other docs we should update first (API/8.18 etc), but the text can be checked.

Any comments on the running order?

kleisauke (approved):
LGTM!

PR #4777, opened 2025-12-05
Regenerate POTFILES.in and function list

kleisauke:
Also, let's hide library-specific implementation details (which I had previously excluded manually from the change list).


jcupitt:
I didn't realize you had done that manually -- yes, the script change is much better!

PR #4782, opened 2025-12-06
rename get_gainmap as prepare_gainmap

jcupitt:
_get_gainmap() implied a safe getter, like _get_width(), but this function can modify the image and needs to be treated with care.

Rename as vips_image_prepare_gainmap().

### DIFF
--- a/libvips/iofuncs/header.c
+++ b/libvips/iofuncs/header.c
@@ -1084,15 +1084,6 @@ vips_image_prepare_gainmap(VipsImage *image)
 {
 	VipsImage *gainmap;
 
-#ifndef NDEBUG
-	printf("bananananan!!!!\n");
-	// allow one ref for the caller, one ref for the operation cache ...
-	// more than that and this is probably a shared image pointer
-	if (G_OBJECT(image)->ref_count > 2)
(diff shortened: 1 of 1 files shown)


jcupitt:
I added a tiny thing (disabled in release mode) to warn if an image you call `_prepare_gainmap()` on is a shared pointer. The tests seem to run cleanly.

jcupitt:
... we'll need to update the release notes too, of course.

kleisauke:
> I added a tiny thing (disabled in release mode) to warn if an image you call `_prepare_gainmap()` on is a shared pointer. The tests seem to run cleanly.

Great idea! I've generalized it in PR #4783 so that any metadata changes on a shared image trigger a warning.

kleisauke (approved):

kleisauke:
On further thought, what if we remove `vips_image_set_image()` from `vips_image_get_gainmap()`? Most callers update the gainmap anyway. This would also allow NetVips to expose the `.Gainmap` property on `Image` class rather than `MutableImage`.

kleisauke:
... I just opened PR #4784 as an alternative.

jcupitt:
Ah sorry, I didn't realize it was either/or, I should have waited.

PR #4783, opened 2025-12-06
Emit warning on metadata changes for shared images

kleisauke:
Reintroduce the code removed in commit e4db747. The warning is now only emitted in debug builds and when the `VIPS_LEAK` mechanism is enabled.

### DIFF
--- a/libvips/iofuncs/header.c
+++ b/libvips/iofuncs/header.c
@@ -1288,6 +1288,14 @@ vips_image_set(VipsImage *image, const char *name, GValue *value)
 	g_assert(name);
 	g_assert(value);
 
+#ifdef DEBUG_LEAK
+	/* Warn if metadata is being set on a shared image.
+	 */
+	if (vips__leak &&
+		G_OBJECT(image)->ref_count > 2)
+		printf("vips_image_set: set \"%s\" on shared image\n", name);
+#endif /*DEBUG_LEAK*/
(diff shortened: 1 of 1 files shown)


jcupitt:
Nice!

`g_warning()` output is hidden by default, maybe just use `printf()`? It'd be harder to ignore.

jcupitt (approved):

jcupitt:
I'll remove the ref count test from that PR I made, this one is better.

jcupitt:
`printf()`s can be annoying to find, maybe change the message to include the function name.

```C
printf("vips_image_set: set \"%s\" on shared image\n", name);
```

kleisauke:
This PR identifies three code paths that may remove or set a metadata item on a shared image:
https://github.com/libvips/libvips/blob/2e494438503f416cfefa5dff2f71eb0337049bdb/libvips/foreign/dzsave.c#L1638
https://github.com/libvips/libvips/blob/2e494438503f416cfefa5dff2f71eb0337049bdb/libvips/foreign/heifsave.c#L318
https://github.com/libvips/libvips/blob/2e494438503f416cfefa5dff2f71eb0337049bdb/libvips/foreign/vips2tiff.c#L1986

jcupitt:
Ah what a pain. Let's merge and fix those in a follow-up.

PR #4784, opened 2025-12-06
Make `_get_gainmap()` transfer ownership to caller

kleisauke:
`_get_gainmap()` now fully transfers ownership of the returned gainmap image to the caller, who is responsible for managing its lifecycle and setting the `"gainmap"` metadata.

Supersedes: #4782.


jcupitt:
Yes, this looks cleaner! 

It's close enough to `vips_image_get_image()` that the name no longer seems jarring.

PR #4786, opened 2025-12-07
Discard existing RGB input profile for CMYK output

kleisauke:
Fixes a regression introduced in commit 1b88a00.

As noticed within sharp's test suite:
https://github.com/lovell/sharp/blob/3609c61a226760a4e9ce8524c1b42e136831b89e/test/unit/metadata.js#L666-L675


jcupitt (approved):

jcupitt:
Oop, that was a silly brain fart. Good catch!

PR #4788, opened 2025-12-07
fix image sharing warning for dzsave

jcupitt:
Fix the dzsave one

See https://github.com/libvips/libvips/pull/4783

### DIFF
--- a/libvips/foreign/dzsave.c
+++ b/libvips/foreign/dzsave.c
@@ -1493,7 +1493,7 @@ direct_strip_init(DirectStrip *strip, Level *level)
 	if (vips_copy(level->image, &strip->image, NULL))
 		return -1;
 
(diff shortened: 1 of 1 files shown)


libvips/foreign/dzsave.c
```diff
@@ -1473,11 +1473,40 @@ image_strip_work(VipsThreadState *state, void *a)
...
+{
+	strip->level = level;
+	strip->x = 0;
+
+	/* We need a private image so we can modify the metadata.
+	 */
+	if (vips_copy(level->image, &strip->image, NULL))
+		return -1;
+
+	/* We don't want threadpoolrun to minimise on completion -- we need to
```
kleisauke:
```suggestion
	/* We don't want threadpool_run to minimise on completion -- we need to
```

kleisauke (approved):
Nice!

PR #4790, opened 2025-12-08
fix metadata change warning for heifsave

jcupitt:
We were removing the orientation tag from the image during save, and this triggered a warning about a shared image pointer.

This PR removes the metadata change -- we were removing the libvips metadata orientation fields, but not regenerating the exif data block, which is what we we were actually writing, so the removal had no effect.

We could call `vips__exif_update()` to rebuild the exif block, but that's not part of the libvips API, so isn't available when heifsave is a module.

See https://github.com/libvips/libvips/pull/4783 for reporting PR.


jcupitt:
Also improve a confusing indent.

PR #4791, opened 2025-12-08
fix the metadata change warning in tiffsave

jcupitt:
Very like the dzsave change -- we need to make a private image to signal the minimise behaviour we want to the thraedpool.

See https://github.com/libvips/libvips/pull/4783

### DIFF
--- a/libvips/foreign/vips2tiff.c
+++ b/libvips/foreign/vips2tiff.c
@@ -1980,20 +1980,26 @@ wtiff_layer_write_tiles(Wtiff *wtiff, Layer *layer, VipsRegion *strip)
 		 */
 		WtiffRow row = { wtiff, strip, layer, 0 };
 
+		VipsImage *x;
+		if (vips_copy(im, &x, NULL))
+			return -1;
+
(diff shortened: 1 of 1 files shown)


libvips/foreign/vips2tiff.c
```diff
@@ -1980,20 +1980,26 @@ wtiff_layer_write_tiles(Wtiff *wtiff, Layer *layer, VipsRegion *strip)
...
 		WtiffRow row = { wtiff, strip, layer, 0 };
 
+		VipsImage *x;
+		if (vips_copy(im, &x, NULL))
+			return -1;
+
 		/* We don't want threadpool_run to minimise on completion -- we need to
 		 * keep the cache on the pipeline before us.
 		 */
 		vips_image_set_int(im, "vips-no-minimise", 1);
```
kleisauke, line 1990:
```suggestion
		vips_image_set_int(x, "vips-no-minimise", 1);
```

(sorry for not seeing this earlier)

jcupitt: :facepalm: Thanks!

PR #4806, opened 2025-12-15
Use correct file extension for issue template config

kleisauke:
I couldn't find a link to libvips discussions when opening a new issue, so it seems the previous config was ineffective.


jcupitt:
How odd!

kleisauke:
Seems to work now! :tada: 

Before:
Screenshot From 2025-12-15 10-50-31

After:
Screenshot From 2025-12-15 10-52-10

Issue #4815, opened 2025-12-19
vips_thumbnail produces meaningless image on s390x


mtasaka:
Using released libvips v8.18.0, https://github.com/libvips/libvips/releases/tag/v8.18.0 
at least `vips_thumbnail` (just tried this) produces meaningless image on s390x.

Note that s390x is big endian.

C source:
`$ cat resize-to-limit-2.c`
```
#include <stdlib.h>
#include <sys/stat.h>

#define MY_ASSERT(expr) \
do { \
	if (!(expr)) { \
		fprintf(stderr, "LINE[%d] : assertion \"" #expr "\" failed.\n", __LINE__); \
		abort(); \
	} \
} while(0)


int main(void) {
	VipsImage *normal_image;
	int ret;

	const char *input_path = "./portrait.jpg";

	ret = vips_thumbnail(
			input_path, &normal_image,
			400,
			"height", 400,
			"size", VIPS_SIZE_DOWN,
			NULL
		);
	MY_ASSERT(!ret);

	vips_image_write_to_file(normal_image, "resized.jpg", NULL);
}
```

mtasaka:
![Image](https://github.com/user-attachments/assets/6bd2d3c7-b52a-4a54-8330-8e48f3e2106d)

Original `portrait.jpg`

mtasaka:
![Image](https://github.com/user-attachments/assets/efa7577c-9694-45e5-b3b0-5d736e809070)

result `resized.jpg` on x86_64 (and actually on i686, ppc64le, aarch64: all same)

mtasaka:
![Image](https://github.com/user-attachments/assets/fd20d1ae-06a4-4c22-b12e-5399af7232fb)
On the other hand, result `resized.jpg` on s390x

kleisauke:
What does the output of `vips --targets` show? Does setting the `VIPS_NOVECTOR` environment variable prevent this?

mtasaka:
on s390x:

```
$ vips --targets
builtin targets:   SCALAR
supported targets: Z15 Z14
```

And yes, `env VIPS_NOVECTOR=1` seems to suppress this issue.

mtasaka:
[...]

PR #4816, opened 2025-12-19
vector: mask supported Highway targets by builtin targets

kleisauke:
Ensure `vips_vector_get_supported_targets()` only reports targets that were actually compiled into libvips by intersecting the CPU supported targets with the builtin targets.

Resolves: #4815.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+date-tbd 8.18.1
+
+- vector: mask supported Highway targets by builtin targets [kleisauke]
+
 17/12/25 8.18.0
 
 - add dcrawload, dcrawload_source, dcrawload_buffer: load raw camera files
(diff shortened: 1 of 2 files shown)


kleisauke:
For reference, tested using:
```console
$ podman run --rm -it --platform=linux/s390x -e QEMU_CPU=max fedora:latest
$ dnf install gcc-c++ meson git glib2-devel expat-devel highway-devel
$ git clone -b issue-4815 https://github.com/kleisauke/libvips.git
$ cd libvips
$ meson setup build --prefix=/usr -D{deprecated,examples}=false
$ meson compile -C build
$ meson install -C build
$ vips --targets
builtin targets:   SCALAR
supported targets: none
```

mtasaka:
And actually this PR fixes https://bugzilla.redhat.com/show_bug.cgi?id=2421762 . :+1:

PR #4818, opened 2025-12-23
Add CICP and HDR support

Starbix:
This is a first go at code for converting image with nclx/CICP signals to scRGB. It's incomplete and was only briefly tested for my use case: HDR AVIF with PQ transfer and Display P3 primaries.

It seems to work reasonably well so far and the output JPG and PNG look much better than using copy.
Specifically the ultrahdr output in the JPG looks pretty good if the gainmap is shown.
[pics.zip](https://github.com/user-attachments/files/24310689/pics.zip)

In the zip you can find the original AVIF, the non-CICP-aware out_copy.jpg, and the CICP2scRGB outputs (both PNG and JPG, but original PNG was too large).

TODO:

- [ ] Add support for other loaders:
  - [x] PNG: read done, write done, bitdepth problematic
  - [ ] JXL: read done, write TODO (no 1:1 match with CICP enums)
  - [x] HEIF/AVIF: read done, write done, bitdepth possibly problematic
  - [ ] Other formats support CICP via ICC v4.4 profiles.
          Design decision needs to be made here: ICC profiles with 'cicp' tag are fine for output as lots of clients support it, but any icc_transform will lead to wrong colours as lcms2 is not aware of that tag
- [ ] Handle other transfer characteristics:
  - [x]  HLG
  - [x] PQ
  - [ ] BT.709/BT.2020 needs a second look
  - [ ] sRGB
  - [ ] other: not sure if needed
- [ ] Add more colour primary matrices
  - [x] BT.2020
  - [x] P3 (Display, DCI)
  - [x] BT.709
  - [ ] other: rarely used for images
- [ ] Make it faster/more elegant
- [ ] Add tests


jcupitt:
Great!

JXL also has CICP, so that would be a good one to add.

I bumped master to 8.19 ready for this to be merged when it's done.

Starbix:
CICP is now its own "colourspace" and transformations always go via scRGB. So "copy" from a CICP-tagged file should now yield somewhat decent colours. The libheif saver now also has support for CICP/nclx values. PNG saver should be relatively straightforward. Formats other than HEIF/AVIF/PNG/JXL could also profit from this, but there we'd need to generate an ICC profile with the "cicp" tag. That ICC profile should only be used for final output as lcms2 wouldn't support any transformations with it. So far conversion to CICP is not implemented. 
JXL is also a special case as it doesn't implement H.273 (rarely used options removed, but support for more colour spaces). This isn't handled super gracefully yet.
I probably broke TIFF support here, as TIFF is usually assumed to support any format. But it only supports CICP via ICC profiles.
Bitdepths also need a second look.

inodentry:
This PR is already highly useful as-is, it enables a lot of common use cases and workflows for HDR images. Do we really need to wait for a "complete" implementation of CICP before merging this? Can this be merged to add limited CICP support to libvips and then potentially have the remaining work in a future follow-up PR?

Are there any deal-breakers that prevent this PR from being merged in its current state?

I read through the code. If it's a matter of some minor fixes and additional testing, I'd be happy to help to bring this PR into a mergeable state.

I don't know if I could help implement the missing functionality though, I am not very familiar with the relevant colorspace math/theory.

I would really like to see this merged, because I would like to use libvips for processing HDR photos.

inodentry:
This PR is already highly useful as-is, it enables a lot of common use cases and workflows for HDR images. Do we really need to wait for a "complete" implementation of CICP before merging this? Can this be merged to add limited CICP support to libvips and then potentially have the remaining work in a future follow-up PR?

Are there any deal-breakers that prevent this PR from being merged in its current state?

I would really like to see this merged, because I would like to use libvips for processing HDR photos.

Starbix:
There are some architectural decisions that need to be made regarding preference of CICP vs ICC, how to handle icc_transform, and whether to treat HDR differently from SDR. Once I have input from the maintainers I can update the PR to make it mergeable.

PR #4643, opened 2025-08-22
Revise sequential access in resampling operations

kleisauke:
More testing is needed, so opened as a draft.

Resolves: #2757.

Context regarding the change to `thumbnail_image`:
https://app.gitter.im/#/room/#libvips_devchat:gitter.im/$x6M0KbuEpkdFjKFVmdRDf0V6m6NvPMD9Q4kwJevRrT8

PR #4823, opened 2025-12-31
experiment with starting the threadpool at size 1

jcupitt:
We currently start threadpools at 50% of max size and let them expand and contract with load. Since this happens pretty quickly, how about starting threadpools at size 1 instead? Simple pipelines would be quicker, heavy pipelines would hardly slow down.

Trivial patch for experimentation.

### DIFF
--- a/libvips/iofuncs/threadpool.c
+++ b/libvips/iofuncs/threadpool.c
@@ -647,10 +647,11 @@ vips_threadpool_run(VipsImage *im,
 	pool->work = work;
 	pool->a = a;
 
-	/* Start with half of the max number of threads, then let it drift up
+	/* Start with a single thread, then let it drift up
 	 * and down with load.
 	 */
 	for (n_working = 0; n_working < 1 + pool->max_workers / 2; n_working++)
+	// for (n_working = 0; n_working < 1; n_working++)
 		if (vips_worker_new(pool)) {
 			vips_threadpool_free(pool);
 			return -1;


jcupitt:
With a simple `vipsthumbnail` and a 6k x 4k RGB image I see before:

```console
$ time vipsthumbnail theo.jpg -o x.jpg --size 800

real	0m0.128s
user	0m0.075s
sys	0m0.074s
```

And after:

```console
$ time vipsthumbnail theo.jpg -o x.jpg --size 800

real	0m0.098s
user	0m0.064s
sys	0m0.041s
```

So it's noticeably quicker. Locking pools at size 1 is slightly quicker again, I'm not sure why:

```console
$ time VIPS_CONCURRENCY=1 vipsthumbnail theo.jpg -o x.jpg --size 800

real	0m0.084s
user	0m0.053s
sys	0m0.033s
```

jcupitt:
Running the code in vips-benchmark as an example of a heavy pipeline I see before:

```console
$ time ./vips-c tmp/x.tif tmp/y.tif

real	0m0.409s
user	0m1.028s
sys	0m0.462s
```

And after:

```console
$ time ./vips-c tmp/x.tif tmp/y.tif

real	0m0.428s
user	0m1.094s
sys	0m0.439s
```

So very slightly slower, since the threadpool will take a little white to expand.

jcupitt:
So a complex pipeline is 5% slower, a simple one is 30% faster, after removing 29ms for libvips startup.

kleisauke:
> Locking pools at size 1 is slightly quicker again, I'm not sure why:
> 
> ```console
> $ time VIPS_CONCURRENCY=1 vipsthumbnail theo.jpg -o x.jpg --size 800
> 
> real	0m0.084s
> user	0m0.053s
> sys	0m0.033s
> ```

IIRC, this may be due to the issue described in #2757. It might also be worth testing with PR #4643.

On older versions of libvips it could also be that `shrink{h,v}` + `reduce{h,v}` was slower than doing `reduce{h,v}` alone. Though, this should no longer be an issue after commit e0854707282ec68d6e20dafe0b781752aa6705c4. If needed, you can disable the `shrink{h,v}` paths with:
```diff
--- a/libvips/resample/thumbnail.c
+++ b/libvips/resample/thumbnail.c
@@ -836,7 +836,9 @@ vips_thumbnail_build(VipsObject *object)
 		in = t[4];
 	}
 
-	if (vips_resize(in, &t[5], 1.0 / hshrink, "vscale", 1.0 / vshrink, NULL))
+	if (vips_resize(in, &t[5], 1.0 / hshrink, "vscale", 1.0 / vshrink,
+			"gap", 0.0,
+			NULL))
 		return -1;
 	in = t[5];
 
```

PR #4826, opened 2026-01-04
Bilinear interpolation: replace loop unroller with for loop

lovell:
As most images only have 1-4 bands, replacing the use of Duff's Device with a simple for loop is ~10% faster with modern compilers.

Before:
```
$ vips resize 500x500.png out.v 4 --kernel=linear

real	0m0.102s
user	0m0.294s
sys		0m0.062s
```
```
712,000,000 (61.63%)  ???:vips_interpolate_bilinear_interpolate [build/libvips/libvips.so.42.21.0]
236,182,625 (20.44%)  ???:vips_affine_gen [build/libvips/libvips.so.42.21.0]
 40,076,920 ( 3.47%)  ???:vips_format_sizeof_unsafe [build/libvips/libvips.so.42.21.0]
 40,000,000 ( 3.46%)  ???:vips_band_format_iscomplex [build/libvips/libvips.so.42.21.0]
```
After:
```
$ vips resize 500x500.png out.v 4 --kernel=linear

real	0m0.098s
user	0m0.261s
sys		0m0.058s
```
```
620,000,000 (58.25%)  ???:vips_interpolate_bilinear_interpolate [build/libvips/libvips.so.42.21.0]
236,182,625 (22.19%)  ???:vips_affine_gen [build/libvips/libvips.so.42.21.0]
 40,076,920 ( 3.77%)  ???:vips_format_sizeof_unsafe [build/libvips/libvips.so.42.21.0]
 40,000,000 ( 3.76%)  ???:vips_band_format_iscomplex [build/libvips/libvips.so.42.21.0]
```

I've been working with gain maps and spotted that `vips_interpolate_bilinear_interpolate` is on the hot path when upscaling these. It's possible to make further small performance gains by creating specific 1, 3 and 4 band paths without any for loop, but these would complicate the logic so I've left this out.

### DIFF
--- a/libvips/resample/interpolate.c
+++ b/libvips/resample/interpolate.c
@@ -428,45 +428,31 @@ G_DEFINE_TYPE(VipsInterpolateBilinear, vips_interpolate_bilinear,
  * p3  p4
  */
 
-#define BILINEAR_INT_INNER \
-	{ \
-		tq[z] = (c1 * tp1[z] + c2 * tp2[z] + \
-					c3 * tp3[z] + c4 * tp4[z] + \
-					(1 << VIPS_INTERPOLATE_SHIFT) / 2) >> \
-			VIPS_INTERPOLATE_SHIFT; \
-		z += 1; \
-	}
-
 /* Fixed-point arithmetic, no tables.
  */
 #define BILINEAR_INT(TYPE) \
 	{ \
 		TYPE *restrict tq = (TYPE *) out; \
 \
-		int X = (x - ix) * VIPS_INTERPOLATE_SCALE; \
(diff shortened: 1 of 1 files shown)


jcupitt:
Alas, poor Duff! I knew him, Horatio. 

A trick whose time seems to have passed.

kmartinez:
I love this! all these speed-up tricks we've used over the years can expire!

lovell:
The one remaining use of Duff's Device in `hist_find` appears to still be useful, presumably as the loop size is much larger.

https://github.com/libvips/libvips/blob/47ffbbb6d70a322bffb64dbd941dd47d1c19e186/libvips/arithmetic/hist_find.c#L309-L321

PR #4827, opened 2026-01-05
thumbnail: add original image size to metadata

pszemus:
The original image size is lost due to the shrink-on-load feature. This PR stores it in the image metadata.

### DIFF
--- a/libvips/resample/thumbnail.c
+++ b/libvips/resample/thumbnail.c
@@ -659,9 +659,9 @@ vips_thumbnail_open(VipsThumbnail *thumbnail)
 	if (!(im = class->open(thumbnail, factor)))
 		return NULL;
 
-	vips_image_set_int(im, "orig-width", thumbnail->input_width);
-	vips_image_set_int(im, "orig-height", thumbnail->input_height);
-	vips_image_set_int(im, "orig-orientation", thumbnail->orientation);
+	vips_image_set_int(im, "original-width", thumbnail->input_width);
+	vips_image_set_int(im, "original-height", thumbnail->input_height);
+	vips_image_set_int(im, "original-orientation", thumbnail->orientation);
 	g_info("pre-shrunk size is %d x %d", im->Xsize, im->Ysize);
 
 	return im;


pszemus:
The original image orientation is also stored in order to correctly interpret the previously stored size.

libvips/resample/thumbnail.c
```diff
@@ -659,6 +659,9 @@ vips_thumbnail_open(VipsThumbnail *thumbnail)
 	if (!(im = class->open(thumbnail, factor)))
 		return NULL;
 
+	vips_image_set_int(im, "orig-width", thumbnail->input_width);
```
jcupitt: We don't use abbreviations in other metadata field names, how about `original-width` instead? It'd be a bit more consistent.

pszemus: Sure. Let me make that change.

jcupitt:
Hi @pszemus, 

You can open the original file and get the width and height that way. It should be very quick. 

It'll even work for pipes, for example:

```python
source = ... a pipe of some kind
image = pyvips.Image.new_from_source(source)
original_width = image.width
thumb = pyvips.Image.thumbnail_source(source, 256)
```

The `new_from_source` will only read the header parts of the image, and the `source` object will save the header bytes in a small buffer. When the `thumbnail_source` operation runs, it'll first see the bytes from this header buffer, then when it starts fetching pixels, `source` will switch to providing bytes from the pipe again.

Do you have a use-case that's not covered by this?

pszemus:
@jcupitt Can i use:

```
VSource image_src = VSource::new_from_file(config.input_path);
VImage input_image = VImage::new_from_source(image_src, "");

int iwidth = input_image.width();
int iheight = input_image.height();
int iorient = input_image.get_int("orientation");
LOG(INFO) << "Input image: " << iwidth << " x " << iheight << ", orient: " << iorient;

VImage image = VImage::thumbnail_source(image_src, 256);
```

?
And still have all thumbnail (e.g. shrink-on-load) optimizations?
Is this (`VSource::new_from_file()`, `VImage::new_from_source()` and then `VImage::thumbnail_source()`) the preferred way to read the input image metadata and then resize it?

jcupitt:
Yes, you'll get all the optimisations, but of course please test it.

`new_from_()` is guaranteed to only read the image header, so it's always quick (except for imagemagick-based loaders, sigh ... disable those if you want to be sure). It's how you're supposed to check for eg. decompression bombs before doing any processing.

pszemus:
Thanks @jcupitt !
It works quite well. I have one more question: why does the C++ API have methods like `VImage::width()`, `VImage::height()` , etc.. but not `VImage::orientation()`? Is calling `image.get_int("orientation")`, after checking if it exists, the only way to access image orientation?

jcupitt:
`width` and `height` are properties of the image object, so they are built into the libvips source code, you can see them here:

https://www.libvips.org/API/current/property.Image.width.html

https://github.com/libvips/libvips/blob/master/libvips/iofuncs/image.c#L1077-L1082

This means they are visible during static introspection. 

Things like `orientation` are managed dynamically by the name -- gvalue mapping that libvips tracks for each image:

https://www.libvips.org/API/current/method.Image.set.html

So they don't exist in the libvips source code.

You can use the `vips_image_set/get_()` API to read and write both image properties and metadata, but behind the scenes they are separate things.

We have `vips_image_get_page_height()` to test, fetch and sanity check `page-height`, and we could add a similar `vips_image_get_orientation()` convenience function, but no one's done it.

eg. `_get_page_height()` is here:

https://github.com/libvips/libvips/blob/master/libvips/iofuncs/header.c#L879-L902

jcupitt:
Oh haha we do have `_get_orientation()`, I'd forgotten:

https://github.com/libvips/libvips/blob/master/libvips/iofuncs/header.c#L977-L999

So the C++ API just needs a stub to wrap that.

pszemus:
I've added a #4829 
Thank you for your help.

PR #4716, opened 2025-11-01
add "bitdepth" to jxlsave

jcupitt:
For 8 and 16-bit uint images.

JXL right-justifies low bitdepth images (libvips is left-justified), so we must shift images down.

jxlload will need a matching change to left-shift low-bitdepth JXL images.

See https://github.com/libvips/libvips/issues/4083

PR #4830, opened 2026-01-08
Fix loading HDR JPEG XL and saving JPEG XL with custom bitdepth

DarthSim:
Hey there 👋 Long time no see 🙂

This PR fixes loading HDR JPEG XL and saving JPEG XL with user-provided `bitdepth`. This regression was introduced in https://github.com/libvips/libvips/pull/4716. libjxl (at least 0.11.1) shifts pixel values under the hood, so libvips should not do this.

Here's a sample of 10-bit JXL: https://github.com/imgproxy/test-images/raw/refs/heads/main/jxl/jxl.jxl. `vips copy jxl.jxl jxl.png` produces the following:

jxl

And here's a sample of 16-bit PNG: https://raw.githubusercontent.com/imgproxy/test-images/refs/heads/main/png/16-bpp.png. `vips copy 16-bpp.png 16-bpp.jxl[bitdepth=10]` produces a black image.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -11,6 +11,7 @@ date-tbd 8.18.1
 - uhdrsave: control generated gainmap scale factor [lovell]
 - source: guard against length truncation [Niebelungen-D]
 - string_to_array_double: ensure delimiter list is consistent [dloebl] [lovell]
+- fix loading/saving of non-8-bit JXL images [DarthSim]
 
 17/12/25 8.18.0
 


DarthSim:
Converted to draft. Need to dig it up a little bit more.

DarthSim:
I guess I got it right now.

The first thing to note is that JPEG XL operates in the XYB color space, which is float. So, bit depth seems to be just a hint to the decoder on which pixel format to use.

In libjxl < 0.8, `JxlBasicInfo.bits_per_sample` doesn't act in the input buffer conversion in any way: https://github.com/libjxl/libjxl/blob/v0.7.2/lib/jxl/enc_external_image.cc#L87. So the input buffer is cast to float based only on the pixel format. Thus, shifting pixel values based on `JxlBasicInfo.bits_per_sample` results in incorrect colors.

libjxl 0.8 introduced a way to specify the bit depth of the input buffer using the `JxlEncoderSetFrameBitDepth` function. Setting `JxlBitDepth.type = JXL_BIT_DEPTH_FROM_CODESTREAM` makes libjxl respect `JxlBasicInfo.bits_per_sample` when casting the input buffer to float. We can use this.

There are changes I made to fix the `jxlsave` behavior:

1. When libjxl < 0.8 is used, `bitdepth` is soft-limited to 8 and 16.
2. When the user-specified `bitdepth` is less than or equal to 8, the image is cast to uchar. libjxl >= 0.8 throws an error when we provide a ushort buffer with bits_per_sample <= 8, so we have to do this. Also, this enables saving 16-bit images as 8-bit with libjxl < 0.8.
3. Because of (2), I moved `vips_cast` after `vips_rshift_const1`.
4. When libjxl >= 0.8 is used, we use `JxlEncoderSetFrameBitDepth` to specify the input buffer bit depth.
5. Since the tests use libjxl 0.7, I changed the `bitdepth` test to save 16-bit and 8-bit images instead of 8-bit and 1-bit. Not sure if it'll work well, though, as JPEG XL size isn't really affected by bit depth. I believe we need to come up with a better test here.

jcupitt:
Hi again @DarthSim, happy 2026!

I'll off for the next few weeks unfortunately :( hospital stuff on one eye. I'll look at this when I'm back, or perhaps someone else can?

DarthSim:
@jcupitt Get well soon!

I changed the test. As I mentioned, bit depth doesn't affect filesize directly, so I offer to only check `format` and `bits-per-sample`.

kleisauke (commented):
Great! Since this is a bug fix, could you retarget this PR to the [`8.18`](https://github.com/libvips/libvips/tree/8.18) branch and include a note in the `ChangeLog`?

kleisauke:
I can also backport it in a follow-up PR, if that's easier. Just let me know. :)

DarthSim:
Hey @kleisauke!
Sorry, I got down with the sickness and then totally forgot about this.
Rebased the branch with 8.18 and changed the PR base.

kleisauke:
Thank you!

Issue #4835, opened 2026-01-16
build fails with Fedora glibc-2.42.9000-13.fc44 and above with glib 2.86.3


mtasaka:
I tried building 
* vips 8.18.0 ( https://github.com/libvips/libvips/tree/v8.18.0 )
* or vips git head ( https://github.com/libvips/libvips/commit/47ffbbb6d70a322bffb64dbd941dd47d1c19e186 )
on Fedora. Now it builds with Fedora 43 (glibc-2.42-6.fc43) but fails with Fedora 44 (currently glibc-2.42.9000-22.fc44.x86_64) like below:

```
[...]

PR #4836, opened 2026-01-17
memory: fix `VIPS_FREEF` for expression-like macros

kleisauke:
Ensure the macro works correctly when F expands to an expression, such as `g_free()` using `free_sized()` (as available in the upcoming glibc 2.43).

Resolves: #4835.
Targets the 8.18 branch.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,6 +3,7 @@ date-tbd 8.18.1
 - vector: mask supported Highway targets by builtin targets [kleisauke]
 - fix Highway paths on big-endian targets [kleisauke]
 - fix build with MSVC [star-hengxing]
+- fix compatibility with glibc 2.43 [mtasaka]
 
 17/12/25 8.18.0
 
--- a/libvips/include/vips/memory.h
+++ b/libvips/include/vips/memory.h
@@ -41,7 +41,7 @@ extern "C" {
 	G_STMT_START \
 	{ \
 		if (S) { \
-			(void) F((S)); \
+			(void) (F((S))); \
 			(S) = 0; \
 		} \
 	} \


TheEssem:
It seems this macro might still have an issue, I keep getting errors like this when trying to build the newest version of nip4 (EndeavourOS, libvips git c9efd305c7b33165d89a3178e6792ed61936f78a, GCC 15.2.1, glibc 2.43, glib 2.86.4):
```
../nip4-9.0.17/src/expr.c: In function ‘expr_strip’:
../nip4-9.0.17/src/expr.c:269:9: error: void value not ignored as it ought to be
  269 |         VIPS_FREE(expr->error_top);
../nip4-9.0.17/src/expr.c:270:9: error: void value not ignored as it ought to be
  270 |         VIPS_FREE(expr->error_sub);
../nip4-9.0.17/src/expr.c: In function ‘expr_error_set’:
../nip4-9.0.17/src/expr.c:467:17: error: void value not ignored as it ought to be
  467 |                 VIPS_SETSTR(expr->error_sub, error_get_sub());
../nip4-9.0.17/src/expr.c:479:17: error: void value not ignored as it ought to be
  479 |                 VIPS_SETSTR(expr->error_top, vips_buf_all(&buf));
```

jcupitt:
Hi @TheEssem, I think Kleis's fix was to the stable 8.18 branch and hadn't been merged to libvips master. 

Could you try current libvips master again? I think it should be fixed.

kleisauke:
It ought to be fixed indeed. Perhaps you have a stray `vips/memory.h` file somewhere?

TheEssem:
Ah, sure enough there was a libvips install in `/usr/local` that I failed to notice. My mistake, sorry for the trouble.

PR #4839, opened 2026-01-27
uhdrload: expose gainmap-downsample-factor metadata

lovell:
I've started integrating all the excellent work on gain maps into sharp and I suspect a few PRs might fall out of this work.

In addition to exposing gain map dimensions from the input metadata, control over the output gain map scale factor might be useful too as this is currently fixed at 2 but most sample images I've been working with appear to use a factor of 4.

### DIFF
--- a/libvips/foreign/uhdrload.c
+++ b/libvips/foreign/uhdrload.c
@@ -475,6 +475,12 @@ vips_foreign_load_uhdr_set_metadata(VipsForeignLoadUhdr *uhdr, VipsImage *out)
 			"gainmap-use-base-cg", gainmap_metadata->use_base_cg);
 	}
 
+	const int gainmap_width = uhdr_dec_get_gainmap_width(uhdr->dec);
+	if (gainmap_width > 0) {
+		vips_image_set_int(out, "gainmap-downsample-factor", VIPS_MAX(1,
+			uhdr_dec_get_image_width(uhdr->dec) / gainmap_width));
+	}
+
 	return 0;
 }
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,7 @@ date-tbd 8.18.1
 - pngload: avoid an expensive check during header read [kleisauke]
 - improve vips7 JPEG load compatibility [kleisauke]
 - fix saving 3-band MATRIX images [kleisauke]
+- uhdrload: expose original gainmap scale factor [lovell]
 
 17/12/25 8.18.0
 
--- a/test/test-suite/test_foreign.py
+++ b/test/test-suite/test_foreign.py
@@ -442,6 +442,8 @@ class TestForeign:
             value = im.get(name)
             assert isinstance(value, (int, float))
 
+        assert im.get("gainmap-downsample-factor") == 4
+
         value = im.get("gainmap-data")
         assert len(value) > 10000
 


jcupitt:
Ah wait, I had a thought, what if the image has a non-zero orientation. Should we swap gainmap width and height?

jcupitt:
Oh, and if you load a uhdr image with a subsample, `uhdrload` will also subsample the gainmap:

https://github.com/libvips/libvips/blob/master/libvips/foreign/uhdrload.c#L425-L437

Maybe it's better to get people to use eg.:

```C
VipsImage *gainmap;
if ((gainmap = vips_image_get_gainmap(image))) {
    int gainmap_width = gainmap->Xsize;
    int gainmap_height = gainmap->Ysize;

    // ... do something with the gainmap dimensions

    g_object_unref(gainmap);
}
```

This should be quick (the result is cached) and now the dimensions are guaranteed to be correct, at least with regard to the actual gainmap on the image.

lovell:
Thanks John. The scenario that led to this PR is less about gain map dimensions and more the original scale factor, as there seems to be a common pattern of using the same factor when re-encoding. Apple seems to use 2, Android prefers 4.

Perhaps we could provide the scale factor via properties like `gainmap_xfactor` or (consolidated) `gainmap_scale_factor`? This could be in addition to (scaled-up and oriented) dimension properties.

jcupitt:
Ah, makes sense.

Maybe `gainmap-downsample-factor`, an int, usually 1, 2, or 4? You choose!

lovell:
Updated as the integral `gainmap-downsample-factor` calculated using the original widths before scaling/orientation, if any.

libvips/foreign/uhdrload.c
```diff
@@ -475,6 +475,12 @@ vips_foreign_load_uhdr_set_metadata(VipsForeignLoadUhdr *uhdr, VipsImage *out)
 			"gainmap-use-base-cg", gainmap_metadata->use_base_cg);
 	}
 
+	const int gainmap_width = uhdr_dec_get_gainmap_width(uhdr->dec);
+	if (gainmap_width > 0) {
+		vips_image_set_int(out, "gainmap-downsample-factor", VIPS_ROUND_UINT(
```
jcupitt: `int / int` is always `int`, so I think we probably don't need a round.

How about `VIPS_MAX(1, uhdr_dec_get_image_width(uhdr->dec) / gainmap_width)`? We shouldn't let it go to zero, which could happen if the gainmap is even one pixel wider than the image.

lovell: Ah yes, I hadn't considered the gainmap having greater dimensions than the base image, updated.

Given their rich history of ignoring standards I'm sure it'll be a future Samsung device that will manage to create one of these.

kleisauke:
> Apple seems to use 2, Android prefers 4.

It appears that Android uses libultrahdr's `JpegR` class directly, which uses different defaults.
https://github.com/google/libultrahdr/blob/v1.4.0/lib/include/ultrahdr/jpegr.h#L34
https://android.googlesource.com/platform/frameworks/av/+/refs/tags/android-16.0.0_r4/services/camera/libcameraservice/api2/JpegRCompositeStream.cpp#300

PR #4844, opened 2026-01-30
Avoid implicit bool-to-enum conversions

kleisauke:
Only the changes to `im_jpeg2vips.c` are noted in the `ChangeLog`. The `vipspng.c` changes are non-functional, since `VipsFailOn` is no-op during header-only reads.

Targets the 8.18 branch.


jcupitt (approved):

jcupitt:
Ooop, that was an oversight.

PR #4851, opened 2026-02-03
Add link to funding via GitHub sponsors

lovell:
This should add a link to https://github.com/sponsors/libvips alongside the existing OpenCollective funding option.


jcupitt:
:+1:

Issue #4834, opened 2026-01-16
vipsthumbnail auto-rotate does not support EXIF orientation in PNG files


pszemus:
**Describe the bug**
vipsthumbnail (and vipsheader) does not apply orientation metadata from EXIF data in PNG files .
EXIF data can be stored in `eXIf`, `zTXt` and `tEXt` PNG chunks.

Sample images: 
- rotated-mordy.png:
  <img width="1600" height="2560" alt="Image" src="https://github.com/user-attachments/assets/b630053f-87ca-4790-a657-586e21bb69b3" />

- Exif_zTXt_Orient_RightTop.png:
  <img width="427" height="640" alt="Image" src="https://github.com/user-attachments/assets/a970af1f-a0c9-4f69-943f-30fee1d875dc" />

GraphicsMagick is able to display the PNG image orientation:
```
$ gm identify -verbose rotated-mordy.png
[...]
  Orientation: RightTop
  Profile-exif: 32 bytes
    Orientation: 6
```
```
$ gm identify -verbose Exif_zTXt_Orient_RightTop.png
[...]
  Orientation: RightTop
  Profile-EXIF: 32 bytes
    Orientation: 6
```

jcupitt:
Sure, let's implement this. I've tagged this as an enhancement issue.

jcupitt:
I had a quick look and libvips already supports PNG rotation.

The problem is that there are two ways to embed EXIF data in PNG. In the bad old days, people would put the EXIF block into a text segment, but since 2017 PNG has supported an official EXIF tag. libvips implements this standardised EXIF in PNG tag, it does not currently support non-standard pre-2017 EXIF in text segments.

I think your test images use non-standard, pre-2017 EXIF in PNG.

kleisauke:
Indeed, inspecting these two images with https://www.nayuki.io/page/png-file-chunk-inspector shows that their EXIF data is stored in the `tEXt` and `zTXt` PNG chunks. However, libvips only supports the standardised `eXIf` PNG chunk.

Note that browsers completely ignore the EXIF orientation tag in PNG files. See e.g. #3269 where this behaviour leads to issues.

pszemus:
I can confirm that vips handles PNG files with EXIF data in `eXIf` chunks correctly in terms of auto-rotation. 
PNG files that have EXIF data in `tEXt` or `zTXt` chunks, as @kleisauke wrote, are not.

[...]

PR #4853, opened 2026-02-04
png: add support for parsing EXIF data from zTXt chunks

pszemus:
As a workaround to the PNG specification prior to 1.5.0,  [that introduced the `eXIf` chunk](https://ftp-osl.osuosl.org/pub/libpng/documents/pngext-1.5.0.html#C.eXIf), some tools ([ImageMagick](https://github.com/ImageMagick/ImageMagick/blob/7.1.2-13/coders/png.c#L3672), [GraphicsMagick](https://sourceforge.net/p/graphicsmagick/code/ci/GraphicsMagick-1_3_46/tree/coders/png.c#l2893), [exiv2](https://dev.exiv2.org/projects/exiv2/wiki/The_Metadata_in_PNG_files), [exiftool](https://exiftool.org/TagNames/PNG.html#TextualData), [libwebp](https://github.com/webmproject/libwebp/blob/v1.6.0/imageio/pngdec.c#L140)) could handle EXIF data, as a compressed string, in the `zTXt` PNG chunks, in  the format described in: https://clanmills.com/exiv2/book/ -> "PNG and the Zlib compression library":
```
\n
exif\n
    number\n
hexEncodedBinary\n
....
```

example file: 
Exif_zTXt_Orient_RightTop

This PR parses the "Raw profile type exif" and "Raw profile type APP1" `zTXt` chunks and stores the EXIF data in `VIPS_META_EXIF_NAME` (only if the image doesn't have EXIF data already stored via reading the `eXIf` chunk).

According to https://zpl.fi/exif-orientation-in-different-formats/ this non-standard format is handled only by Safari, but this change could be used by vips  to auto-rotate according to the embedded orientation for example.

Before this PR:
```
$ vipsheader -a Exif_zTXt_Orient_RightTop.png 
Exif_zTXt_Orient_RightTop.png: 427x640 uchar, 3 bands, srgb, pngload
width: 427
height: 640
bands: 3
format: uchar
coding: none
interpretation: srgb
xoffset: 0
yoffset: 0
xres: 2,834
yres: 2,834
filename: /home/pszemus/projects/imageresizer-worker/test/resources/Exif_zTXt_Orient_RightTop.png
vips-loader: pngload
png-comment-0-Raw profile type exif: 
exif
      32
45786966000049492a0008000000010012010300010000000600000000000000

png-comment-1-date:modify: 2021-09-13T17:14:24+00:00
bits-per-sample: 8
background: 255 255 255 
```

After applying this PR:
```
$ buildDir/tools/vipsheader -a Exif_zTXt_Orient_RightTop.png 
Exif_zTXt_Orient_RightTop.png: 427x640 uchar, 3 bands, srgb, pngload
width: 427
height: 640
bands: 3
format: uchar
coding: none
interpretation: srgb
xoffset: 0
yoffset: 0
xres: 2,834
yres: 2,834
filename: /home/pszemus/projects/imageresizer-worker/test/resources/Exif_zTXt_Orient_RightTop.png
vips-loader: pngload
exif-data: 32 bytes of binary data
exif-ifd0-Orientation: 6 (Prawo - góra, Short, 1 components, 2 bytes)
exif-ifd0-XResolution: 71983/1000 (71,983, Rational, 1 components, 8 bytes)
exif-ifd0-YResolution: 71983/1000 (71,983, Rational, 1 components, 8 bytes)
exif-ifd0-ResolutionUnit: 2 (Cal, Short, 1 components, 2 bytes)
exif-ifd2-ExifVersion: Exif w wersji 2.1 (Exif w wersji 2.1, Nieokreślony, 4 components, 4 bytes)
exif-ifd2-FlashpixVersion: FlashPix w wersji 1.0 (FlashPix w wersji 1.0, Nieokreślony, 4 components, 4 bytes)
exif-ifd2-ColorSpace: 65535 (Nieskalibrowana, Short, 1 components, 2 bytes)
orientation: 6
png-comment-1-date:modify: 2021-09-13T17:14:24+00:00
bits-per-sample: 8
background: 255 255 255 
```
Fixes: #4834


libvips/foreign/vipspng.c
```diff
@@ -356,6 +357,89 @@ read_new(VipsSource *source, VipsImage *out,
 	return read;
 }
 
+static uint8_t *
+vips__hex_string_to_bytes(const char *hex_string, size_t hex_length)
+{
+	const char *src = hex_string;
+	size_t actual_length = 0;
+	uint8_t *const raw_data = (uint8_t *) g_malloc(hex_length);
```
jcupitt: `g_malloc()` can't fail, you probably want `g_try_malloc()`, which can return NULL for memory full.

libvips has `VIPS_ARRAY` which is slightly neater, eg.:

```C
uint8_t *raw_data = VIPS_ARRAY(NULL, hex_length, uint8_t);
```

libvips/foreign/vipspng.c
```diff
@@ -356,6 +357,89 @@ read_new(VipsSource *source, VipsImage *out,
...
+		if (*src == '\n')
+			continue;
+		val[0] = *src++;
+		val[1] = *src;
+		val[2] = '\0';
+		*dst++ = (uint8_t) strtol(val, &end, 16);
+		if (end != val + 2)
+			break;
+		actual_length++;
+	}
```
jcupitt: How about looping along the output buffer instead? You could get the loop counters and increments all in one place then. Perhaps:

```C
static uint8_t *
vips__hex_string_to_bytes(const char *hex_string, size_t hex_length)
{
    uint8_t *raw_data = VIPS_ARRAY(NULL, hex_length, uint8_t);
    if (!raw_data)
        return NULL;

    const char *p;
    size_t i;
    for (i = 0, p = hex_string; i < hex_length && p[0] && p[1]; i++, p += 2) {
        char val[3];

        val[0] = p[0];
        val[1] = p[1];
        val[2] = '\0';
        if (!(raw_data[i] = g_ascii_strtoll(val, NULL, 16)))
            break;
    }

    if (i < hex_length) {
        g_free(raw_data);
        return NULL;
    }

    return raw_data;
}
```

`g_ascii_strtoll()` is a `strtoll()` that avoids locale issues and is usually better for reading file data.

libvips/foreign/vipspng.c
```diff
@@ -356,6 +357,89 @@ read_new(VipsSource *source, VipsImage *out,
...
+		g_warning("Malformed raw profile");
+		return NULL;
+	}
+	p++;
+
+	// Skip profile type (e.g. "exif")
+	while (*p != '\0' && *p != '\n')
+		p++;
+	// Extract EXIF data length
+	expected_length = (int) strtol(p, &hex_payload, 10);
```
jcupitt: Perhaps use `g_ascii_strtoll()` here too.

libvips/foreign/vipspng.c
jcupitt: Ooops, it'd need something to skip `\n` too, of course.

jcupitt:
This is great! Thank you for doing this work, @pszemus 

I saw a couple of minor things. This is probably worth a ChangeLog note too.

libvips/foreign/vipspng.c
jcupitt: Drat that still won't work. I think it would be better to structure this like a parser, so have little funcs to read the various terminals, then the parse just has the file structure.

So your terminals are:

```C
static const char *
skip_whitespace(const char *p)
{
    if (p)
        p += strcspn(p, " \n");

    return p;
}

static const char *
skip_line(const char *p)
{
    while (p && *p && *p != '\n')
        p++;
    if (p && *p == '\n')
        p++;

    return p;
}

static const char *
read_hex_pair(const char *p, int *value)
{
    if (!p || !p[0] || !p[1])
        return NULL;

    char val[3] = { p[0], p[1], '\0' };
    char *q;
    gint64 i = g_ascii_strtoll(val, &q, 16);
    if (errno || q == val)
        return NULL;

    *value = i;

    return p + 2;
}

static const char *
read_length(const char *p, size_t *value)
{
    if (!p)
        return NULL;

    char *q;
    gint64 i = g_ascii_strtoll(p, &q, 10);
    // limit the length to 10mb for sanity
    if (errno || q == p || i <= 0 || i > 10 * 1024 * 1024)
        return NULL;

    *value = i;

    return q;
}
```

Now `vips__parse_raw_profile()` is pretty simple:

```C
static uint8_t *
vips__parse_raw_profile(const char *text, size_t *data_length)
{
    const char *p = text;

    // Raw profile should start with a new line
    p = skip_line(p);
    if (!p) {
        g_warning("pngload: malformed raw profile");
        return NULL;
    }

    // Skip profile type (e.g. "exif")
    p = skip_line(p);

    // number of hex pairs to read
    size_t length;
    p = read_length(p, &length);
    if (!p) {
        g_warning("pngload: malformed raw profile");
        return NULL;
    }

    uint8_t *data = VIPS_ARRAY(NULL, length, uint8_t);
    if (!data)
        return NULL;

    size_t i;
    for (i = 0; i < length; i++) {
        int value;

        p = skip_whitespace(p);
        p = read_hex_pair(p, &value);
        if (!p)
            break;

        data[i] = value;
    }

    if (i < length) {
        g_warning("pngload: malformed raw profile");
        VIPS_FREE(data);
        return NULL;
    }

    *data_length = length;

    return data;
}
```

pszemus:
> This is great! Thank you for doing this work, @pszemus
> 
> I saw a couple of minor things. This is probably worth a ChangeLog note too.

Thanks @jcupitt ! I've applied your suggestions, tested and pushed the changes.

jcupitt (approved):

jcupitt:
Great! Thanks again!

PR #4855, opened 2026-02-06
uhdrsave: expose control over gain map scale factor

lovell:
A follow-up to #4839 for the encoding side, plus adds some docs and tests as well.

### DIFF
--- a/libvips/foreign/uhdrsave.c
+++ b/libvips/foreign/uhdrsave.c
@@ -77,6 +77,8 @@ typedef struct _VipsForeignSaveUhdr {
 
 	int Q;
 
+	int gainmap_scale_factor;
+
 	uhdr_codec_private_t *enc;
 
 } VipsForeignSaveUhdr;
@@ -373,7 +375,7 @@ vips_foreign_save_uhdr_hdr(VipsForeignSaveUhdr *uhdr, VipsImage *image)
 	g_info("saving scRGB as UltraHDR");
 
 	uhdr_enc_set_output_format(uhdr->enc, UHDR_CODEC_JPG);
-	uhdr_enc_set_gainmap_scale_factor(uhdr->enc, 2);
+	uhdr_enc_set_gainmap_scale_factor(uhdr->enc, uhdr->gainmap_scale_factor);
 	uhdr_enc_set_using_multi_channel_gainmap(uhdr->enc, 0);
 
 	// attach the gainmap, if we have one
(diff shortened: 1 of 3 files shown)


libvips/foreign/uhdrsave.c
```diff
@@ -373,7 +375,7 @@ vips_foreign_save_uhdr_hdr(VipsForeignSaveUhdr *uhdr, VipsImage *image)
 	g_info("saving scRGB as UltraHDR");
 
 	uhdr_enc_set_output_format(uhdr->enc, UHDR_CODEC_JPG);
-	uhdr_enc_set_gainmap_scale_factor(uhdr->enc, 2);
+	uhdr_enc_set_gainmap_scale_factor(uhdr->enc, uhdr->gainmap_scale_factor);
 	uhdr_enc_set_using_multi_channel_gainmap(uhdr->enc, 0);
```
jcupitt, line 379: Oh no! I hadn't realised this was the official uhdr name for this value. We should probably change `gainmap-downsample-factor` in uhdrload to match this.

I can make a PR if you like (since it's my fault).

libvips/foreign/uhdrsave.c
```diff
@@ -542,12 +544,20 @@ vips_foreign_save_uhdr_class_init(VipsForeignSaveUhdrClass *class)
 		G_STRUCT_OFFSET(VipsForeignSaveUhdr, Q),
 		1, 100, 75);
 
+	VIPS_ARG_INT(class, "gainmap_scale_factor", 11,
```
jcupitt, line 547: This is an API change so I suppose strictly it shouldn't happen in a micro version and ought to wait for 8.19, but that would mean delaying sharp's uhdr integration by a few months, so I can see why we'd want to have this.

@kleisauke, any thoughts?

jcupitt: We could always push out 8.19 sooner than the usual six month window, of course, perhaps as soon as sharp needs it. Maybe that's the best solution?

kleisauke: While this isn't a strict bug fix, turning the hardcoded scale factor into a configurable option adds enough value that I'm fine including it in a micro/patch release.

libvips/foreign/uhdrsave.c
lovell: https://github.com/libvips/libvips/pull/4856

Issue #4857, opened 2026-02-10
Heap-buffer-overflow in `vips_source_read_to_memory()`  (integer truncation)


Niebelungen-D:
## Questions, enhancements, tips, etc.
A potential heap overflow may occur in vips_source_read_to_memory due to a lack of input validation.

## Bug report

**Describe the bug**
`vips_source_read_to_memory()` (`libvips/libvips/iofuncs/source.c`) allocates a `GByteArray` using `source->length` (`gint64`), but `g_byte_array_set_size()` takes a 32-bit `guint`. When `source->length > G_MAXUINT`, the allocation truncates while the subsequent read loop still writes up to the full 64-bit length bound, causing a heap-buffer-overflow.

**To Reproduce**
1. Build and install libvips with ASAN/UBSAN:
   - `meson setup build-asan -Ddebug=true -Db_sanitize=address,undefined --prefix "$PWD/_inst" --libdir=lib`
   - `meson compile -C build-asan`
   - `meson install -C build-asan`
   - `export PKG_CONFIG_PATH="$PWD/_inst/lib/pkgconfig:$PKG_CONFIG_PATH"`
   - `export LD_LIBRARY_PATH="$PWD/_inst/lib:$LD_LIBRARY_PATH"`
   - `export VIPSHOME="$PWD/_inst"`
2.  poc:
```C
#include <vips/vips.h>

#include <glib.h>

#include <stdint.h>
#include <stdio.h>
#include <string.h>

typedef struct {
	gint64 pos;
	gint64 len;
	int reads;
} SourceState;

static gint64
seek_cb(VipsSourceCustom *source_custom, gint64 offset, int whence, SourceState *state)
{
	(void) source_custom;

	switch (whence) {
	case SEEK_SET:
		state->pos = offset;
		break;
	case SEEK_CUR:
		state->pos += offset;
		break;
	case SEEK_END:
		state->pos = state->len + offset;
		break;
	default:
		return -1;
	}

	if (state->pos < 0)
		return -1;

	return state->pos;
}

static gint64
read_cb(VipsSourceCustom *source_custom, void *buffer, gint64 size, SourceState *state)
{
	(void) source_custom;

	/* Overflow should happen on the first call: libvips will pass a large
	 * @size, but the destination buffer is truncated by guint.
	 */
	if (state->reads++ > 0)
		return 0;

	const gint64 to_write = size > 4096 ? 4096 : size;
	memset(buffer, 0x41, (size_t) to_write);
[...]

PR #4858, opened 2026-02-10
source: guard against length truncation

kleisauke:
GByteArray stores its length as guint, while libvips uses a 64-bit length. Passing values larger than UINT_MAX could silently truncate the length. Add checks to prevent overflow.

Resolves: #4857.
Targets the 8.18 branch.

### DIFF
--- a/libvips/iofuncs/source.c
+++ b/libvips/iofuncs/source.c
@@ -954,6 +954,12 @@ vips_source_read_to_memory(VipsSource *source)
 	g_assert(!source->header_bytes);
 	g_assert(source->length >= 0);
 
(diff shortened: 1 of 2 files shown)


libvips/iofuncs/source.c
```diff
@@ -954,14 +954,20 @@ vips_source_read_to_memory(VipsSource *source)
 	g_assert(!source->header_bytes);
 	g_assert(source->length >= 0);
 
+	if (G_UNLIKELY(source->length > UINT_MAX)) {
```
lovell, line 957: Good use of `G_UNLIKELY` 👍

PR #4846, opened 2026-02-02
smartcrop: introduce cropping by a specific point of interest

pszemus:
This is the first step toward cropping around a specific point using the vips_thumbnail() function. In this PR, vips_smartcrop() has been improved.

A full description can be found here: https://github.com/libvips/libvips/discussions/4828

PR #4860, opened 2026-02-12
thumbnail: introduce cropping by a specific point of interest

pszemus:
This is the second and final step toward cropping around a specific point using the `vips_thumbnail()` function.
In this PR, `vips_thumbnail()` has been improved.

A full description can be found here: https://github.com/libvips/libvips/discussions/4828

### DIFF
--- a/libvips/resample/thumbnail.c
+++ b/libvips/resample/thumbnail.c
@@ -988,8 +988,8 @@ vips_thumbnail_build(VipsObject *object)
 		gboolean rotated = thumbnail->swap && thumbnail->auto_rotate;
 		int rotated_input_width = rotated ? thumbnail->input_height : thumbnail->input_width;
 		int rotated_input_height = rotated ? thumbnail->input_width : thumbnail->input_height;
-		double overall_hshrink = (double) in->Xsize / rotated_input_width;
-		double overall_vshrink = (double) in->Ysize / rotated_input_height;
+		double overall_hshrink = (double) rotated_input_width / in->Xsize;
+		double overall_vshrink = (double) rotated_input_height / in->Ysize;
 
 		/* Need to copy to memory, we have to stay seq.
 		 *
@@ -998,8 +998,8 @@ vips_thumbnail_build(VipsObject *object)
 		if (!(t[13] = vips_image_copy_memory(in)) ||
 			vips_smartcrop(t[13], &t[14], crop_width, crop_height,
 				"interesting", thumbnail->crop,
-				"interesting_x", VIPS_ROUND_UINT((double) thumbnail->interesting_x * overall_vshrink),
-				"interesting_y", VIPS_ROUND_UINT((double) thumbnail->interesting_y * overall_hshrink),
+				"interesting_x", VIPS_ROUND_UINT((double) thumbnail->interesting_x / overall_vshrink),
+				"interesting_y", VIPS_ROUND_UINT((double) thumbnail->interesting_y / overall_hshrink),
 				NULL))
 			return -1;
 		in = t[14];


pszemus:
Test command:
(let's use the same parameters as in #4846 )
```
vips thumbnail 0491b5b7-55e3-4b65-a94a-bb5b155568f0 /tmp/vips.jpg 400 --height=600 --crop=specific --interesting-x=500 --interesting-y=200
```
(input: https://wptv.wpcdn.pl/devel/0491b5b7-55e3-4b65-a94a-bb5b155568f0)

As a result an image cropped to 400x600 and focused on a face located at (500,200) in the original picture is returned:
vips

pszemus:
@jcupitt can you take a look at this?

libvips/resample/thumbnail.c
```diff
@@ -990,6 +992,8 @@ vips_thumbnail_build(VipsObject *object)
 		if (!(t[13] = vips_image_copy_memory(in)) ||
 			vips_smartcrop(t[13], &t[14], crop_width, crop_height,
 				"interesting", thumbnail->crop,
+				"interesting_x", VIPS_ROUND_UINT(thumbnail->interesting_x / vshrink),
```
jcupitt: Is this correct? I think it doesn't include (for example) the JPEG shrink-on-load scale.

How about calculating this as eg.:

```C
(int) ((double) thumbnail->interesting_x * in->Xsize / thumbnail->input_width)
```

ie. find the resize ratio between the original image (`interesting_x` will be in this coordinate space) and the image we are passing to `smartcrop`, and scale the position by that factor.

`(int)` is probably safer than round -- the +0.5 could move the x/y out of image bounds, though perhaps it doesn't matter.

jcupitt:
Hello again! Sorry for the delay.

libvips/resample/thumbnail.c
pszemus: vshrink(hshrink) is calculated by `vips_thumbnail_calculate_shrink()` function in https://github.com/pszemus/libvips/blob/thumbnail_crop_specific/libvips/resample/thumbnail.c#L422-L423 as input_width(height) / target_width(height) and it is OK in my opinion, but your way explains the intention better - to scale the interesting position by an in/out resize factor.

As for the potential rounding issue - the interesting point coordinates are clipped by the `vips_smartcrop()` function to the image dimensions: https://github.com/pszemus/libvips/blob/thumbnail_crop_specific/libvips/conversion/smartcrop.c#L384-L387

jcupitt: I'm not sure -- it's called here:

https://github.com/pszemus/libvips/blob/thumbnail_crop_specific/libvips/resample/thumbnail.c#L806

And it's given the size of the image after any shrink-on-load. 

I think this will miss the crop point for large JPEG images -- try cropping a large (eg. 10k pixels across) image, selecting a point on the right. It'll crop left of where you want.

pszemus: You're right. I've found my mistake. I'll send a fix in a minute. 
Do you think calculating the overall resize ratio every time, even if it's only used for interesting=specific, is an overkill?

pszemus: Also, cropping EXIF-rotated images now fails, because in rotated images `in->Xsize` and `in->Ysize` are swapped, while `thumbnail->input_width` and `thumbnail->input_height` are not.

Fixing this is easy, but first, we need to ask ourselves an API question: Should the user specify `interesting_x` and `interesting_y` in relation to the auto-rotated image or the raw/input image?

- The first option is easier to implement, as it doesn't involve rotating the interesting coordinates. These coordinates correspond to the image as it is presented in a browser or image editor (auto-oriented).
- The latter could use `vips_autorot` output parameters (`flip` and `angle`) to adjust the interesting coordinates.

jcupitt: `smartcrop` is called on the rotated image, so I think I'd use that orientation.

Maybe (untested!):

```C
int rotated_input_width = thumbnail->swap ? thumbnail->input_height : thumbnail->input_width;
int rotated_input_height = thumbnail->swap ? thumbnail->input_width : thumbnail->input_height;
double overall_hshrink = in->Xsize / rotated_input_width;
double overall_vshrink = in->Ysize / rotated_input_height;
...
                "interesting_x", VIPS_ROUND_UINT(thumbnail->interesting_x / overall_hshrink),
                "interesting_y", VIPS_ROUND_UINT(thumbnail->interesting_y / overall_vshrink),
```

pszemus: Thanks @jcupitt . I've followed your advice and also handled the case when the auto-orientation is disabled.
Now all my cases have been addressed.

libvips/resample/thumbnail.c
```diff
@@ -983,13 +985,21 @@ vips_thumbnail_build(VipsObject *object)
 
 		g_info("cropping to %dx%d", crop_width, crop_height);
 
+		gboolean rotated = thumbnail->swap && thumbnail->auto_rotate;
+		int rotated_input_width = rotated ? thumbnail->input_height : thumbnail->input_width;
+		int rotated_input_height = rotated ? thumbnail->input_width : thumbnail->input_height;
+		double overall_hshrink = (double) in->Xsize / rotated_input_width;
+		double overall_vshrink = (double) in->Ysize / rotated_input_height;
```
jcupitt: Sorry, this is such a silly thing, but I got these names wrong (my fault).

libvips is supposed to use `shrink` to mean how much to shrink by (so 2 means make it half the size, ie. a division), and `scale` to mean how much to scale by (ie. a multiplicative factor).

These should be called `overall_hscale` and `overall_vscale`, or alternatively do:

```C
		double overall_hshrink = (double) rotated_input_width / in->Xsize;
		double overall_vshrink = (double) rotated_input_height / in->Ysize;
```

and swap the `*` for a `/` below.

jcupitt (approved):
I messed up, sorry, otherwise let's merge!

libvips/resample/thumbnail.c
pszemus: Done

jcupitt:
Great! Thank you for doing this work, @pszemus

PR #4862, opened 2026-02-13
meson: add `tests`, `tools` and `fuzz` build options

pszemus:
Add `tests`, `tools` and `fuzz` meson build options (default: true) to disable building unwanted targets.

The disabling logic was added to the included `meson.build` files due to: 
```
# these lines removed by a regexp for oss-fuzz builds, don't touch!
subdir('man')
subdir('po')
subdir('tools')
subdir('test')
subdir('fuzz')
```
at the end of the main `meson.build` file.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -764,26 +764,26 @@ endforeach
 
 subdir('libvips')
 if get_option('docs')
-  subdir('doc')
+    subdir('doc')
 endif
 if get_option('examples')
-  subdir('examples')
+    subdir('examples')
 endif
 if get_option('cplusplus')
-  subdir('cplusplus')
+    subdir('cplusplus')
 endif
 if get_option('man') and get_option('tools')
-  subdir('man')
+    subdir('man')
 endif
 if get_option('po')
-  subdir('po')
+    subdir('po')
 endif
 if get_option('tools')
-  subdir('tools')
+    subdir('tools')
 endif
 if get_option('tests')
     subdir('test')
 endif
 if get_option('fuzz')
-  subdir('fuzz')
+    subdir('fuzz')
 endif


lovell:
Thanks for the PR. Did you see the following script?

https://github.com/libvips/libvips/blob/7fab325d2d48c1b4eae4ad66de8f9ba8a1cad7ad/fuzz/oss_fuzz_build.sh#L238-L239

This file used to be in a separate OSS Fuzz repo but now it's within libvips this is probably a good time to check these options in the top-level `meson.build` file and remove the `sed` and `don't touch` comments :sweat_smile:

pszemus:
> Thanks for the PR. Did you see the following script?
> 
> https://github.com/libvips/libvips/blob/7fab325d2d48c1b4eae4ad66de8f9ba8a1cad7ad/fuzz/oss_fuzz_build.sh#L238-L239
> 
> This file used to be in a separate OSS Fuzz repo but now it's within libvips this is probably a good time to check these options in the top-level `meson.build` file and remove the `sed` and `don't touch` comments 😅

OK, so do you recommend adding meson options for all five subdirs (`man`, `po`, `tools`, `test` and `fuzz`) and disabling the first four in [fuzz/oss_fuzz_build.sh](https://github.com/libvips/libvips/blob/7fab325d2d48c1b4eae4ad66de8f9ba8a1cad7ad/fuzz/oss_fuzz_build.sh#L240)?

If so, I think `man`, `po`, `tools` and `test` should be enabled by default but `fuzz` should not. Also, building `man` when `tools` are disabled seems pointless.

lovell:
@pszemus That sounds perfect, thank you.

meson.build
```diff
@@ -772,10 +772,18 @@ endif
...
+  subdir('man')
+endif
+if get_option('po')
+  subdir('po')
+endif
+if get_option('tools')
+  subdir('tools')
+endif
+if get_option('tests')
+    subdir('test')
```
kleisauke, line 785: nit:
```suggestion
  subdir('test')
```

pszemus: The entire `meson.build` file is indented by four spaces. Maybe I should amend lines 766-789?

kleisauke: Ah, indeed. Yes, please re-indent lines 766 to 789 using four spaces. 👍

kleisauke:
Thank you!

Issue #4864, opened 2026-02-13
potential heap-buffer-overflow in vips_shrinkh_uchar_hwy


dloebl:
Found with #4863

### Reproducer

`shrinkh_repro.c`:
```c
#include <string.h>
#include <stdlib.h>
#include <vips/vips.h>

int
main(int argc, char **argv)
{
	static const char ppm[] =
		"P6\n2 2\n255\n"
		"\xff\x00\x00" "\x00\xff\x00"
		"\x00\x00\xff" "\xff\xff\x00";
	const size_t ppm_len = sizeof(ppm) - 1;
	VipsImage *in, *out;

	if (VIPS_INIT(argv[0]))
		return 1;

	void *buf = malloc(ppm_len);
	memcpy(buf, ppm, ppm_len);

	if (!(in = vips_image_new_from_buffer(buf, ppm_len, "", NULL)))
		vips_error_exit(NULL);
	if (vips_shrinkh(in, &out, 2, NULL))
		vips_error_exit(NULL);

	double d;
	vips_min(out, &d, NULL);

	g_object_unref(out);
	g_object_unref(in);
	free(buf);
	return 0;
}
```

Both libvips and the reproducer must be built with ASAN:
```
meson setup build -Db_sanitize=address
meson compile -C build

clang -fsanitize=address -g -o shrinkh_repro shrinkh_repro.c \
  $(PKG_CONFIG_PATH=build/meson-uninstalled pkg-config --cflags --libs vips)

# Linux
LD_LIBRARY_PATH=build/libvips ./shrinkh_repro
# macOS
DYLD_LIBRARY_PATH=build/libvips ./shrinkh_repro
```

### ASAN output

```
==57235==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x603000007c77 at pc 0x000105e6fba8 bp 0x00016afbe5c0 sp 0x00016afbe5b8
READ of size 4 at 0x603000007c77 thread T1
    #0 0x000105e6fba4 in N_NEON_BF16::vips_shrinkh_uchar_hwy(unsigned char*, unsigned char*, int, int, int)+0x508 (libvips.42.dylib:arm64+0x287ba4)
    #1 0x000105e6c0d0 in vips_shrinkh_uchar_vector_gen+0x460 (libvips.42.dylib:arm64+0x2840d0)
    #2 0x0001060b8b2c in vips_region_generate+0x138 (libvips.42.dylib:arm64+0x4d0b2c)
[...]

PR #4866, opened 2026-02-13
shrinkh: fix possible OOB read in Highway path

kleisauke:
Input regions are not guaranteed to be padded. The Highway path used `LoadU` for every output pixel, which can read past the valid pixel data when the number of band elements in a scanline (`width * bands`) is smaller than the vector-length.

Request one extra input pixel on the right so the Highway path can safely process a full SIMD vector.

While we're here, constrain the Highway path to <= 4 bands, since images with more than 4 bands are not handled correctly.

Resolves: #4864.
Targets the 8.18 branch.

### DIFF
--- a/libvips/resample/shrinkh.c
+++ b/libvips/resample/shrinkh.c
@@ -374,6 +374,7 @@ vips_shrinkh_build(VipsObject *object)
 	 */
 #ifdef HAVE_HWY
 	if (in->BandFmt == VIPS_FORMAT_UCHAR &&
+		in->Bands <= 4 &&
 		vips_vector_isenabled()) {
 		generate = vips_shrinkh_uchar_vector_gen;
 		g_info("shrinkh: using vector path");


kleisauke:
Marking this as a draft, since some of these analyses were incorrect. The reproducer seems to pass when doing:
```diff
--- a/shrinkh_repro.c
+++ b/shrinkh_repro.c
@@ -6,7 +6,7 @@ int
 main(int argc, char **argv)
 {
 	static const char ppm[] =
-		"P6\n2 2\n255\n"
+		"P6\n2 2\n255"
 		"\xff\x00\x00" "\x00\xff\x00"
 		"\x00\x00\xff" "\xff\xff\x00";
 	const size_t ppm_len = sizeof(ppm) - 1;
```

So, it looks like a bug in `ppmload` instead :sweat_smile:. I could also not reproduce this when doing:
```python
#!/usr/bin/env python3
import pyvips

data = bytes([
    255, 0, 0,
    0, 255, 0,
    0, 0, 255,
    255, 255, 0
])
im = pyvips.Image.new_from_memory(data, 2, 2, 3, 'uchar')
im2 = im.shrinkh(2)
print(im2.avg())
```

kleisauke:
Ah, please ignore my previous comment. That isn't a valid PPM, and you'll need to build Python with ASAN instrumentation to reproduce this using `pyvips.Image.new_from_memory()`. 

Here's a reproducer without `ppmload`:
```c
#include <vips/vips.h>

int
main(int argc, char **argv)
{
	static const unsigned char data[] = "\xff\x00\x00" "\x00\xff\x00"
										"\x00\x00\xff" "\xff\xff\x00";
	const size_t length = sizeof(data) - 1;
	VipsImage *in, *out;

	if (VIPS_INIT(argv[0]))
		return 1;

	if (!(in = vips_image_new_from_memory(data, length, 2, 2, 3, VIPS_FORMAT_UCHAR)))
		vips_error_exit(NULL);
	if (vips_shrinkh(in, &out, 2, NULL))
		vips_error_exit(NULL);

	double d;
	vips_avg(out, &d, NULL);
	printf("%.2f\n", d);

	g_object_unref(out);
	g_object_unref(in);
	return 0;
}

```

kleisauke:
The regression test (commit 83a4caeda8fa2ad3670db43ce439f411c9b55add) uncovered an issue with AVX2. https://github.com/google/highway/blob/1.3.0/hwy/ops/x86_256-inl.h#L3636-L3642

I'll revert back to commit a94c8c0e0f735ee0528de82a1ae94d1a10342899 and update the commit message accordingly.

kleisauke:
... actually, it's probably easier to just request one extra input pixel on the right. I've updated the PR to do that.

Issue #4872, opened 2026-02-14
potential heap-buffer-overflow in vips_invertlut


dloebl:
Found with #4863

### Reproducer
libvips must be built with ASAN:
```
meson setup build -Db_sanitize=address
meson compile -C build
```
And then the following `invertlut` command crashes:
```
printf 'P4\n2UTF-8\0\0\0\0\0\0\0\0-\xb6\xb6\0\xff\0\0nPhist_cum\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\xb6X' > input.dat
./build/tools/vips invertlut input.dat out.v
```

### ASAN output
```
==85596==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602000082f58 at pc 0x0001036801d4 bp 0x00016d8eea60 sp 0x00016d8eea58
READ of size 8 at 0x602000082f58 thread T0
    #0 0x0001036801d0 in vips_invertlut_build+0x9a8 (libvips.42.dylib:arm64+0x5541d0)
    #1 0x0001035a26ec in vips_object_build+0xd8 (libvips.42.dylib:arm64+0x4766ec)
    #2 0x0001035d1cbc in vips_call_argv+0x168 (libvips.42.dylib:arm64+0x4a5cbc)
    #3 0x0001025132b0 in main+0xb30 (vips:arm64+0x1000032b0)
    #4 0x000182279d50  (<unknown module>)
```

### Environment
- libvips master (7fab325d2)
- macOS, M3 Pro (aarch64)
-  Highway enabled

dloebl:
Resolved with #4882

PR #4882, opened 2026-02-15
invertlut: check height before interpolating values #4872

lovell:
A (not particularly useful) lookup table with only 1 row cannot use interpolation so instead copy the single value.

Fixes https://github.com/libvips/libvips/issues/4872

### DIFF
--- a/libvips/create/invertlut.c
+++ b/libvips/create/invertlut.c
@@ -229,14 +229,19 @@ vips_invertlut_build_create(VipsInvertlut *lut)
 			if (j == -1)
 				j = 0;
 
-			/* Interpolate k as being between row data[j] and row
-			 * data[j + 1].
-			 */
-			irange = lut->data[j + 1][b + 1] - lut->data[j][b + 1];
-			orange = lut->data[j + 1][0] - lut->data[j][0];
+			if (height > 1) {
+				/* Interpolate k as being between row data[j] and row
+				 * data[j + 1].
+				 */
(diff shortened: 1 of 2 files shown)


libvips/create/invertlut.c
```diff
@@ -229,14 +229,18 @@ vips_invertlut_build_create(VipsInvertlut *lut)
...
-			 */
-			irange = lut->data[j + 1][b + 1] - lut->data[j][b + 1];
-			orange = lut->data[j + 1][0] - lut->data[j][0];
-
-			lut->buf[b + k * bands] = lut->data[j][0] +
-				orange * ((ki - lut->data[j][b + 1]) / irange);
+			if (height > 1) {
+				/* Interpolate k as being between row data[j] and row
+				* data[j + 1].
+				*/
```
kleisauke:
```suggestion
				 * data[j + 1].
				 */
```

libvips/create/invertlut.c
```diff
@@ -229,14 +229,18 @@ vips_invertlut_build_create(VipsInvertlut *lut)
...
+				irange = lut->data[j + 1][b + 1] - lut->data[j][b + 1];
+				orange = lut->data[j + 1][0] - lut->data[j][0];
+
+				lut->buf[b + k * bands] = lut->data[j][0] +
+					orange * ((ki - lut->data[j][b + 1]) / irange);
+			} else {
```
kleisauke:
```suggestion
			}
			else {
```

Issue #4871, opened 2026-02-14
potential heap-buffer-overflow in vips_maplut_gen


dloebl:
Found with #4863

### Reproducer
libvips must be built with ASAN:
```
meson setup build -Db_sanitize=address
meson compile -C build
```
And then the following `hist_equal` command crashes:
```
printf 'Pfthe\xb7\xb7\xb7\xb7\xb7\xb7\xb7\xb7\xb7\xb7\xb7\xb7\xb7\xb7\xb7\xb7\xb7\xb7\xb7\xb7\xb7te\nP48P\x8ad\xff\xff\x01\0\0\xcd\xf5\0*\0\x042\n\xff\n+1\0UTF-8ynd6t' > input.dat
./build/tools/vips hist_equal input.dat out.v
```

### ASAN output
```
==83882==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x00010e9547fc at pc 0x000105836030 bp 0x00016b8c6340 sp 0x00016b8c6338
READ of size 4 at 0x00010e9547fc thread T3
    #0 0x00010583602c in vips_maplut_gen+0x25998 (libvips.42.dylib:arm64+0x41602c)
    #1 0x0001058f0ba4 in vips_region_generate+0x138 (libvips.42.dylib:arm64+0x4d0ba4)
    #2 0x0001058f0924 in vips_region_prepare+0x1f8 (libvips.42.dylib:arm64+0x4d0924)
    #3 0x0001058a68fc in vips_image_write_gen+0x30 (libvips.42.dylib:arm64+0x4868fc)
    #4 0x0001058f0ba4 in vips_region_generate+0x138 (libvips.42.dylib:arm64+0x4d0ba4)
    #5 0x0001058f0924 in vips_region_prepare+0x1f8 (libvips.42.dylib:arm64+0x4d0924)
    #6 0x00010574d838 in vips_copy_gen+0x30 (libvips.42.dylib:arm64+0x32d838)
    #7 0x0001058f0ba4 in vips_region_generate+0x138 (libvips.42.dylib:arm64+0x4d0ba4)
    #8 0x0001058f0924 in vips_region_prepare+0x1f8 (libvips.42.dylib:arm64+0x4d0924)
    #9 0x0001058a68fc in vips_image_write_gen+0x30 (libvips.42.dylib:arm64+0x4868fc)
    #10 0x0001058f0ba4 in vips_region_generate+0x138 (libvips.42.dylib:arm64+0x4d0ba4)
[...]

PR #4883, opened 2026-02-15
maplut: ensure lookup table index is unsigned #4871

lovell:
This is a possible fix for #4871 and related code paths that deal with LUT indexes.

This change relies on the existing overflow check to clip the index to the LUT size. An alternative might be to continue to use a signed int but add an underflow check. (There are probably other solutions too.)


libvips/histogram/maplut.c
jcupitt: I think this needs to be unsigned as well :(

jcupitt: And this one. These two cases can be expanded for `IN` being `UINT` and trigger the same error.

Issue #4881, opened 2026-02-15
unpremultiply: unchecked `--alpha-band` causes out-of-bounds access in `vips_unpremultiply_gen()`


Niebelungen-D:
## Bug report

**Describe the bug**
In `libvips/libvips/conversion/unpremultiply.c`, `alpha_band` is user-controllable and is used as a direct pixel index in the hot path, but `vips_unpremultiply_build()` does not validate `alpha_band < in->Bands`.

1) The operation exposes a very wide range for `alpha_band`:

```c
/* vips_unpremultiply_class_init() */
VIPS_ARG_INT(class, "alpha_band", 116,
    _("Alpha band"),
    _("Unpremultiply with this alpha"),
    VIPS_ARGUMENT_OPTIONAL_INPUT,
    G_STRUCT_OFFSET(VipsUnpremultiply, alpha_band),
    0, 100000000, 3);
```

2) Build path sets a default value only when the argument is unset, but performs no bounds check for explicit input:

```c
/* vips_unpremultiply_build() */
if (!vips_object_argument_isset(object, "alpha_band"))
    unpremultiply->alpha_band = in->Bands - 1;
/* no check that alpha_band is in [0, in->Bands - 1] */
```

3) Generator path dereferences `p[alpha_band]` and writes `q[alpha_band]` directly (`UNPRE_MANY` / `FUNPRE_MANY`):

```c
#define UNPRE_MANY(TYPE, IN, OUT) \
{ \
    ...
    TYPE alpha = p[alpha_band]; \
    ...
    for (j = 0; j < alpha_band; j++) \
        q[j] = alpha == 0 ? 0 : (p[j] * max_alpha) / alpha; \
    ...
}
```

Trigger condition: use a normal 3-band image and pass `--alpha-band 10`.  
The code enters `UNPRE_MANY` (non-4-band path), reads `p[10]` and later writes `q[10]`, causing out-of-bounds access. ASAN reports `heap-buffer-overflow` in `vips_unpremultiply_gen()`.

**To Reproduce**
1. Build ASAN:
```bash
cd /home/niebelungen/workspace/bug_hunting/libvips
meson setup build-asan-only -Ddebug=true -Db_sanitize=address
meson compile -C build-asan-only
```
2. Prepare a normal 3-band input:
```bash
LSAN_OPTIONS=detect_leaks=0 ASAN_OPTIONS=detect_leaks=0 \
[...]

PR #4886, opened 2026-02-17
unpremultiply: check alpha_band is in range #4881

lovell:
Fixes https://github.com/libvips/libvips/issues/4881


libvips/conversion/unpremultiply.c
```diff
@@ -288,6 +288,11 @@ vips_unpremultiply_build(VipsObject *object)
 	if (!vips_object_argument_isset(object, "alpha_band"))
 		unpremultiply->alpha_band = in->Bands - 1; // FIXME: Invalidates operation cache
 
+	if (unpremultiply->alpha_band >= in->Bands) {
```
kleisauke:
```suggestion
	else if (unpremultiply->alpha_band >= in->Bands) {
```

PR #4870, opened 2026-02-14
Ensure delimiter list is consistent when parsing strings

lovell:
The loops to count tokens were using a different set of delimiters than the loops to process them, leading to heap overflow.

Fixes #4868

PR #4892, opened 2026-02-18
define delimiters once for vips_break_token()

jcupitt:
Following on from #4870, define the delimiters once, so they can't get out of step again. A little safer.


jcupitt:
My thought was to avoid touching the "stable" branch more than absolutely necessary.

PR #4863, opened 2026-02-13
add vips fuzzer covering the operation API

dloebl:
Adds a fuzz target that covers the complete libvips operation API, similar to how the standalone `vips` tool works. It parses an operation name and arguments from the fuzzer input and runs the operation.

**ToDo:**
- [x] Include optional arguments as well
- [x] Add a script to generate the fuzzer dictionary at runtime

PR #4893, opened 2026-02-20
CI: run CIFuzz with both ASAN and UBSAN

dloebl:
Noticed while working on #4863: https://github.com/libvips/libvips/actions/runs/22092771573/job/64029163205#step:5:48
Previously, CIFuzz only ran with the default sanitizer (`address`). This adds a matrix strategy so that fuzzing is run separately with both ASAN and UBSAN.
For reference: https://google.github.io/oss-fuzz/getting-started/continuous-integration/

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -4,22 +4,26 @@ permissions: {}
 jobs:
   Fuzzing:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        sanitizer: [address, undefined]
     steps:
     - name: Build Fuzzers
       id: build
       uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
       with:
         oss-fuzz-project-name: 'libvips'
-        dry-run: false
+        sanitizer: ${{ matrix.sanitizer }}
     - name: Run Fuzzers
       uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
       with:
         oss-fuzz-project-name: 'libvips'
+        sanitizer: ${{ matrix.sanitizer }}
         fuzz-seconds: 600
-        dry-run: false
     - name: Upload Crash
       uses: actions/upload-artifact@v6
       if: failure() && steps.build.outcome == 'success'
       with:
-        name: artifacts
+        name: artifacts-${{ matrix.sanitizer }}
         path: ./out/artifacts


.github/workflows/cifuzz.yml
```diff
@@ -4,22 +4,26 @@ permissions: {}
...
         oss-fuzz-project-name: 'libvips'
-        dry-run: false
+        sanitizer: ${{ matrix.sanitizer }}
     - name: Run Fuzzers
       uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
       with:
         oss-fuzz-project-name: 'libvips'
+        sanitizer: ${{ matrix.sanitizer }}
         fuzz-seconds: 600
-        dry-run: false
```
dloebl, line 19: `dry-run: false` is the default: https://google.github.io/oss-fuzz/getting-started/continuous-integration/#optional-configuration

.github/workflows/cifuzz.yml
```diff
@@ -4,22 +4,26 @@ permissions: {}
 jobs:
   Fuzzing:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
dloebl, line 8: `fail-fast: false` so the other CIFuzz job keeps running (even if the other one failed).

kleisauke:
Note to self: ThreadSanitizer isn't supported by CIFuzz yet, see:
https://github.com/google/oss-fuzz/issues/5272#issuecomment-3456966272

dloebl:
> Note to self: ThreadSanitizer isn't supported by CIFuzz yet, see: https://github.com/google/oss-fuzz/issues/5272#issuecomment-3456966272

Ok I see! It's interesting that they recommend running TSAN on the corpus instead of fuzzing - maybe something we can try in libvips CI as well. So basically fetch libvips public OSS Fuzz corpus from GCS and then run all fuzz targets against it with TSAN (I can try that in a follow-up PR).

Also related: Have we considered MSAN and i386 for the [OSS Fuzz project config](https://github.com/google/oss-fuzz/blob/master/projects/libvips/project.yaml)? It's both disabled by default, but might catch some more bugs. For reference:
https://google.github.io/oss-fuzz/getting-started/new-project-guide/#sanitizers
https://google.github.io/oss-fuzz/getting-started/new-project-guide/#architectures

kleisauke:
> Have we considered MSAN and i386 for the [OSS Fuzz project config](https://github.com/google/oss-fuzz/blob/master/projects/libvips/project.yaml)?

I looked into MSan a while ago, but it requires building all libraries from source ([which we currently don't do](https://github.com/google/oss-fuzz/blob/88d88978c8d3ff34afe46f8d8835882d04c53e6e/projects/libvips/Dockerfile#L24-L27)), see e.g. issue https://github.com/google/oss-fuzz/issues/6294.

Enabling i386 fuzzing is more straightforward, as it only requires installing packages like `glib2.0-dev:i386`. However, Debian (and therefore Ubuntu) is likely to phase out i386 support in the near future.

PR #4905, opened 2026-02-24
add VIPS_UNLIMITED to disable DoS checks

### LINKED ISSUE #4903: Add VIPS_UNLIMITED env var
Many loaders now have `unlimited` flags you can set to remove denial of service limits, for example `tiffload`, `svgload`, etc.

These flags are very useful for web services, but will prevent some desktop programs from opening some files. OME-TIFF images, for example, will often trigger the `tiffload` limit.

Rather than modifying all these programs to always set the `unlimited` flag if the load operation supports it, it would be simpler to have a `VIPS_UNLIMITED` env var which disables all denial of service protection. Programs which assume that the user trusts the image can simply set this on startup.

jcupitt:
VIPS_UNLIMITED env var, --vips-unlimited CLI arg, and vips_unlimited_set()/_get() to disable denial of service checks

Closes #4903


jcupitt:
Found during nip4 dev -- the `unlimited` checks we've added were stopping a lot of useful images from loading, especially the horrible OME-TIFF variants.

libvips/resample/thumbnail.c
```diff
@@ -970,6 +970,13 @@ vips_thumbnail_build(VipsObject *object)
 		}
 	}
 
+
+	printf("vips_thumbnail_build:\n");
+	printf("\tvshrink = %g\n", vshrink);
+	printf("\thshrink = %g\n", hshrink);
+	printf("\tinput_width = %d\n", thumbnail->input_width);
+	printf("\tinput_height = %d\n", thumbnail->input_height);
+
```
kleisauke: Stray debug prints.
```suggestion
```

libvips/histogram/maplut.c
kleisauke, line 1: The changes in this file were already applied in commit ea7e608844b85f4d296821e59b2483369b301b53, so we may want to merge the `8.18` branch back into `master`.

jcupitt: Good idea, it's been a while.

PR #4906, opened 2026-02-24
vips_window_take: int underflow

jcupitt:
This could underflow for very small mapped images on 64-bit machines, causing images to become offset and triggering assert fails in dev builds.

### DIFF
--- a/libvips/iofuncs/window.c
+++ b/libvips/iofuncs/window.c
@@ -392,7 +392,8 @@ vips_window_take(VipsWindow *window, VipsImage *im, int top, int height)
 	gint64 window_bytes = height * line_bytes;
 	gint64 margin_bytes =
 		(VIPS_MAX(window_bytes, vips__window_bytes) - window_bytes) / 2;
-	gint64 margin_lines = VIPS_CLIP(0, margin_bytes / line_bytes, 10000);
+	gint64 margin_lines =
+		VIPS_CLIP(0, margin_bytes / line_bytes, VIPS_MAX_COORD);
 
 	top -= margin_lines;
 	height += margin_lines * 2;


jcupitt:
I should have posted a reproducer. With git 8.18:

```
$ vips black x.v 2 2
memory: high-water mark 225 bytes
$ vips invert x.v x2.v
**
VIPS:ERROR:../libvips/iofuncs/region.c:1708:vips_region_prepare: assertion failed: (vips_rect_includesrect(&reg->valid, &save))
Bail out! VIPS:ERROR:../libvips/iofuncs/region.c:1708:vips_region_prepare: assertion failed: (vips_rect_includesrect(&reg->valid, &save))
Aborted (core dumped)
```

The window margin is overflowing int32, so the window top is moved down and misses the target. 

Asserts fail in debug mode. There's no crash in release mode, but calculated results will be wrong.

jcupitt:
Oh, good idea!

This bug is only for mmapped files, so I think binary ppm is only format we fuzz and which might trigger this, but asserts might catch other stuff, of course.

jcupitt:
I swapped the 10,000 limit on the window margin for `VIPS_MAX_COORD`. It feels less arbitrary, and will still prevent int over/underflow.

jcupitt:
It's interesting how many of these recent bugs would have been impossible in Rust. An idea for another day!

kleisauke:
> It's interesting how many of these recent bugs would have been impossible in Rust. An idea for another day!

Indeed, that's an interesting idea. I've been following the development of https://github.com/libjxl/jxl-rs recently, and it's great to see what's possible there. AFAIK, it's still about 2-3x slower than libjxl, but it's probably too early to tell, since improvements are ongoing.

On another note, Meson is also making good progress in supporting Rust <--> C interoperability. :)

jcupitt:
The vips8 rewrite was 2010 -- 2015 and I think I couldn't face another one so soon. It's a shame, it was just before rust became the obvious choice. If I had my time again ...

PR #4909, opened 2026-02-26
revise update centre sampling

jcupitt:
Scale the input offset by the scale factor.

Thanks @chcunningham

See https://github.com/libvips/libvips/discussions/4843


jcupitt:
With this 3x3 pixel test PNG:

3x3

And this upscaler:

```python
#!/usr/bin/env python3

import sys

from PIL import Image
import pyvips
import numpy as np

if len(sys.argv) != 3:
    print(f"usage: {sys.argv[0]}: image-file percent-scale")
    sys.exit(1)

filename = sys.argv[1]
scale = int(sys.argv[2]) / 100.0

pil = Image.open(filename)
width, height = pil.size
new_width = int(width * scale)
new_height = int(height * scale)
pil = pil.resize((new_width, new_height), Image.LANCZOS)
pil = pyvips.Image.new_from_array(np.array(pil))

pil.write_to_file("pil.png")

vips = pyvips.Image.new_from_file(filename)
vips = vips.resize(scale, kernel="lanczos3")

vips.write_to_file("vips.png")

comp = pil.join(vips, "vertical")
# scale up to make pixels visible
comp = comp.zoom(50, 50)
comp.set_type(pyvips.GValue.gint_type, "loop", 0)
comp.set_type(pyvips.GValue.array_int_type, "delay", [500] * 2)
comp.set_type(pyvips.GValue.gint_type, "page-height", pil.height * 50)
print(f"writing compare.gif ...")
```

I can run:

```
$ ./upscale.py ~/pics/3x3.png 300
writing compare.gif ...
$
```

With git master libvips I see:

compare

There's a residual shift. With this PR I see:

compare

You can compare various scaling percentages to verify behaviour.

They still differ slightly -- libvips seems to have better abyss handling (you can see the PIL version "spreads" the white pixel a bit at the bottom) and lanczos2 vs lanzos3 makes a small difference to the shape of the black pixel.

jcupitt:
At 700% you can see the change in the shape of the black pixel more clearly:

compare

lovell (approved):
LGTM, thanks John.

PR #4916, opened 2026-02-26
OSS-Fuzz: enable GLib cast checks and asserts

kleisauke:
Since cast and assertion failures often indicate underlying bugs.

Context: https://github.com/libvips/libvips/pull/4906#pullrequestreview-3852809998.


kleisauke:
Marked as draft due to dependence on #4917.

lovell (approved):

PR #4924, opened 2026-03-02
feat: expose progressive jxl settings

mertalev:
libjxl has not one but several knobs to configure how progressive the output is, but none of them are exposed in libvips. This PR exposes all the relevant settings for downstream use. It's worth noting that cjxl exposes these individually, but also has a higher level `--progressive` flag that (as of [this](https://github.com/libjxl/libjxl/pull/4258) PR) sets good values for these internally. I'm not sure if libvips should have something similar or if it should just expose the actual encoding parameters.

The defaults are the same as what libjxl normally uses with a minor caveat: `responsive` is set true by default in the encoder if it's using lossy modular mode (as opposed to lossless modular or VarDCT, which is always lossy). However, modular mode is only ever used for lossless encoding with libvips since there's no knob exposed to enable it for lossy encoding. I set it to 0 since this seems consistent with how libvips handles defaults, but can set it to -1 instead if preferred (e.g. if an option to force modular mode is exposed in libvips in the future).

These options were all added in libjxl 0.7 in 2021, which is also the minimum for libvips, so there should be no version check needed.

### DIFF
--- a/libvips/foreign/jxlsave.c
+++ b/libvips/foreign/jxlsave.c
@@ -1211,7 +1211,7 @@ vips_foreign_save_jxl_class_init(VipsForeignSaveJxlClass *class)
 	VIPS_ARG_BOOL(class, "progressive", 17,
 		_("Progressive"),
 		_("Enable progressive encoding"),
-		VIPS_ARGUMENT_OPTIONAL_INPUT | VIPS_ARGUMENT_DEPRECATED,
+		VIPS_ARGUMENT_OPTIONAL_INPUT,
 		G_STRUCT_OFFSET(VipsForeignSaveJxl, interlace),
 		FALSE);
 


jcupitt:
Hi @mertalev!

This looks great. 

I think my instinct would be for a higher level control, if possible. libvips is not aiming to duplicate the whole of the cjxl interface.

Ideally `jxlsave` should be stable over many versions of libjxl and not affected by API changes there. For example, when the new Rust based lib comes along, libvips users should be able to swap to it without code changes. Hopefully.

mertalev:
That's fair! I was happy with the new values that cjxl's `--progressive` uses and those values are what I'd actually use in practice. I could do the same here with a single `progressive` option that sets these values internally? It'd make it a lot easier to use as well.

If you'd like to test as well, [this](https://jxl-oxide.tirr.dev/demo/index.html) demo is super useful to see exactly how it changes the decode process.

jcupitt:
That sounds great. Yes, the `jxlsave` `--Q` setting works like that, the rules are adapted from the ones in the `cjxl` sources.

lovell:
This is great, thank you Mert. Other libvips savers (e.g. JPEG, PNG) refer to "progressive" as "interlace", but I'm unsure if we should be consistent with the libvips API or libjxl API?

mertalev:
My preference is "progressive" since this is the common and canonical term for it - nobody really uses "interlaced" for JPEG XL images. That being said, I'm fine with either if you do prefer calling it "interlace".

jcupitt:
> Other libvips savers (e.g. JPEG, PNG) refer to "progressive" as "interlace", but I'm unsure if we should be consistent with the libvips API or libjxl API?

Oh good point, we should probably call it `interlace` for consistency with other savers.

How about also having `progressive`, but tagging it as deprecated? That way people could use either.

mertalev:
Good idea! I changed it to use interlace and accept progressive as an alias.

libvips/foreign/foreign.c
```diff
@@ -2611,13 +2611,17 @@ vips_jxlload_source(VipsSource *source, VipsImage **out, ...)
...
+ *
  * ::: tip "Optional arguments"
  *     * @tier: `gint`, decode speed tier
  *     * @distance: `gdouble`, maximum encoding error
  *     * @effort: `gint`, encoding effort
  *     * @lossless: `gboolean`, enables lossless compression
  *     * @Q: `gint`, quality setting
  *     * @bitdepth: `gint`, image bitdepth
+ *     * @interlace: `gboolean`, enables progressive encoding
+ *     * @progressive: `gboolean`, alias of @interlace
```
jcupitt, line 2624: We don't document deprecated options, so I think we should either remove the `DEPRECATED` flag from the arg, or remove the doc line. Either is fine with me, you choose.

mertalev: I think it's useful to document it so people are aware of the alias, so I removed the deprecated flag

goodusername123:
There is something here I feel might be important to consider and that is `progressive` + `lossless` = noticeably bloated file sizes. Maybe this should be noted somewhere in some way? also AFAIK lossy Modular and VarDCT are both fine file size wise it's exclusively lossless Modular which causes size bloat, and it's not an issue with libjxl but just something inherent to doing the transforms required for progressiveness in a fully lossless capacity.

jcupitt:
I forgot, this is an API change so we'll need a note in the changelog. I'll add it in another PR.

libvips/foreign/jxlsave.c
```diff
@@ -719,6 +720,16 @@ vips_foreign_save_jxl_save_page(VipsForeignSaveJxl *jxl,
 		JXL_ENC_FRAME_SETTING_EFFORT, jxl->effort);
 	JxlEncoderSetFrameLossless(frame_settings,
 		jxl->lossless);
+	if (jxl->interlace) {
+		JxlEncoderFrameSettingsSetOption(frame_settings,
+			JXL_ENC_FRAME_SETTING_PROGRESSIVE_DC, 1);
```
goodusername123, line 725: This also supports being set to a value of `2` however setting it to such won't make images appear/render sooner compared to `1` it will just result in the very early render looking nicer (proper downsample preview rather then a blocky looking one) for the cost of extra file size. Just noting this here as possibly something to consider, although `-p` in cjxl didn't consider `2` to be a valuable enough tradeoff.

mertalev: I think there's even been discussion on not exposing 2 in cjxl because it's so rarely worth it.

Issue #4321, opened 2024-12-17
Random loss of ICC profiless due to PR #3623


DarthSim:
Hey team 👋

I found a pretty nasty libvips behavior appeared because of https://github.com/libvips/libvips/pull/3623 (more specifically: https://github.com/libvips/libvips/pull/3623/commits/0d665ba5a39b6d1f70b6c57e8c25745ce1ca9222). This PR adds stripping ICC profiles when converting from/to `scRGB`. While this seems correct at first glance, this also makes vips behave unpredictably.

For example, you import `ICC` with `vips_icc_import` and then convert the image to `sRGB`. Since the conversion from `XYZ`/`LAB` to `sRGB` goes through `vips_sRGB2scRGB_build`, it removes the ICC profile from the image. The same happens if you apply `vips_sharpen` and convert the image back to `sRGB`.

DarthSim:
It indeed is fixed 👍
I see that you got rid of ICC removal during scRGB->sRGB conversion in this commit: https://github.com/libvips/libvips/commit/1db77accf3e2e2a16eb413a9b2519febf8100f7f

PR #4928, opened 2026-03-04
feat: cicp and hdr support

mertalev:
This is a continuation of @Starbix's great work in #4818. It uses that PR as a base, implementing jxlsave, making a number of bug fixes that affected color accuracy and HDR, implementing more CICP values and adding tests.

Regarding the design questions noted in the PR, I think libjxl's CICP support is mostly a non-issue since `JXL_PRIMARIES_CUSTOM` exists. I also decided that the "Other formats support CICP via ICC v4.4 profiles." bullet point is out of scope for this PR because it's hard to get that right and shouldn't block the savers that don't need it. There is still a question about how to handle images that have both ICC and CICP metadata. The current implementation prioritizes CICP if it's using an HDR curve and otherwise prefers ICC.

You can see [pics.zip](https://github.com/user-attachments/files/25735884/pics.zip) for the same AVIF original and the outputs as of this PR. Ultra HDR in particular is significantly improved, though it added some overhead in calculating the peak brightness.

I'd really love to land this feature and am happy to make any changes needed to make it happen.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -12,6 +12,12 @@ date-tbd 8.19.0
 - conversion: non-separable blending modes added [kstanikviacbs]
 - dECMC: fix incorrect results [jcupitt]
 - jxlsave: add `interlace` option [mertalev]
+- add vips_CICP2scRGB [Starbix] [mertalev]
+- jxlload, heifload, pngload: read CICP metadata [Starbix] [mertalev]
+- jxlsave, heifsave, pngsave: write CICP metadata and pixel data
+  [Starbix] [mertalev]
+- jpegsave, uhdrsave: convert CICP image with HDR transfer to scRGB for HDR
+  [Starbix] [mertalev]
 
 18/3/26 8.18.1
 


mertalev:
I made some more fixes for edge-cases and I think it's working well for the files I've tested so far across different output formats. This is of course a large PR with a lot of things that can go wrong (though the test coverage helps), but I wanted to make sure the overall approach is sound and working correctly. I'm happy to break it up if we're aligned on the approach.

jcupitt:
Nice! I'll try to find some time tomorrow for testing.

libvips/foreign/pngsave.c
```diff
@@ -140,9 +139,23 @@ vips_foreign_save_png_build(VipsObject *object)
...
+	else {
+		VipsBandFormat target_format =
+			png->bitdepth > 8 ? VIPS_FORMAT_USHORT : VIPS_FORMAT_UCHAR;
+
+		/* Cast in down to target format if we can.
+		 */
+		if (vips_cast(in, &x, target_format, NULL)) {
+			g_object_unref(in);
+			return -1;
+		}
```
kleisauke: Could you revert this? It seems the rebase may have gone wrong, since this was removed in commit f18cb4b04ea35c2f3066d69d10a00a412804cddc.

libvips/colour/colourspace.c
```diff
@@ -494,6 +495,24 @@ static VipsColourRoute vips_colour_routes[] = {
 	{ OKLCH, OKLAB, { vips_Oklch2Oklab, NULL } },
 	{ OKLCH, OKLCH, { vips_cast_float, NULL } },
 
+	{ CICP, XYZ, {vips_CICP2scRGB, vips_scRGB2XYZ, NULL}, },
+	{ CICP, LAB, {vips_CICP2scRGB, vips_scRGB2XYZ, vips_XYZ2Lab, NULL} },
+	{ CICP, LABQ, {vips_CICP2scRGB, vips_scRGB2XYZ, vips_XYZ2Lab, vips_Lab2LabQ, NULL} },
+	{ CICP, LCH, {vips_CICP2scRGB, vips_scRGB2XYZ, vips_XYZ2Lab, vips_Lab2LCh, NULL} },
+	{ CICP, CMC, {vips_CICP2scRGB, vips_scRGB2XYZ, vips_XYZ2Lab, vips_Lab2LCh, vips_LCh2CMC, NULL} },
+	{ CICP, LABS, {vips_CICP2scRGB, vips_scRGB2XYZ, vips_XYZ2Lab, vips_Lab2LabS, NULL} },
+	{ CICP, CMYK, {vips_CICP2scRGB, vips_scRGB2XYZ, vips_XYZ2CMYK, NULL} },
```
kleisauke: Formatting looks odd here.
```suggestion
	{ CICP, XYZ, { vips_CICP2scRGB, vips_scRGB2XYZ, NULL }, },
	{ CICP, LAB, { vips_CICP2scRGB, vips_scRGB2XYZ, vips_XYZ2Lab, NULL } },
	{ CICP, LABQ, { vips_CICP2scRGB, vips_scRGB2XYZ, vips_XYZ2Lab, vips_Lab2LabQ, NULL } },
	{ CICP, LCH, { vips_CICP2scRGB, vips_scRGB2XYZ, vips_XYZ2Lab, vips_Lab2LCh, NULL } },
	{ CICP, CMC, { vips_CICP2scRGB, vips_scRGB2XYZ, vips_XYZ2Lab, vips_Lab2LCh, vips_LCh2CMC, NULL } },
	{ CICP, LABS, { vips_CICP2scRGB, vips_scRGB2XYZ, vips_XYZ2Lab, vips_Lab2LabS, NULL } },
	{ CICP, CMYK, { vips_CICP2scRGB, vips_scRGB2XYZ, vips_XYZ2CMYK, NULL } },
```

libvips/colour/colourspace.c
```diff
@@ -494,6 +495,24 @@ static VipsColourRoute vips_colour_routes[] = {
...
+	{ CICP, scRGB, { vips_CICP2scRGB, NULL } },
+	{ CICP, sRGB, { vips_CICP2scRGB, vips_scRGB2sRGB, NULL } },
+	{ CICP, HSV, { vips_CICP2scRGB, vips_scRGB2sRGB, vips_sRGB2HSV, NULL } },
+	{ CICP, BW, { vips_CICP2scRGB, vips_scRGB2BW, NULL } },
+	{ CICP, RGB16, { vips_CICP2scRGB, vips_scRGB2RGB16, NULL } },
+	{ CICP, GREY16, { vips_CICP2scRGB, vips_scRGB2BW16, NULL } },
+	{ CICP, YXY, { vips_CICP2scRGB, vips_scRGB2XYZ, vips_XYZ2Yxy, NULL } },
+	{ CICP, OKLAB, { vips_CICP2scRGB, vips_scRGB2XYZ, vips_XYZ2Oklab, NULL } },
+	{ CICP, OKLCH, { vips_CICP2scRGB, vips_scRGB2XYZ, vips_XYZ2Oklab, vips_Oklab2Oklch, NULL } },
+	{ CICP, CICP, { NULL } },
```
kleisauke: Perhaps add a comment explaining why this is needed? I noticed this is done for CMYK -> CMYK above, but I'm not sure whether the same reasoning applies here.
https://github.com/libvips/libvips/blob/fe420cf3ac49c8dbf7b8093d988a4a7e14d8beeb/libvips/colour/colourspace.c#L332-L333

libvips/colour/uhdr2scRGB.c
kleisauke, line 1: Do you think it might make sense to split this change into a separate PR? It looks like a bug fix.

kleisauke (commented):
I found a couple of things, I'll do another round of review later.

libvips/colour/uhdr2scRGB.c
mertalev: Yup! The change was to make the Ultra HDR output look like the other HDR outputs, but it can be a separate PR.

libvips/colour/CICP2scRGB.c
```diff
@@ -0,0 +1,640 @@
...
+	colour->n = 1;
+	colour->in = (VipsImage **) vips_object_local_array(object, 1);
+
+	if (cicp->in) {
+
+		if (cicp->in->BandFmt != VIPS_FORMAT_UCHAR &&
+			cicp->in->BandFmt != VIPS_FORMAT_USHORT) {
+			vips_error(class->nickname, "%s", _("image must be uchar or ushort"));
+			return -1;
+		}
```
jcupitt: You can use `vips_check_u8or16()` for this.

```suggestion
        if (vips_check_u8or16(class->nickname, cicp->in))
            return -1;
```

jcupitt:
I experimented with this PR a bit. 

I think we should probably change approach slightly. Having a CICP colourspace feels awkward, since it can only do 8 and 16-bit images (unlike the other colourspaces, which can take any format). It's much more like a device image, eg, ICC or UHDR. And the image is supposed to look OK-ish without interpretation (like uhdr).

So I suggest:

- load CICP as a uchar / ushort device image with attached metadata (as now)
- users must use `CICP2scRGB()` to make a device independent image if they want to do colour processing

Sorry, this is my fault, I suggested making a cicp colourspace before implementing the uhdr stuff and realising what a bad idea it was.

mertalev:
Yeah, I also thought the way CICP is a "colorspace" is a bit awkward, but didn't have a good alternative in mind. Let me give that suggestion a try

mertalev:
I changed it and it seems cleaner than the colorspace approach. One thing is that not calling CICP2scRGB for a CICP image can make the output look completely wrong if the saver doesn't support CICP since it'll treat anything as sRGB / RGB16. I kind of handled that in foreign.c with a conversion step there, but it's kind of limited since calling `.colorspace` or applying ICC would still be wrong. Would you prefer removing the fallback or handling the other cases as well?

kleisauke:
I think we should also remove the logic related to `VIPS_FOREIGN_SAVEABLE_CICP`, since we don't have equivalent flags like `VIPS_FOREIGN_SAVEABLE_ICC` or `VIPS_FOREIGN_SAVEABLE_GAINMAP`. However, introducing a `VIPS_FOREIGN_KEEP_CICP` flag could make sense.

libvips/colour/colourspace.c
```diff
@@ -217,7 +217,6 @@ typedef struct _VipsColourRoute {
 #define YXY VIPS_INTERPRETATION_YXY
 #define OKLAB VIPS_INTERPRETATION_OKLAB
 #define OKLCH VIPS_INTERPRETATION_OKLCH
-
```
kleisauke, line 220: nit: revert this.

libvips/foreign/jpegsave.c
kleisauke, line 1: Perhaps this logic should be implemented in `uhdrsave` itself?

libvips/foreign/vipspng.c
```diff
@@ -1357,6 +1375,27 @@ write_vips(Write *write,
...
+			int transfer_characteristics;
+			int matrix_coefficients;
+			int full_range_flag;
+
+			if (vips_image_get_int(in, "cicp-colour-primaries", &colour_primaries) == 0 &&
+				vips_image_get_int(in, "cicp-transfer-characteristics", &transfer_characteristics) == 0 &&
+				vips_image_get_int(in, "cicp-matrix-coefficients", &matrix_coefficients) == 0 &&
+				vips_image_get_int(in, "cicp-full-range-flag", &full_range_flag) == 0) {
+
+			png_set_cICP(write->pPng, write->pInfo,
```
kleisauke: nit: indentation looks odd here.

test/test-suite/test_cicp.py
```diff
@@ -0,0 +1,435 @@
...
+            f"{name}: primaries {out.get('cicp-colour-primaries')} != {primaries}"
+
+    # -- HEIF/AVIF regression tests --
+
+    @skip_if_no("heifsave")
+    def test_heif_cicp_identity_matrix_saves(self):
+        im = make_cicp_image(128, 100, 80,
+                             primaries=PRIMARIES_DISPLAY_P3,
+                             transfer=TRANSFER_PQ)
+        buf = im.heifsave_buffer()
```
kleisauke: We should probably use `compression="av1"` for all the `heifsave_buffer()` calls in this file. Otherwise, we'll need to mark these as `xfail`, since we can't assume that the patent-encumbered HEVC dependencies are enabled. See for example commit 8f9520506d7ff2f09004f2a4b5d290fbbe793517.

test/test-suite/test_foreign.py
kleisauke, line 1: nit: these test thresholds can also be reverted now (I think).

drhead:
> There is still a question about how to handle images that have both ICC and CICP metadata. The current implementation prioritizes CICP if it's using an HDR curve and otherwise prefers ICC.

For PNG, per standards you should be prioritizing cICP always if both cICP and iCCP chunks are present: https://www.w3.org/TR/png-3/#color-chunk-precendence.

libvips/foreign/jpegsave.c
mertalev: This checks if it should route to uhdrsave in the first place, e.g. if the CICP metadata is just SDR, there's no reason for it to go through uhdrsave.

kleisauke: Sorry, I only meant the `vips_CICP2scRGB()` call. For example, so that the following works from the CLI:
```console
$ vips uhdrsave CICP.avif x.jpg
```

mertalev: Ah gotcha, just made that change :)

libvips/colour/CICP2scRGB.c
```diff
@@ -0,0 +1,656 @@
...
+			cicp->ootf_lut = vips_hlg_ootf_build_lut();
+
+		colour->in[0] = cicp->in;
+		g_object_ref(cicp->in);
+	}
+
+	if (VIPS_OBJECT_CLASS(vips_CICP2scRGB_parent_class)->build(object))
+		return -1;
+
+	return 0;
```
kleisauke, line 592:
```suggestion
	return VIPS_OBJECT_CLASS(vips_CICP2scRGB_parent_class)->
		build(object);
```

libvips/colour/CICP2scRGB.c
```diff
@@ -0,0 +1,656 @@
...
+}
+
+static void
+vips_CICP2scRGB_dispose(GObject *gobject)
+{
+	VipsCICP2scRGB *cicp = (VipsCICP2scRGB *) gobject;
+
+	g_clear_pointer(&cicp->transfer_lut, g_free);
+	g_clear_pointer(&cicp->ootf_lut, g_free);
```
kleisauke: There's a dedicated macro for this.
```suggestion
	VIPS_FREE(cicp->transfer_lut);
	VIPS_FREE(cicp->ootf_lut);
```

mertalev:
> > There is still a question about how to handle images that have both ICC and CICP metadata. The current implementation prioritizes CICP if it's using an HDR curve and otherwise prefers ICC.
> 
> 
> 
> For PNG, per standards you should be prioritizing cICP always if both cICP and iCCP chunks are present: https://www.w3.org/TR/png-3/#color-chunk-precendence.

Good to know! CICP is actually not applied automatically anymore though, so it's left to the caller to decide which to prefer.

kleisauke:
I just pushed a couple of trivial commits since it was quicker for me to fix them myself, hope that's OK.

mertalev:
> I just pushed a couple of trivial commits since it was quicker for me to fix them myself, hope that's OK.

Totally fine!

In other news, I started work on making sharp use this branch. It converts CICP inputs to scRGB unless `keepCicp` is set, which makes it do pass-through instead without linearization.

~~I think the removal of the interpretation is making things harder though. It all works when `keepCicp` is set, but when it isn't I get black images for some outputs. I think it's converting somewhere and treating the image as srgb or rgb16 even though it isn't, or maybe double converting. `vips thumbnail` seems to have the same issue. I think it was easier to use when it was automatic and handled internally.~~

Actually, this issue existed even before the change to make the CICP handling explicit. I fixed vipsthumbnail and the [sharp branch](https://github.com/lovell/sharp/pull/4508) is producing accurate HDR output now.

kleisauke:
Commit 2fb4a8931a823e93c0b23d9a21a422c1cc01069b and e1a76af8b9df6a487bfcb8513c2a9d5f3c9d5cc0 should be unnecessary, since `colour->out` is a newly created image.
https://github.com/libvips/libvips/blob/aa93a37d9b3aaeccc0ba28feabffc01cf446254f/libvips/colour/colour.c#L227

Starbix:
> In other news, I started work on making sharp use this branch. It converts CICP inputs to scRGB unless `keepCicp` is set, which makes it do pass-through instead without linearization.

Keeping CICP should be the default. Keeping pixel-data intact is always superior in my opinion. Unless of course the saver doesn't support CICP or a colourspace change is requested.

mertalev:
> > In other news, I started work on making sharp use this branch. It converts CICP inputs to scRGB unless `keepCicp` is set, which makes it do pass-through instead without linearization.
> 
> 
> 
> Keeping CICP should be the default. Keeping pixel-data intact is always superior in my opinion. Unless of course the saver doesn't support CICP or a colourspace change is requested.

The current state of the branch actually doesn't linearize unless you call pipelineColorspace, but calling `keepCicp` overrides it so CICP images skip that conversion regardless.

The behavior I'm going for is basically having the same sharp command get the best result for the input image:
* CICP images should preserve their pixels and metadata when possible by default
* A CICP HDR input targeting JPEG with uhdrsave available should result in a uhdr image (this is handled in libvips internally)
* Non-CICP wide-gamut images should be processed in rgb16 with Display P3 output color profile (achieved with `keepCicp` making these only take effect for non-CICP images)

That said, this is all very tentative and would need discussion in the sharp side.

mertalev:
> Commit [2fb4a89](https://github.com/libvips/libvips/commit/2fb4a8931a823e93c0b23d9a21a422c1cc01069b) and [e1a76af](https://github.com/libvips/libvips/commit/e1a76af8b9df6a487bfcb8513c2a9d5f3c9d5cc0) should be unnecessary, since `colour->out` is a newly created image.
> 
> https://github.com/libvips/libvips/blob/aa93a37d9b3aaeccc0ba28feabffc01cf446254f/libvips/colour/colour.c#L227

The metadata gets copied to colour->out in the `vips_image_pipeline_array` call right after.

kleisauke:
> > Commit [2fb4a89](https://github.com/libvips/libvips/commit/2fb4a8931a823e93c0b23d9a21a422c1cc01069b) and [e1a76af](https://github.com/libvips/libvips/commit/e1a76af8b9df6a487bfcb8513c2a9d5f3c9d5cc0) should be unnecessary, since `colour->out` is a newly created image.
> > https://github.com/libvips/libvips/blob/aa93a37d9b3aaeccc0ba28feabffc01cf446254f/libvips/colour/colour.c#L227
> 
> The metadata gets copied to colour->out in the `vips_image_pipeline_array` call right after.

Oops, you're right. It seems I made a mistake when debugging this locally.

FWIW, we ran into some issues with this approach w.r.t. ICC profiles (see e.g. #4321). That said, I don't expect it to cause problems here, since `CICP2scRGB` has no reverse operation. I do wonder whether we should also remove `VIPS_META_BITS_PER_SAMPLE` from other color conversion operations, especially now that we've modified some interpretation checks with `vips_image_get_bits_per_sample()`.

mertalev:
> > > Commit [2fb4a89](https://github.com/libvips/libvips/commit/2fb4a8931a823e93c0b23d9a21a422c1cc01069b) and [e1a76af](https://github.com/libvips/libvips/commit/e1a76af8b9df6a487bfcb8513c2a9d5f3c9d5cc0) should be unnecessary, since `colour->out` is a newly created image.
> > > https://github.com/libvips/libvips/blob/aa93a37d9b3aaeccc0ba28feabffc01cf446254f/libvips/colour/colour.c#L227
> > 
> > 
> > The metadata gets copied to colour->out in the `vips_image_pipeline_array` call right after.
> 
> Oops, you're right. It seems I made a mistake when debugging this locally.
> 
> FWIW, we ran into some issues with this approach w.r.t. ICC profiles (see e.g. #4321). That said, I don't expect it to cause problems here, since `CICP2scRGB` has no reverse operation. I do wonder whether we should also remove `VIPS_META_BITS_PER_SAMPLE` from other color conversion operations, especially now that we've modified some interpretation checks with `vips_image_get_bits_per_sample()`.

That's a good point, but I think a future `scrgb2CICP` operation would probably take the target CICP as parameters and set the according metadata based on that. Since it's called manually, the caller decides when they call `CICP2scRGB` and what they want `scrgb2CICP` to produce. There's no intermediary conversion happening behind the scenes like in that issue.

mertalev:
Re: `VIPS_META_BITS_PER_SAMPLE`, I reverted the `vips_image_get_bits_per_sample` change to be on the safe side.

libvips/colour/CICP2scRGB.c
```diff
@@ -0,0 +1,685 @@
...
+		return -1;
+
+	/* Strip CICP and bits-per-sample metadata from the output — they
+	 * no longer describe the pixel values after linearization to scRGB.
+	 */
+	vips_image_remove(colour->out, "cicp-colour-primaries");
+	vips_image_remove(colour->out, "cicp-transfer-characteristics");
+	vips_image_remove(colour->out, "cicp-matrix-coefficients");
+	vips_image_remove(colour->out, "cicp-full-range-flag");
+	vips_image_remove(colour->out, VIPS_META_BITS_PER_SAMPLE);
```
kleisauke: Stripping `VIPS_META_BITS_PER_SAMPLE` should be unnecessary now, as it no longer control saver behavior.
```suggestion
	/* Strip CICP metadata from the output - it no longer describes
	 * the pixel values after linearization to scRGB.
	 */
	vips_image_remove(colour->out, "cicp-colour-primaries");
	vips_image_remove(colour->out, "cicp-transfer-characteristics");
	vips_image_remove(colour->out, "cicp-matrix-coefficients");
	vips_image_remove(colour->out, "cicp-full-range-flag");
```

libvips/resample/thumbnail.c
```diff
@@ -726,6 +726,14 @@ vips_thumbnail_build(VipsObject *object)
 		(thumbnail->input_profile ||
 			vips_image_get_typeof(in, VIPS_META_ICC_NAME));
 
+	/* CICP-tagged images should not have their
+	 * colourspace converted — the pixel values are in a specific
```
kleisauke: nit:
```suggestion
	 * colourspace converted - the pixel values are in a specific
```
(as unfortunately the em dash may now suggest LLM involvement in 2026)

libvips/resample/thumbnail.c
```diff
@@ -888,8 +901,13 @@ vips_thumbnail_build(VipsObject *object)
 	 *
 	 * We always export as depth 8, to match the no profile case which
 	 * uses vips_colourspace(sRGB|B_W).
+	 *
+	 * Skip for CICP images — their pixel encoding must be preserved.
```
kleisauke: ditto:
```suggestion
	 * Skip for CICP images - their pixel encoding must be preserved.
```

jcupitt:
I suppose we'll also need some expanded docs -- libvips colour and HDR handling is becoming quite complex now. That can wait for a later PR, and probably after the inevitable API adjustments we'll need to make.

Nice work @mertalev and @Starbix !

mertalev:
I'm not sure what's causing the TSan data race failure in `vips_foreign_load_jxl_dispose`. It seems like an issue with the cache, but this PR doesn't seem to touch anything that relates to it. It does add tests that load JXL images, so maybe that exposed an issue with eviction?

kleisauke:
The TSan CI job is [somewhat flaky](https://github.com/libvips/libvips/actions/metrics/performance?sort=failureRate&tab=jobs) (AFAIK, this started after the added test case in commit d55db5cfd4356e630b8d0fbe180ebf15e791a937) and restarting it (as I just did) usually resolves the issue.

Note to self: investigate whether setting `TSAN_OPTIONS="history_size=7"` would fix this flakiness, see:
https://github.com/google/sanitizers/wiki/ThreadSanitizerFlags#runtime-flags

kleisauke:
Thank you very much for doing this work, @mertalev and @Starbix!

PR #4929, opened 2026-03-04
min: fix possible OOB read with complex images

kleisauke:
<details>
  <summary>Reproducer</summary>

```console
$ printf 'complex\npolar\n' > repro
$ printf 'PF\n1 3\n-1.0\n' >> repro
$ printf '\xFF\xFF\xFF\xFF%.0s' {1..8} >> repro
$ printf '\x7A\x67\x2E\xFF' >> repro
$ ./build/fuzz/vips_fuzzer -rss_limit_mb=2560 -timeout=60 -runs=100 repro
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 1551465937
INFO: Loaded 2 modules   (271573 inline 8-bit counters): 271106 [0x7fb401bdd0e8, 0x7fb401c1f3ea), 467 [0x62a638, 0x62a80b), 
INFO: Loaded 2 PC tables (271573 PCs): 271106 [0x7fb401c1f3f0,0x7fb402042410), 467 [0x5e0258,0x5e1f88), 
./build/fuzz/vips_fuzzer: Running 1 inputs 100 time(s) each.
Running: repro
=================================================================
==553502==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x7c33f9be8470 at pc 0x7fb3ffcf667d bp 0x7bb3f4bac630 sp 0x7bb3f4bac628
READ of size 4 at 0x7c33f9be8470 thread T2
    #0 0x7fb3ffcf667c in vips_min_scan /home/kleisauke/libvips/build/../libvips/arithmetic/min.c:420:3
    #1 0x7fb3ffda46bd in vips_statistic_scan /home/kleisauke/libvips/build/../libvips/arithmetic/statistic.c:86:7
    #2 0x7fb4006d9180 in sink_work /home/kleisauke/libvips/build/../libvips/iofuncs/sink.c:417:12
    #3 0x7fb40061136c in vips_worker_work_unit /home/kleisauke/libvips/build/../libvips/iofuncs/threadpool.c:368:6
    #4 0x7fb40060fb70 in vips_thread_main_loop /home/kleisauke/libvips/build/../libvips/iofuncs/threadpool.c:393:3
    #5 0x7fb40060a9ae in vips_threadset_work /home/kleisauke/libvips/build/../libvips/iofuncs/threadset.c:187:3
    #6 0x7fb400605924 in vips_thread_run /home/kleisauke/libvips/build/../libvips/iofuncs/thread.c:108:11
[...]

PR #4930, opened 2026-03-04
max: fix possible OOB read with complex images

kleisauke:
Same as #4929, but for `vips_max()`.

Targets the 8.18 branch.


libvips/arithmetic/max.c
```diff
@@ -359,7 +359,7 @@ vips_max_stop(VipsStatistic *statistic, void *seq)
 			TYPE mod2 = p[0] * p[0] + p[1] * p[1]; \
 \
 			if (!isnan(mod2)) \
-				vips_values_add(values, p[i], x + i / bands, y); \
+				vips_values_add(values, mod2, x + i / bands, y); \
```
jcupitt, line 362: Ouch, looks like we had a typo here, nice find.

PR #4936, opened 2026-03-08
sign: use ready image instead of `unary->in`

kleisauke:
Buffer functions should operate on the ready images provided by the arithmetic operation rather than the original input stored on the `VipsUnary` base struct.

Resolves: https://issues.oss-fuzz.com/issues/490707025.
Targets the 8.18 branch.


kleisauke:
Note to self: occurrences of `vips_image_get_bands(statistic->in)` also need to be checked, as this may produce incorrect results when processing RAD/LABQ images.

PR #4944, opened 2026-03-11
Add APNG R/W Support via libpng

felixbuenemann:
## Summary

Add APNG (animated PNG) load/save support via libpng.

Adds animated PNG to the existing libpng backend, gated behind `PNG_APNG_SUPPORTED` (libpng 1.6+APNG patch or libpng 1.8+). The loader supports `page`/`n` parameters matching the GIF/WebP loaders, full frame compositing (all dispose and blend ops, 8-bit and 16-bit), and sequential access. The saver writes APNG via `sink_disc` with palette mode support. Non-animated PNGs are unaffected — the only overhead is an `acTL` chunk check.


felixbuenemann:
Btw. I also opened a PR for spng (which libvips also supports) to add APNG R/W support at https://github.com/randy408/libspng/pull/283, but it looks like the library is unmaintained since 2023.

felixbuenemann:
The quantiser is actually quite good for its simplicity. It doesn't quite match libimageqaunt, but it is faster and only required ~14MB RAM, while libimagequant memory use depends on image size.

**Here's a sample original 24-Bit PNG:**
dice_original

**Converted with libimagequant:**
dice_liq

**Converted with the new built-in quantiser:**
dice_wu_alphafs

jcupitt:
This is great @felixbuenemann!

Yes, libspng seems to have gone, sadly, so libpng only is fine.

I'm trapped in a horrible deadline right now (I must submit a paper in a month presenting work I've still not finished), but after that, and a beer, I'll be back on libvips.

felixbuenemann:
@jcupitt That's fine. Do you prefer if I split the quantiser into a separate PR?

jcupitt:
I'm ok with one PR, but maybe someone else will review. I'll defer to them.

jcupitt:
Getting rid of a dependency sounds great, especially if we save time and memory.

There's an issue somewhere asking for a "find $N most significant colours" operation, which is difficult with libimagequant because of the API. If we had our own cluster finder we could (probably?) add stuff like this relatively easily.

felixbuenemann:
> Getting rid of a dependency sounds great, especially if we save time and memory.
> 
> There's an issue somewhere asking for a "find $N most significant colours" operation, which is difficult with libimagequant because of the API. If we had our own cluster finder we could (probably?) add stuff like this relatively easily.

That shouldn't be too hard to do, but currently the built-in quantizer is only used as a fallback, when neither quantizr nor libimagequant are found.

I also have another optimization in the pipeline, that remaps pixels with a fully transparent alpha, but different rgb values to a single 0,0,0,0 alpha palette entry to ensure remaining palette entries are not wasted. I just haven;t come around to generating good test images to exercise the code.

felixbuenemann:
I pushed the optimisation that replaces all fully transparent pixels with a single `{0,0,0,0}` palette entry at position 0, which also improves PNG compression.

felixbuenemann:
I also discovered #4971 while working on this.

felixbuenemann:
I fixed the fuzzer errors by checking sub-frame dimensions and frame counts before allocating.

felixbuenemann:
@jcupitt you said:
> There's an issue somewhere asking for a "find $N most significant colours" operation, which is difficult with libimagequant because of the API. If we had our own cluster finder we could (probably?) add stuff like this relatively easily.

We can get just the palette with all three quantizers (built-in, libimagequant, quantizr) and return it as RGBA strip or what was the intended API?
Or is "most significant colours" something else than the palette, when quantizing to $N colors?

felixbuenemann:
I amended the quantizer commit, because cgif was disabled with the built-in quantizer and also added an error handling bugfix for the quantizr lib in a separate commit.

felixbuenemann:
@jcupitt I've split the quantizer to PR https://github.com/libvips/libvips/pull/4987 since both changes are independent of each other.

jcupitt:
Now I look more carefully, this pr is HUGE, and very difficult to review properly.

I'll split out a bite-size chunk (just apng load?) and make a small PR for that.

felixbuenemann:
> Now I look more carefully, this pr is HUGE, and very difficult to review properly.
> 
> I'll split out a bite-size chunk (just apng load?) and make a small PR for that.

I already cut this PR down to just apng support with the quantizer in another PR. Or do you refer to the comment noise on this PR?

jcupitt:
> I already cut this PR down to just apng support with the quantizer in another PR.

Yes, it needs to be smaller still, ideally under 1,000 lines. It's a huge effort to review large PRs :( 

I'm also concerned about code quality. The apng load PR I've started is based on your loader, but I'm being forced to rewrite almost the entire thing by hand. 

The LLM generated code is not maintainable --- anyone trying to fix an issue in it will get a bad headache. I think merging code like that would seriously damage the long-term health of the project.

felixbuenemann:
If you have specific feedback, maybe the llm can fix most if it by itself, saving you some work.

jcupitt:
For example, this was the first function I looked at:

https://github.com/felixbuenemann/libvips/blob/apng-support/libvips/foreign/vipspng.c#L618-L683

It scans the file looking for delay chunks to build the delay array.  

It's hard to read and verify:

- common code in then and else parts
- things are incremented, and then "var - 1" appears later, creating
complex dependencies between different parts of the function, and
making maintenance much harder
- conceptually this should be a for loop along the chunks, why is this a while?
- (minor) libvips is 4-space tabs, so a lot of linebreaks can go 
- silently ignores malformed chunks -- it should probably be a warning and optionally throw an error
- lots of unnecessary ternaries
- unnecessary apng_delay_to_ms() function

The rest of the PR has similar problems. The code is tortuous, unclear, and would be difficult to maintain.

There are larger structural problems too. APNG load hasn't been built in layers, instead it's a general melange of mixed up concepts. Everything dissolves into a uniform fog of vague confusion, with no light of intelligence.

For example, there should be four functions for manipulating areas of pixels: clear a buffer, take a copy of a buffer, paste at position, and blend at position. They should be easily verifiable, including slightly tricky things like clipping and handling of out of range values. Later parts of apng load would use those four to do any pixel manipulation, and would not contain any direct memory writes.

Without strict layers, verifiability is difficult. Can you *prove* (informally anyway!) that apng load can never write out of bounds? It's much harder if writes happen everywhere rather than being isolated in single functions.

I think for non-library projects, code like this is probably fine. If there's a problem, you can just regenerate the whole thing. Maintainability doesn't really matter.

For long-lived libraries though, it's just not going to work. Maintainability over *many* years is almost the key quality. You have to be able to present a stable and predictable API to downstream users, and that means code has to evolve. You can't evolve code that no one can read.

I feel bad, I know you spent a lot of time on this PR and must be disappointed and annoyed, but now I've spent time reading the code I think it all has to be rewritten, and by a human. LLMs are great, but for projects like libvips, only as very junior assistants, not as (almost) final authors.

I'll write an AI-POLICY.md to try to stop this happening again.

felixbuenemann:
@jcupitt I have pushed a new commit that refactors the code according to your comments. Let me know, if it is an improvement.

jcupitt:
Hi again, I spent some time trying to review this, and I just can't. It's not fun to read something as convoluted as this, and I don't really want to work on something that's not fun in my spare time.

I have a new plan: I'll make a pair of PRs based on this PR (so you get the credit in the git log), but I'll rewrite the whole thing by hand. It'll be less work than reviewing this, and more fun. I'll probably do APNG save first.

In future, if I see a PR that's hard to read, I'll probably just close it. Life's too short!

PR #4957, opened 2026-03-20
Improve the vips_fuzzer seed corpus

dloebl:
Expand the vips_fuzzer seed corpus with entries not covered well by the current (2026-03-19) OSS Fuzz corpus.
**Note:** This adds a lot of loose files. We could move it to a single zip file if that's preferred.

**Before:**
| Function Coverage | Line Coverage | Region Coverage | Branch Coverage |
| - | - | - | - |
| 53.37% | 49.36% | 46.11% | 31.69% |

**With the improved seed corpus:**
| Function Coverage | Line Coverage | Region Coverage | Branch Coverage |
| - | - | - | - |
| 57.86%  | 56.66%  | 51.89% | 38.67% |

This should help OSS Fuzz to discover more interesting issues quicker.

**The approach I took:**
I ran https://github.com/davebcn87/pi-autoresearch (an LLM-driven experimentation loop) with `claude-opus-4-6` on the current OSS Fuzz corpus ([libvips_vips_fuzzer/2026-03-19.zip](https://storage.googleapis.com/libvips-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/libvips_vips_fuzzer/2026-03-19.zip)) + the seed corpus in `/fuzz/vips_fuzzer_corpus/` with the goal to improve the coverage. It generated a coverage report on each pass and either kept or discarded the changes depending on if the coverage improved.

### DIFF
--- a/fuzz/test_fuzz.sh
+++ b/fuzz/test_fuzz.sh
@@ -19,19 +19,6 @@ for fuzzer in *_fuzzer; do
       ret=1
     fi
   done
-
-  # Run fuzzer-specific corpus if it exists.
-  corpus_dir="$top_srcdir/fuzz/${fuzzer}_corpus"
-  if [ -d "$corpus_dir" ]; then
-    for file in "$corpus_dir"/*; do
-      exit_code=0
-      ./$fuzzer $file || exit_code=$?
-      if [ $exit_code -ne 0 ]; then
-        echo FAIL $fuzzer $file
-        ret=1
-      fi
-    done
-  fi
 done
 
 exit $ret


kleisauke:
1973 files added! We might consider moving this seed corpus to a separate repository, similar to https://github.com/libjpeg-turbo/seed-corpora.

jcupitt:
> We might consider moving this seed corpus to a separate repository,

I had a look and this takes us up to about 1mb for the fuzz directory, which isn't too bad. Having them in the main repo lets us avoid annoyances like git subprojects. For, now anyway!

kleisauke:
That's true. I just opened PR #4958, since we probably don't want to include this in the dist tarball.

Note that we want to preserve `common_fuzzer_corpus`, as it is used for regression tests.
https://github.com/libvips/libvips/blob/d8faae7aba8c4175c95bcae35b49e56d07f9ea40/fuzz/test_fuzz.sh#L14

dloebl:
Yes, it's a lot of files 😅 
Excluding them from the dist tarball makes sense to me.

> [libvips/fuzz/test_fuzz.sh](https://github.com/libvips/libvips/blob/d8faae7aba8c4175c95bcae35b49e56d07f9ea40/fuzz/test_fuzz.sh#L14)

btw I just noticed that the `vips_fuzzer_corpus` isn't run as part of the tests. I addressed this with a1c0d09a5b4183907f1aa4500070c59b84175282
Edit: Looks like this makes the CI very slow, I've reverted for now - maybe I'll try to make it faster with some follow-up work.

Also the improved seed corpus seems to already include cases that fail with UBSAN (I can create individual issues later on).

dloebl:
Can we get this merged or is it too many files? If it's too many, I'm happy to add all the loose files to a tarball/zip or so.

jcupitt:
Sure, let's do it. Thank you for these improvements, @dloebl !

PR #4959, opened 2026-03-21
shrink the RAD test image

jcupitt:
`sample.hdr` was 11mb! Crop it down to 400kb, we don't need a large file for this test.


jcupitt:
Ooop! I always forget to set the target, thanks!

PR #4962, opened 2026-03-22
Update API docs for 8.18.1

kleisauke:
Context: https://github.com/google/libultrahdr/issues/380#issuecomment-4106166581.


gregbenz:
@kleisauke I'm moving the compression discussion over here to stay off that other thread.

Thank you for sharing that Q argument. Unless I've missed something deeper, it exposes a single quality factor. A gain map is 2 images and the base image and map (auxiliary image) can be set seprately in libultrahdr. When I reduce the quality factor down to 1 for transcoding via wasm-vips, the SDR image shows heavy compression - yet the map itself remains very high quality. I believe there is a missing argument to control the full result.

Ideally, requesting quality sets the base and automatically sets a reasonable quality factor for the map as well based on that single compression value (a single input is ideal for a lot of use). However, it should also allow an optional argument to directly control the map compression (as allows for greater optimization of quality vs size).

jcupitt:
Ooops, you're right, we were not setting Q when we recompressed the gainmap, my fault. I made a PR.

jcupitt:
> However, it should also allow an optional argument to directly control the map compression

libvips will often not decompress and recompress the gainmap. For example:

```
$ vips copy input-uhdr.jpg output-uhdr.jpg[Q=95]
```

Will decompress and recompress the main image, but the compressed gainmap data will be simply copied over, since it is never used.

If you do (perhaps):

```
$ vipsthumbnail uhdr.jpg --suffix .jpg[Q=95]
```

Now the gainmap must be decompressed for the resize and possible crop. It will be recompressed with `Q=95`.

gregbenz:
@jcupitt Agree that the image should not be recompressed if not necessary.

Can we pass an argument to allow the user to set Q for the base and map separately? Of course, ideal to set a default value if not provided by the user. I've found compression needs for base and map can be quite different when you want to fully optimize the encoding.

jcupitt:
> Can we pass an argument to allow the user to set Q for the base and map separately? 

You can compress the gainmap yourself, of course. For example (untested):

```python
gainmap = image.get_gainmap()
image.remove("gainmap")
data = gainmap.jpegsave_buffer(Q=20)
image.set("gainmap-data", data)
image.write_to_file("x.jpg")
```

So set the `gainmap-data` metadata item yourself and make sure there's no `gainmap` item.

You can do a range of manipulations, like changing the image-image / gainmap-image size ratio, making the gainmap monochrome, using a different quantisation table, and so on.

gregbenz:
@jcupitt Thank you for that, very helpful. To confirm, if you do not do that, the gain map would use a 95% quality factor otherwise (when re-encoding of the map is required, such as when resizing or cropping)?

gregbenz:
@jcupitt I'm seeing some exports where the gain map is encoded at less than 444. I need to confirm that isn't coming from me, but that's a watchout. Sub-sampling assumptions apply to photos based on human vision. But the map itself is not a photograph and those assumptions are not valid. The gain map should be encoded at 444 if it has three channels.

As an example, the min/max/gamma of each channel is unique. They are just meant to multiply the base pixels and there should be no cross-talk between the channels. This can result in edge cases where there are significant quality issues in the HDR rendering after transcoding.

jcupitt:
Do many people use RGB gainmaps? I think the standard only allows them by accident. Mono gainmaps must be much more common.

You can set Q to 90 to prevent chroma subsampling on reencode, of course, but if you want detailed control over the gainmap compression it'd be best to handle the compress yourself, as outlined above.

jcupitt:
> To confirm, if you do not do that, the gain map would use a 95% quality factor otherwise 

Current stable will use default jpeg compression, so Q 75.

gregbenz:
@jcupitt RGB maps are a core part of the spec. Adobe uses them for just about everything. I almost never share images with an RGB map.  The best images will almost always use an RGB map (unless the image is monochrome).

Without them, the HDR image quality will suffer significantly. It will limit color (the base SDR often has desaturated highlights as values push towards the peak brightness of that limited encoding). And there is value in other corrections when building the HDR from the SDR. 

Mobile phones use low quality maps (1/2-1/4 resolution without color). This likely because (a) mobile data / storage make size reduction a high priority and (b) there is no artistic input from the user into the SDR + HDR relationship, which limits the value of the map - it's fairly comparable to tone mapping that way.

So there are a lot of luminosity-only maps due to the volume of phone captures, but not because they are artistically competitive with RGB maps.

gregbenz:
@jcupitt it would be best to never allow transcoding of an existing gain map to use sub sampling less than the source. 

You could probably avoid these risks with sub sampling done in the original encode in combination with careful controls to avoid the sneaky issues created by it. In transcoding, you could solve it, but it would require full decoding, recalculating the gain map, etc. 

Fundamentally, the use of the map violates the assumptions which make sub-sampling a valid thing to do. The map is not an image, it is just math encoded in an image format where there is often no relationship between the channels.

I've seen some pretty nasty artifacts with sub sampling of the map (such as blue skies showing heavy levels of noise due to near zero channels mixing with others and causing a lot of noise in the map multipliers created by that scenario). They tend to be edge cases, so you can easily miss it in testing, but it's a problem which will come too often up at scale.

jcupitt:
> Without [RGB maps], the HDR image quality will suffer significantly.

I'm very sceptical. I think even expert users would struggle to see any difference between a mono and RGB gainmap in a blind test on a real display. Camera manufacturers use mono maps because they've done a lot of testing. HDR formats like Radiance (surely one of the highest quality HDR file types), are all based on mono gain maps.

... but I agree that chroma subsampling an RGB gainmap is always wrong, I'll add something to stop it in that PR, thanks!

gregbenz:
> ... but I agree that chroma subsampling an RGB gainmap is always wrong, I'll add something to stop it in that PR, thanks!

Thank you! This will help many deva avoid unexpected issues. Gain maps gets tricky because a software developer (who may have no HDR background and may not have a good HDR display) needs to understand both the art and math deeply enough to anticipate the many edge cases gain maps can create if not used properly.

PR #4963, opened 2026-03-22
uhdrsave: set gainmap Q on recompress

jcupitt:
We were not setting `Q` when we recompressed the gainmap.

see https://github.com/libvips/libvips/pull/4962#issuecomment-4106361192


jcupitt:
\o/ thanks!

PR #4969, opened 2026-03-25
perf: optimize conversion module hot paths

JakeChampion:
- flatten: replace per-pixel float division with 256-entry LUTs for uchar alpha blending
- composite: precompute `inv_max_band[]` reciprocals instead of dividing per pixel; initialize blend arrays with `= {0}` instead of per-pixel zero-fill loop
- rot/flip/zoom/subsample: replace byte-at-a-time pixel copies with typed stores

Measured with [hyperfine](https://github.com/sharkdp/hyperfine) on an arm64 Apple M2, clang 21

Test images: 8000x8000 or 4000x4000 uchar, created with `vips gaussnoise + cast + bandjoin`

```
$ hyperfine --warmup 5 --runs 15 \
    -n master 'VIPS_CONCURRENCY=1 vips <op> input.v output.v' \
    -n branch 'VIPS_CONCURRENCY=1 vips <op> input.v output.v'
```

| Operation              | master       | branch      | speedup          |
|------------------------|--------------|-------------|------------------|
| flatten 8k RGBA        | 170.0 ms     | 96.0 ms     | **1.77x faster** |
| rot90 4k RGBA          | 86.0 ms      | 52.4 ms     | **1.64x faster** |
| rot90 4k RGB           | 80.6 ms      | 50.7 ms     | **1.59x faster** |
| zoom 2x 4k RGB         | 121.1 ms     | 75.9 ms     | **1.59x faster** |
| rot270 4k RGB          | 79.8 ms      | 55.0 ms     | **1.45x faster** |
| flip horiz 4k RGBA     | 68.7 ms      | 47.6 ms     | **1.44x faster** |
| flip horiz 4k RGB      | 63.3 ms      | 47.7 ms     | **1.33x faster** |
| rot180 4k RGB          | 63.3 ms      | 48.8 ms     | **1.30x faster** |
| subsample 2x 8k RGB    | 78.2 ms      | 66.6 ms     | **1.17x faster** |

All affected operations produced bit-identical output to master, verified by running each operation on a 512x512 RGBA/RGB test image and comparing raw `.v` files with `cmp -s`.

## Test plan

- [x] `meson test -C build` passes
- [x] Output images are bit-identical to master
- [ ] Test on x86_64 / gcc

### DIFF
--- a/libvips/conversion/composite.cpp
+++ b/libvips/conversion/composite.cpp
@@ -591,23 +591,19 @@ vips_composite_base_blend(VipsCompositeBase *composite,
 {
 	const int bands = composite->bands;
 
-	double A[MAX_BANDS + 1];
+	double A[MAX_BANDS + 1] = { 0 };
 	double aA;
 	double aB;
 	double aR;
 	double t1;
 	double t2;
 	double t3;
-	double f[MAX_BANDS + 1];
+	double f[MAX_BANDS + 1] = { 0 };
 
 	/* Load and scale the pixel to 0 - 1.
 	 */
 	for (int b = 0; b <= bands; b++)
 		A[b] = p[b] * composite->inv_max_band[b];
-	/* Not necessary, but it stops a compiler warning.
-	 */
-	for (int b = bands + 1; b < MAX_BANDS + 1; b++)
-		A[b] = 0.0;
 
 	aA = A[bands];
 	aB = B[bands];
--- a/libvips/conversion/composite.cpp
+++ b/libvips/conversion/composite.cpp
@@ -125,6 +125,11 @@ typedef struct _VipsCompositeBase {
 	 */
 	double max_band[MAX_BANDS + 1];
 
+	/* Precomputed reciprocals of max_band, to replace per-pixel
+	 * division with multiplication.
+	 */
+	double inv_max_band[MAX_BANDS + 1];
+
 	/* TRUE if all our modes are skippable, ie. we can avoid compositing
 	 * the whole stack for every pixel request.
 	 */
@@ -168,6 +173,7 @@ typedef struct {
 	 * on a 16-byte boundary.
 	 */
 	v4f max_band_vec;
+	v4f inv_max_band_vec;
 #endif /*HAVE_VECTOR_ARITH*/
 
 	VipsCompositeBase *composite;
@@ -332,13 +338,20 @@ vips_composite_start(VipsImage *out, void *a, void *b)
 #ifdef HAVE_VECTOR_ARITH
 	/* We need a float version for the vector path.
 	 */
-	if (composite->bands == 3)
+	if (composite->bands == 3) {
 		seq->max_band_vec = (v4f){
 			(float) composite->max_band[0],
 			(float) composite->max_band[1],
 			(float) composite->max_band[2],
 			(float) composite->max_band[3]
 		};
+		seq->inv_max_band_vec = (v4f){
+			(float) composite->inv_max_band[0],
+			(float) composite->inv_max_band[1],
+			(float) composite->inv_max_band[2],
+			(float) composite->inv_max_band[3]
+		};
+	}
 #endif
 
 	return seq;
@@ -590,7 +603,7 @@ vips_composite_base_blend(VipsCompositeBase *composite,
 	/* Load and scale the pixel to 0 - 1.
 	 */
 	for (int b = 0; b <= bands; b++)
-		A[b] = p[b] / composite->max_band[b];
+		A[b] = p[b] * composite->inv_max_band[b];
 	/* Not necessary, but it stops a compiler warning.
 	 */
 	for (int b = bands + 1; b < MAX_BANDS + 1; b++)
@@ -881,7 +894,7 @@ vips_composite_base_blend3(VipsCompositeSequence *seq,
 	A[2] = p[2];
 	A[3] = p[3];
 
-	A /= seq->max_band_vec;
+	A *= seq->inv_max_band_vec;
 
 	aA = A[3];
 	aB = B[3];
@@ -1152,7 +1165,7 @@ vips_combine_pixels(VipsCompositeSequence *seq, VipsPel *q)
 	/* Load and scale the base pixel to 0 - 1.
 	 */
 	for (int b = 0; b <= bands; b++)
-		B[b] = tp[0][b] / composite->max_band[b];
+		B[b] = tp[0][b] * composite->inv_max_band[b];
 
 	aB = B[bands];
 	if (!composite->premultiplied)
@@ -1220,7 +1233,7 @@ vips_combine_pixels3(VipsCompositeSequence *seq, VipsPel *q)
 
 	/* Scale the base pixel to 0 - 1.
 	 */
-	B /= seq->max_band_vec;
+	B *= seq->inv_max_band_vec;
 	aB = B[3];
 
 	if (!composite->premultiplied) {
@@ -1651,6 +1664,9 @@ vips_composite_base_build(VipsObject *object)
 			"%s", _("unsupported compositing space"));
 		return -1;
 	}
+
+	for (int b = 0; b <= composite->bands; b++)
+		composite->inv_max_band[b] = 1.0 / composite->max_band[b];
 
 	/* Transform the input images to match in format. We may have
 	 * mixed float and double, for example.
--- a/libvips/conversion/flip.c
+++ b/libvips/conversion/flip.c
@@ -65,6 +65,7 @@
 
 #include "pconversion.h"
 
+
 typedef struct _VipsFlip {
 	VipsConversion parent_instance;
 
@@ -169,13 +170,8 @@ vips_flip_horizontal_gen(VipsRegion *out_region,
 		q = VIPS_REGION_ADDR(out_region, le, y);
 
 		for (x = le; x < ri; x++) {
-			/* Copy the pel.
-			 */
-			for (z = 0; z < ps; z++)
-				q[z] = p[z];
+			VIPS_MEMCPY(q, p, ps);
 
-			/* Skip forwards in out, back in in.
-			 */
 			q += ps;
 			p -= ps;
 		}
--- a/libvips/conversion/flatten.c
+++ b/libvips/conversion/flatten.c
@@ -164,6 +164,66 @@ G_DEFINE_TYPE(VipsFlatten, vips_flatten, VIPS_TYPE_CONVERSION);
 		} \
 	}
 
+/* Fast UCHAR path for black background: replace per-pixel double division
+ * with a 256-entry LUT.
+ */
+static int
+vips_flatten_black_gen_uchar(VipsRegion *out_region,
+	void *vseq, void *a, void *b, gboolean *stop)
+{
+	VipsRegion *ir = (VipsRegion *) vseq;
+	VipsFlatten *flatten = (VipsFlatten *) b;
+	VipsRect *r = &out_region->valid;
+	int width = r->width;
+	int bands = ir->im->Bands;
+	double max_alpha = flatten->max_alpha;
+
+	int x, y, i;
+
+	float factor_lut[256];
+	for (i = 0; i < 256; i++)
+		factor_lut[i] = (float) ((double) i / max_alpha);
+
+	if (vips_region_prepare(ir, r))
+		return -1;
+
+	for (y = 0; y < r->height; y++) {
+		unsigned char *restrict p =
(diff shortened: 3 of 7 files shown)


libvips/conversion/composite.cpp
```diff
@@ -578,23 +591,19 @@ vips_composite_base_blend(VipsCompositeBase *composite,
...
 
 	/* Load and scale the pixel to 0 - 1.
 	 */
 	for (int b = 0; b <= bands; b++)
-		A[b] = p[b] / composite->max_band[b];
-	/* Not necessary, but it stops a compiler warning.
-	 */
-	for (int b = bands + 1; b < MAX_BANDS + 1; b++)
-		A[b] = 0.0;
+		A[b] = p[b] * composite->inv_max_band[b];
```
lovell, line 606: I guess it's possible that this change might introduce possible (tiny) rounding differences, unsure if we need to worry about that.

libvips/conversion/flip.c
```diff
@@ -65,6 +65,34 @@
 
 #include "pconversion.h"
 
+#define COPY_PEL(Q, P, PS) \
```
lovell: Probably worth moving the `COPY_PEL` macro to `util.h` so we define it once.

lovell (commented):
Thanks for this Jake, those timings look great. Would be great to see the impact on x64 CPUs too. I've left a couple of comments inline.

libvips/conversion/flip.c
jcupitt: Maybe make it into `VIPS_MEMCPY()`?

libvips/conversion/composite.cpp
JakeChampion: I verified bit-identical output for all affected operations with the below:

Prepare for vips from master and this branch:
```bash
git worktree add /tmp/vips_master upstream/master
git worktree add /tmp/vips_branch jake/perf-conversion
cd /tmp/vips_master && meson setup build --buildtype=release && ninja -C build
cd /tmp/vips_branch && meson setup build --buildtype=release && ninja -C build
```


Create a test image with varied pixel values:
```bash
VIPS=/tmp/vips_master/build/tools/vips
DYLD_LIBRARY_PATH=/tmp/vips_master/build/libvips
$VIPS gaussnoise /tmp/n1.v 512 512 --mean 128 --sigma 60
$VIPS gaussnoise /tmp/n2.v 512 512 --mean 100 --sigma 80
$VIPS gaussnoise /tmp/n3.v 512 512 --mean 180 --sigma 40
$VIPS gaussnoise /tmp/n4.v 512 512 --mean 128 --sigma 100
$VIPS cast /tmp/n1.v /tmp/c1.v uchar
$VIPS cast /tmp/n2.v /tmp/c2.v uchar
$VIPS cast /tmp/n3.v /tmp/c3.v uchar
$VIPS cast /tmp/n4.v /tmp/c4.v uchar
$VIPS bandjoin "/tmp/c1.v /tmp/c2.v /tmp/c3.v /tmp/c4.v" /tmp/rgba.v
$VIPS bandjoin "/tmp/c1.v /tmp/c2.v /tmp/c3.v" /tmp/rgb.v
```

Run each operation with both builds, compare raw output
```bash
for op in \
  "flatten /tmp/rgba.v" \
  "rot /tmp/rgb.v d90" \
  "rot /tmp/rgb.v d180" \
  "rot /tmp/rgb.v d270" \
  "flip /tmp/rgb.v horizontal" \
  "subsample /tmp/rgb.v 2 2" \
  "zoom /tmp/rgb.v 2 2" \
  "rot /tmp/rgba.v d90" \
  "flip /tmp/rgba.v horizontal"
do
  name=$(echo $op | awk '{print $1}')
  DYLD_LIBRARY_PATH=/tmp/vips_master/build/libvips \
    /tmp/vips_master/build/tools/vips $op /tmp/master_${name}.v
  DYLD_LIBRARY_PATH=/tmp/vips_branch/build/libvips \
    /tmp/vips_branch/build/tools/vips $op /tmp/branch_${name}.v
  if cmp -s /tmp/master_${name}.v /tmp/branch_${name}.v; then
    echo "PASS  $name"
  else
    echo "FAIL  $name"
  fi
done
```

libvips/conversion/flip.c
JakeChampion: Definitely - I've made that change now 👍

libvips/conversion/composite.cpp
lovell: Fantastic, thank you for checking.

JakeChampion:
> Thanks for this Jake, those timings look great. Would be great to see the impact on x64 CPUs too. I've left a couple of comments inline.

Are we waiting on me to come back with x64 numbers for this pr before we think about landing it? If so, I can look to get those numbers 👍

jcupitt:
Sorry Jake, I've just been drowning in stuff. I'm at a conference this week (the deadline I've been fighting), but I promise I'll look at this next week when I'm back.

libvips/conversion/flip.c
```diff
@@ -169,13 +170,8 @@ vips_flip_horizontal_gen(VipsRegion *out_region,
 		q = VIPS_REGION_ADDR(out_region, le, y);
 
 		for (x = le; x < ri; x++) {
-			/* Copy the pel.
-			 */
-			for (z = 0; z < ps; z++)
```
jcupitt, line 174: The declaration of z above (line 139 I think) also needs to go. 

libvips is now finally C99, so we could maybe move the loop vars into the for() at the same time, eg.:

```C
		for (int x = le; x < ri; x++) {
```

rot, subsample and zoom need the same change (removal of the inner loop var decl).

jcupitt:
Hi again, I tried on x64. 

I made a test 8500 x 8500 uchar RGBA image, then ran eg.:

```
$ VIPS_CONCURRENCY=1 time vips composite2 input.v input.v output.v over
3.55user 0.26system 0:03.61elapsed 105%CPU (0avgtext+0avgdata 325284maxresident)k
0inputs+564464outputs (0major+4880minor)pagefaults 0swaps
```

Then took the best user time of three runs.

- For composite2, I see maybe a 2% speedup, which I think is probably not worthwhile for the extra complexity.
- For flatten I see about a 3x speedup, which is very nice! Though that was for a RGBA image, which is probably the best case.
- rot, flip, zoom, subsample are a little over about 2x faster, lovely! 

So I think we should probably remove the composite changes, but the rest seem great.

Thank you for doing this work @JakeChampion

jcupitt:
I had two more thoughts in the shower this morning:

First. we should check on musl too, `memcpy()` is just the kind of thing they'd optimise. We should disable `VIPS_MEMCPY()` on musl if they have a better implementation.

Second, `VIPS_MEMCPY()` should probably handle all cases of 8 and under, including zero, otherwise it'll just fall back to `memcpy()`. Maybe (untested!):

```C
#define VIPS_MEMCPY(Q, P, N) \
	G_STMT_START \
	{ \
		switch (N) { \
		case 8: \
			*((guint64 *) (Q)) = *((guint64 *) (P)); \
			break; \
\
		case 7: \
			(Q)[6] = (P)[6]; \
\
		case 6: \
			(Q)[5] = (P)[5]; \
\
		case 5: \
			(Q)[4] = (P)[4]; \
\
		case 4: \
			*((guint32 *) (Q)) = *((guint32 *) (P)); \
			break; \
\
		case 3: \
			(Q)[2] = (P)[2]; \
\
		case 2: \
			*((guint16 *) (Q)) = *((guint16 *) (P)); \
			break; \
\
		case 1: \
			(Q)[0] = (P)[0]; \
\
		case 0: \
			break; \
\
		default: \
			memcpy((Q), (P), (N)); \
			break; \
		} \
	} \
	G_STMT_END
```

That will copy bytes backwards for some cases, I've not timed it to see if that's an issue (I assume not?). The fall throughs in the case are a bit tricksy, maybe that's bad too.

jcupitt:
I suppose looking at pointer alignment would be the next step. We could copy until P is 8-byte aligned, then loop copying gint64 chunks, then copy a few bytes to use up N. It should give a nice further speedup.

That's probably too complex for a macro.

We should benchmark on windows too, I've no idea what the MSCRT memcpy() is like.

jcupitt:
Just for fun I made a version that tries to copy in pointer-aligned chunks:

```C
#define VIPS_MEMCPY0(Q, P, N) \
    G_STMT_START \
    { \
        switch (N) { \
        case 8: \
            *((guint64 *) (Q)) = *((guint64 *) (P)); \
            break; \
\
        case 7: \
            (Q)[6] = (P)[6]; \
\
        case 6: \
            (Q)[5] = (P)[5]; \
\
        case 5: \
            (Q)[4] = (P)[4]; \
\
        case 4: \
            *((guint32 *) (Q)) = *((guint32 *) (P)); \
            break; \
\
        case 3: \
            (Q)[2] = (P)[2]; \
\
        case 2: \
            *((guint16 *) (Q)) = *((guint16 *) (P)); \
            break; \
\
        case 1: \
            (Q)[0] = (P)[0]; \
\
        case 0: \
            break; \
\
        default: \
            g_assert(FALSE); \
            break; \
        } \
    } \
    G_STMT_END

#define VIPS_MEMCPY(Q, P, N) \
    G_STMT_START \
    { \
        unsigned char *_p = (P); \
        unsigned char *_q = (Q); \
        size_t _n = (N); \
 \
        size_t _head = VIPS_MIN(_n, 8 - (((uintptr_t) _p) & 0x7)); \
        VIPS_MEMCPY0(_q, _p, _head); \
        _q += _head; \
        _p += _head; \
        _n -= _head; \
 \
        if (_n > 7) { \
            size_t _chunks = _n >> 3; \
            guint64 *_p64 = (guint64 *) _p; \
            guint64 *_q64 = (guint64 *) _q; \
            for (size_t _i = 0; _i < _chunks; _i++) \
                _q64[_i] = _p64[_i]; \
            _p += _chunks << 3; \
            _q += _chunks << 3; \
            _n -= _chunks << 3; \
        } \
 \
        g_assert(_n < 8); \
        VIPS_MEMCPY0(_q, _p, _n); \
    } \
    G_STMT_END
```

But for larger values of N (I tried 45), it's slower than glibc memcpy() by about 50%. Assuming I didn't mess it up heh.

It looks like the win is only for very small values of N (less than about 10?).

jcupitt:
I ended up with this, it seems to work OK:

```C
#define VIPS_MEMCPY(Q, P, N) \
    G_STMT_START \
    { \
        switch (N) { \
        case 8: \
            *((guint64 *) (Q)) = *((guint64 *) (P)); \
            break; \
\
        case 7: \
            ((guint8 *) (Q))[6] = ((guint8 *) (P))[6]; \
\
        case 6: \
            ((guint8 *) (Q))[5] = ((guint8 *) (P))[5]; \
\
        case 5: \
            ((guint8 *) (Q))[4] = ((guint8 *) (P))[4]; \
\
        case 4: \
            *((guint32 *) (Q)) = *((guint32 *) (P)); \
            break; \
\
        case 3: \
            ((guint8 *) (Q))[2] = ((guint8 *) (P))[2]; \
\
        case 2: \
            *((guint16 *) (Q)) = *((guint16 *) (P)); \
            break; \
\
        case 1: \
            ((guint8 *) (Q))[0] = ((guint8 *) (P))[0]; \
\
        case 0: \
            break; \
\
        default: \
            memcpy((Q), (P), (N)); \
            break; \
        } \
    } \
    G_STMT_END
```

jcupitt:
I tested musl. The memcpy() there is quicker, but the macro still beats it for small N.

With memcpy():

```
# vips black x.v 8000 8000 --bands 7
# time vips rot x.v y.v d90
user	0m 0.72s
```

With the macro I posted just above:

```
# time vips rot x.v y.v d90
user	0m 0.50s
```

jcupitt:
Hey @JakeChampion, are you OK to make these changes? I can do it if you're short on time.

JakeChampion:
> Hey @JakeChampion, are you OK to make these changes? I can do it if you're short on time.

Apologies for not replying sooner, I'm a bit busy with family life right now to get time for anything else, we recently had our second child and I've taken time away from all work until July (already had a long time off until now) - if you are able to take this task, that'd be amazing 🙏🙇‍♀️

jcupitt:
Ah wow, well done! Mine are 19 and 21 so what's happening to your sleep is only a distant but terrible memory hehe.

I'll make another PR based on this and hack that around a little.

jcupitt:
https://github.com/libvips/libvips/pull/5072 is this PR with the suggested changes, I'll close this PR in favour of that.

Thank you for pushing for these changes, @JakeChampion, the improvement is very welcome.

PR #4970, opened 2026-03-25
perf: optimize resample module hot paths

JakeChampion:
- thumbnail: for uchar RGBA images, premultiply/unpremultiply directly in uchar using a 256x256 LUT, avoiding the UCHAR->FLOAT->UCHAR round-trip that the generic `vips_premultiply`/`vips_unpremultiply` path requires
- shrinkh/shrinkv: replace per-pixel integer division by `hshrink`/`vshrink` with a precomputed multiply-shift reciprocal for unsigned short types (the uchar path already used this technique)
- shrinkh/shrinkv: precompute `1.0 / shrink` for the float accumulator path instead of dividing per pixel
- jp2ksave: replace per-pixel division by `n_pels` in the SHRINK macro with a precomputed multiply-shift reciprocal

Measured with [hyperfine](https://github.com/sharkdp/hyperfine) on arm64 Apple M2, clang 21.

Test images: 4000x4000 uchar/ushort/float, created with `vips gaussnoise + cast + bandjoin`.

```
$ hyperfine --warmup 5 --runs 15 \
    -n master 'VIPS_CONCURRENCY=1 vips <op> input.v output.v' \
    -n branch 'VIPS_CONCURRENCY=1 vips <op> input.v output.v'
```

| Operation                       | master       | branch      | speedup          |
|---------------------------------|--------------|-------------|------------------|
| shrink 8x ushort RGBA 4000     | 62.8 ms      | 56.0 ms     | **1.12x faster** |
| shrink 4x float RGB 4000       | 77.4 ms      | 72.5 ms     | **1.07x faster** |
| resize 0.125x ushort RGBA 4000 | 110.8 ms     | 103.3 ms    | **1.07x faster** |
| thumbnail RGBA PNG 4000->500   | 236.3 ms     | 224.0 ms    | **1.06x faster** |
| resize 0.25x ushort RGB 4000   | 145.7 ms     | 140.8 ms    | **1.04x faster** |

### DIFF
--- a/libvips/foreign/jp2ksave.c
+++ b/libvips/foreign/jp2ksave.c
@@ -307,6 +307,8 @@ vips_foreign_save_jp2k_rgb_to_ycc(VipsRegion *region,
 		ACC_TYPE *acc = (ACC_TYPE *) accumulate; \
 		OUTPUT_TYPE *tq = (OUTPUT_TYPE *) q; \
 		const int n_pels = comp->dx * comp->dy; \
+		const guint64 n_pels_recip = \
+			((1ULL << 32) + n_pels - 1) / n_pels; \
 \
 		PIXEL_TYPE *tp; \
 		ACC_TYPE *ap; \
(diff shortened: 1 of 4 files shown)


jcupitt:
Hi again, sorry for being so slow on this.

For premultiply, how about adding a `uchar` option to the `premultiply` operation, meaning that the output image should be uchar, not the default float (other operations like `linear` have a flag like this). 

We can then add a special uchar -> uchar path using a LUT to `premultiply`, and set this new `uchar` flag in `thumbnail`. It should be a bit neater than adding a special premultiply thing to thumbnail.

jcupitt:
I've made a PR based off this to incorporate review comments.

https://github.com/libvips/libvips/pull/5077

Thanks for doing this work!

PR #4981, opened 2026-03-28
CIFuzz: integrate ThreadSanitizer

kleisauke:
Replaces the TSan CI workflow, which required numerous suppressions
because GLib was not built with ThreadSanitizer instrumentation.

See also: https://github.com/google/oss-fuzz/pull/15235.
Context: https://github.com/libvips/libvips/pull/4713#issuecomment-3467458167.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -16,3 +16,4 @@ libvips/foreign/libnsgif/*	linguist-vendored
 .gitignore					export-ignore
 .gitkeep					export-ignore
 .github/					export-ignore
+subprojects/				export-ignore
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,19 +14,24 @@ jobs:
         include:
           - name: "Linux x64 (Ubuntu 24.04) - GCC 14"
             os: ubuntu-24.04
-            build: { cc: gcc-14, cxx: g++-14, linker: ld, docs: true }
+            build: { cc: gcc-14, cxx: g++-14, linker: ld }
+            extra_meson_args: -Ddocs=true
 
           - name: "Linux x64 (Ubuntu 24.04) - Clang 19 with ASan and UBSan"
             os: ubuntu-24.04
             build: { cc: clang-19, cxx: clang++-19, linker: ld.lld-19, sanitize: 'address,undefined' }
+            extra_meson_args: -Dintrospection=disabled
 
-          - name: "Linux x64 (Ubuntu 24.04) - Clang 19 with TSan"
-            os: ubuntu-24.04
-            build: { cc: clang-19, cxx: clang++-19, linker: ld.lld-19, sanitize: 'thread' }
+          # We cannot reliably test this if GLib is built without TSan instrumentation.
+          #- name: "Linux x64 (Ubuntu 24.04) - Clang 19 with TSan"
+          #  os: ubuntu-24.04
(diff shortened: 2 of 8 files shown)


kleisauke:
This is starting to take shape. I've opened PR #4985 and https://github.com/google/oss-fuzz/pull/15235, which this PR depends on.

TSan is also far more reliable when GLib is built with it, which allows me to remove most of the previously needed suppressions. :)

kleisauke:
The changeset at https://github.com/kleisauke/libvips/compare/8ed96ff...6967b16 (which I'll split out later) fixes all known data races found in this PR. The TSan suppressions now apply only to GLib:
https://github.com/libvips/libvips/blob/6967b161e1aed4b3a8abb58a3428a1c958954891/suppressions/tsan.supp#L3-L16

Note that accessing a shared variable from multiple threads without synchronization (where at least one access is a write) is _always_ a data race and causes UB.

dloebl:
That's awesome! Thanks Kleis!
How severe were the TSan findings so far? I'm asking because we still are on libvips v8.16.0 because of https://github.com/libvips/libvips/discussions/4622

kleisauke:
I think(?) the only severe issue was the one fixed in commit 6967b161e1aed4b3a8abb58a3428a1c958954891 (the code was introduced in v8.16.1), where the eval callback could race with the one in `vips_sink_base_progress()`. However, this would likely only affect cases where `vips_image_set_progress()` is enabled and the [eval signal](https://www.libvips.org/API/current/signal.Image.eval.html) is connected via `g_signal_connect()`.

Other than that, the remaining data races involved only flags, which are likely not severe.

jcupitt:
> Note that accessing a shared variable from multiple threads without synchronization (where at least one access is a write) is always a data race and causes UB.

libvips does that quite often for events which are not time-critical. Do we need to lock all these, or can we just suppress them?

kleisauke:
> > Note that accessing a shared variable from multiple threads without synchronization (where at least one access is a write) is _always_ a data race and causes UB.
> 
> libvips does that quite often for events which are not time-critical. Do we need to lock all these, or can we just suppress them?

I think this falls under what are referred to as ["Benign" data races](https://www.usenix.org/legacy/event/hotpar11/tech/final_files/Boehm.pdf). This blog post also discusses why such races can still be problematic:
https://bartoszmilewski.com/2020/08/11/benign-data-races-considered-harmful/

I'm aware that GLib uses [`__ATOMIC_SEQ_CST` semantics for its `g_atomic` API](https://github.com/GNOME/glib/blob/2.88.0/glib/gatomic.h#L315-L342), so it's not using weak atomics. Let me run benchmarks on the previously mentioned changeset using https://github.com/kleisauke/vips-microbench, to see whether it has a measurable impact on performance.

jcupitt:
> https://bartoszmilewski.com/2020/08/11/benign-data-races-considered-harmful/

Ah interesting, I guess that's true. Then I suppose we should use glib atomics for these cross-thread "signal" variables.

It'd be good to avoid mutexes or cond vars if possible, they can limit SMP scaling.

kleisauke:
Marking this as ready for review. The `CIFUZZ_TEST=1` env variable is still needed until https://github.com/google/oss-fuzz/pull/15235 lands, but I don't think that's worth blocking on (as it would also hold up the follow-up PRs that fix these data races).

jcupitt (approved):

jcupitt:
This looks like a really nice improvement!

PR #4986, opened 2026-03-30
Fix quantizr backend returning success when quantization fails

felixbuenemann:
`quantizr_quantize()` returns `NULL` on failure instead of an error code, but `vips__quantise_image_quantize()` unconditionally returned `0` (success), causing callers to dereference a `NULL` result pointer.


jcupitt (approved):

jcupitt:
Ooops!

PR #3122, opened 2022-10-28
WIP -- add vips_palette

jcupitt:
See https://github.com/libvips/libvips/issues/3010 and https://github.com/libvips/libvips/issues/2841

So far this is just a simple wrapper around the quantiser.

```
$ vips palette
compute image palette
usage:
   palette in out [--option-name option-value ...]
where:
   in           - Input image, input VipsImage
   out          - Output image, output VipsImage
optional arguments:
   bitdepth     - Compute an N bit palette, input gint
			default: 8
			min: 0, max: 16
   effort       - Quantisation effort, input gint
			default: 7
			min: 1, max: 10
operation flags: sequential 
$ vips palette k2.jpg x.v
$ vipsheader x.v
x.v: 256x1 uchar, 4 bands, srgb
```

### To do

- [ ] expose the dither code as well
- [ ] make `pngload` attach the load palette? tiff could too
- [ ] deprecate `gif-palette` and get that to set plain `palette` as well
- [ ] pngsave needs an option that means "reuse any attached palette rather than regenerating"

PR #4987, opened 2026-03-30
Add built-in color quantizer (Wu's optimal 4D RGBA algorithm)

felixbuenemann:
Built-in palette quantization that works without external libraries and is used when libimagequant/quantizr are not available.
Pipeline: Wu partition → k-means refinement → Floyd-Steinberg dithering.

Wu's optimal 4D RGBA quantizer:
- 5-bit RGB + 4-bit alpha histogram (33^3 × 17 = 610K entries)
- 4D prefix-sum with 16-term inclusion-exclusion
- Greedy variance-maximizing partition into up to 256 boxes
- 10-bit perceptual gamma LUT (exponent 1.25) with channel weights (R=4, G=9, B=3, A=5)
- K-means palette refinement (9 passes, sort-means accelerated)

Nearest-color distance map:
- For each quantized cell, finds the palette entry with minimum weighted perceptual distance
- Precomputed finer dither map (6-bit RGB + 5-bit alpha) doubles bin resolution at palette boundaries

Floyd-Steinberg error diffusion with serpentine scanning:
- 4-neighbor error distribution (7/16, 3/16, 5/16, 1/16)
- Alpha-scaled RGB error propagation prevents bright-dot artifacts at transparency edges
- Inherited error clamping and output error dampening suppress cascade artifacts
- Parallel horizontal strip threading with 16-row overlap

Transparency handling:
- Fully transparent pixels (alpha=0) are excluded from the histogram so their garbage RGB does not pollute the palette
- Palette index 0 is reserved for a canonical {0,0,0,0} entry, improving compression for large transparent regions
- Few-colours fast path: lossless output (DSSIM=0) when image has ≤ max_colors unique values

Streaming and memory:
- Two-pass pipeline via `vips_sink` — no `vips_image_copy_memory` needed
- Sparse per-thread GHashTable histograms (-48% peak memory vs dense arrays)
- libimagequant backend: row callback via `liq_image_create_custom` (-72% peak memory)

### DIFF
--- a/libvips/conversion/reduce_colours.c
+++ b/libvips/conversion/reduce_colours.c
@@ -1,0 +1,206 @@
+/* reduce an image to n quantised colours
+ *
+ * 14/4/26
+ *	- initial version
+ */
+
+/*
+
+	This file is part of VIPS.
+
+	VIPS is free software; you can redistribute it and/or modify
+	it under the terms of the GNU Lesser General Public License as published by
+	the Free Software Foundation; either version 2 of the License, or
+	(at your option) any later version.
+
+	This program is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU Lesser General Public License for more details.
+
+	You should have received a copy of the GNU Lesser General Public License
+	along with this program; if not, write to the Free Software
+	Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+	02110-1301  USA
+
+ */
+
+#ifdef HAVE_CONFIG_H
+#include <config.h>
+#endif /*HAVE_CONFIG_H*/
+#include <glib/gi18n-lib.h>
+
+#include <vips/vips.h>
+
+#include "../foreign/quantise.h"
+#include "pconversion.h"
+
+typedef struct _VipsReduceColours {
+	VipsConversion parent_instance;
+
+	VipsImage *in;
+	int n;
+	double dither;
+	int effort;
+} VipsReduceColours;
+
+typedef VipsConversionClass VipsReduceColoursClass;
+
+G_DEFINE_TYPE(VipsReduceColours, vips_reduce_colours, VIPS_TYPE_CONVERSION);
+
+static int
+vips_reduce_colours_build(VipsObject *object)
+{
+	VipsConversion *conversion = VIPS_CONVERSION(object);
+	VipsReduceColours *rc = (VipsReduceColours *) object;
+	VipsImage **t = (VipsImage **) vips_object_local_array(object, 4);
+
+	VipsImage *index;
+	VipsImage *palette;
+	VipsImage *lut;
+	gboolean all_opaque;
+
+	if (VIPS_OBJECT_CLASS(vips_reduce_colours_parent_class)->build(object))
+		return -1;
+
+	/* Q hardcoded to 100; threshold_alpha hardcoded to FALSE.
+	 */
+	if (vips__quantise_image(rc->in, &index, &palette,
+			rc->n, 100, rc->dither, rc->effort, FALSE))
+		return -1;
+	t[0] = index;
+	t[1] = palette;
+
+	/* Detect a fully-opaque palette by walking its alpha column. The
+	 * palette is at most 256 entries, so this is trivially cheap.
+	 */
+	all_opaque = TRUE;
+	{
+		VipsRect r = { 0, 0, palette->Xsize, 1 };
+		VipsRegion *region = vips_region_new(palette);
+		const VipsPel *p;
+		int i;
+
+		if (vips_region_prepare(region, &r)) {
+			g_object_unref(region);
+			return -1;
+		}
+		p = VIPS_REGION_ADDR(region, 0, 0);
+		for (i = 0; i < palette->Xsize; i++)
+			if (p[i * 4 + 3] != 255) {
+				all_opaque = FALSE;
+				break;
+			}
+		g_object_unref(region);
+	}
+
+	/* When fully opaque, trim the LUT to RGB so vips_maplut produces a
+	 * 3-band output. Otherwise keep the 4-band RGBA LUT.
+	 */
+	lut = palette;
+	if (all_opaque) {
+		if (vips_extract_band(palette, &t[2], 0, "n", 3, NULL))
+			return -1;
+		lut = t[2];
+	}
+
+	if (vips_maplut(index, &t[3], lut, NULL) ||
+		vips_image_write(t[3], conversion->out))
+		return -1;
+
+	return 0;
+}
+
+static void
+vips_reduce_colours_class_init(VipsReduceColoursClass *class)
+{
+	GObjectClass *gobject_class = G_OBJECT_CLASS(class);
+	VipsObjectClass *vobject_class = VIPS_OBJECT_CLASS(class);
+	VipsOperationClass *operation_class = VIPS_OPERATION_CLASS(class);
+
+	gobject_class->set_property = vips_object_set_property;
+	gobject_class->get_property = vips_object_get_property;
+
+	vobject_class->nickname = "reduce_colours";
+	vobject_class->description =
+		_("reduce an image to n quantised colours");
+	vobject_class->build = vips_reduce_colours_build;
+
+	operation_class->flags = VIPS_OPERATION_SEQUENTIAL;
+
+	VIPS_ARG_IMAGE(class, "in", 1,
+		_("Input"),
+		_("Input image"),
+		VIPS_ARGUMENT_REQUIRED_INPUT,
+		G_STRUCT_OFFSET(VipsReduceColours, in));
+
+	VIPS_ARG_INT(class, "n", 2,
+		_("N"),
+		_("Number of colours"),
+		VIPS_ARGUMENT_REQUIRED_INPUT,
+		G_STRUCT_OFFSET(VipsReduceColours, n),
+		2, 256, 256);
+
+	VIPS_ARG_DOUBLE(class, "dither", 3,
+		_("Dither"),
+		_("Floyd-Steinberg dithering level"),
+		VIPS_ARGUMENT_OPTIONAL_INPUT,
+		G_STRUCT_OFFSET(VipsReduceColours, dither),
+		0.0, 1.0, 1.0);
+
+	VIPS_ARG_INT(class, "effort", 4,
+		_("Effort"),
+		_("Quantisation effort"),
+		VIPS_ARGUMENT_OPTIONAL_INPUT,
+		G_STRUCT_OFFSET(VipsReduceColours, effort),
+		1, 10, 7);
+}
+
+static void
+vips_reduce_colours_init(VipsReduceColours *rc)
+{
+	rc->dither = 1.0;
+	rc->effort = 7;
+}
+
+/**
+ * vips_reduce_colours: (method)
+ * @in: input image
+ * @out: (out): output image
+ * @n: number of colours
+ * @...: `NULL`-terminated list of optional named arguments
+ *
+ * Optional arguments:
+ *
+ * * @dither: dithering level
+ * * @effort: quantisation effort
+ *
+ * Quantise @in to a palette of at most @n colours and remap every pixel to
+ * its nearest entry, returning the same-size sRGB(A) uchar image. Uses
+ * libimagequant, quantizr, or the built-in Wu quantiser, whichever is
+ * available at compile time. @n may be in the range 1 to 256.
+ *
+ * If every entry of the chosen palette is fully opaque, the output is a
+ * 3-band sRGB image; otherwise it has the alpha channel and is 4-band.
+ *
+ * @dither sets the Floyd-Steinberg dithering level (0 = none, 1 = full).
+ * @effort sets the quantiser's speed/quality trade-off (1 = fastest,
+ * 10 = best palette).
+ *
+ * ::: seealso
+ *     [method@Image.dominant_colours], [method@Image.maplut].
+ *
+ * Returns: 0 on success, -1 on error
+ */
+int
+vips_reduce_colours(VipsImage *in, VipsImage **out, int n, ...)
+{
+	va_list ap;
+	int result;
+
+	va_start(ap, n);
+	result = vips_call_split("reduce_colours", ap, in, out, n);
+	va_end(ap);
+
+	return result;
+}
--- a/libvips/foreign/quantise-builtin.c
+++ b/libvips/foreign/quantise-builtin.c
@@ -1,0 +1,2019 @@
+/* Built-in color quantizer: Wu's optimal + Floyd-Steinberg dithering.
+ *
+ * Provides baseline palette quantization when libimagequant/quantizr
+ * are not available.
+ *
+ * Pipeline: Wu partition → nearest-color map → F-S dithering
+ *
+ * References:
+ *   Xiaolin Wu, "Efficient Statistical Computations for Optimal Color
+ *   Quantization", Graphics Gems II, 1991.
+ */
+
+/*
+
+	This file is part of VIPS.
+
+	VIPS is free software; you can redistribute it and/or modify
+	it under the terms of the GNU Lesser General Public License as published by
+	the Free Software Foundation; either version 2 of the License, or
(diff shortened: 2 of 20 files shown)


felixbuenemann:
The quantiser is actually quite good for its simplicity. It doesn't quite match libimageqaunt, but it is faster and only required ~14MB RAM, while libimagequant memory use depends on image size.

**Here's a sample original 24-Bit PNG:**
dice_original

**Converted with the new built-in quantiser:**
dice_builtin


**Converted with libimagequant:**
dice_liq

**Converted with quantizr:**
dice_quantizr

felixbuenemann:
@jcupitt I have one optimization for this in the pipeline, that I'm unsure wether to commit: All the current quantizers use a full memory image of the source due to api restrictions. I have a branch that instead uses `vips_sink_disc` to use streaming quantization and dithering, which saves around 50% memory and is also faster. But because some of the low level vips quantiser api's work on raw pixels, I end up converting those back to VipsImage for processing (ori I would break the third party quantizers). Any ideas on how to handle that? I could push the optimization if you'd like to take a look at it.

jcupitt:
Hi again @felixbuenemann,

I like the idea of our own quantise and dither. We could expose them as first-class libvips operations and remove the quantise and dither from at least GIF and PNG save. And removing a dependency sounds very good too. Plus, as you say, a more vipsy quantiser has potential memory savings.

master is supposed to be for experimentation, so how about merging this PR more or less as is, then looking at ways to exploit it in new operations and refactoring of the associated savers? Then, if it all works out, removing the `vips__quant*()` stuff before 8.19.

What do you think? @lovell, @kleisauke, ok with you?

felixbuenemann:
@jcupitt I think you misunderstood my comment. All of the quantizers (built-in, libimagequant, quantizr) do both quantization and dithering. But the vips_quantise_* operations implemented by each quantizer usually take raw pixels as input, so the whole source id decoded to an uncompressed image in memory first. For the built-in quantizer I can slip this and instead work on slices provides by `vips_sink_disc`, but to be compatible with the existing quantizer function it means must sometimes convert the passed in raw pixels back to a VipsImage, eg. for gifsave path. It works and is still faster, but feels a bit hacky.

jcupitt:
I understand, I'm proposing as a longer term goal:

- split quant and dither into separate libvips operations
- quant would work like one of the histogram operations
- dither would take an RGB(A) image plus the quant table and output an index image
- the GIF and PNG savers would call these things and not use the current quantise API
- we remove all the `vips__quant*()` stuff

To get there, we would:

- merge this PR mostly as is
- another PR (maybe?) does the split and implements quant and dither
- another reworks the savers
- perhaps a final one removes the quantiser dependency

felixbuenemann:
Ah OK, I get your idea now: splitting quantising and dithering into different operations. But given how libimagequant and quantizr work, we can only split quantization (palette generation) and remapping into separate steps. dithering is done as part of the remapping (if requested).

jcupitt:
Yes, I meant remap + dither as an operation.

lovell:
@jcupitt Happy to (eventually) remove the dependency on libimagequant. We probably need to think about [quantizr](https://github.com/DarthSim/quantizr) too - perhaps its API could adapt?

felixbuenemann:
I‘d like to add that libimagequant has better quality on challenging pictures. I haven’t fully explored how much adding eg. k-means passes could bring the built-in quantizer closer. I haven’t compared quantizr at all.

felixbuenemann:
I have made significant updates to the quantizer code:
* Added streaming for built-in wu and libimagequant (reduces memory usage)
* Added threading and error dampening to Floyd-Steinberg dithering
* Added 0-9 k-means passes on palette based on `effort` value
* Changed built-in quantizer to internally work in perceptual colorspace
* Added RGBA weights based on human eye cone sensitivity
* Added Floyd-Steinberg adaptive-dithering

**The DSSIM scores improved a lot:**
The built-in wu's quantizer is now ~20% better quality than quantizr and ~22% worse than libimagequant.

felixbuenemann:
@jcupitt I am very happy with the quantization and dithering results. The only thing that libimagequant does notably better is transparent edges, because it does edge detection. Notice in the dice example above, how the edge of the green dice is smooth in liq, but has a bit jagged edges with the built-in quantizer.

felixbuenemann:
I updated the perceptual color processing in the built-in quantizer to 10-Bit to recover some distinct values lost due to gamma curve compression. This gives a slight quality boost and we're now within 20% of libimagequant quality.

jcupitt:
Wow @felixbuenemann, that's very cool!

I'm sorry, I'm still trapped in a terrible deadline -- I have to finish a big project for a conference in May. I'll be coming back to libvips after that, or perhaps another maintainer can review your PR.

felixbuenemann:
Another 9% DSSIM improvement by adding a refinement pass on the bin assignment and tuning k-means passes. We're now within 11% of libimagequant  and 27% better than quantizr. Also reduced memory usage by 48%.

felixbuenemann:
## Quantizer Benchmark Results

256 colours, quantette CQ100 + Unsplash image set, 111 images.
Earth12k.jpg (144 MP) for peak memory measurement.
Default effort=7, dither=1.0, VIPS_CONCURRENCY=10.

| Quantizer | Avg DSSIM | Images | Time | Peak Memory |
|---|---|---|---|---|
| builtin | 0.001319 | 111 | 38s | 317 MB |
| libimagequant | 0.001184 | 111 | 34s | 238 MB |
| quantizr | 0.001799 | 111 | 43s | 826 MB |

felixbuenemann:
@lovell Would you like to review this?

lovell:
This looks great, thanks Felix, I'm currently focussing any time I have on the next version of sharp. Once that's done I'll be able to review your good work here.

jcupitt:
Related issue: https://github.com/libvips/php-vips/issues/256 php-vips would like a `reduceColors()` method (just the quantisation step).

felixbuenemann:
I believe all three quantizers (buil-tin, libimagerquant, quantizr) can quantize down to an arbitrary number of colors. We just need to add a new operation to return the quantization result as a new image. I could certainly add that.

kipcole9:
> Related issue: [libvips/php-vips#256](https://github.com/libvips/php-vips/issues/256) php-vips would like a `reduceColors()` method (just the quantisation step).

[image](https://github.com/elixir-image/image) would very much like a way to return the palette generated from quantisation too.  Just the quantisation part like it sounds `php-vips` would like.

felixbuenemann:
@kipcole9 That sounds a lot like the dominant colors feature request related to this feature. Would a `1px x {number colors}` palette image work?

kipcole9:
It most certainly would!  Thanks for the consideration. At the moment I save to  GIF and then manually extract the palette which is definitely not a great approach - even if it works!

felixbuenemann:
@kipcole9 I pushed a new `dominant_colours(n)` operation in this branch. Can you test it?

**pyvips-Example:**
```python
import pyvips

im = pyvips.Image.new_from_file("input.jpg")

# Get up to 5 dominant colours
palette = im.dominant_colours(5)

# palette is an N x 1 RGBA uchar image (N <= 5)
print(f"got {palette.width} colours")

# Read each colour out as [R, G, B, A]
for x in range(palette.width):
    r, g, b, a = palette(x, 0)
    print(f"  #{int(r):02x}{int(g):02x}{int(b):02x} (alpha={int(a)})")

# Save the palette strip as a tiny PNG for inspection
palette.write_to_file("palette.png")
```

felixbuenemann:
> Related issue: [libvips/php-vips#256](https://github.com/libvips/php-vips/issues/256) php-vips would like a `reduceColors()` method (just the quantisation step).

I added a `reduce_colours(n, dither)` operation that does this. It takes the quantises the source image, then takes the resulting 1-band and the palette as a LUT to re-generate an sRGB(A) image, similar to `falsecolour()`.

**pyvips-Example:**
```python
import pyvips

im = pyvips.Image.new_from_file("input.jpg")

# n=1..256 (required), dither=0..1 (default 1.0), effort=1..10 (default 7)
out = im.reduce_colours(16, dither=0.5, effort=10)

out.write_to_file("output_16colours.png")
```

felixbuenemann:
I had to add a TSan suppression for `vips_dominant_colours()` because it can't quite understand the vips worker thread management. It flagged access to a LUT that was built once on the main thread then accessed on `vips_sink` worker threads.

felixbuenemann:
@jcupitt I have to emulate libimagequant's and quantizr's respective algorithms to support quantization to 1 color, since both require at least 2 colors. I'm not really happy with it and I noticed that 1 color quantization is pretty useless (usually a grayish color), so its use for `vips_docminant_colours()` is also limited.

Do you think we should just go with minimum two colors for `vips_dominant_colors()` and drop that extra code?

jcupitt:
I agree, extra complexity for a corner case no one will use is a bad idea. Delete away!

felixbuenemann:
Removed all the special code added to handle 1 colour quantisation, so now both `vips_dominant_colours()` and `vips_reduce_colours()` require at least 2 colours.

felixbuenemann:
@jcupitt I'm not sure if quantisation is the right answer for the dominant colours operation: during quantisation the colours change, because they get averaged. So if you ask for the 2 most dominant colours in a photo, you might end up with a darkish gray and a brighter gray, that shows kind of a silhouette of the original picture, but I wouldn't call those the most dominant colours. The most dominant colours would have to be determined by counting each rgba value and selecting by count or not?

Then again @kipcole9 asked for a way to get the quantisation palette, which is what the current `vips_dominant_colours()` provides.

Maybe we should give the operation a better name? Ideas?

felixbuenemann:
@kipcole9 What is your use case for getting the quantisation palette?

kipcole9:
@felixbuenemann sorry for the slow responses - just other project priorities the last couple of days.

Two primary cases I'm working on:
* A implementation in my [Image](https://github.com/elixir-image/image) lib of something similar to https://github.com/axe312ger/sqip
* Using the dominant colours of an existing image to drive palette creation in my [Color](https://github.com/elixir-image/color) library (which has a very simple visualizer at https://elixir-color.fly.dev

I can't justify a special feature request for this in libvips. My thinking was that since Felix is doing tremendous work on quantisation, dominant colours is a very natural outcome of that existing work.

felixbuenemann:
@kipcole9 wrote:
> Two primary cases I'm working on:
> 
> * A implementation in my [Image](https://github.com/elixir-image/image) lib of something similar to https://github.com/axe312ger/sqip

This is a use case where the quantisation palette works very well.

> * Using the dominant colours of an existing image to drive palette creation in my [Color](https://github.com/elixir-image/color) library

This sound like a use case where you want the real dominant colours (the quantisation already produces averaged colours, which can differ a lot, especially if you quantise to eg. 8 colours instead of 256)

jcupitt:
This snoozing PR has a possible API (edit: should be https://github.com/libvips/libvips/pull/3122, thanks Kleis)

It has:

```C

/**
 * vips_palette: (method)
 * @in: input image 
 * @out: (out): output image
 * @...: %NULL-terminated list of optional named arguments
 *
 * Optional arguments:
 *
 * * @effort: %gint, how much effort to spend on the search
 * * @bitdepth: %gint, bitdepth for generated palette
 *
 * Find the most common colours in an 8-bit RGB or RGBA image.
 *
 * Set @bitdepth to control the size of the computed palette. By default it
 * finds an 8-bit palette, or 256 colours. 
 *
 * Set @effort to control the CPU effort (1 is the fastest,
 * 10 is the slowest, 7 is the default).
 *
 * See also: vips_dither(), vips_maplut().
 *
 * Returns: 0 on success, -1 on error
 */
int
vips_palette( VipsImage *in, VipsImage **out, ... )
```

And:

```C

/**
 * vips_dither: (method)
 * @in: input image 
 * @out: (out): output image
 * @palette: (in): palette image
 * @...: %NULL-terminated list of optional named arguments
 *
 * Optional arguments:
 *
 * * @dither: %gdouble, dithering level
 *
 * Dither @in using @palette. 
 *
 * Use @dither to set the degree of Floyd-Steinberg dithering.
 *
 * See also: vips_palette().
 *
 * Returns: 0 on success, -1 on error
 */
int
vips_dither( VipsImage *in, VipsImage **out, VipsImage *palette, ... )
```

jcupitt:
I've finished my deadline! Phew.

My feeling is that we should review and merge this PR without many more changes (it's already large enough), then in follow-up PRs remove support for other quantisers and perhaps polish the API.

Any thoughts?

kleisauke:
PR #3122 is the snoozing PR. I'll have a review of this PR next week.

@felixbuenemann Just wondering, did you use an LLM to assist with this PR? I'm asking because I noticed the commit messages in this PR include Unicode symbols that are commonly produced by LLMs (i.e. `×`, `±`, `→`, `—`, `≤`, etc.). I also came across this earlier commit: https://github.com/felixbuenemann/libvips/commit/1006451c7363df384217f9edb5e936e2a364ebf0, which seems consistent with that pattern.

> I had to add a TSan suppression for `vips_dominant_colours()` because it can't quite understand the vips worker thread management. It flagged access to a LUT that was built once on the main thread then accessed on `vips_sink` worker threads.

PR #4981 will likely fix this.

felixbuenemann:
@jcupitt I can back out the two new operations and submit them as new PRs, after (if) this is merged.

@kleisauke Yes, I am. Is there a policy against it on this project or do you just want the commit messages cleaned up?

kleisauke:
@felixbuenemann Thanks for disclosing the use of an LLM. It means I'll review this PR more carefully than most other contributions.

AFAIK, libvips don't have an LLM/GenAI policy, and I'm not sure where the other libvips maintainers stand on this yet. Personally, I'd prefer something similar to GTK's policy:
https://gitlab.gnome.org/GNOME/gtk/-/blob/main/CONTRIBUTING.md#ai-contribution-policy
(especially the "Disclosure" section, which led me to add [this comment](https://github.com/llvm/llvm-project/pull/194862#issuecomment-4451566131).)

libvips/foreign/quantise-builtin.c
```diff
@@ -0,0 +1,2019 @@
...
+		nearest_map = g_new(VipsPel, DITHER_MAP_SIZE);
+		build_dither_map(pal, coarse_map, nearest_map);
+		g_free(coarse_map);
+	}
+	else {
+		nearest_map = g_new(VipsPel, CACHE_SIZE);
+		build_nearest_map(pal, palette->count, nearest_map);
+	}
+
+	n_threads = vips_concurrency_get();
```
lovell, line 1857: Are you able to explain a little more about the approach to threading here? It looks like `vips_concurrency_get()` new threads are created per operation and each of these relies on `vips_region_prepare()` for its strip of data. Is there any co-ordination here? Can we test with more complex operations known to affect sequential behaviour such as 90 degree rotation as part of the same pipeline?

felixbuenemann: The threads here are a one-shot fork/join for the dithered remap pass only, not a general threadpool — Floyd–Steinberg is row-recursive (row y consumes the error buffer produced by row y−1), so it can't be expressed as a vips_sink over tiles. The only parallelism available is between independent horizontal strips, each processed strictly top-to-bottom by a single thread with its own error buffer.

I did test `rot90` which works fine, however I _did_ notice that `access-sequential` is broken, due to the two-pass read (histogram, then remap), however I'm not sure of the best solution. Existing libvips code does either:

* nothing (fails with out of order read, like the quantizer)
* vips_copy_memory (buffer to memory)
* buffer to mmaped temp file or memory based on threshold (only loaders)

I like the last approach best, but currently only loaders use it, so I'm not sure it is ok.

libvips/conversion/reduce_colours.c
```diff
@@ -0,0 +1,206 @@
...
+	if (vips__quantise_image(rc->in, &index, &palette,
+			rc->n, 100, rc->dither, rc->effort, FALSE))
+		return -1;
+	t[0] = index;
+	t[1] = palette;
+
+	/* Detect a fully-opaque palette by walking its alpha column. The
+	 * palette is at most 256 entries, so this is trivially cheap.
+	 */
+	all_opaque = TRUE;
```
jcupitt: I'd swap this loop for eg.:

```C
    double min;
    if (vips_extract_band(t[1], &t[2], 3, NULL) ||
        vips_min(t[2], &min, NULL))
        return -1;
    all_opaque = min == 255;
```

As you say, it's just 256 entries.

kleisauke:
As of today, libvips does not accept AI-generated contributions:

PR #4994, opened 2026-04-08
Improve macOS vips concurrency detection

felixbuenemann:
On Apple Silicon, use `hw.perflevel0.logicalcpu` to detect the number of performance cores and use that as the default thread count, avoiding scheduling work on slower efficiency cores.

Falls back to `g_get_num_processors()` on Intel Macs where the sysctl key does not exist.

This fixes poor performance with automatic vips concurrency detection on Apple Silicon Macs, since with the default of all cpu cores, some threads must be scheduled on slower efficiency cores (eg. M4 Pro chip has 10 performance and 4 efficiency cores).

### DIFF
--- a/libvips/iofuncs/thread.c
+++ b/libvips/iofuncs/thread.c
@@ -56,6 +56,10 @@
 #include <windows.h>
 #endif /*G_OS_WIN32*/
 
+#ifdef __APPLE__
+#include <sys/sysctl.h>
+#endif /*__APPLE__*/
+
 /* Maximum value we allow for VIPS_CONCURRENCY. We need to stop huge values
  * killing the system.
  */
@@ -175,8 +179,24 @@ vips__concurrency_get_default(void)
 				) &&
 		(x = atoi(str)) > 0)
 		nthr = x;
-	else
+	else {
 		nthr = g_get_num_processors();
+
+#ifdef __APPLE__
+		/* On Apple Silicon, prefer the number of performance cores to
+		 * avoid scheduling work on slower efficiency cores.
+		 */
+		{
+			int pcore_count = 0;
+			size_t len = sizeof(pcore_count);
+
+			if (sysctlbyname("hw.perflevel0.logicalcpu",
+					&pcore_count, &len, NULL, 0) == 0 &&
+				pcore_count > 0)
+				nthr = pcore_count;
+		}
+#endif /*__APPLE__*/
+	}
 
 	if (nthr < 1 ||
 		nthr > MAX_THREADS) {


felixbuenemann:
Theoretically we could check the same on Windows and Linux for hybrid CPUs:

- **Windows:** `GetSystemCpuSetInformation` returns per-core `EfficiencyClass`. Higher = faster. Works for both Intel hybrid and ARM. On homogeneous systems all cores report the same class.
- **Linux:**
  - `/sys/devices/cpu_core/cpus` – Intel hybrid only (Alder Lake+). A cpumask of P-cores. Doesn't exist on ARM or non-hybrid Intel.
  - `/sys/devices/system/cpu/cpu*/cpu_capacity` – Works on both Intel hybrid (5.18+) and ARM big.LITTLE. P-cores get 1024, E-cores get a lower value. On homogeneous systems all cores report 1024 or the file is absent.

Hoewever I don't have hardware I could test that on.

jcupitt:
Oh, nice! We could also consider making a PR on `g_get_num_processors()`, I'm sure they'd like this improvement too.

felixbuenemann:
@jcupitt I think `g_get_num_processors()` returning the total number of cores is correct, it is just not useful for us, because we want to know at which vips concurrency we get the best performance.

felixbuenemann:
@jcupitt Is this OK to merge? I'm reluctant to target more platforms without the ability to test if it works.

jcupitt:
This special apple code is a bit ugly, but I can see the purpose. 

I made a quick PR for glib based on this (I hope that's OK): https://gitlab.gnome.org/GNOME/glib/-/merge_requests/5153 

If that gets merged, perhaps we can remove this code. In the meantime, let's go ahead.

jcupitt (approved):

kleisauke:
For context, #3802 includes a prior discussion on limiting libvips' concurrency to performance cores (though only on Windows). The same CPU type is also discussed in Intel MPI [here](https://community.intel.com/t5/Intel-MPI-Library/Efficient-Core-Selection-Slowing-Down-Analysis/m-p/1725207) (where it can be worked around using the [`I_MPI_PIN_PROCESSOR_EXCLUDE_LIST=ecores` env variable](https://www.intel.com/content/www/us/en/developer/articles/release-notes/mpi-library-release-notes-linux.html#inpage-nav-3-2:~:text=CPU%20pinning.-,New%20options%20added,pcores%29)).

That said, Rust's [`thread::available_parallelism()`](https://doc.rust-lang.org/std/thread/fn.available_parallelism.html) and [`navigator.hardwareConcurrency`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/hardwareConcurrency) on the web do _not_ distinguish between performance and efficiency cores, though that may be comparing apples to oranges.

felixbuenemann:
@jcupitt I think `g_get_num_processors()` should return the total core count, because that's what the function name implies. But for vips concurrency, we know the work is usually cpu bound, so we know using only p-cores improves performance.

jcupitt:
> @jcupitt I think g_get_num_processors() should return the total core count, because that's what the function name implies. 

I don't like the name much :( 

It was originally a libvips utility function to set the threadpool size that we pushed upstream to glib. If it's not picking a good threadpool size, it needs fixing, or it's useless. 

The alternative would be to copy the whole function back into our tree and make it do exactly what we want, but I think fixing the glib version would be better.

jcupitt:
@felixbuenemann your patch has been merged into glib! So we can remove this workaround in libvips in ooo hmmmm maybe 2035.

https://gitlab.gnome.org/GNOME/glib/-/merge_requests/5153

PR #4996, opened 2026-04-10
fix: luminance scaling for uhdrsave

mertalev:
### Description

libultrahdr uses a different scaling convention from scRGB (1.0 = 203 nits rather than 1.0 = 80 nits). This PR rescales to and from 203 nits for more accurate luminance.

Additionally, libultrahdr assumes a peak of 10000 nits by default. Setting the peak to the 99th percentile luminance produces more accurate gain.

This was initially in #4928 but split out to keep the scope of changes smaller.

Resolves [this](https://github.com/libvips/libvips/pull/4954#issuecomment-4218024992) comment

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -22,6 +22,7 @@ date-tbd 8.19.0
 - uhdrsave: add existing JPEG options for base image [lovell]
 - csvload: add maximum limit on column count [lovell]
 - pdfload_buffer: only search the first few bytes in is_a [violetviolinist]
+- add percent_lum: find threshold for percent of pixels by BT.709 luminance [mertalev]
 
 tbd 8.18.3
 
--- a/doc/libvips-histogram.md
+++ b/doc/libvips-histogram.md
@@ -24,6 +24,7 @@ CIELAB images, but might be useful elsewhere.
 
 * [method@Image.maplut]
 * [method@Image.percent]
+* [method@Image.percent_lum]
 * [method@Image.stdif]
 * [method@Image.hist_cum]
 * [method@Image.hist_norm]
--- a/libvips/colour/uhdr2scRGB.c
+++ b/libvips/colour/uhdr2scRGB.c
@@ -69,6 +69,11 @@ G_DEFINE_TYPE(VipsUhdr2scRGB, vips_uhdr2scRGB, VIPS_TYPE_COLOUR);
 /* Derived from the apache-licensed applyGain() method of libuhdr.
  */
 
+/* Rescale from libuhdr's linear convention (1.0 = 203 nits, the ITU-R
+ * BT.2408 SDR reference white) to scRGB (1.0 = 80 nits).
+ */
+#define UHDR_TO_SCRGB (203.0f / 80.0f)
+
 /* Monochrome gainmap, colour image. Probably the most common case.
  */
 static void
@@ -97,9 +102,9 @@ vips_uhdr2scRGB_mono(VipsUhdr2scRGB *uhdr,
 
 		float gaing = exp2(boostg);
 
-		q[0] = ((r + uhdr->offset_sdr[1]) * gaing) - uhdr->offset_hdr[1];
-		q[1] = ((g + uhdr->offset_sdr[1]) * gaing) - uhdr->offset_hdr[1];
-		q[2] = ((b + uhdr->offset_sdr[1]) * gaing) - uhdr->offset_hdr[1];
+		q[0] = (((r + uhdr->offset_sdr[1]) * gaing) - uhdr->offset_hdr[1]) * UHDR_TO_SCRGB;
+		q[1] = (((g + uhdr->offset_sdr[1]) * gaing) - uhdr->offset_hdr[1]) * UHDR_TO_SCRGB;
+		q[2] = (((b + uhdr->offset_sdr[1]) * gaing) - uhdr->offset_hdr[1]) * UHDR_TO_SCRGB;
 		q += 3;
 	}
 }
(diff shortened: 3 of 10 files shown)


dshatz:
Can confirm, looks great!

Input:
https://drive.google.com/file/d/1Yl5x7gxwO93DL9CW4eyzE9UQ6ZXqJEIy/view?usp=sharing

Output:
out

jcupitt:
Hi @mertalev,

This looks nice! But scanning the input image to find the peak brightness will break things like thumbnail, where you can only scan once. 

`uhdrsave` would need to render the whole image into memory (at 3 x float per pixel), then compute the peak nits, and then save.

A lower memory solution would be to either add an optional `peak_nits` argument to `uhdrsave`, or perhaps a new metadata item like `gainmap-peak-nits`. We could possibly have `vips_peak_nits()` to compute this value.

What do you think?

mertalev:
> Hi @mertalev,
> 
> This looks nice! But scanning the input image to find the peak brightness will break things like thumbnail, where you can only scan once.
> 
> `uhdrsave` would need to render the whole image into memory (at 3 x float per pixel), then compute the peak nits, and then save.
> 
> A lower memory solution would be to either add an optional `peak_nits` argument to `uhdrsave`, or perhaps a new metadata item like `gainmap-peak-nits`. We could possibly have `vips_peak_nits()` to compute this value.
> 
> What do you think?

Oh, that's a great point I hadn't considered! I agree the general idea of having the caller set the peak and making it easy to compute would be the solution then. That would be partly resolved with #4954, which allows setting the target peak brightness among other things.

On the nit calculation, using the *actual* peak led to the gain being a bit lower than expected since a stray pixel is enough to dampen the overall gain, so I think an API like `vips_nit_percentile` (naming is hard) is best. I also think we could default this value to something lower when not provided since I imagine the vast majority of content peaks at much lower than 10000 nits in practice.

jcupitt:
libvips has `vips_percent()`, which could maybe be reused. In python, something like (untested):

```python
scrgb = ... some scrgb image
lum = 0.2126 * scrgb[0] + 0.7152 * scrgb[1] + 0.0722 * scrgb[2]

# scale to ushort for histogram ... over 6500 will be clipped, but that's probably ok 
int_lim = (lum * 10).cast("ushort")
peak_nits = int_lum.percent(99) / 10.0
```

mertalev:
> libvips has `vips_percent()`, which could maybe be reused. In python, something like (untested):
> 
> ```python
> scrgb = ... some scrgb image
> lum = 0.2126 * scrgb[0] + 0.7152 * scrgb[1] + 0.0722 * scrgb[2]
> 
> # scale to ushort for histogram ... over 6500 will be clipped, but that's probably ok 
> int_lim = (lum * 10).cast("ushort")
> peak_nits = int_lum.percent(99) / 10.0
> ```

I tested
```python
lum = im.recomb([[0.2126, 0.7152, 0.0722]])
int_lum = (lum * 10).cast("ushort")
int_lum.percent(99) / 10.0 * 80.0
```
vs. a theoretical
```python
im.percent_lum(99.0, max=10000.0 / 80.0) * 80.0
```
`percent` works, but there are some issues:
1. The output is less accurate. I tested with 65535 buckets for the nit histogram as ground truth (it uses 1024 buckets in practice since it's accurate enough)

| Approach                           | 99th (nits) | 99.9th (nits) |
|------------------------------------|------------:|--------------:|
| `vips_percent_lum` (65536 buckets)      |     2578.02 |       2624.99 |
| `vips_percent_lum` (1024 buckets)       |     2574.61 |       2625.39 |
| `vips_percent`                     |     2624.00 |       2880.00 |

For another image, it got 80 nits instead of the expected 68 nits, which is a fairly big relative error. I'm not sure if this is a difference in `percent`'s bucketing or because of the ushort scaling.

2. It's slower, testing on a 6048×8064 image

| Approach                           | Latency (ms) |
|------------------------------------|----------:|
| `vips_percent_lum` (65536 buckets)      |     11.17 |
| `vips_percent_lum` (1024 buckets)       |      8.76 |
| `vips_percent`                     |     31.94 |

Given providing the peak is the intended way to get the right gain for uhdrsave, my preference would be the separate API since it's easier to use, more accurate and faster.

mertalev:
I've tentatively added the API and reverted the change to calculate the peak in uhdrsave, but can of course remove it depending on preference. I kept the 80<->203 rescaling since it's needed regardless. From there, I could either add the peak brightness parameter in this PR or defer to #4954 which also adds it.

jcupitt:
OK, let's add `percent_lum` and defer the extra uhdr API for #4954, as you suggest.

We just need to fix the tests (the uhdr roundtrip has changed).

jcupitt:
Great! The docs check is failing:

> ERROR: Symbol '[method@Image.percent_lum]' is not listed in '/home/runner/work/libvips/libvips/build/../doc/libvips-histogram.md'

ie. `doc/libvips-histogram.md` needs a note about the new `vips_percent_lum()` method.

This is an API addition, so a one-line note in changelog too, eg. perhaps:

```
- add percent_lum: find threshold for percent of pixels by BT.709 luminance [mertalev]
```

mertalev:
> Great! The docs check is failing:
> 
> > ERROR: Symbol '[[method@Image.percent_lum](mailto:method@Image.percent_lum)]' is not listed in '/home/runner/work/libvips/libvips/build/../doc/libvips-histogram.md'
> 
> ie. `doc/libvips-histogram.md` needs a note about the new `vips_percent_lum()` method.
> 
> This is an API addition, so a one-line note in changelog too, eg. perhaps:
> 
> ```
> - add percent_lum: find threshold for percent of pixels by BT.709 luminance [mertalev]
> ```
Oops, I keep forgetting to do that haha

dshatz:
Hi 
Waiting for this to proceed with #4954!

jcupitt (approved):

jcupitt:
Oooof finally back on this, I reread it and it looks fine. Let's go!

Thank you for fixing this @mertalev

PR #4998, opened 2026-04-15
object: guard against invalid interpolators

kleisauke:
Resolves: https://issues.oss-fuzz.com/issues/502697518.

### DIFF
--- a/libvips/iofuncs/object.c
+++ b/libvips/iofuncs/object.c
@@ -1941,10 +1941,29 @@ vips_object_set_argument_from_string(VipsObject *object,
 		g_value_init(&gvalue, VIPS_TYPE_SOURCE);
 		g_value_set_object(&gvalue, source);
 
-		/* Setting gvalue will have upped @out's count again,
+		/* Setting gvalue will have upped @source's count again,
 		 * go back to 1 so that gvalue has the only ref.
 		 */
 		g_object_unref(source);
+	}
+	else if (g_type_is_a(otype, VIPS_TYPE_INTERPOLATE)) {
+		VipsInterpolate *interpolate;
+
+		if (!value) {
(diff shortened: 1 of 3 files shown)


libvips/iofuncs/object.c
kleisauke, line 2032: I think we may be able to remove this block on `master` after this PR, since all optional and required `VIPS_TYPE_OBJECT`-derived arguments are already handled by the if statements earlier.

libvips/iofuncs/object.c
```diff
@@ -2676,7 +2695,7 @@ vips_object_to_string_optional(VipsObject *object,
  * @buf: write string here
  *
  * The inverse of [ctor@Object.new_from_string]: turn @object into eg.
- * `"VipsInterpolateSnohalo1(blur=.333333)"`.
```
kleisauke, line 2679: For context, `snohalo1` was renamed to `snohalo` with commit 415ccb5c08e944c88b74bd035d3e3b3509ae64d4, which was later removed in commit 910df771611a41d7e0faeeab3d3f66694210e700.

libvips/iofuncs/object.c
kleisauke: Actually, `VIPS_TYPE_TARGET` is not handled by the earlier if statements, so this code path is also used when running:
```console
$ vips ppmsave_target x.v VipsTarget[filename=x.ppm]
```

By adding a special case for `VIPS_TYPE_TARGET`  (using e.g. `vips_target_new_from_options()`), we can instead support:
```console
$ vips ppmsave_target x.v [filename=x.ppm]
```

This makes the syntax consistent with:
```console
$ vips ppmload_source [filename=x.ppm] x.v
```

lovell (approved):

PR #4999, opened 2026-04-15
foreign: guard against NULL filenames

kleisauke:
Resolves:
https://issues.oss-fuzz.com/issues/502697518#comment4
https://issues.oss-fuzz.com/issues/503132532


lovell (approved):
💯

PR #5003, opened 2026-04-19
vips_fuzzer: allow filename-based load/save paths

dloebl:
Route the `filename` arg of foreign-load/save ops through a `/tmp` file instead of a random path from the input (load) or `/dev/null` (save).

Unlocks coverage of code that memory sources are unable to reach: Mostly `libvips/iofuncs/source.c` (e.g. mmap
fast paths).

Temp files are written lazily and `g_unlink`ed per iteration, so ops without a `filename` arg aren't slowed down.

**ToDo:**
- [x] Improve the seed corpus for `filename` operations further
- [x] Check for operations that aren't cleaned up correctly (`VipsForeignSaveDz` might be a candidate) and block them
- [x] Local test fuzzing run


kleisauke:
Great! This also means `vipsload` is now covered (which is file-based only).

Could you omit the seed corpus changes from this PR? They should now be added to https://github.com/libvips/seed-corpora instead.

dloebl:
@kleisauke Sounds good! I've opened a PR here with the seed corpus for this work: https://github.com/libvips/seed-corpora/pull/2

fuzz/vips_fuzzer.cc
```diff
@@ -74,8 +79,40 @@ typedef struct _FuzzCtx {
 	int n_string_args;
 	int string_idx; // Next string argument to consume
 	gboolean failed;
+	const guint8 *pending_data; // Bytes to lazily write to input_filename
+	size_t pending_size;
+	char *input_filename;  // /tmp path holding the raw fuzz bytes, or NULL
+	gboolean input_written; // TRUE iff input_filename was written ok
```
kleisauke:
```suggestion
	gboolean input_written; // TRUE if input_filename was written ok
```

kleisauke (approved):
Great!

Issue #3339, opened 2023-02-15
PDFs with offset header are not being processed


JannKleen:
**Describe the bug**
The foreign loader for pdfium and Doppler seems to expect the header of the file to start with the first character, unfortunately that doesn't seem to be a requirement (s. https://github.com/pdf-association/pdf20examples/blob/master/PDF%202.0%20with%20offset%20start.pdf).
Even though I think PDFs formatted like that are mostly created by accident, it stops us from processing PDFs that at least pdfium (I don't know about poppler) itself should be able to handle.

I believe these are the relevant sections of code: https://github.com/libvips/libvips/blob/870f9e5ee5f30f4cdb39c98e400150591ccab078/libvips/foreign/pdfiumload.c#L792-L817
https://github.com/libvips/libvips/blob/870f9e5ee5f30f4cdb39c98e400150591ccab078/libvips/foreign/popplerload.c#L678-L703

**To Reproduce**
1. Download `https://github.com/pdf-association/pdf20examples/blob/master/PDF%202.0%20with%20offset%20start.pdf`
2. Run `vipsthumbnail PDF\ 2.0\ with\ offset\ start.pdf` 
3. See error

**Expected behavior**
PDFs that pdfium is able to parse should be passed to pdfium for processing.

**Actual behavior**
PDFs with an offset magic header are not being processed.

**Environment**
- Debian with vips-8.14.1 / pdfium
- MacOS with vips-8.14.1 / poppler

jcupitt:
Indeed, the file works:

```
$ vips pdfload ~/pics/PDF\ 2.0\ with\ offset\ start.pdf x.jpg
```

It's the autodetect that fails. Having to search a (perhaps?) very large file for the magic `%PDF-2.0` string sounds terrible, and perhaps unreliable.

Maybe it should try to use the filename suffix if sniffing fails?

lovell:
> the sample looks from the outside like an HTML document with a nested SVG

LOL, it looks like I was served the GitHub 500 error page rather than the sample file. Trying again I see the Linux `file` command interprets this sample as "ISO-8859 text, with CRLF, CR line terminators". I guess we could use a similar approach to SVG, where the first 1000 chars are tested (whilst never forgetting the WebP file with `<svg` in the compressed pixel data).

Do we know if there's any limitation on what can appear before the magic string, e.g. UTF-8 only? That might help filter out obviously-invalid files.

jcupitt:
The file says:

```
[...]

PR #5013, opened 2026-04-26
pdfload_buffer: only search first 32 bytes

### LINKED ISSUE #4885: vips_image_new_from_buffer() incorrectly tries to read HEIF buffer as PDF
## Bug report

When `vips_image_new_from_buffer()` is used on an HEIF input buffer, it attempts to load the buffer as PDF. The error raised is:
```
vips_image_new_from_buffer() failed: pdfload: file not in PDF format or corrupted
pdfload: source: unable to load
```

I think this happens because when libvips tries to determine the correct loader class for the buffer, the `vips__pdf_is_a_buffer()` function is called before its HEIF equivalent. That alone is not a problem. The problem is that as the second argument to `vips__pdf_is_a_buffer()` is the entire length of the buffer is passed, when it should be 32 as per https://github.com/libvips/libvips/issues/3339#issuecomment-1814212565

**To Reproduce**

I have created (AI-generated) a dockerised project with a single .cpp file and the input image - https://github.com/violetviolinist/libvips-bug-report

**Expected behavior**
`vips_image_new_from_buffer()` should correctly read the input buffer as a HEIF buffer

**Actual behavior**
`vips_image_new_from_buffer()` tries to read the input buffer as a PDF buffer

**Environment**
 - OS: debian 12 bookworm
 - Vips: latest (built from source from the main branch with PDFium and libheif support enabled)

**Edit**: Added context about libvips build

jcupitt:
The `is_a` detector for pdf buffer load was being passed and then searching the whole input buffer for the `%PDF` marker. This could lead to binary files being incorrectly recognised as PDF.

Closes: https://github.com/libvips/libvips/issues/4885

Thanks @violetviolinist

### DIFF
--- a/libvips/foreign/pdf.c
+++ b/libvips/foreign/pdf.c
@@ -51,21 +51,6 @@ const char *vips__pdf_suffs[] = {
 	NULL
 };
 
-gboolean
-vips__pdf_is_a_buffer(const void *buf, size_t len)
-{
-	const char *str = (const char *) buf;
-
-	if (len < 4)
-		return FALSE;
-
-	for (size_t i = 0; i < len - 4; i++)
-		if (vips_isprefix("%PDF", str + i))
-			return TRUE;
-
-	return FALSE;
-}
-
 /* PDF v2 allows for offset headers, ie. there may be any number of
  * characters of padding before the "%PDF" file marker. These can be
  * arbitrary printer control characters, whitespace, etc.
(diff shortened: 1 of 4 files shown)


jcupitt:
This only shows up as a problem with pdfium load, curiously. pdfium and poppler load are the same priority, but pdfium will sort earlier in the priority order.

I read the other `is_a_buffer` testers, and they all limit the search window correctly.

libvips/foreign/pdf.c
```diff
@@ -59,7 +59,8 @@ vips__pdf_is_a_buffer(const void *buf, size_t len)
 	if (len < 4)
 		return FALSE;
 
-	for (size_t i = 0; i < len - 4; i++)
+	// expect the %PDF marker in the first 32 bytes
+	for (size_t i = 0; i < VIPS_MIN(32, len - 4); i++)
```
lovell: Perhaps use the existing `MAX_OFFSET` value here instead of a "magic" number?

lovell (commented):
See also https://github.com/libvips/libvips/issues/3339

libvips/foreign/pdf.c
jcupitt: Oop! Duh, fixed, thanks!

kleisauke (approved):

PR #5014, opened 2026-04-26
stdif: expand input by one column

lovell:
~~Follow-up~~ Alternative to https://github.com/libvips/libvips/pull/5002

Fixes https://issues.oss-fuzz.com/issues/506473566 and https://issues.oss-fuzz.com/issues/503995698

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -5,7 +5,7 @@ tbd 8.18.3
 - edge: ensure parent object is built [lovell]
 - object: guard against invalid interpolators [kleisauke]
 - foreign: guard against NULL filenames [kleisauke]
-- stdif: verify window dimensions are smaller than image [lovell]
+- stdif: expand input by one column [lovell]
 - dcrawload: add support for non-square pixels [jcupitt]
 
 31/3/26 8.18.2
--- a/libvips/histogram/stdif.c
+++ b/libvips/histogram/stdif.c
@@ -231,8 +231,8 @@ vips_stdif_build(VipsObject *object)
 	if (vips_check_format(class->nickname, in, VIPS_FORMAT_UCHAR))
 		return -1;
 
-	if (stdif->width >= in->Xsize ||
-		stdif->height >= in->Ysize) {
+	if (stdif->width > in->Xsize ||
+		stdif->height > in->Ysize) {
 		vips_error(class->nickname, "%s", _("window too large"));
 		return -1;
 	}
@@ -245,7 +245,7 @@ vips_stdif_build(VipsObject *object)
 	 */
 	if (vips_embed(in, &t[1],
 			stdif->width / 2, stdif->height / 2,
-			in->Xsize + stdif->width - 1, in->Ysize + stdif->height - 1,
+			in->Xsize + stdif->width, in->Ysize + stdif->height - 1,
 			"extend", VIPS_EXTEND_COPY,
 			NULL))
 		return -1;
@@ -259,7 +259,7 @@ vips_stdif_build(VipsObject *object)
 	if (vips_image_pipelinev(stdif->out,
 			VIPS_DEMAND_STYLE_FATSTRIP, in, NULL))
 		return -1;
-	stdif->out->Xsize -= stdif->width - 1;
+	stdif->out->Xsize -= stdif->width;
 	stdif->out->Ysize -= stdif->height - 1;
 
 	if (vips_image_generate(stdif->out,


kleisauke:
Looking at commit e60f11eb4549d4cccda3b3fcab08130a0bbc1d85 (and its linked issue), which was effectively reverted by 2bc60c5e764b3de3af8a9146f17c8e67e6e0cf3e, I think(?) `rank` may require a similar fix.

kleisauke:
I just opened PR #5015 and #5016 as follow-up.

kleisauke:
Checking this locally with the OSS-Fuzz reproducers, it looks like we should also apply the following change:
```diff
--- a/ChangeLog
+++ b/ChangeLog
@@ -5,7 +5,7 @@ tbd 8.18.3
 - edge: ensure parent object is built [lovell]
 - object: guard against invalid interpolators [kleisauke]
 - foreign: guard against NULL filenames [kleisauke]
-- stdif: verify window dimensions are smaller than image [lovell]
+- stdif: exclude last col when adapting sums [lovell]
 - dcrawload: add support for non-square pixels [jcupitt]
 
 31/3/26 8.18.2
--- a/libvips/histogram/stdif.c
+++ b/libvips/histogram/stdif.c
@@ -233,8 +233,8 @@ vips_stdif_build(VipsObject *object)
 	if (vips_check_format(class->nickname, in, VIPS_FORMAT_UCHAR))
 		return -1;
 
-	if (stdif->width >= in->Xsize ||
-		stdif->height >= in->Ysize) {
+	if (stdif->width > in->Xsize ||
+		stdif->height > in->Ysize) {
 		vips_error(class->nickname, "%s", _("window too large"));
 		return -1;
 	}
```

In other words, this PR appears to be the correct fix (+ if you `grep` for `window too large`, you'll see that other functions don't use `>=`).

lovell:
@kleisauke Thanks, I've tested locally too and can confirm this PR is the better approach so have (partially) reverted the window size check and updated the changelog.

jcupitt:
Rather than adding the `if()`:

```C
if (x < r->width - 1) {
...
```

How about expanding the input by another column? Perhaps:

```C
    if (vips_embed(in, &t[1],
            stdif->width / 2, stdif->height / 2,
            in->Xsize + stdif->width, in->Ysize + stdif->height - 1,
            "extend", VIPS_EXTEND_COPY,
```

`conv` does this to avoid having to clip every mask against the available pixels.

jcupitt:
Also, another stupid off-by-one error groan.

kleisauke:
I confirm that John's suggestion works as well.

FWIW, I could also reproduce this in `hist_local` when doing:
```diff
--- a/libvips/histogram/hist_local.c
+++ b/libvips/histogram/hist_local.c
@@ -310,7 +310,7 @@ vips_hist_local_build(VipsObject *object)
 	if (vips_embed(in, &t[1],
 			local->width / 2, local->height / 2,
 			in->Xsize + local->width - 1, in->Ysize + local->height - 1,
-			"extend", VIPS_EXTEND_MIRROR,
+			"extend", VIPS_EXTEND_COPY,
 			NULL))
 		return -1;
 	in = t[1];
@@ -355,7 +355,7 @@ vips_hist_local_class_init(VipsHistLocalClass *class)
 	object_class->description = _("local histogram equalisation");
 	object_class->build = vips_hist_local_build;
 
-	operation_class->flags = VIPS_OPERATION_SEQUENTIAL;
+	//operation_class->flags = VIPS_OPERATION_SEQUENTIAL;
 
 	VIPS_ARG_IMAGE(class, "in", 1,
 		_("Input"),
```

kleisauke:
... the same applies to `rank`, which can only be reproduced there when doing:
```diff
--- a/libvips/morphology/morphology.c
+++ b/libvips/morphology/morphology.c
@@ -65,7 +65,7 @@ vips_morphology_class_init(VipsMorphologyClass *class)
 	vobject_class->nickname = "morphology";
 	vobject_class->description = _("morphological operations");
 
-	operation_class->flags = VIPS_OPERATION_SEQUENTIAL;
+	//operation_class->flags = VIPS_OPERATION_SEQUENTIAL;
 
 	/* Inputs set by subclassess.
 	 */
```

Though, `VIPS_OPERATION_SEQUENTIAL`  is only a hint, removing it doesn't guarantee random access.

lovell:
Thanks both, I've updated to expand the input by one column instead.

kleisauke (approved):

jcupitt (approved):

PR #5016, opened 2026-04-26
stdif: support many band images

kleisauke:
While we're here, re-sync this operation and `hist_local` with `rank`.

### DIFF
--- a/libvips/histogram/hist_local.c
+++ b/libvips/histogram/hist_local.c
@@ -102,13 +102,11 @@ vips_hist_local_stop(void *vseq, void *a, void *b)
 	VipsImage *in = (VipsImage *) a;
 
 	VIPS_UNREF(seq->ir);
-	if (seq->hist &&
-		in) {
-		int i;
 
-		for (i = 0; i < in->Bands; i++)
+	if (seq->hist &&
+		in)
+		for (int i = 0; i < in->Bands; i++)
 			VIPS_FREE(seq->hist[i]);
-	}
 	VIPS_FREE(seq->hist);
 	VIPS_FREE(seq);
 
@@ -121,8 +119,6 @@ vips_hist_local_start(VipsImage *out, void *a, void *b)
 	VipsImage *in = (VipsImage *) a;
 	VipsHistLocalSequence *seq;
 
-	int i;
-
 	if (!(seq = VIPS_NEW(NULL, VipsHistLocalSequence)))
 		return NULL;
 	seq->ir = NULL;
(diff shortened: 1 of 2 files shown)


libvips/histogram/hist_local.c
jcupitt: This is a silly thing, but how about making all the allocations local to `out`? It feels a little strange to have the main allocation local, but these subarrays freed in `_stop()`. All or nothing!

```suggestion
       !(seq->hist = VIPS_ARRAY(out, in->Bands, unsigned int *))) {
```

libvips/histogram/stdif.c
```diff
@@ -86,72 +86,112 @@ typedef VipsOperationClass VipsStdifClass;
...
+
+	if (!(seq = VIPS_NEW(out, VipsStdifSequence)))
+		return NULL;
+	seq->ir = NULL;
+	seq->sum = NULL;
+	seq->sum2 = NULL;
+
+	if (!(seq->ir = vips_region_new(in)) ||
+		!(seq->sum = VIPS_ARRAY(NULL, in->Bands, unsigned int)) ||
+		!(seq->sum2 = VIPS_ARRAY(NULL, in->Bands, unsigned int))) {
```
jcupitt: Possibly make these local too.

jcupitt (approved):

jcupitt:
That was some cheesy old code, nice job fixing it!

libvips/histogram/hist_local.c
kleisauke: Good idea. Done with commit 541e6741d9998eb939ce8b78e061b74d69272e99.

IIRC, the `g_free()` now occurs slightly later. I also vaguely recall an older commit that did the opposite (i.e. free earlier and avoid allocating anything local to the output image). So, while this approach is much simpler, I'm not entirely sure if it's better.

libvips/histogram/stdif.c
kleisauke: Done with commit bbccf137d6f955b6be60151574eea60432a06e3d.

libvips/histogram/hist_local.c
kleisauke: ... I think it was commit 3b2c77dd354720d34d4cba5fe680dfdf27720a0d (though I have no idea how I managed to remember that :man_shrugging:).

jcupitt: Yes, I've certainly gone back and forth on this. 

Local allocations are neater, but they push memory use up, and there's an overhead in running all the close callbacks. We have a fast and early `_stop()` mechanism, why not just use that.

kleisauke: Great! I've opened PR #5028 for this and applied the same change to the "re-sync with rank" commits in this PR.

There are probably other operations that could benefit from this as well.

jcupitt: Ooops, sorry, I was just talking rhetorically about the pros and cons, I wasn't trying to make you do any work. 

I don't really mind either way, except that it felt a little odd to mix the two styles.

kleisauke: No worries, it wasn't much work.

I quickly skimmed through the codebase and noticed that `reducev` could also benefit from this change, which is a hot code path. :)

PR #5024, opened 2026-04-29
fuzz: prefer use of `__*_default_suppressions` hooks

kleisauke:
Works with ASan, LSan and TSan. Support for UBSan depends on:
llvm/llvm-project#194862

Context: while commit 87cd335c0024bf158b08f5717567f46af388e29b works locally and when reproducing bugs using https://google.github.io/oss-fuzz/advanced-topics/reproducing/#reproducing-bugs, it does not work in ClusterFuzz for some reason.

### DIFF
--- a/fuzz/config.h
+++ b/fuzz/config.h
@@ -1,30 +1,52 @@
 #pragma once
 
-#include <unistd.h>
-
-/* Only supported since clang >= 14.0.
- */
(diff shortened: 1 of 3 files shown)


jcupitt (approved):

jcupitt:
Fingers crossed!

PR #5028, opened 2026-05-01
Free sequence structs earlier

kleisauke:
Free the sequence structs at the end of eval, not on image dispose. This helps keep memory use down in some long-running operations.

In line with commit 3b2c77dd354720d34d4cba5fe680dfdf27720a0d.

Context: https://github.com/libvips/libvips/pull/5016#discussion_r3174165792.


jcupitt (approved):

jcupitt:
It does look neater, I think!

PR #5020, opened 2026-04-27
Fix OOB read in jp2ksave

dloebl:
Noticed while fuzzing with #5003 locally, there is a small OOB read in `jp2ksave`:
```
./build-asan/tools/vips black test.png 22 3 --bands 3
./build-asan/tools/vips jp2ksave test.png out.jp2 --subsample-mode on
```

```
==44563==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x61700000069a at pc 0x000103c8e0a8 bp 0x00016d459990 sp 0x00016d459988
READ of size 1 at 0x61700000069a thread T6
    #0 0x000103c8e0a4 in vips_foreign_save_jp2k_unpack_subsample jp2ksave.c:379
    #1 0x000103c8c9a4 in vips_foreign_save_jp2k_write_tiles jp2ksave.c:517
    #2 0x000103c8bbc0 in vips_foreign_save_jp2k_write_block jp2ksave.c:577
    #3 0x000104109cc8 in wbuffer_write sinkdisc.c:186
    #4 0x000104109af0 in wbuffer_write_thread sinkdisc.c:211
    #5 0x0001040a8258 in vips_threadset_work threadset.c:187
    #6 0x0001040a5cb0 in vips_thread_run thread.c:112
    #7 0x000102ea18dc in g_thread_proxy+0x50 (libglib-2.0.0.dylib:arm64+0x558dc)
    #8 0x000104941d64 in asan_thread_start(void*)+0x4c (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x3dd64)
    #9 0x000198b0fc04 in _pthread_start+0x84 (libsystem_pthread.dylib:arm64e+0x6c04)
    #10 0x000198b0aba4 in thread_start+0x4 (libsystem_pthread.dylib:arm64e+0x1ba4)
```

Subsampling should clamp each block to the remaining pixels, otherwise the final row/column can try to read a full `comp->dy`/`comp->dx` block past the image bounds.
libvips isn't built with JPEG2000 support in OSS Fuzz, so this wasn't found yet.

PR #5030, opened 2026-05-02
jp2ksave: prevent heap buffer overflow on non-RGB images

dloebl:
Similar to #5020, there is an OOB read/write (heap buffer overflow) in the `save_as_ycc` path if the input image isn't RGB. The `Bands == 3` check that is already present in `vips__foreign_save_jp2k_compress` has to be applied to the regular `jp2ksave` build as well: https://github.com/libvips/libvips/blob/9bde113b9dd0eeae39a9c9d5b92a957490586dc7/libvips/foreign/jp2ksave.c#L1338

**Reproducer:**
```
./build-asan/tools/vips black --bands 1 out.jp2"[subsample-mode=on]" 32 32
```

```
==74562==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x61a000000ac7 at pc 0x00010523b4d8 bp 0x00016bd76980 sp 0x00016bd76978
READ of size 1 at 0x61a000000ac7 thread T11
    #0 0x00010523b4d4 in vips_foreign_save_jp2k_rgb_to_ycc+0xd48 (libvips.42.dylib:arm64+0x474d4)
    #1 0x00010523d6e0 in vips_foreign_save_jp2k_write_block+0x36c (libvips.42.dylib:arm64+0x496e0)
    #2 0x000105695e6c in wbuffer_write_thread+0x134 (libvips.42.dylib:arm64+0x4a1e6c)
    #3 0x00010565c294 in vips_threadset_work+0x168 (libvips.42.dylib:arm64+0x468294)
    #4 0x00010565b230 in vips_thread_run+0x50 (libvips.42.dylib:arm64+0x467230)
    #5 0x0001048b58dc in g_thread_proxy+0x50 (libglib-2.0.0.dylib:arm64+0x558dc)
    #6 0x000105e02418 in asan_thread_start(void*)+0x4c (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x3a418)
    #7 0x00018d939c04 in _pthread_start+0x84 (libsystem_pthread.dylib:arm64e+0x6c04)
    #8 0x00018d934ba4 in thread_start+0x4 (libsystem_pthread.dylib:arm64e+0x1ba4)
```

### DIFF
--- a/libvips/foreign/jp2ksave.c
+++ b/libvips/foreign/jp2ksave.c
@@ -852,6 +852,11 @@ vips_foreign_save_jp2k_build(VipsObject *object)
 	if (jp2k->subsample)
 		jp2k->save_as_ycc = TRUE;
 
+	/* Our rgb->ycc only works for exactly 3 bands.
+	 */
+	jp2k->save_as_ycc = jp2k->save_as_ycc && save->ready->Bands == 3;
+	jp2k->subsample = jp2k->subsample && jp2k->save_as_ycc;
+
 	/* Set parameters for compressor.
 	 */
 	opj_set_default_encoder_parameters(&jp2k->parameters);
--- a/libvips/foreign/jp2ksave.c
+++ b/libvips/foreign/jp2ksave.c
@@ -837,7 +837,7 @@ vips_foreign_save_jp2k_build(VipsObject *object)
 		break;
 
 	case VIPS_FOREIGN_SUBSAMPLE_ON:
-		jp2k->subsample = TRUE;
+		jp2k->subsample = save->ready->Bands == 3;
 		break;
 
 	case VIPS_FOREIGN_SUBSAMPLE_OFF:
@@ -851,11 +851,6 @@ vips_foreign_save_jp2k_build(VipsObject *object)
 
 	if (jp2k->subsample)
 		jp2k->save_as_ycc = TRUE;
-
-	/* Our rgb->ycc only works for exactly 3 bands.
-	 */
-	jp2k->save_as_ycc = jp2k->save_as_ycc && save->ready->Bands == 3;
-	jp2k->subsample = jp2k->subsample && jp2k->save_as_ycc;
 
 	/* Set parameters for compressor.
 	 */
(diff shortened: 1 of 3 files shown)


jcupitt (dismissed):

libvips/foreign/jp2ksave.c
```diff
@@ -852,6 +852,11 @@ vips_foreign_save_jp2k_build(VipsObject *object)
 	if (jp2k->subsample)
 		jp2k->save_as_ycc = TRUE;
 
+	/* Our rgb->ycc only works for exactly 3 bands.
+	 */
+	jp2k->save_as_ycc = jp2k->save_as_ycc && save->ready->Bands == 3;
+	jp2k->subsample = jp2k->subsample && jp2k->save_as_ycc;
```
jcupitt: Sorry, thinking about this again, the logic here is a little ugly. 

How about this instead?

```C
    case VIPS_FOREIGN_SUBSAMPLE_ON:
+       jp2k->subsample = save->ready->Bands == 3;
        break;

    case VIPS_FOREIGN_SUBSAMPLE_OFF:
        jp2k->subsample = FALSE;
        break;

    default:
        g_assert_not_reached();
        break;
    }

    if (jp2k->subsample)
        jp2k->save_as_ycc = TRUE;

    /* Set parameters for compressor.
     */
    opj_set_default_encoder_parameters(&jp2k->parameters);
```

ie. check that `Bands == 3` before turning subsample on, same as the auto path. Now the logic is just in one place.

dloebl: Thanks! Addressed with 38b8beba087c56dd2dd82264231bf114dc54f08a

dloebl:
btw should I make this PR target the `master` branch or `8.18` now that you do automatic backports?

jcupitt (approved):

jcupitt:
I think we're moving to the new git strategy for 8.19, it'd be too confusing to swap half-way through a cycle.

Thanks for this PR!

PR #5034, opened 2026-05-08
heifload: tighten restrictions on heif memory use

jcupitt:
Tighten up the limits on heifload memory use. These new limits are still loose enough to allow a 16k x 16k pixel image to load successfully.

With this test image (32k x 32k pixels, don't click to view, right click and download instead):

https://upload.wikimedia.org/wikipedia/commons/d/d7/Giovanni_Bellini_-_Saint_Francis_in_the_Desert_-_Google_Art_Project.jpg

I can run:

```
$ # the largest avif we allow
$ vips crop st-francis.jpg st-francis.avif 0 0 16384 16384
$ ls -l st-francis.avif 
-rw-r--r-- 1 john john 24918121 May  8 14:15 st-francis.avif
$ /usr/bin/time -f %M:%e vips copy st-francis.avif x.jpg
2831204:4.63
$
```

So it loads in 280mb of memory and without triggering a security limit.

Thanks @provos !

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -13,6 +13,7 @@ tbd 8.18.3
 - fix possible null-pointer dereference in quantizr backend [felixbuenemann]
 - jp2ksave: fix OOB read during chroma subsampling [dloebl]
 - jp2ksave: prevent heap buffer overflow on non-RGB images [dloebl]
+- heifload: tighten limits on memory use [nprovos]
 
 31/3/26 8.18.2
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -13,7 +13,7 @@ tbd 8.18.3
 - fix possible null-pointer dereference in quantizr backend [felixbuenemann]
 - jp2ksave: fix OOB read during chroma subsampling [dloebl]
 - jp2ksave: prevent heap buffer overflow on non-RGB images [dloebl]
-- heifload: tighten limits on memory use [nprovos]
+- heifload: tighten limits on memory use [provos]
 
 31/3/26 8.18.2
 
--- a/libvips/foreign/heifload.c
+++ b/libvips/foreign/heifload.c
@@ -356,9 +356,14 @@ vips_foreign_load_heif_build(VipsObject *object)
 		heif_context_set_maximum_image_size_limit(heif->ctx,
 			heif->unlimited ? USHRT_MAX : 0x4000);
 #ifdef HAVE_HEIF_MAX_TOTAL_MEMORY
-		if (!heif->unlimited)
-			heif_context_get_security_limits(heif->ctx)
-				->max_total_memory = 2UL * 1024 * 1024 * 1024;
+		if (!heif->unlimited) {
+			heif_security_limits *limits =
+				heif_context_get_security_limits(heif->ctx);
+
+			limits->max_total_memory = 2L * 1024 * 1024 * 1024;
+			limits->max_memory_block_size = 1024 * 1024 * 1024;
+			limits->max_items = 16;
+		}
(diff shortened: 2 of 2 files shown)


ChangeLog
```diff
@@ -13,6 +13,7 @@ tbd 8.18.3
 - fix possible null-pointer dereference in quantizr backend [felixbuenemann]
 - jp2ksave: fix OOB read during chroma subsampling [dloebl]
 - jp2ksave: prevent heap buffer overflow on non-RGB images [dloebl]
+- heifload: tighten limits on memory use [nprovos]
```
```suggestion
- heifload: tighten limits on memory use [provos]
```

kleisauke (approved):

ChangeLog
jcupitt: Ooops! Thank you!

caseywebdev:
I have a basic 2.1 MB iPhone HEIC failing to convert because it contains 48 references.

```
Error: Input buffer has corrupt header: heif: Invalid input: Security limit exceeded: Number of references in iref box (48) exceeds the security limits of 16 references. (2.1000)
```

This is user-submitted content, so I don't trust it and want to keep reasonable safe limits, but this hardcoded 16 item limit is problematic if I can't accept user iPhone pictures.

[snow.heic.zip](https://github.com/user-attachments/files/28842115/snow.heic.zip)

Please consider increasing this limit or allowing it to be configured without removing all memory limits.

jcupitt:
Thanks for the test image @caseywebdev. I agree, 64 is probably a better default, I'll make a PR.

caseywebdev:
Awesome, thanks!

jcupitt:
I credited you on the changelog, hope that's ok.

caseywebdev:
Of course, thanks for the quick fix!

d--j:
With the increased limits of #5075 I get
```
heif: Invalid input: Security limit exceeded: ipma box wants to define properties for 115 items, but the security limit has been set to 64 items (2.1000)
```
when trying to load this [random iPhone 17 HEIC](https://github.com/user-attachments/files/28999038/IMG_1397.HEIC.zip).

I have patched my vips to use 256 as `max_items` limit. With that the HEIC can be loaded. That's still approx 1/4 of the [global/default limit](https://github.com/strukturag/libheif/blob/3021ff4efd897f9d66fd6287dad33e138de27dd4/libheif/security_limits.cc#L37) as far as I understand. Unfortunately I do not know anything about `libheif` and what these limits all mean… So I do not know if 256 `ipma` boxes in a HEIF is a big value.
`iinf` (Item Information), `iloc` (Item Location), `ipma` (Item Property Association), and `iref` (Item Reference) boxes get these limits (see [box.cc](https://github.com/strukturag/libheif/blob/master/libheif/box.cc))

jcupitt:
How annoying! Thanks for reporting this.

I'll raise libvips to 256 as well.

jcupitt:
I pushed the change directly to 8.18 since it seems very uncontroversial.

jonashaag:
Hello, we just hit this too with 34 items. I can't find the change to 256 anywhere, has it been forgotten? Shall I submit a PR?

jcupitt:
It's in the 8.18 branch and will be released as 8.18.4. I'm on holiday this week, so maybe on Monday.

https://github.com/libvips/libvips/blob/8.18/libvips/foreign/heifload.c#L365

https://github.com/libvips/libvips/blob/8.18/ChangeLog#L3

jcupitt:
Released as 8.18.4: https://github.com/libvips/libvips/releases/tag/v8.18.4

PR #5040, opened 2026-05-12
magickload: check animation height

jcupitt:
check final image height for imagemagick animation files

thanks Himanshu Anand


jcupitt:
Sorry, this PR is larger than it needs to be. As well as adding the check, I tried to make parsing the imagemagick image data clearer and simpler.

lovell (approved):

PR #5047, opened 2026-05-15
revise analyze2vips

jcupitt:
- check for dimension overflow earlier (this was caught before, but it's better to catch sooner)
- improve formatting for new layout and C99
- tiny typo in insert

Thanks 007bsd!

### DIFF
--- a/libvips/foreign/analyze2vips.c
+++ b/libvips/foreign/analyze2vips.c
@@ -78,10 +78,10 @@ typedef enum {
 /* A field in the dsr header.
  */
 typedef struct {
-	const char *name; /* Eg. "header_key.sizeof_hdr" */
+	const char *name;		/* Eg. "header_key.sizeof_hdr" */
 	Type type;
-	glong offset; /* Offset in struct */
-	int len;	  /* Sizeof ... useful for string types */
+	glong offset;			/* Offset in struct */
+	int len;				/* Sizeof ... useful for string types */
 } Field;
 
 static Field dsr_header[] = {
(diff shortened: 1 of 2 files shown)


libvips/foreign/analyze2vips.c
```diff
@@ -237,19 +237,18 @@ generate_filenames(const char *path, char *header, char *image)
 static char *
 getstr(int mx, const char *str)
 {
-	static char buf[256];
-	int i;
-
 	g_assert(mx < 256);
 
+	static char buf[256];
 	g_strlcpy(buf, str, mx);
```
lovell, line 245: I checked the [g_strlcpy](https://docs.gtk.org/glib/func.strlcpy.html) docs and it says it will copy `dest_size - 1` bytes. The comment says `str` might not be null-terminated so should we use `mx + 1` for `dest_size` to avoid potentially removing the last char?

It also looks like `mx` is defined by the (untrusted) input, so perhaps the `g_assert` should be a proper runtime check?

jcupitt: > It also looks like mx is defined by the (untrusted) input, 

I think `mx` comes from `dsr_header`, eg.:

```C
                getstr(dsr_header[i].len,
                    &G_STRUCT_MEMBER(char, d, dsr_header[i].offset)));
```

so it's compile-time. It's max 80 chars, so there's a lot of headroom, which I think makes the copy safe.

jcupitt: I improved the comment, it was NOT very clear heh.

lovell (approved):

PR #5048, opened 2026-05-15
CIFuzz: fix build after google/oss-fuzz#15324

kleisauke:
See: https://github.com/libvips/libvips/actions/runs/25921929218/job/76192907533?pr=5042


lovell (approved):

kleisauke:
Thanks!

@jcupitt Note that when you merge [`8.18`](https://github.com/libvips/libvips/tree/8.18) back into the master branch, it will likely conflict in `‎fuzz/oss_fuzz_build.sh` due to commit 179d03d082ec1087e5ac86a85794f38e072557ee. In that case, these changes should be preferred (i.e. the check for the `$SRC/seed-corpora` directory and its fallback should be removed).

PR #5053, opened 2026-05-26
buildlut: check upper bound of x range

lovell:
Fixes https://issues.oss-fuzz.com/issues/516503785

~~EDIT: GitHub Actions [is broken](https://www.githubstatus.com/incidents/gnftqj9htp0g) again so CI hasn't, possibly won't, run for this PR.~~

### DIFF
--- a/libvips/create/buildlut.c
+++ b/libvips/create/buildlut.c
@@ -119,11 +119,11 @@ vips_buildlut_build_init(VipsBuildlut *lut)
 	VipsObjectClass *class = VIPS_OBJECT_GET_CLASS(lut);
 
 	int y;
-	int xlow, xhigh;
+	double xlow, xhigh;
 
 	/* Need xlow and xhigh to get the size of the LUT we build.
 	 */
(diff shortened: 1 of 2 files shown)


libvips/create/buildlut.c
```diff
@@ -142,8 +142,15 @@ vips_buildlut_build_init(VipsBuildlut *lut)
 		if (v > xhigh)
 			xhigh = v;
 	}
-	lut->xlow = xlow;
-	lut->lut_size = xhigh - xlow + 1;
+
+	if (xlow < INT_MIN || xhigh > INT_MAX ||
+		xhigh - xlow + 1 > INT_MAX) {
```
jcupitt: I suppose we could check against `VIPS_MAX_COORD` instead. We'll fail with an error later if we're over max coord, of course, but it'd stop a large memory allocation.

jcupitt (approved):

libvips/create/buildlut.c
lovell: Thanks, I've updated to use this for the `lut_size` check.

PR #5057, opened 2026-06-05
ppmload: check image dimensions

lovell:
Fixes https://issues.oss-fuzz.com/issues/519999735


jcupitt (approved):

jcupitt:
Reading through ppmload, it seems to have a few issues! I made a PR with some cleanups https://github.com/libvips/libvips/pull/5058

PR #5058, opened 2026-06-05
revise ppmload

jcupitt:
Revise ppmload following https://github.com/libvips/libvips/pull/5057

- move the dimension check into header read
- improve have_read_header use
- propagate any errors from set_image

### DIFF
--- a/libvips/foreign/ppmload.c
+++ b/libvips/foreign/ppmload.c
@@ -79,8 +79,8 @@
  */
 
 /*
-#define DEBUG
  */
+#define DEBUG
 
 #ifdef HAVE_CONFIG_H
 #include <config.h>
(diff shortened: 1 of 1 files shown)


libvips/foreign/ppmload.c
```diff
@@ -79,8 +79,8 @@
  */
 
 /*
-#define DEBUG
  */
+#define DEBUG
```
lovell, line 82: Does this need to go back inside the comment block?

lovell (commented):
Should we use `8.18` as the target branch?

jcupitt:
> Should we use `8.18` as the target branch?

It's just a cleanup (mostly), so I think master is OK. It'll give us more testing time too.

lovell (approved):

jcupitt:
Thanks fo9r the review!

PR #5059, opened 2026-06-05
block very deep pyramids

jcupitt:
Thanks Niels Provos.

### DIFF
--- a/libvips/resample/thumbnail.c
+++ b/libvips/resample/thumbnail.c
@@ -271,9 +271,11 @@ vips_thumbnail_get_pyramid_page(VipsThumbnail *thumbnail)
 	printf("vips_thumbnail_get_pyramid_page:\n");
 #endif /*DEBUG*/
 
-	/* Single-page docs can't be pyramids.
+	/* Single-page docs can't be pyramids, more than 30 levels will int
+	 * overflow.
 	 */
-	if (thumbnail->n_pages < 2)
+	if (thumbnail->n_pages < 2 ||
+		thumbnail->n_pages > 29)
 		return;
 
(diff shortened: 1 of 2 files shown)


libvips/resample/thumbnail.c
```diff
@@ -271,9 +271,11 @@ vips_thumbnail_get_pyramid_page(VipsThumbnail *thumbnail)
 	printf("vips_thumbnail_get_pyramid_page:\n");
 #endif /*DEBUG*/
 
-	/* Single-page docs can't be pyramids.
+	/* Single-page docs can't be pyramids, more than 30 levels will int
+	 * overflow.
 	 */
-	if (thumbnail->n_pages < 2)
+	if (thumbnail->n_pages < 2 ||
+		thumbnail->n_pages > 29)
```
kleisauke, line 278: Shifting a 32-bit integer by 31 or more is UB, so this could be adjusted to:
```suggestion
		thumbnail->n_pages > 30)
```
(this also matches the comment above the if statement)

jcupitt: I think I wanted to match the subifd case below, or it's puzzling that the two use different limits.

kleisauke: Ah, I see. I've also never come across a TIFF pyramid image with more than 29 levels, so it's probably fine.

kleisauke (approved):

libvips/resample/thumbnail.c
jcupitt: Yes, `VIPS_MAX_COORD` is 10m pixels, or 2^24-ish, so 29 levels is far more than we could ever use.

PR #5061, opened 2026-06-06
Fix merge conflict

kleisauke:
Fixes a build failure when compiling against libjxl < 0.9.


jcupitt (approved):

jcupitt:
Oooop! Thanks!

We should make 0.9 our minimum at some point.

PR #5068, opened 2026-06-08
fix: jpegsave without uhdrsave

jcupitt:
jpegsave should not defer to uhdrsave if uhdr support is missing

Right now, if you build without uhdr support, you can trigger this error:

	$ vips copy k2.jpg x.hdr
	$ vips copy x.hdr x.jpg
	VipsOperation: class "uhdrsave_target" not found

Because jpegsave will see the HDR image and call uhdrsave.

It should not do this if uhdr support is disabled.

### DIFF
--- a/libvips/foreign/jpegsave.c
+++ b/libvips/foreign/jpegsave.c
@@ -132,6 +132,12 @@ static VipsBandFormat bandfmt_jpeg[10] = {
 	/* Promotion: */ UC, UC, UC, UC, UC, UC, UC, UC, UC, UC
 };
 
+#ifdef HAVE_UHDR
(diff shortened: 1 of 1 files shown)


jcupitt:
I suppose this should maybe go into 8.18 as well.

kleisauke:
> I suppose this should maybe go into 8.18 as well.

Good idea. Feel free to target the 8.18 branch, this looks like a safe change.

jcupitt:
OK, done with  https://github.com/libvips/libvips/commit/75d82d12581c04fda78eeb7f51c18b967b095c6a

Just squeezed into 8.18.3, phew

kleisauke (approved):

PR #5070, opened 2026-06-08
Add apng load

jcupitt:
Split from https://github.com/libvips/libvips/pull/4944 for easier reviewing.

I've started working through it doing some reformatting and cleanups, currently at `apng_blend_pixel8`


jcupitt:
Parking for now.

PR #5072, opened 2026-06-10
perf: optimize conversion module hot paths

jcupitt:
A copy of https://github.com/libvips/libvips/pull/4969 with review comments implemented.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -22,12 +22,9 @@ date-tbd 8.19.0
 - uhdrsave: add existing JPEG options for base image [lovell]
 - csvload: add maximum limit on column count [lovell]
 - pdfload_buffer: only search the first few bytes in is_a [violetviolinist]
-<<<<<<< perf-conversion
 - add VIPS_MEMCPY() [kleisauke]
-=======
 - heifload: allow mif2 and mif3 "brand" images e.g. mini box [lovell]
 - add vips_scRGB2CICP [mertalev]
->>>>>>> master
 
 6/6/26 8.18.3
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -22,6 +22,7 @@ date-tbd 8.19.0
 - uhdrsave: add existing JPEG options for base image [lovell]
 - csvload: add maximum limit on column count [lovell]
 - pdfload_buffer: only search the first few bytes in is_a [violetviolinist]
+- add VIPS_MEMCPY() [kleisauke]
 
 tbd 8.18.3
 
--- a/libvips/include/vips/util.h
+++ b/libvips/include/vips/util.h
@@ -64,6 +64,36 @@ extern "C" {
 #define VIPS_ABS(V) (((V) >= 0) ? (V) : -(V))
 #define VIPS_FABS(V) fabs((V)) VIPS_DEPRECATED_MACRO_FOR(fabs)
 
+/* Copy a single pixel, optimised for common pixel sizes.
+ */
+#define VIPS_MEMCPY(Q, P, N) \
+	G_STMT_START \
+	{ \
+		switch (N) { \
+		case 1: \
+			(Q)[0] = (P)[0]; \
+			break; \
+		case 2: \
+			*((guint16 *) (Q)) = *((guint16 *) (P)); \
+			break; \
+		case 3: \
+			(Q)[0] = (P)[0]; \
+			(Q)[1] = (P)[1]; \
+			(Q)[2] = (P)[2]; \
+			break; \
+		case 4: \
+			*((guint32 *) (Q)) = *((guint32 *) (P)); \
+			break; \
+		case 8: \
+			*((guint64 *) (Q)) = *((guint64 *) (P)); \
+			break; \
+		default: \
+			memcpy((Q), (P), (N)); \
+			break; \
+		} \
+	} \
+	G_STMT_END
+
 // is something (eg. a pointer) N aligned
 #define VIPS_ALIGNED(P, N) ((((guint64) (P)) & ((N) - 1)) == 0)
 
--- a/libvips/include/vips/util.h
+++ b/libvips/include/vips/util.h
@@ -64,35 +64,48 @@ extern "C" {
 #define VIPS_ABS(V) (((V) >= 0) ? (V) : -(V))
 #define VIPS_FABS(V) fabs((V)) VIPS_DEPRECATED_MACRO_FOR(fabs)
 
-/* Copy a single pixel, optimised for common pixel sizes.
+/* memcpy(), but with fast paths for small copies.
  */
 #define VIPS_MEMCPY(Q, P, N) \
-	G_STMT_START \
-	{ \
-		switch (N) { \
-		case 1: \
-			(Q)[0] = (P)[0]; \
-			break; \
-		case 2: \
-			*((guint16 *) (Q)) = *((guint16 *) (P)); \
-			break; \
-		case 3: \
-			(Q)[0] = (P)[0]; \
-			(Q)[1] = (P)[1]; \
-			(Q)[2] = (P)[2]; \
-			break; \
-		case 4: \
-			*((guint32 *) (Q)) = *((guint32 *) (P)); \
-			break; \
-		case 8: \
-			*((guint64 *) (Q)) = *((guint64 *) (P)); \
-			break; \
-		default: \
-			memcpy((Q), (P), (N)); \
-			break; \
-		} \
-	} \
-	G_STMT_END
+    G_STMT_START \
+    { \
+        switch (N) { \
+        case 8: \
+            *((guint64 *) (Q)) = *((guint64 *) (P)); \
+            break; \
+\
+        case 7: \
+            ((guint8 *) (Q))[6] = ((guint8 *) (P))[6]; \
+\
+        case 6: \
+            ((guint8 *) (Q))[5] = ((guint8 *) (P))[5]; \
+\
+        case 5: \
+            ((guint8 *) (Q))[4] = ((guint8 *) (P))[4]; \
+\
+        case 4: \
+            *((guint32 *) (Q)) = *((guint32 *) (P)); \
+            break; \
+\
+        case 3: \
+            ((guint8 *) (Q))[2] = ((guint8 *) (P))[2]; \
+\
+        case 2: \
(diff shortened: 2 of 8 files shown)


jcupitt:
So this PR is #4969 but with:

- revise MEMCPY macro to include all N of 8 and under
- remove composite code
- revise loops in other functions

Tested against musl and glibc. For flatten I see about a 3x speedup (thanks to the new int8 path); rot, flip, zoom, subsample are a little over 2x faster (thanks to the memcpy macro).

libvips/include/vips/util.h
```diff
@@ -64,6 +64,49 @@ extern "C" {
...
+\
+        case 4: \
+            *((guint32 *) (Q)) = *((guint32 *) (P)); \
+            break; \
+\
+        case 3: \
+            ((guint8 *) (Q))[2] = ((guint8 *) (P))[2]; \
+\
+        case 2: \
+            *((guint16 *) (Q)) = *((guint16 *) (P)); \
```
kleisauke: UBSan seems to complain on this line:
```
../libvips/conversion/flip.c:172:4: runtime error: load of misaligned address 0x52200029d231 for type 'guint16' (aka 'unsigned short'), which requires 2 byte alignment
0x52200029d231: note: pointer points here
 04 02 03  04 02 03 04 02 03 04 02  03 04 02 03 04 02 03 04  02 03 04 02 03 04 02 03  04 02 03 04 02
```
https://github.com/libvips/libvips/actions/runs/27304449375/job/80658412065?pr=5072#step:16:498

jcupitt: Yes, it'll do unaligned copies. They work fine on x64 (though slower than aligned copies) but are officially UB. Does arm allow unaligned pointers? I've not tested it.

I suppose we either add extra code to check alignment before doing a pointer copy, or add something to keep ubsan quiet.

jcupitt: Oh, I guess wasm is the other platform we care about. 

Or we could swap case 3 for:

```C
        case 4: \
            *((guint32 *) (Q)) = *((guint32 *) (P)); \
            break; \
\
        case 3: \
            ((guint8 *) (Q))[0] = ((guint8 *) (P))[0]; \
            ((guint8 *) (Q))[1] = ((guint8 *) (P))[1]; \
            ((guint8 *) (Q))[2] = ((guint8 *) (P))[2]; \
            break; \
\
        case 2: \
            *((guint16 *) (Q)) = *((guint16 *) (P)); \
```

And assume that if we are copying N bytes, the pointers will be N byte-aligned.

`VIPS_MEMCPY()` would not be a good name then, since it wouldn't be a general memcpy. 

Perhaps `VIPS_PELCPY()`, since it'll only work for large aligned arrays of equal sized elements?

jcupitt: ... actually that's not enough, we'd need:

```C
/* Copy N bytes within a large array which is N byte aligned, like an image.
 */
#define VIPS_PELCPY(Q, P, N) \
    G_STMT_START \
    { \
        switch (N) { \
        case 8: \
            ((guint64 *) (Q))[0] = ((guint64 *) (P))[0]; \
            break; \
\
        case 7: \
            ((guint8 *) (Q))[0] = ((guint8 *) (P))[0]; \
            ((guint8 *) (Q))[1] = ((guint8 *) (P))[1]; \
            ((guint8 *) (Q))[2] = ((guint8 *) (P))[2]; \
            ((guint8 *) (Q))[3] = ((guint8 *) (P))[3]; \
            ((guint8 *) (Q))[4] = ((guint8 *) (P))[4]; \
            ((guint8 *) (Q))[5] = ((guint8 *) (P))[5]; \
            ((guint8 *) (Q))[6] = ((guint8 *) (P))[6]; \
            break; \
\
        case 6: \
            ((guint16 *) (Q))[0] = *((guint16 *) (P))[0]; \
            ((guint16 *) (Q))[1] = *((guint16 *) (P))[1]; \
            ((guint16 *) (Q))[2] = *((guint16 *) (P))[2]; \
            break; \
\
        case 5: \
            ((guint8 *) (Q))[0] = ((guint8 *) (P))[0]; \
            ((guint8 *) (Q))[1] = ((guint8 *) (P))[1]; \
            ((guint8 *) (Q))[2] = ((guint8 *) (P))[2]; \
            ((guint8 *) (Q))[3] = ((guint8 *) (P))[3]; \
            ((guint8 *) (Q))[4] = ((guint8 *) (P))[4]; \
            break; \
\
        case 4: \
            ((guint32 *) (Q))[0] = ((guint32 *) (P))[0]; \
            break; \
\
        case 3: \
            ((guint8 *) (Q))[0] = ((guint8 *) (P))[0]; \
            ((guint8 *) (Q))[1] = ((guint8 *) (P))[1]; \
            ((guint8 *) (Q))[2] = ((guint8 *) (P))[2]; \
            break; \
\
        case 2: \
            ((guint16 *) (Q))[0] = ((guint16 *) (P))[0]; \
            break; \
\
        case 1: \
            ((guint8 *) (Q))[0] = ((guint8 *) (P))[0]; \
\
        case 0: \
            break; \
\
        default: \
            memcpy((Q), (P), (N)); \
            break; \
        } \
    } \
    G_STMT_END
```

Which I think will always use aligned pointers.

kleisauke: The compiler is likely able to optimize fixed-size `memcpy()` calls on its own. So, you could write something like:
```c
/* Inline fast paths for small, fixed-size memcpy() operations.
 *
 * Use only when N is not a compile-time constant. For constant sizes,
 * modern compilers can usually optimize plain memcpy() calls directly.
 *
 * N must not have side effects (e.g. N++).
 */
#define VIPS_MEMCPY(Q, P, N) \
	G_STMT_START \
	{ \
		switch (N) { \
		case 8: memcpy((Q), (P), 8); break; \
		case 7: memcpy((Q), (P), 7); break; \
		case 6: memcpy((Q), (P), 6); break; \
		case 5: memcpy((Q), (P), 5); break; \
		case 4: memcpy((Q), (P), 4); break; \
		case 3: memcpy((Q), (P), 3); break; \
		case 2: memcpy((Q), (P), 2); break; \
		case 1: memcpy((Q), (P), 1); break; \
		case 0: break; \
		default: memcpy((Q), (P), (N)); break; \
		} \
	} \
	G_STMT_END
```
(untested)

jcupitt: Oh, very clever! Sure, let's do that.

I timed on this PC and see eg.:

```
$ vips replicate ~/pics/lion.svg rgba.v 20 20
$ VIPS_CONCURRENCY=1 time vips rot rgba.v x.v d90 
0.40user 0.58system 0:01.54elapsed 63%CPU (0avgtext+0avgdata 1182188maxresident)k
```

ie. a single-threaded 90 degree rotate of a 16k x 16k pixel uint8 RGBA image:

`memcpy()`: 1.26
the `VIPS_PELCPY()` above: 0.40
your `VIPS_MEMCPY()`: 0.40

gcc, release mode, fastest of five runs, this SVG:

lion

jcupitt: This is such a simple optimisation you wonder why compilers don't do this automatically as well.

jcupitt: I tested with gcc15 and clang21 and both see the same large jump in performance for things like 90 degree rotate.

kleisauke: Great! I'll mark this as resolved, that should appease UBSan.

jcupitt:
I added a changelog line for the addition of `VIPS_MEMCPY()` and credited you, btw.

kleisauke:
Thanks! I noticed there's a unresolved merge conflict in `ChangeLog`.

kleisauke:
(I updated the PR title to match #4969)

ChangeLog
```diff
@@ -22,8 +22,12 @@ date-tbd 8.19.0
 - uhdrsave: add existing JPEG options for base image [lovell]
 - csvload: add maximum limit on column count [lovell]
 - pdfload_buffer: only search the first few bytes in is_a [violetviolinist]
+<<<<<<< perf-conversion
+- add VIPS_MEMCPY() [kleisauke]
+=======
 - heifload: allow mif2 and mif3 "brand" images e.g. mini box [lovell]
 - add vips_scRGB2CICP [mertalev]
+>>>>>>> master
```
```suggestion
- add VIPS_MEMCPY() [kleisauke]
- heifload: allow mif2 and mif3 "brand" images e.g. mini box [lovell]
- add vips_scRGB2CICP [mertalev]
```

kleisauke (approved):
LGTM!

ChangeLog
jcupitt: Stupid thing!! I swear pressed "both", sigh.

jcupitt:
@JakeChampion, merged! And thank you very much for doing this work.

jcupitt:
For reference, I think this is best version of the memcpy macro just using pointer copies:

```C
#define VIPS_MEMCPY(Q, P, N) \
    G_STMT_START \
    { \
        switch (N) { \
        case 8: \
            ((guint64 *) (Q))[0] = ((guint64 *) (P))[0]; \
            break; \
\
        case 4: \
            ((guint32 *) (Q))[0] = ((guint32 *) (P))[0]; \
            break; \
\
        case 6: \
            ((guint16 *) (Q))[2] = ((guint16 *) (P))[2]; \
            ((guint16 *) (Q))[1] = ((guint16 *) (P))[1]; \
        case 2: \
            ((guint16 *) (Q))[0] = ((guint16 *) (P))[0]; \
            break; \
 \
        case 7: \
            ((guint8 *) (Q))[6] = ((guint8 *) (P))[6]; \
            ((guint8 *) (Q))[5] = ((guint8 *) (P))[5]; \
        case 5: \
            ((guint8 *) (Q))[4] = ((guint8 *) (P))[4]; \
            ((guint8 *) (Q))[3] = ((guint8 *) (P))[3]; \
        case 3: \
            ((guint8 *) (Q))[2] = ((guint8 *) (P))[2]; \
            ((guint8 *) (Q))[1] = ((guint8 *) (P))[1]; \
        case 1: \
            ((guint8 *) (Q))[0] = ((guint8 *) (P))[0]; \
            break; \
\
        case 0: \
            break; \
\
        default: \
            memcpy((Q), (P), (N)); \
            break; \
        } \
    } \
    G_STMT_END
```

But it has identical performance and binary size to the 8-way memcpy version, so memcpy wins.

PR #5077, opened 2026-06-12
perf: optimize resample module hot paths

jcupitt:
From https://github.com/libvips/libvips/pull/4970 but incorporating review comments.

### DIFF
--- a/libvips/resample/thumbnail.c
+++ b/libvips/resample/thumbnail.c
@@ -157,6 +157,11 @@ typedef struct _VipsThumbnail {
 	 */
 	gboolean page_pyramid;
 
+	/* Max alpha value for premultiply/unpremultiply, stored here so
+	 * gen functions can access it.
+	 */
+	double premultiply_max_alpha;
+
 } VipsThumbnail;
 
 typedef struct _VipsThumbnailClass {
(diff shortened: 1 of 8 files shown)


jcupitt:
I added a fast uchar path to premultiply / unpremultiply:

```
$ VIPS_CONCURRENCY=1 time vips premultiply rgba.v x.v
0.98user 1.86system 0:05.68elapsed 50%CPU (0avgtext+0avgdata 1233144maxresident)k
392inputs+9031256outputs (0major+20384minor)pagefaults 0swaps

$ VIPS_CONCURRENCY=1 time vips premultiply rgba.v x.v --uchar
0.76user 0.72system 0:01.82elapsed 82%CPU (0avgtext+0avgdata 1182132maxresident)k
96inputs+2257824outputs (0major+7535minor)pagefaults 0swaps
```

So 0.98 down to 0.76 for premultiply. Unpremultiply is only very slightly faster, but it'll save a float -> uchar later, and a bit of memory.

Accuracy is 1 or less.

jcupitt:
Edit: updated timings after review corrections, it's not as good heh

I flipped thumbnail to use the uchar path in premultiply / unpremultiply.

git master with a 17k x 17k rgba image:

```
$ VIPS_CONCURRENCY=1 time vipsthumbnail rgba.v -o x.v --size 512
1.10user 0.14system 0:01.24elapsed 100%CPU (0avgtext+0avgdata 1265024maxresident)k
0inputs+2056outputs (0major+30208minor)pagefaults 0swaps
```

This PR:

```
$ VIPS_CONCURRENCY=1 time vipsthumbnail rgba.v -o x.v --size 512
0.74user 0.55system 0:01.76elapsed 73%CPU (0avgtext+0avgdata 1182436maxresident)k
0inputs+2257824outputs (0major+7647minor)pagefaults 0swaps
```

About 50% faster.

libvips/resample/thumbnail.c
```diff
@@ -853,10 +851,18 @@ vips_thumbnail_build(VipsObject *object)
...
 
-		if (vips_premultiply(in, &t[3], NULL) ||
-			vips_cast(t[3], &t[4], unpremultiplied_format, NULL))
-			return -1;
-		in = t[4];
+		if (in->BandFmt == VIPS_FORMAT_UCHAR) {
+			/* Fast path: stay in uchar.
+			 */
+			if (vips_premultiply(in, &t[3], "uchar", TRUE, NULL))
+				return -1;
```
kleisauke: I think there's a missing `in = t[3];` here.

libvips/resample/thumbnail.c
```diff
@@ -884,10 +890,19 @@ vips_thumbnail_build(VipsObject *object)
...
-		if (vips_unpremultiply(in, &t[6], NULL) ||
-			vips_cast(t[6], &t[7], unpremultiplied_format, NULL))
-			return -1;
-		in = t[7];
+
+		if (unpremultiplied_format == VIPS_FORMAT_UCHAR) {
+			/* Fast path: unpremultiply in UCHAR directly.
+			 */
+			if (vips_unpremultiply(in, &t[6], "uchar", TRUE, NULL))
+				return -1;
```
kleisauke, line 893: Same as above, it looks like there's a missing `in = t[6];` here.

libvips/resample/thumbnail.c
```diff
@@ -853,10 +851,18 @@ vips_thumbnail_build(VipsObject *object)
...
-		in = t[4];
+		if (in->BandFmt == VIPS_FORMAT_UCHAR) {
+			/* Fast path: stay in uchar.
+			 */
+			if (vips_premultiply(in, &t[3], "uchar", TRUE, NULL))
+				return -1;
+		}
+		else {
+			if (vips_premultiply(in, &t[3], NULL) ||
+				vips_cast(t[3], &t[4], unpremultiplied_format, NULL))
```
kleisauke: For context, this was introduced in commit 2ece8c2d22b8bffa569defdf29309868a87247d8 from PR #3618. So, I think you can remove this `vips_cast()` and below now.

kleisauke: Sorry, I meant only this `vips_cast()`, below needs to stay. I need more coffee.

kleisauke: I fiddled with this a while back by adding optimized float paths using Highway, which would avoid the need for this. IIRC, that didn't end up working out. Let me see if I can revive that changeset.

kleisauke: > Let me see if I can revive that changeset.

I found it on one of my local branches and pushed it here:
https://github.com/libvips/libvips/compare/master...kleisauke:simd-highway-float

jcupitt: I guess, though it would mean we'd be shrinking ushort alpha images as floats. It probably doesn't matter though, you're right.

libvips/resample/thumbnail.c
jcupitt: Ooops!

libvips/resample/thumbnail.c
```diff
@@ -853,10 +851,18 @@ vips_thumbnail_build(VipsObject *object)
...
+			 */
+			if (vips_premultiply(in, &t[3], "uchar", TRUE, NULL))
+				return -1;
+			in = t[3];
+		}
+		else {
+			if (vips_premultiply(in, &t[3], NULL))
+				return -1;
+			in = t[4];
+		}
```
kleisauke:
```suggestion
		if (vips_premultiply(in, &t[3],
				/* Fast path: stay in uchar.
				 */
				"uchar", in->BandFmt == VIPS_FORMAT_UCHAR,
				NULL))
			return -1;
		in = t[3];
```

libvips/resample/thumbnail.c
```diff
@@ -884,10 +885,20 @@ vips_thumbnail_build(VipsObject *object)
...
-			vips_cast(t[6], &t[7], unpremultiplied_format, NULL))
-			return -1;
-		in = t[7];
+
+		if (unpremultiplied_format == VIPS_FORMAT_UCHAR) {
+			/* Fast path: unpremultiply in UCHAR directly.
+			 */
+			if (vips_unpremultiply(in, &t[6], "uchar", TRUE, NULL))
+				return -1;
+			in = t[3];
```
kleisauke:
```suggestion
			in = t[6];
```

libvips/resample/thumbnail.c
jcupitt: We could just set `uchar` all the time, it only has an effect for uchar input images. But maybe we don't want to link these two functions together that closely.

libvips/resample/thumbnail.c
kleisauke: Indeed, I'm not sure how common ushort alpha images are either.

jcupitt:
Ah I'm rushing again, I need to slow down and think a little.

Thanks for reviewing this half-baked thing!

kleisauke:
No worries, I think it's in good shape now, still need to check the timings myself.

(I updated the PR title to match #4970)

kleisauke:
Test image:
```console
$ vips bandjoin_const st-francis.jpg x.v 255
$ vipsheader x.v
x.v: 30000x26319 uchar, 4 bands, srgb, jpegload
```

On my AMD Ryzen 9 7900 workstation, the timings are:

`master` branch:
```console
$ time vipsthumbnail x.v -o x2.v --size 512 --vips-concurrency=1

real	0m3.559s
user	0m3.329s
sys	0m0.218s
$ sha256sum x2.v
0e6783d90aa99380a64c6f973ef35355aae91bad098dd8805685323652d426c1  x2.v
```

This PR:
```console
$ time vipsthumbnail x.v -o x2.v --size 512 --vips-concurrency=1

real	0m1.443s
user	0m1.237s
sys	0m0.201s
$ sha256sum x2.v
0e6783d90aa99380a64c6f973ef35355aae91bad098dd8805685323652d426c1  x2.v
```

So, ~2.5x faster while producing identical images, nice! :tada:

@JakeChampion Thanks for doing this initial work. Much appreciated!

kleisauke (approved):

PR #2380, opened 2021-08-03
Fix a few `-Wstrict-aliasing=1` warnings

kleisauke:
GCC 11.x could produce bad assembly code due to this. See for example: https://godbolt.org/z/8vT7W8d1o.

```
Lab2XYZ.c: In function ‘vips_col_Lab2XYZ’:
Lab2XYZ.c:250:43: warning: dereferencing type-punned pointer might break strict-aliasing rules [-Wstrict-aliasing]
  250 |         vips_Lab2XYZ_line( (VipsColour *) &Lab2XYZ,
      |                                           ^~~~~~~~
Lab2XYZ.c:251:47: warning: dereferencing type-punned pointer might break strict-aliasing rules [-Wstrict-aliasing]
  251 |                 (VipsPel *) out, (VipsPel **) &x, 1 );
      |                                               ^~
LCh2Lab.c: In function ‘vips_col_Ch2ab’:
LCh2Lab.c:113:64: warning: dereferencing type-punned pointer might break strict-aliasing rules [-Wstrict-aliasing]
  113 |         vips_LCh2Lab_line( NULL, (VipsPel *) out, (VipsPel **) &x, 1 );
      |                                                                ^~
XYZ2Lab.c: In function ‘vips_col_XYZ2Lab’:
XYZ2Lab.c:185:43: warning: dereferencing type-punned pointer might break strict-aliasing rules [-Wstrict-aliasing]
  185 |         vips_XYZ2Lab_line( (VipsColour *) &XYZ2Lab,
      |                                           ^~~~~~~~
XYZ2Lab.c:186:47: warning: dereferencing type-punned pointer might break strict-aliasing rules [-Wstrict-aliasing]
  186 |                 (VipsPel *) out, (VipsPel **) &x, 1 );
      |                                               ^~
```

PR #5079, opened 2026-06-12
CI: various upgrades

kleisauke:

~~Commit d716bd68ab8767daf93e018f4fd7eda924ccf987 was needed due to:~~

<details>
  <summary>Details</summary>

```console
=================================== FAILURES ===================================
_________________________ TestColour.test_colourspace __________________________

self = <test_colour.TestColour object at 0x7221ba1d9f90>

    def test_colourspace(self):
        # mid-grey in Lab ... put 42 in the extra band, it should be copied
        # unmodified
        test = pyvips.Image.black(100, 100) + [50, 0, 0, 42]
        test = test.copy(interpretation=pyvips.Interpretation.LAB)
    
        # a long series should come in a circle
        im = test
        for col in colour_colourspaces + [pyvips.Interpretation.LAB]:
            im = im.colourspace(col)
            assert im.interpretation == col
    
            for i in range(0, 4):
                min_l = im.extract_band(i).min()
                max_h = im.extract_band(i).max()
>               assert pytest.approx(min_l) == max_h
E               assert 1.25693418340...e-06 ± 1.3e-12 == 1.2837916756325285e-06
E                 
E                 comparison failed
E                 Obtained: 1.2837916756325285e-06
E                 Expected: 1.2569341834023362e-06 ± 1.3e-12

test/test-suite/test_colour.py:24: AssertionError
```
</details>

~~which seems to be a GCC 15 issue, as Clang 22 appears to be OK.~~ **Edit:** see triage below.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,8 +28,8 @@ jobs:
           #  build: { cc: clang-22, cxx: clang++-22, linker: ld.lld-22, sanitize: 'thread' }
           #  extra_meson_args: -Dintrospection=disabled
 
-          - name: "macOS arm64 (15) - Xcode 16"
-            os: macos-15
+          - name: "macOS arm64 (26) - Xcode 26"
+            os: macos-26
             build: { cc: clang, cxx: clang++, linker: ld.lld }
             extra_meson_args: -Dintrospection=disabled
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,20 +12,20 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - name: "Linux x64 (Ubuntu 24.04) - GCC 14"
-            os: ubuntu-24.04
-            build: { cc: gcc-14, cxx: g++-14, linker: ld }
+          - name: "Linux x64 (Ubuntu 26.04) - GCC 15"
+            os: ubuntu-26.04
+            build: { cc: gcc-15, cxx: g++-15, linker: ld }
             extra_meson_args: -Ddocs=true
 
-          - name: "Linux x64 (Ubuntu 24.04) - Clang 19 with ASan and UBSan"
-            os: ubuntu-24.04
-            build: { cc: clang-19, cxx: clang++-19, linker: ld.lld-19, sanitize: 'address,undefined' }
+          - name: "Linux x64 (Ubuntu 26.04) - Clang 22 with ASan and UBSan"
+            os: ubuntu-26.04
+            build: { cc: clang-22, cxx: clang++-22, linker: ld.lld-22, sanitize: 'address,undefined' }
             extra_meson_args: -Dintrospection=disabled
 
           # We cannot reliably test this if GLib is built without TSan instrumentation.
-          #- name: "Linux x64 (Ubuntu 24.04) - Clang 19 with TSan"
-          #  os: ubuntu-24.04
-          #  build: { cc: clang-19, cxx: clang++-19, linker: ld.lld-19, sanitize: 'thread' }
+          #- name: "Linux x64 (Ubuntu 26.04) - Clang 22 with TSan"
+          #  os: ubuntu-26.04
+          #  build: { cc: clang-22, cxx: clang++-22, linker: ld.lld-22, sanitize: 'thread' }
           #  extra_meson_args: -Dintrospection=disabled
 
           - name: "macOS arm64 (15) - Xcode 16"
@@ -48,16 +48,17 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install \
-            meson gi-docgen pkg-config \
+            meson pkgconf \
+            gi-docgen gobject-introspection \
             libarchive-dev libcfitsio-dev libcgif-dev \
             libexif-dev libexpat1-dev libffi-dev \
             libfftw3-dev libheif-dev libheif-plugin-aomenc \
-            libheif-plugin-x265 libhwy-dev libimagequant-dev \
-            libjpeg-dev libjxl-dev liblcms2-dev \
-            libmatio-dev libnifti-dev libopenexr-dev \
-            libopenjp2-7-dev libopenslide-dev libpango1.0-dev \
-            libpng-dev libpoppler-glib-dev librsvg2-dev \
-            libtiff5-dev libwebp-dev
+            libhwy-dev libimagequant-dev libjpeg-dev \
+            libjxl-dev liblcms2-dev libmatio-dev \
+            libnifti-dev libopenexr-dev libopenjp2-7-dev \
+            libopenslide-dev libpango1.0-dev libpng-dev \
+            libpoppler-glib-dev librsvg2-dev libtiff5-dev \
+            libwebp-dev
 
       - name: Install macOS dependencies
         if: runner.os == 'macOS'
@@ -76,9 +77,9 @@ jobs:
             mozjpeg openexr openjpeg openslide pango \
             poppler webp
 
-      - name: Install Clang 19
-        if: runner.os == 'Linux' && matrix.build.cc == 'clang-19'
-        run: sudo apt-get install clang-19 libclang-rt-19-dev lld-19 llvm-19
+      - name: Workaround -march=x86-64-v3 GCC bug
+        if: runner.os == 'Linux' && matrix.build.cc == 'gcc-15'
+        run: echo "CPPFLAGS=-Wall -march=x86-64" >> $GITHUB_ENV
 
       - name: Prepare macOS environment
         if: runner.os == 'macOS'
@@ -88,7 +89,9 @@ jobs:
       - name: Prepare sanitizers environment
         if: matrix.build.sanitize
         run: |
-          echo "CPPFLAGS=-g -fno-omit-frame-pointer -fno-sanitize=function -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION" >> $GITHUB_ENV
+          # FIXME: remove `-DHWY_BROKEN_TARGETS=(HWY_AVX3|HWY_AVX3_DL)` once Highway is updated to 1.4.0
+          # https://github.com/google/highway/issues/2705
+          echo "CPPFLAGS=-g -fno-omit-frame-pointer -fno-sanitize=function -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -DHWY_BROKEN_TARGETS=(HWY_AVX3|HWY_AVX3_DL)" >> $GITHUB_ENV
           echo "LDFLAGS=-g -shared-libsan" >> $GITHUB_ENV
           # Glib is built without -fno-omit-frame-pointer. We need
           # to disable the fast unwinder to get full stacktraces.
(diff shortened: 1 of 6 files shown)


jcupitt:
How about changing test_colour to use a bigger epsilon instead? Eg. maybe:

```python
         assert abs (max_h - min_l) < 0.00001
```

It's strange that gcc differs from clang here. Exact float precision depends on the float instruction set used, which can vary with autovectorisation. Maybe the compilers are using different vector instruction sets for some operations?

kleisauke:
A bigger epsilon would fix this, but I suspect it's a GCC-specific issue (since I couldn't reproduce it locally with GCC 16.1.1).

I'll run a reverse `git bisect` on GCC, I'm too curious to find out which commit fixed it.

kleisauke:
It's the [optimized amd64v3 APT repo](https://discourse.ubuntu.com/t/introducing-architecture-variants-amd64v3-now-available-in-ubuntu-25-10/71312) that is causing this issue.

<details>
  <summary>Reproducer</summary>

```bash
docker run -it ubuntu:26.04

export DEBIAN_FRONTEND=noninteractive

echo 'APT::Architecture-Variants "amd64v3";' | tee /etc/apt/apt.conf.d/99enable-amd64v3
apt-get update -y
apt-get upgrade -y

apt-get update -y
apt-get install -y \
  python3-pip \
  git meson pkgconf \
  libarchive-dev libcfitsio-dev libcgif-dev \
  libexif-dev libexpat1-dev libffi-dev \
  libfftw3-dev libheif-dev libheif-plugin-aomenc \
  libhwy-dev libimagequant-dev libjpeg-dev \
  libjxl-dev liblcms2-dev libmatio-dev \
  libnifti-dev libopenexr-dev libopenjp2-7-dev \
  libopenslide-dev libpango1.0-dev libpng-dev \
  libpoppler-glib-dev librsvg2-dev libtiff5-dev \
  libwebp-dev

git clone https://github.com/libvips/libvips.git
cd libvips
meson setup build -Ddebug=true
meson compile -C build
meson test -C build
meson install -C build
ldconfig

pip3 install pyvips[test] --break-system-packages

VIPS_LEAK=1 python3 -m pytest -sv --log-cli-level=WARNING test/test-suite/test_colour.py -k test_colourspace
```

</details>

This repo configures [GCC with `--with-arch-64=x86-64-v3`](https://git.launchpad.net/ubuntu/+source/gcc-15/tree/debian/rules2?h=import/15.2.0-16ubuntu1&id=e83786c68583101d742bfcc7f165fb963df803be#n728), which sets the default for the `-march` option. I was able to workaround this by explicitly compiling libvips with `CPPFLAGS=-march=x86-64` (and I've removed the `xfail` marker in favor for this workaround).

Note that this also affects GCC 16.1.1 (as available in Fedora 44) and GCC trunk (commit https://github.com/gcc-mirror/gcc/commit/bd4b33d265d44bfb3222c7bb962818179242bf4b) when compiling with `CPPFLAGS=-march=x86-64-v3`, so it might need further investigation.

jcupitt:
Ah how interesting. Float math is very unreliable :( A larger epsilon might be a safer fix.

(if we ever have to revisit this)

jcupitt (approved):

kleisauke:
> A larger epsilon might be a safer fix.

Indeed, I still suspect this is a GCC-specific optimization bug, but let's revisit once it also reproduces with Clang when compiling with `CPPFLAGS=-march=x86-64-v3`.

> I'll run a reverse `git bisect` on GCC, I'm too curious to find out which commit fixed it.

For future reference, the gist at https://gist.github.com/kleisauke/aa13a0031f92f933a0e0bbfcb5e798fb could be useful for a reverse bisection.

kleisauke:
Ah, GCC 14.3.0 doesn't have this issue. A standard `git bisect` seems to point to https://github.com/gcc-mirror/gcc/commit/77bd23a3e247555ba427d01af8e5713934be8d5b.

<details>
  <summary>Details</summary>

```console
$ git bisect start
status: waiting for both good and bad commits
$ git bisect bad releases/gcc-15.1.0
status: waiting for good commit(s), bad commit known
$ git bisect good releases/gcc-14.3.0
Bisecting: a merge base must be tested
[b8e9fd535d6093e3a24af858364d8517a767b0d7] amdgcn: Add gfx90c target
$ git bisect run ~/gcc-bisect.sh
...
77bd23a3e247555ba427d01af8e5713934be8d5b is the first bad commit
...
bisect found first bad commit
$ git bisect log
git bisect start
# status: waiting for both good and bad commits
# bad: [1b306039ac49f8ad91ca71d3de3150a3c9fa792a] Update ChangeLog and version files for release
git bisect bad 1b306039ac49f8ad91ca71d3de3150a3c9fa792a
# status: waiting for good commit(s), bad commit known
# good: [c9cd41fba9ebd288c4f101e4b99da934bcb96a11] Update ChangeLog and version files for release
git bisect good c9cd41fba9ebd288c4f101e4b99da934bcb96a11
# good: [b8e9fd535d6093e3a24af858364d8517a767b0d7] amdgcn: Add gfx90c target
git bisect good b8e9fd535d6093e3a24af858364d8517a767b0d7
# bad: [acba8b3d8dec0124c8b3a7e112b3a784a5091214] [PATCH] Fix SLP when ifcvt versioned loop is not vectorized
git bisect bad acba8b3d8dec0124c8b3a7e112b3a784a5091214
# good: [f9c7775f58798a051b57356ad321b758a2ee837d] xtensa: Make use of scaled [U]FLOAT/TRUNC.S instructions
git bisect good f9c7775f58798a051b57356ad321b758a2ee837d
# good: [323291c29c77e3214f4850129bb8a3d0d8da6a45] ipa: Rename ipa_supports_p to ipa_vr_supported_type_p
git bisect good 323291c29c77e3214f4850129bb8a3d0d8da6a45
# bad: [e5f5cffb8c8243896a9d3bd0e2b8f14c70f8df1e] tree-optimization/116024 - simplify C1-X cmp C2 for wrapping signed types
git bisect bad e5f5cffb8c8243896a9d3bd0e2b8f14c70f8df1e
# bad: [27003e5d6eadcddde617b89f11bab47ab75cc203] doc: Remove MinGW note on binutils 2.16
git bisect bad 27003e5d6eadcddde617b89f11bab47ab75cc203
# bad: [0189ab205aa86b8e67ae982294f0fe58aa9c4774] testsuite: Update commandline for PR116628.c to use neoverse-v2 [PR116628]
git bisect bad 0189ab205aa86b8e67ae982294f0fe58aa9c4774
# good: [952df9c50b30cc6f849c422b84592a81524f8ef7] riscv: Fix duplicate assmbler label in @tlsdesc<mode> insn
git bisect good 952df9c50b30cc6f849c422b84592a81524f8ef7
# good: [2b7b8d3bb52a23aa8b1d6e9a2d57c83db2078f73] i386: Add missing avx512f-mask-type.h include
git bisect good 2b7b8d3bb52a23aa8b1d6e9a2d57c83db2078f73
# bad: [2ec49f3907a6e079567d49dd3bb4cb617526e2a1] Update gcc zh_CN.po
git bisect bad 2ec49f3907a6e079567d49dd3bb4cb617526e2a1
# good: [e917a251d82acf7c7c8577bd5dcb448deacb6f66] Match: Support form 3 for scalar signed integer .SAT_ADD
git bisect good e917a251d82acf7c7c8577bd5dcb448deacb6f66
# good: [d3a7302ec5985abcda561886cc724d388c7143cb] libstdc++: add #pragma diagnostic
git bisect good d3a7302ec5985abcda561886cc724d388c7143cb
# bad: [5b5a36b122e1205449f1512bf39521b669e713ef] tree-optimization/116768 - wrong dependence analysis
git bisect bad 5b5a36b122e1205449f1512bf39521b669e713ef
# bad: [77bd23a3e247555ba427d01af8e5713934be8d5b] Fall back to single-lane SLP before falling back to no SLP
git bisect bad 77bd23a3e247555ba427d01af8e5713934be8d5b
# first bad commit: [77bd23a3e247555ba427d01af8e5713934be8d5b] Fall back to single-lane SLP before falling back to no SLP
```

`gcc-bisect.sh`: https://gist.github.com/kleisauke/f54326e7852f8c173b6109c1af9bdefd
</details>

Given the bisect result, this likely relates to changes in the vectorizer. I couldn't find any bug reports mentioning this specific commit, apart from the one at https://gcc.gnu.org/PR116784 (which appears to be just a test expectation that needed updating).

@thesamesam Sorry for the ping, but since you helped with the previous GCC issue, do you think it's appropriate to report it as-is at https://gcc.gnu.org/bugzilla/? Or should I reduce it further first? Happy to dig deeper if needed.

kleisauke:
Further investigation shows that the issue is only reproducible with `-O3` (i.e. `--buildtype=release`, which libvips uses by default), whereas building with `--buildtype=debugoptimized` (i.e. `-O2`) prevents the assertion failure.

More precisely, adding `__attribute__((optimize("O2")))` to `vips_scRGB2XYZ_line()` seems to fix this.
https://github.com/libvips/libvips/blob/79be1c3b89d2709ef57eaa761b319fa52d0ed0ad/libvips/colour/scRGB2XYZ.c#L57-L79
(see draft PR https://github.com/libvips/libvips/pull/5087)

When comparing `-O3`  builds between GCC commits https://github.com/gcc-mirror/gcc/commit/d3a7302ec5985abcda561886cc724d388c7143cb and https://github.com/gcc-mirror/gcc/commit/77bd23a3e247555ba427d01af8e5713934be8d5b, I also observed assembly-level differences in `LabS2Lab.c`, `Oklab2XYZ.c` and `XYZ2Lab.c`.

I've included some preliminary analysis here: https://gist.github.com/kleisauke/6dcc2df09863f6e45233bb6b1c81a1fe.

thesamesam:
> @thesamesam Sorry for the ping, but since you helped with the previous GCC issue, 

Please always feel free to ask. I enjoy these!

> More precisely, adding `__attribute__((optimize("O2")))` to `vips_scRGB2XYZ_line()` seems to fix this.

Can you `__attribute__((noipa))` it first and then if that works, try `__attribute__((noinline))`? It'll tell us whether that function by itself gets miscompiled or if it requires information from a caller (say if it gets partially inlined or cloned with some fixed value where a caller lets it make an assumption).

I think the function itself is broken because it's called via a function pointer and GCC doesn't propagate much across those (even speculatively) yet, though it is being worked on.

> Or should I reduce it further first? Happy to dig deeper if needed.

Let's try get it a little bit further first. Also, I'm not an expert on FP, but does `-ffp-contract=off` help? In the past, I've debugged some of these and it ended up being quasi-expected because of FP contraction, though the plan is to fix GCC (hopefully for 17 finally) to default to safe contraction. The current scheme can sometimes accumulate significant errors.

jcupitt:
O2 would disable autovectorisation, of course, could that be relevant? Perhaps removing the `restrict` would also fix (or side-step!) this.

kleisauke:
> > More precisely, adding `__attribute__((optimize("O2")))` to `vips_scRGB2XYZ_line()` seems to fix this.
> 
> Can you `__attribute__((noipa))` it first and then if that works, try `__attribute__((noinline))`? 

Those attributes don't seem to make any difference. I also investigated whether it could be a strict aliasing violation, as we previously had in this component (see e.g. PR #2380), but that also appears to be fine.

> Also, I'm not an expert on FP, but does `-ffp-contract=off` help?

Ah, it turns out that `-ffp-contract=off` does fix the issue! You can also notice this on Godbolt:
https://godbolt.org/z/TMjsvnd4Y

So, we should probably increase the epsilon in the testsuite to account for differences introduced by FMA optimizations.

PR #5084, opened 2026-06-13
add '@' size limit modifier to vipsthumbnail

jcupitt:
So eg.:

	$ vipsthumbnail k2.jpg --size 10000@
	$ vipsheader tn_k2.jpg
	tn_k2.jpg: 83x118 uchar, 3 bands, srgb, jpegload

83 * 118 is 9794, so just less than 10,000.o

The syntax comes from imagemagick:

https://imagemagick.org/command-line-processing/#geometry

See https://github.com/libvips/libvips/discussions/5082

Thanks @clairem-sl for the suggestion


clairem-sl:
This would be perfectly suitable for my use case!
Hoping this gets merged 😍

kleisauke (approved):

PR #5086, opened 2026-06-15
dcrawload: add preview jpegs

### LINKED ISSUE #5026: dcrawload: improve thumbnail filtering
dcrawload tries to select one thumbnail image from the raw file to attach as metadata. In some cases, the current implementation does not always make the best choice. 

Revisit this code and see if we can improve it.

### Discussed in https://github.com/libvips/libvips/discussions/5001

<sup>Originally posted by **kiefermat** April 17, 2026</sup>
Hi,

I am currently looking into raw support of vips. Let me first thank you all for this project. I really like it!

I am curious what is the reason behind this code in dcrawload.c:
```c++
		// useless thumbnails the same size as the main image are very
		// common
		if (this->twidth >= sizes->iwidth &&
			this->theight >= sizes->iheight)
			continue;
```

The code loops over all embedded previews and puts the best fitting into the field "raw-thumbnail-data".  Many raw images have a full scale jpeg image embedded which can be used for previewing the file content.

Let me explain my use case: My target is to get a full HD preview of a raw file as fast as possible. For this, I am experimenting with the following python code:
```python
def open_image_with_preview_if_available(file_path: str, size: int) -> Image:
    vips_image = Image.new_from_file(file_path, access="sequential")
    result = None
    if vips_image.get_typeof("raw-thumbnail-data"):
        thumbnail_data = vips_image.get("raw-thumbnail-data")
        embedded_image = Image.thumbnail_buffer(thumbnail_data, size, height=size, size=Size.DOWN, auto_rotate=False)
        if embedded_image.width < size and embedded_image.height < size:
            LOG.info("skipping embedded thumbnail because it is smaller than the target size: %dx%d", embedded_image.width, embedded_image.height)
        else:        
            original_orientation = vips_image.get("orientation") if vips_image.get_typeof("orientation") else None
            if original_orientation:
                embedded_image = transform_according_to_orientation(embedded_image, original_orientation)
            result = embedded_image
    if not result:
        result = Image.thumbnail(file_path, size, height=size, size=Size.DOWN)
    return result
```

I am currently using a similar mechanism using dcraw to extract the embedded preview image and using the embedded jpeg is usually much faster than processing the raw file directly.

However, since vips filters out the full sized jpeg preview, this approach does currently not work since the other embedded thumbnails are often much smaller.

So my questions are:
* Is this reasonable code to accomplish my use case with vips?
* Would it be possible to include the full sized jpeg as thumbnail data into the image fields (either as separate field or as "raw-thumbnail-data")?
* Is there a possibility to extend vipsthumbnail to also use the embedded image data? 😄 
</div>

jcupitt:
And rename `raw-thumbnail-data` as `jpeg-thumbnail-data`.

	$ vipsheader -f jpeg-thumbnail-data RAW_NIKON_E5700_SRGB.NEF | base64 -d > x.jpg
	$ vipsheader -f jpeg-preview-data RAW_NIKON_E5700_SRGB.NEF | base64 -d > x.jpg

closes #5026

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -29,6 +29,7 @@ date-tbd 8.19.0
 - add VIPS_UHDR_TO_SCRGB
 - add uchar options to premultiply and unpremultiply [jcupitt]
 - vipsthumbnail: add '@' modifier for size to pixel count
+- dcrawload: attach preview jpeg as metadata
 
 6/6/26 8.18.3
 
--- a/libvips/foreign/dcrawload.c
+++ b/libvips/foreign/dcrawload.c
@@ -350,6 +350,11 @@ vips_foreign_load_dcraw_set_metadata(VipsForeignLoadDcRaw *raw,
 		vips_image_set_blob_copy(image, "jpeg-thumbnail-data",
 			raw->raw_processor->thumbnail.thumb,
 			raw->raw_processor->thumbnail.tlength);
+
+		// old deprecated name for compat
+		vips_image_set_blob_copy(image, "raw-thumbnail-data",
+			raw->raw_processor->thumbnail.thumb,
+			raw->raw_processor->thumbnail.tlength);
 	}
 
 	if (preview_index != -1 &&
--- a/libvips/foreign/dcrawload.c
+++ b/libvips/foreign/dcrawload.c
@@ -2,6 +2,8 @@
  *
  * 15/8/25 dvdkon
  *	- handle image rotation
+ * 15/6/26
+ *	- add raw-preview-data, the largest embedded preview jpeg
  */
 
 /*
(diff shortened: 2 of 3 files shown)


jcupitt:
There two obvious issues with this PR:

### Orientation

With `RAW_NIKON_E5700_SRGB.NEF`, for example, the main image has orientation 8, but the thumbnail jpeg has no orientation metadata and has been spun upright. We can't apply the main image orientation to the thumbnail. Other cameras and RAW formats seem to do different things here. 

Perhaps there's not much libvips can do and this should just get punted downstream.

### Metadata names

This PR attaches two JPEGs:

1. `jpeg-thumbnail-data` is a JPEG-formatted blob of the biggest JPEG image in the RAW file that's under 1mb and which matches the main image in number of bands. Current stable libvips called this field `raw-thumbnail-data`, so this is a (small) API change. 
2. `jpeg-preview-data` is a JPEG-formatted blob of the largest JPEG image in the RAW file that is not the thumbnail selected above.

One / both are optional, depending on the image.

We could also attach `raw-thumbnail-data` but tag it as deprecated, if the name change would break anyone's code.

jcupitt:
Test with eg.:

```
$ vipsheader -a IMG_3260.dng 
IMG_3260.dng: 5496x3670 uchar, 3 bands, srgb, dcrawload
width: 5496
height: 3670
bands: 3
format: uchar
coding: none
interpretation: srgb
xoffset: 0
yoffset: 0
xres: 1
yres: 1
filename: IMG_3260.dng
vips-loader: dcrawload
raw-make: Canon
raw-model: EOS 70D
raw-software: Adobe DNG Converter 17.2 (Windows)
raw-iso: 400
raw-shutter: 0.001
raw-aperture: 9
raw-focal-length: 840
raw-timestamp: 2025-06-22T11:31:16Z
xmp-data: 6556 bytes of binary data
raw-lens: 150-600mm F5-6.3 DG OS HSM | Contemporary 015 +1.4x
orientation: 1
jpeg-thumbnail-data: 38892 bytes of binary data
jpeg-preview-data: 1700958 bytes of binary data
```

Then:

```
$ vipsheader -f jpeg-preview-data IMG_3260.dng | base64 -d > x.jpg
```

To make:

```
$ vipsheader x.jpg
x.jpg: 5472x3648 uchar, 3 bands, srgb, jpegload
```

x

jcupitt:
... we should also look at integrating this with `vips_thumbnail()`.

Perhaps looking at the scale factor, picking one of the JPEG images, then using the thumbnail JPEG path.

PR #5087, opened 2026-06-15
test: relax tolerance for colourspace min/max comparison

kleisauke:
To account for floating-point variation introduced by FMA optimizations.

Context: https://github.com/libvips/libvips/pull/5079#issuecomment-4710936434.


lovell (approved):
+100.001%

PR #5090, opened 2026-06-16
tiffload: squash read errors

### LINKED ISSUE #5089: libvips openin_source_read() propagates -1 to libtiff TIFFReadAndRealloc() causing heap OOB write
## Summary

libvips is a natural victim of a heap out-of-bounds write in libtiff's
`TIFFReadAndRealloc()` because its TIFF read callback (`openin_source_read`)
uses the `read()` syscall directly, which returns `-1` on `ECONNRESET`.

Unlike GDAL and ImageMagick — which normalize read errors to `0` — libvips
propagates the `-1` directly into libtiff, triggering the OOB write.

The root bug is in libtiff and has been reported upstream:
https://gitlab.com/libtiff/libtiff/-/work_items/854

---

## Affected Component in libvips

**File:** `libvips/foreign/tiff.c` — function `openin_source_read()`

```c
static tmsize_t
openin_source_read( thandle_t st, void *data, tmsize_t size )
{
    ...
    bytes_read = vips_source_read( source, data, size );
    return (tmsize_t) bytes_read;   /* returns -1 on ECONNRESET — no masking */
}
```

Call chain:

```
vips_source_read()
  → vips_source_read_real()
    → read()  ← syscall, returns -1 on ECONNRESET
```

When a network-backed `VipsSource` receives a TCP RST during TIFF strip
loading, `read()` returns `-1`. This value is returned directly from
`openin_source_read()` into `TIFFReadAndRealloc()` in libtiff, triggering
the heap OOB write.

---

## Why libvips is Uniquely Affected

| Consumer    | Read implementation | Can return `-1`? | Natural victim? |
|-------------|---------------------|-----------------|-----------------|
| **libvips** | `read()` syscall    | **YES**         | **YES**         |
| GDAL        | `VSIFReadL()`       | No (`size_t`)   | No              |
| ImageMagick | `fread()`           | No (`size_t`)   | No              |

---

## Attack Scenario

An attacker targets a service that uses libvips for automatic TIFF decoding
(e.g. thumbnail generation, document conversion):

1. Attacker serves a valid TIFF header (IFD with `StripOffsets` pointing
   past the sent data).
2. Attacker resets the TCP connection (RST) before the strip payload is sent.
3. libvips `openin_source_read()` calls `read()` → returns `-1` (ECONNRESET).
4. `-1` propagates into libtiff `TIFFReadAndRealloc()`.
5. `TIFFReadAndRealloc()` computes a negative offset and calls `memset()`
   **one byte before the heap allocation**, corrupting adjacent heap memory.

---

## ASAN Confirmation (Local Reproduction)

Tested with a self-contained C PoC using a local TCP loopback
(no external server required):

```
[read] pos=200   req=2   got=-1 (ECONNRESET -> -1 returned to libtiff!)

=================================================================
ERROR: AddressSanitizer: heap-buffer-overflow on address 0x51900000007f
WRITE of size 1025 at 0x51900000007f thread T0
    #0 __asan_memset
    #1 TIFFReadAndRealloc        libtiff/tif_read.c:133
    #2 TIFFReadRawStripOrTile2   libtiff/tif_read.c:743
    #3 TIFFFillStrip             libtiff/tif_read.c:978
    #4 TIFFReadEncodedStrip      libtiff/tif_read.c:579

0x51900000007f is located 1 bytes before 1024-byte region
  [0x519000000080, 0x519000000480)
allocated by thread T0:
    #0 realloc
    #1 TIFFReadAndRealloc        libtiff/tif_read.c:108
```

**Environment:**

| Field        | Value                                            |
|--------------|--------------------------------------------------|
| libvips      | 8.15.1                                           |
| libtiff      | upstream master `a94d1ded` (also reproduced on system libtiff 4.5.1 via LD_PRELOAD) |
| OS           | WSL Linux / x86\_64                              |
| Build flags  | `gcc -fsanitize=address,undefined`               |

---

## Recommended Fix for libvips

**Option A — Guard in `openin_source_read()` (immediate):**

```c
static tmsize_t
openin_source_read( thandle_t st, void *data, tmsize_t size )
{
    ...
    bytes_read = vips_source_read( source, data, size );
+   if( bytes_read < 0 )
+       bytes_read = 0;   /* never return negative to libtiff */
    return (tmsize_t) bytes_read;
}
```

**Option B — Update libtiff dependency (long term):**

Once the fix lands in libtiff upstream (issue #854), update the libtiff
dependency to a version that includes the fix in `TIFFReadAndRealloc()`.

After the guard is applied, observed behavior becomes a clean error:

```
TIFFFillStrip: Read error at scanline 0; got 0 bytes, expected 2.
```

No ASAN `heap-buffer-overflow` observed.

[poc_natural_trigger_TIFFReadAndRealloc_OOB.zip](https://github.com/user-attachments/files/28994172/poc_natural_trigger_TIFFReadAndRealloc_OOB.zip)

---

## Attachment

`poc_natural_trigger_TIFFReadAndRealloc_OOB.zip` contains:
[](url)

| File                          | Description                        |
|-------------------------------|------------------------------------|
| `natural_trigger_local_tcp.c` | Self-contained PoC (single C file) |
| `asan_output.txt`             | Full captured ASAN output          |
| `README.txt`                  | Build and run instructions         |

jcupitt:
libtiff read handlers can't return error codes, so squash any read errors into EOF (0 byte read).

Thanks aleens-labs

closes #5089

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,7 @@
 date tbd 8.18.4
 
 - heifload: raise security limits on references [caseywebdev]
+- tiffload: squash out read errors in openin_source_read [aleens-labs]
 
 6/6/26 8.18.3
 
--- a/libvips/foreign/tiff.c
+++ b/libvips/foreign/tiff.c
@@ -112,9 +112,10 @@ openin_source_read(thandle_t st, tdata_t data, tsize_t size)
 		gint64 bytes_read;
 
 		bytes_read = vips_source_read(source, data, size - total_read);
-		if (bytes_read == -1)
-			return -1;
-		if (bytes_read == 0)
+		/* libtiff does not allow error returns from read, so any -1s need to
+		 * be squashed out as EOF.
+		 */
+		if (bytes_read <= 0)
 			break;
 
 		total_read += bytes_read;


kleisauke (approved):

aleens-labs:
Thank you @jcupitt and @kleisauke — great to see this merged so quickly!

For context, the root cause is also in libtiff itself: `TIFFReadAndRealloc()` 
in `tif_read.c:133` does not validate negative return values before using them 
in pointer arithmetic and `memset()`. I reported it upstream at 
https://gitlab.com/libtiff/libtiff/-/work_items/854 — contributor 4augustus 
has opened MR !901 there as well.

Do you plan to request a GitHub Security Advisory (GHSA) or CVE for the 
libvips side? Happy to assist with coordination if needed.

jcupitt:
> Do you plan to request a GitHub Security Advisory (GHSA) or CVE for the libvips side?

I would think probably not. This is a relatively minor error and it's not clear if it's even a DoS problem.

I'm very happy to be overruled if someone can demonstrate that it can be exploited in some way, of course.

aleens-labs:
Thank you for reviewing and merging the fix.

I understand your position regarding GHSA/CVE for the libvips side, and I agree that the root cause is in libtiff rather than libvips.

The reason I considered this security-relevant is that ASAN confirms a heap out-of-bounds write: memset() writes 1025 bytes starting 1 byte before the allocated heap region. Without ASAN, this may corrupt heap metadata or adjacent allocations and may lead to a crash depending on allocator behavior and heap layout.

I am not claiming code execution from the libvips side. My plan is to request the CVE against libtiff itself, where the vulnerable TIFFReadAndRealloc() logic exists. I will reference this libvips fix only as downstream/consumer hardening evidence.

Thanks again for the quick review and fix.

PR #5091, opened 2026-06-16
Disable CMYK in jxlsave

DarthSim:
Hey there 👋

I faced an issue where vips can't save CMYK images as JPEG XL. I tried to fix this (https://github.com/DarthSim/libvips/commit/f242c745bd0aa3590205d2798d15d317cdd9b65c), but there are a few problems:

1. The colors are off. I tried multiple tools that support exporting CMYK JPEG XL, and all of them produce the same result. So it's not an implementation problem.
2. CMYK + alpha is saved incorrectly.
3. Lossless encoding doesn't work at all, producing a broken image.

I had to admit that CMYK support in JPEG XL is not mature enough. So I added code to convert CMYK images to sRGB.

### DIFF
--- a/libvips/foreign/jxlsave.c
+++ b/libvips/foreign/jxlsave.c
@@ -946,7 +946,7 @@ vips_foreign_save_jxl_build(VipsObject *object)
 {
 	VipsForeignSave *save = (VipsForeignSave *) object;
 	VipsForeignSaveJxl *jxl = (VipsForeignSaveJxl *) object;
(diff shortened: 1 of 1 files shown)


jcupitt (approved):

jcupitt:
Sure, looks great!

I also fought libjxl CMYK a few years ago and came away bruised :( This is probably the best solution for now.

kleisauke (approved):
LGTM! For reference, PR #3883 is John's earlier attempt.

PR #5092, opened 2026-06-17
Assorted ThreadSanitizer fixes

kleisauke:

Original changeset: https://github.com/libvips/libvips/compare/master...kleisauke:tsan-fixes-original.
Benchmark results: https://github.com/kleisauke/vips-microbench/commit/be59e7f99382b5a88861c7fd1c9b50f6fabb4911 (no measurable performance impact).
Context: https://github.com/libvips/libvips/pull/4981#issuecomment-4160767512.

### DIFF
--- a/libvips/arithmetic/avg.c
+++ b/libvips/arithmetic/avg.c
@@ -158,7 +158,7 @@ vips_avg_stop(VipsStatistic *statistic, void *seq)
  */
 static int
 vips_avg_scan(VipsStatistic *statistic, void *seq,
-	int x, int y, void *in, int n)
+	int x, int y, void *in, int n, gboolean *stop)
 {
 	const int sz = n * vips_image_get_bands(statistic->ready);
 
--- a/libvips/arithmetic/deviate.c
+++ b/libvips/arithmetic/deviate.c
@@ -164,7 +164,7 @@ vips_deviate_stop(VipsStatistic *statistic, void *seq)
 
 static int
 vips_deviate_scan(VipsStatistic *statistic, void *seq,
-	int x, int y, void *in, int n)
+	int x, int y, void *in, int n, gboolean *stop)
 {
 	const int sz = n * vips_image_get_bands(statistic->ready);
 
--- a/libvips/arithmetic/hist_find.c
+++ b/libvips/arithmetic/hist_find.c
@@ -342,7 +342,7 @@ vips_hist_find_stop(VipsStatistic *statistic, void *seq)
 
 static int
 vips_hist_find_scan(VipsStatistic *statistic, void *seq,
-	int x, int y, void *in, int n)
+	int x, int y, void *in, int n, gboolean *stop)
 {
 	VipsHistFind *hist_find = (VipsHistFind *) statistic;
 	Histogram *hist = (Histogram *) seq;
--- a/libvips/arithmetic/hist_find_indexed.c
+++ b/libvips/arithmetic/hist_find_indexed.c
@@ -424,7 +424,7 @@ typedef void (*VipsHistFindIndexedScanFn)(VipsHistFindIndexed *indexed,
 
 static int
 vips_hist_find_indexed_scan(VipsStatistic *statistic, void *seq,
-	int x, int y, void *in, int n)
+	int x, int y, void *in, int n, gboolean *stop)
 {
 	Histogram *hist = (Histogram *) seq;
 	VipsHistFindIndexed *indexed = (VipsHistFindIndexed *) statistic;
--- a/libvips/arithmetic/hist_find_ndim.c
+++ b/libvips/arithmetic/hist_find_ndim.c
@@ -247,7 +247,7 @@ vips_hist_find_ndim_stop(VipsStatistic *statistic, void *seq)
 
 static int
 vips_hist_find_ndim_scan(VipsStatistic *statistic, void *seq,
-	int x, int y, void *in, int n)
+	int x, int y, void *in, int n, gboolean *stop)
 {
 	Histogram *hist = (Histogram *) seq;
 	VipsHistFindNDim *ndim = (VipsHistFindNDim *) statistic;
--- a/libvips/arithmetic/hough.c
+++ b/libvips/arithmetic/hough.c
@@ -131,7 +131,7 @@ vips_hough_stop(VipsStatistic *statistic, void *seq)
 
 static int
 vips_hough_scan(VipsStatistic *statistic,
-	void *seq, int x, int y, void *in, int n)
+	void *seq, int x, int y, void *in, int n, gboolean *stop)
 {
 	VipsHough *hough = (VipsHough *) statistic;
 	VipsHoughClass *class = VIPS_HOUGH_GET_CLASS(hough);
--- a/libvips/arithmetic/max.c
+++ b/libvips/arithmetic/max.c
@@ -319,7 +319,7 @@ vips_max_stop(VipsStatistic *statistic, void *seq)
 				m = values->value[0]; \
 \
 				if (m >= UPPER) { \
-					statistic->stop = TRUE; \
+					g_atomic_int_set(stop, TRUE); \
 					break; \
 				} \
 			} \
@@ -381,7 +381,7 @@ vips_max_stop(VipsStatistic *statistic, void *seq)
  */
 static int
 vips_max_scan(VipsStatistic *statistic, void *seq,
-	int x, int y, void *in, int n)
+	int x, int y, void *in, int n, gboolean *stop)
 {
 	VipsValues *values = (VipsValues *) seq;
 	const int bands = vips_image_get_bands(statistic->ready);
--- a/libvips/arithmetic/min.c
+++ b/libvips/arithmetic/min.c
@@ -319,7 +319,7 @@ vips_min_stop(VipsStatistic *statistic, void *seq)
 				m = values->value[0]; \
 \
 				if (m <= LOWER) { \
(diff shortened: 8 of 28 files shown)


kleisauke:
This is ready for review now.

> Benchmark results: https://github.com/kleisauke/vips-microbench/commit/be59e7f99382b5a88861c7fd1c9b50f6fabb4911 (no measurable performance impact).

These benchmarks were done on a previous changeset of this PR (i.e. with PR #5097 included), though I don't expect that to negatively affect performance.

kleisauke:
Build notes:

<details>
  <summary>Details</summary>

```bash
export CC="clang"
export CXX="clang++"
export LD="lld"

export TSAN_DSO=$($CC -print-file-name=libclang_rt.tsan.so)
export TSAN_SYMBOLIZER_PATH=$($CC -print-prog-name=llvm-symbolizer)

export CPPFLAGS="-g -fno-omit-frame-pointer -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION"

export TSAN_OPTIONS="suppressions=$PWD/suppressions/tsan.supp:halt_on_error=1:print_suppressions=1:ignore_noninstrumented_modules=1"

export LD_LIBRARY_PATH=$(dirname $TSAN_DSO)

# -Dglib:default_library=static
rm -rf build/
meson setup build --prefix=$HOME/tsan-test -Ddebug=true -Dexamples=false -Dfuzz=true -Dfuzzing_engine=libfuzzer -Dmodules=disabled -Dintrospection=disabled -Db_sanitize=thread -Db_lundef=false \
  --force-fallback-for=glib -Dglib:glib_debug=disabled -Dglib:nls=disabled -Dglib:sysprof=disabled -Dglib:tests=false
meson compile -Cbuild
ldd build/libvips/libvips.so
meson test -Cbuild --timeout-multiplier=0 --print-errorlogs
meson install -Cbuild

export PKG_CONFIG_PATH=$HOME/tsan-test/lib64/pkgconfig${PKG_CONFIG_PATH:+:$PKG_CONFIG_PATH}
export LD_LIBRARY_PATH=$HOME/tsan-test/lib64${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}
export PATH=$HOME/tsan-test/bin${PATH:+:$PATH}
LD_PRELOAD=$TSAN_DSO python3 -m pytest -sv --log-cli-level=WARNING test/test-suite > test-suite.txt 2>&1
```
</details>

jcupitt:
Sorry for being slow on this. 

I tried two benchmarks: the vips-bench repo (IO bound), and a sigma 100 gaussblur on a large jpg (CPU bound) and couldn't detect a speed difference! So that's great.

However this seems to break progress reporting in nip4 and vipsdisp -- try loading a large jpg and saving as png, for example.

I think the problem is moving progress reporting from the main thread to the worker. This is happening:

- the main nip4 thread (the one that must do all gtk calls) attaches signal handlers to the image, then calls "pngsave"
- this call will not return until the save completes, so the main nip4 thread is stuck inside the loop in `vips_threadpool_run()`
- a worker signals progress and this runs `save_options_eval()`
- `save_options_eval()` updates the UI, but of course the UI thread is blocked in libvips

The previous implementation would signal the eval from the threadpool loop when workers sent ticks, so the gtk4 UI thread was not blocked.

I think we probably need to return to this model. It should be possible to make this tsan friendly, you'd think, since the thing that woke up the main thread was done cleanly with a semaphore.

kleisauke:
That sounds bad. It means we may need to revisit commit https://github.com/kleisauke/libvips/commit/e3e80b46c68404f57c1f0ab583792566390f7ced (moving progress reporting off the main thread probably wasn't a good idea for UI applications :sweat_smile:), I'll have a look.

IIRC, the issue with the previous model is that the eval callback could race with the one in `vips_sink_base_progress()`. Moving the progress reporting to the worker would fix this (and would avoid the need for commit https://github.com/kleisauke/libvips/commit/094ed9ea325190277d71dd76a94c071ad55c4f7e and https://github.com/kleisauke/libvips/commit/facb2d5a94cf19cf6c6388cc5bde36049e13a4c0).

kleisauke:
I was able to reproduce it. Commit https://github.com/kleisauke/libvips/commit/e3e80b46c68404f57c1f0ab583792566390f7ced was dropped in favor of https://github.com/kleisauke/libvips/commit/3b0904fb7eb432bae759982b3c82913d57979a3e. Of course, this doesn't fix the issue, but it makes the race condition a bit clearer. 

I'll take another look tomorrow.

kleisauke:
Given that eval callbacks could already be invoked from a non-main thread.
https://github.com/libvips/libvips/blob/411b8da562ab8b1ae21ceba56e1a2a35ce27a4ac/libvips/foreign/cgifsave.c#L597
https://github.com/libvips/libvips/blob/411b8da562ab8b1ae21ceba56e1a2a35ce27a4ac/libvips/foreign/jxlsave.c#L430
https://github.com/libvips/libvips/blob/411b8da562ab8b1ae21ceba56e1a2a35ce27a4ac/libvips/foreign/webpsave.c#L216

I think we need to fix vipsdisp in this case. I've just opened PR https://github.com/libvips/vipsdisp/pull/63 for that, which works with the latest changeset  of this PR (which reintroduces commit https://github.com/kleisauke/libvips/commit/e3e80b46c68404f57c1f0ab583792566390f7ced).

jcupitt:
Ouch! You're right, those non-main eval signals are not correct!

But does this really fix the problem? The main GUI thread will still be trapped in the threadpool loop and it won't see these idle events until that loop returns, when they'll all be processed in a rush.

Maybe a better fix is to give the cgif / jxl / webp savers a way to signal progress to the main thread in the threadpool loop, just like workers do on tile finish? That would keep all signals in the main thread as per the docs, and enable immediate processing of progress events.

kleisauke:
> But does this really fix the problem? The main GUI thread will still be trapped in the threadpool loop and it won't see these idle events until that loop returns, when they'll all be processed in a rush.

I did indeed run into this issue initially (`save_options_eval_idle()` was never called). However, with PR https://github.com/libvips/vipsdisp/pull/63, I also moved the main threadpool save loop to a separate thread (i.e. `g_task_run_in_thread()`), which would fix that (and remove the need to call `process_events()` for GUI updates during `save_options_eval()`).

> Maybe a better fix is to give the cgif / jxl / webp savers a way to signal progress to the main thread in the threadpool loop, just like workers do on tile finish? That would keep all signals in the main thread as per the docs, and enable immediate processing of progress events.

I need to think about this further, but `vips_image_eval()` never had the documented restriction that it may only be called from the main thread. That applied only to `VipsThreadpoolProgressFn`. Also, `vips_threadpool_run()` is unlikely to be used outside libvips, so it is probably(?) still safe to revise this.

kleisauke:
Actually, there wasn't a race in `jxlsave.c`, as that is not called in a `vips_sink_*()` iterator. 
https://github.com/libvips/libvips/blob/7ac5d0eaf2a930e155a7dbcb35caab775fc2814a/libvips/foreign/jxlsave.c#L1286-L1288
(but I think libjxl may still call `get_color_channel_data_at` from a non-main thread)

For context, moving the progress signalling to a worker thread (while serializing it with `->allocate_lock`) would fix two issues:

1. The `cgifsave.c` and `webpsave.c` eval callbacks mentioned above could race with the eval callback in `vips_sink_base_progress()` invoked from the main thread:
   https://github.com/libvips/libvips/blob/7ac5d0eaf2a930e155a7dbcb35caab775fc2814a/suppressions/tsan.supp#L11-L15
2. A data race on the `->processed` field.
   https://github.com/libvips/libvips/blob/7ac5d0eaf2a930e155a7dbcb35caab775fc2814a/suppressions/tsan.supp#L34-L35
   The field is read from the main thread at:
   https://github.com/libvips/libvips/blob/7ac5d0eaf2a930e155a7dbcb35caab775fc2814a/libvips/iofuncs/sink.c#L437
   and updated by a `VipsThreadpoolAllocateFn` running on a worker thread at:
   https://github.com/libvips/libvips/blob/7ac5d0eaf2a930e155a7dbcb35caab775fc2814a/libvips/iofuncs/sink.c#L240

The first issue only affected cases where `vips_image_set_progress()` was enabled and the [eval signal](https://www.libvips.org/API/current/signal.Image.eval.html) was connected via `g_signal_connect()`. The second issue was always present in `sink.c`, `sinkdisc.c` and `sinkmemory.c`.

jcupitt:
Ouch, I still think it's important to keep start/eval/stop signalling to the thread that calls eg. `write_to_file()`. Otherwise bindings will have to handle cross-thread callbacks.

It'd be cleaner to have a way for workers to signal progress to the thread waiting in `_threadpool_run()` by ticking the progress semaphore, and for that to emit the progress signals.

Maybe we should split this PR? Merge as is, then have a followup that tries to put the guarantees about no cross-thread signals back.

kleisauke:
I need to think about the best approach moving forward.

Are there any bindings that do not handle cross-thread callbacks? AFAIK, only php-vips has a quirk related to this, requiring `zend.max_allowed_stack_size=-1` to be set in `php.ini` (see e.g. comment https://github.com/libvips/php-vips/pull/237#issuecomment-1944031826 for context). If there are bindings that do not support this, it would also mean they likely do not support `VipsTargetCustom`, since `vips_target_write()` (and consequently `vips_target_custom_write_real()`) is usually called from a non-main thread as well.

kleisauke:
The last commit from this PR has been split out into #5165.

PR #5093, opened 2026-06-17
threadpool: avoid double check

kleisauke:
Also exit early when `progress()` returns an error, since
`pool->stop` and `pool->error` will be made atomic later.

Split out from #5092.

### DIFF
--- a/libvips/iofuncs/threadpool.c
+++ b/libvips/iofuncs/threadpool.c
@@ -663,13 +663,11 @@ vips_threadpool_run(VipsImage *im,
 
 		VIPS_DEBUG_MSG("vips_threadpool_run: tick\n");
 
-		if (pool->stop ||
-			pool->error)
-			break;
-
 		if (progress &&
-			progress(pool->a))
+			progress(pool->a)) {
 			pool->error = TRUE;
+			break;
+		}
 
 		if (pool->stop ||
 			pool->error)


libvips/iofuncs/threadpool.c
```diff
@@ -668,12 +668,10 @@ vips_threadpool_run(VipsImage *im,
 			break;
 
 		if (progress &&
-			progress(pool->a))
+			progress(pool->a)) {
 			pool->error = TRUE;
-
-		if (pool->stop ||
-			pool->error)
 			break;
```
jcupitt, line 669: I think this means we're not longer testing `->stop` after signalling progress, so shutdown might be slower (it'll wait for one more tick). 

This can happen in eg. nip4 compute or vipsdisp save-as:

- tile completes
- main thread wakes up and tests "stop"
- signals progress
- nip4 GUI sees progress signal and updates UI
- user presses "cancel"
- cancel callback sets stop
- progress callback returns
- libvips should test stop to exit the tile loop
- ... otherwise it won't be tested again until the next tile completes

Maybe only remove the stop test before the progress signal?

```C
        VIPS_DEBUG_MSG("vips_threadpool_run: tick\n");
    
        if (progress &&
            progress(pool->a))
            pool->error = TRUE;

        if (pool->stop ||
            pool->error)
            break;
        
        n_waiting = g_atomic_int_get(&pool->n_waiting);
```

jcupitt (approved):

libvips/iofuncs/threadpool.c
kleisauke: You're right, it seems I removed the wrong if statement. The latest changeset should fix this (and also mentions why we need to exit early).

jcupitt:
Thanks! 

Not that I did any testing or benchmarking :( 

Testing downstream applications will probably need to wait until this set of changes are merged.

kleisauke:
I'll remove "Non-functional change." from the commit message, as `progress()` may now be called even when `->stop` or `->error` is signaled, though this particular issue will be (hopefully) fixed in commit e3e80b46c68404f57c1f0ab583792566390f7ced from PR #5092.

PR #5096, opened 2026-06-17
dzsave: check cancellation on `save->ready`

kleisauke:
Split out from #5092.

Probably also required for PR #5097, but still need to verify this.


jcupitt (approved):

jcupitt:
Deprecating the time struct is great.

I'm a little unsure how these changes will interact with nip4's and vipsdisp's progress and cancellation systems, but we can do any tweaking required later.

PR #5101, opened 2026-06-18
uhdrsave: expose peak-brightness and max-content-boost parameters

dshatz:
Successor to #4954

libultrahdr accepts some arguments which were so far not possible to control through vips.

 - target_display_peak_brightness
 - max_content_boost

### DIFF
--- a/libvips/foreign/uhdrsave.c
+++ b/libvips/foreign/uhdrsave.c
@@ -408,13 +408,13 @@ vips_foreign_save_uhdr_hdr(VipsForeignSaveUhdr *uhdr, VipsImage *image)
 	uhdr_enc_set_gainmap_scale_factor(uhdr->enc, uhdr->gainmap_scale_factor);
 	uhdr_enc_set_using_multi_channel_gainmap(uhdr->enc, 0);
 
-	if (uhdr->target_display_peak_brightness > 0) {
-		uhdr_enc_set_target_display_peak_brightness(uhdr->enc, (float) uhdr->target_display_peak_brightness);
-	}
+	if (uhdr->target_display_peak_brightness > 0)
+		uhdr_enc_set_target_display_peak_brightness(uhdr->enc,
+			(float) uhdr->target_display_peak_brightness);
 
-	if (uhdr->max_content_boost > 0) {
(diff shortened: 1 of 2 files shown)


jcupitt:
Hi @dshatz, thanks for this.

- the parameter names are very long, can we shorten them? perhaps `peak_brightness`, for example?
- we'll need a line in the CHANGELOG too, since this is an API addition

dshatz:
> Hi @dshatz, thanks for this.
> 
>     * the parameter names are very long, can we shorten them? perhaps `peak_brightness`, for example?
> 
>     * we'll need a line in the CHANGELOG too, since this is an API addition

Added a line to changelog, and renamed to `peak_brightness`.

jcupitt (approved):

libvips/foreign/uhdrsave.c
```diff
@@ -406,6 +408,14 @@ vips_foreign_save_uhdr_hdr(VipsForeignSaveUhdr *uhdr, VipsImage *image)
 	uhdr_enc_set_gainmap_scale_factor(uhdr->enc, uhdr->gainmap_scale_factor);
 	uhdr_enc_set_using_multi_channel_gainmap(uhdr->enc, 0);
 
+	if (uhdr->target_display_peak_brightness > 0)
+		uhdr_enc_set_target_display_peak_brightness(uhdr->enc,
+			(float) uhdr->target_display_peak_brightness);
+
+	if (uhdr->max_content_boost > 0)
+		uhdr_enc_set_min_max_content_boost(uhdr->enc, -FLT_MIN,
```
kleisauke, line 416: According to https://github.com/google/libultrahdr/blob/v1.4.0/examples/ultrahdr_app.cpp#L1590, the `min_boost` argument should be `FLT_MIN`.
```suggestion
		uhdr_enc_set_min_max_content_boost(uhdr->enc, FLT_MIN,
```

jcupitt:
Great! Thanks for this @dshatz

libvips/foreign/uhdrsave.c
kleisauke: Ah, forgot to apply this suggestion. PR #5106 will fix that.

Issue #5051, opened 2026-05-18
16-bit HDR PNG to UltraHDR JPEG results in severe colour banding


Auxx:
**Describe the bug**
There is a severe colour banding when converting 16-bit HDR PNG files into UltraHDR JPEG files.

**To Reproduce**
1. Original image can be found here - https://images.hexmode.org/bug-reports/libvips/hdr-jpeg-banding/original.png (please note it's a huge 23MB file).
2. Convert using CLI `$ vips.exe jpegsave original.png jpeg100.jpg --Q 100` (or using any bindings programmatically).
3. The end result can be seen here - https://images.hexmode.org/bug-reports/libvips/hdr-jpeg-banding/jpeg100.jpg
4. Here's an example of the same conversion done through Adobe Lightroom Classic with no banding issue present - https://images.hexmode.org/bug-reports/libvips/hdr-jpeg-banding/lrc.jpg

**Expected behavior**
I expect converted JPEG not to have colour banding and look like the image produced by LrC.

**Actual behavior**
Severe colour banding is present at any JPEG quality level, including Q = 100.

**Environment**
 - OS: any
 - Vips: any

jcupitt:
Hi @Auxx,

libvips 8.19 is adding CICP colour handling, which should improve this sort of conversion. Right now, 8.18 will see your PNG file as a plain RGB16 image (no HDR metadata).

Running:

```
$ vips.exe jpegsave original.png jpeg100.jpg --Q 100
```

Will not save as UltraJPEG, just as plain JPEG (SDR). You need:

```
$ vips uhdrsave original.png x.jpg 
```

To make:

<img width="3840" height="2160" alt="Image" src="https://github.com/user-attachments/assets/45f09f46-d3c5-424b-b355-3f2143f4a6f5" />

I'd guess there's some gamma issue as well, perhaps CICP support will fix this.

jcupitt:
(8.19 is due in a couple of months, I should have said)

Auxx:
@jcupitt OK, will have to wait. Just one note.

> Will not save as UltraJPEG, just as plain JPEG (SDR)

The end result is definitely an HDR image, not SDR. Might not be UltraHDR with a gain map. I guess it's just an 8-bit image with a PQ gamma and that's what causes colour banding.

jcupitt:
Oh, maybe. Your image has a funny ICC profile and some XMP, perhaps there's something going on there.

Auxx:
[...]

PR #5104, opened 2026-06-25
heif: preserve CLLI signaling

gregbenz:
## What

HEIF/AVIF images can carry CLLI (Content Light Level Information), which records MaxCLL and MaxFALL for HDR content.

This PR preserves complete CLLI pairs through HEIF/AVIF load and save, including when ancillary metadata is stripped. libvips copies existing values; it does not synthesize or recompute them.

## Why

CLLI provides content light-level information that displays and tone-mapping systems can use when adapting HDR content. libvips currently loses this signaling during HEIF/AVIF transcoding.

## Details

- Detect the typed libheif CLLI load/save APIs at build time.
- Load and save CLLI when both values are present and within the unsigned 16-bit range.
- Preserve CLLI whether or not the input image has CICP metadata, including with `keep=none`.
- Omit missing, partial, or out-of-range CLLI values.
- Remove stale CLLI after explicit CICP/scRGB color transformations.
- Add no new public option or keep flag.

## Test Plan

- CLLI round-trips with `keep=none`.
- CLLI round-trips when the input has no CICP metadata.
- Missing, partial, and out-of-range CLLI values are not written.
- Explicit CICP/scRGB transformations remove stale CLLI.
- CLLI tests use `xfail` when libheif lacks the required APIs.

Local verification:

- `git diff --check`
- `python3 -m py_compile test/test-suite/test_cicp.py`
- Meson configure and compile
- Targeted CICP/HEIF tests with CLLI-capable libheif
- Manual 50% HDR AVIF resize confirming CLLI survives `keep=none`

### DIFF
--- a/libvips/foreign/heifload.c
+++ b/libvips/foreign/heifload.c
@@ -573,7 +573,7 @@ vips_foreign_load_heif_set_page(VipsForeignLoadHeif *heif,
 }
 
 static gboolean
-vips_foreign_load_heif_get_cicp_current(VipsForeignLoadHeif *heif,
+vips_foreign_load_heif_get_cicp(VipsForeignLoadHeif *heif,
 	VipsHeifCicp *cicp)
 {
 	struct heif_error error;
@@ -597,43 +597,9 @@ vips_foreign_load_heif_get_cicp_current(VipsForeignLoadHeif *heif,
 	return FALSE;
 }
 
-static int
-vips_foreign_load_heif_set_cicp_metadata(VipsForeignLoadHeif *heif,
-	VipsImage *out)
-{
-	VipsHeifCicp cicp;
-
-	if (vips_foreign_load_heif_set_page(heif, heif->page, FALSE))
-		return -1;
-
-	if (vips_foreign_load_heif_get_cicp_current(heif, &cicp)) {
-#ifdef DEBUG
-		printf("\tnclx->color_primaries: %d\n", cicp.colour_primaries);
-		printf("\tnclx->transfer_characteristics: %d\n",
-			cicp.transfer_characteristics);
-		printf("\tnclx->matrix_coefficients: %d\n",
-			cicp.matrix_coefficients);
-		printf("\tnclx->full_range_flag: %d\n", cicp.full_range_flag);
-#endif
-
-		g_info("heifload: setting CICP from nclx");
-
-		vips_image_set_int(out, "cicp-colour-primaries",
-			cicp.colour_primaries);
-		vips_image_set_int(out, "cicp-transfer-characteristics",
-			cicp.transfer_characteristics);
-		vips_image_set_int(out, "cicp-matrix-coefficients",
-			cicp.matrix_coefficients);
-		vips_image_set_int(out, "cicp-full-range-flag",
-			cicp.full_range_flag);
-	}
-
-	return 0;
-}
-
 #ifdef HAVE_HEIF_CONTENT_LIGHT_LEVEL
 static gboolean
-vips_foreign_load_heif_get_clli_current(VipsForeignLoadHeif *heif,
+vips_foreign_load_heif_get_clli(VipsForeignLoadHeif *heif,
 	VipsHeifClli *clli)
 {
 	heif_content_light_level content_light_level;
@@ -649,27 +615,6 @@ vips_foreign_load_heif_get_clli_current(VipsForeignLoadHeif *heif,
 	}
 
 	return FALSE;
-}
-
-static int
-vips_foreign_load_heif_set_clli_metadata(VipsForeignLoadHeif *heif,
-	VipsImage *out)
-{
-	VipsHeifClli clli;
-
-	if (vips_foreign_load_heif_set_page(heif, heif->page, FALSE))
-		return -1;
-
-	if (vips_foreign_load_heif_get_clli_current(heif, &clli)) {
-		g_info("heifload: setting CLLI from content light level");
-
-		vips_image_set_int(out, "clli-max-content-light-level",
-			clli.max_content_light_level);
-		vips_image_set_int(out, "clli-max-frame-average-light-level",
-			clli.max_frame_average_light_level);
-	}
-
-	return 0;
 }
 #endif /*HAVE_HEIF_CONTENT_LIGHT_LEVEL*/
 
@@ -846,19 +791,42 @@ vips_foreign_load_heif_set_header(VipsForeignLoadHeif *heif, VipsImage *out)
 		vips_image_set_blob(out, VIPS_META_ICC_NAME,
 			(VipsCallbackFn) vips_area_free_cb, data, length);
 	}
+
 	/* Always try to fetch the NCLX profile, even when ICC is present.
 	 * CICP takes priority over ICC for colour interpretation.
 	 */
-	if (vips_foreign_load_heif_set_cicp_metadata(heif, out))
-		return -1;
+	VipsHeifCicp cicp;
+	if (vips_foreign_load_heif_get_cicp(heif, &cicp)) {
+#ifdef DEBUG
+		printf("\tnclx->color_primaries: %d\n", cicp.colour_primaries);
+		printf("\tnclx->transfer_characteristics: %d\n",
+			cicp.transfer_characteristics);
+		printf("\tnclx->matrix_coefficients: %d\n",
+			cicp.matrix_coefficients);
+		printf("\tnclx->full_range_flag: %d\n", cicp.full_range_flag);
+#endif
+
+		g_info("heifload: setting CICP from nclx");
+
+		vips_image_set_int(out, "cicp-colour-primaries",
+			cicp.colour_primaries);
+		vips_image_set_int(out, "cicp-transfer-characteristics",
+			cicp.transfer_characteristics);
+		vips_image_set_int(out, "cicp-matrix-coefficients",
+			cicp.matrix_coefficients);
+		vips_image_set_int(out, "cicp-full-range-flag",
+			cicp.full_range_flag);
+	}
 
 #ifdef HAVE_HEIF_CONTENT_LIGHT_LEVEL
-	if (vips_image_get_typeof(out, "cicp-colour-primaries") &&
-		vips_image_get_typeof(out, "cicp-transfer-characteristics") &&
-		vips_image_get_typeof(out, "cicp-matrix-coefficients") &&
-		vips_image_get_typeof(out, "cicp-full-range-flag")) {
-		if (vips_foreign_load_heif_set_clli_metadata(heif, out))
-			return -1;
+	VipsHeifClli clli;
+	if (vips_foreign_load_heif_get_clli(heif, &clli)) {
+		g_info("heifload: setting CLLI from content light level");
+
+		vips_image_set_int(out, "clli-max-content-light-level",
+			clli.max_content_light_level);
+		vips_image_set_int(out, "clli-max-frame-average-light-level",
+			clli.max_frame_average_light_level);
 	}
 #endif
 
--- a/libvips/foreign/heifload.c
+++ b/libvips/foreign/heifload.c
@@ -211,6 +211,20 @@ typedef struct _VipsForeignLoadHeif {
 
 } VipsForeignLoadHeif;
 
+typedef struct {
+	int colour_primaries;
+	int transfer_characteristics;
+	int matrix_coefficients;
+	int full_range_flag;
+} VipsHeifCicp;
+
+#ifdef HAVE_HEIF_CONTENT_LIGHT_LEVEL
+typedef struct {
+	int max_content_light_level;
+	int max_frame_average_light_level;
+} VipsHeifClli;
+#endif
+
 #ifdef HAVE_HEIF_INIT
 static void *
 vips__heif_init_once(void *client)
@@ -558,6 +572,107 @@ vips_foreign_load_heif_set_page(VipsForeignLoadHeif *heif,
 	return 0;
 }
 
+static gboolean
+vips_foreign_load_heif_get_cicp_current(VipsForeignLoadHeif *heif,
+	VipsHeifCicp *cicp)
+{
+	struct heif_error error;
+	struct heif_color_profile_nclx *nclx = NULL;
+
+	error = heif_image_handle_get_nclx_color_profile(heif->handle, &nclx);
+	if (error.code == 0 && nclx) {
+		cicp->colour_primaries = nclx->color_primaries;
+		cicp->transfer_characteristics = nclx->transfer_characteristics;
+		cicp->matrix_coefficients = nclx->matrix_coefficients;
+		cicp->full_range_flag = nclx->full_range_flag;
+
+		heif_nclx_color_profile_free(nclx);
+
+		return TRUE;
+	}
+
+	if (nclx)
+		heif_nclx_color_profile_free(nclx);
+
+	return FALSE;
+}
+
+static int
+vips_foreign_load_heif_set_cicp_metadata(VipsForeignLoadHeif *heif,
+	VipsImage *out)
+{
+	VipsHeifCicp cicp;
+
+	if (vips_foreign_load_heif_set_page(heif, heif->page, FALSE))
+		return -1;
+
+	if (vips_foreign_load_heif_get_cicp_current(heif, &cicp)) {
+#ifdef DEBUG
+		printf("\tnclx->color_primaries: %d\n", cicp.colour_primaries);
+		printf("\tnclx->transfer_characteristics: %d\n",
+			cicp.transfer_characteristics);
+		printf("\tnclx->matrix_coefficients: %d\n",
+			cicp.matrix_coefficients);
+		printf("\tnclx->full_range_flag: %d\n", cicp.full_range_flag);
+#endif
+
+		g_info("heifload: setting CICP from nclx");
+
+		vips_image_set_int(out, "cicp-colour-primaries",
+			cicp.colour_primaries);
+		vips_image_set_int(out, "cicp-transfer-characteristics",
+			cicp.transfer_characteristics);
+		vips_image_set_int(out, "cicp-matrix-coefficients",
+			cicp.matrix_coefficients);
+		vips_image_set_int(out, "cicp-full-range-flag",
+			cicp.full_range_flag);
+	}
+
+	return 0;
+}
+
+#ifdef HAVE_HEIF_CONTENT_LIGHT_LEVEL
+static gboolean
+vips_foreign_load_heif_get_clli_current(VipsForeignLoadHeif *heif,
+	VipsHeifClli *clli)
+{
+	heif_content_light_level content_light_level;
+
+	if (heif_image_handle_get_content_light_level(heif->handle,
+			&content_light_level)) {
+		clli->max_content_light_level =
+			content_light_level.max_content_light_level;
+		clli->max_frame_average_light_level =
+			content_light_level.max_pic_average_light_level;
+
+		return TRUE;
+	}
+
+	return FALSE;
+}
+
+static int
+vips_foreign_load_heif_set_clli_metadata(VipsForeignLoadHeif *heif,
+	VipsImage *out)
+{
+	VipsHeifClli clli;
+
+	if (vips_foreign_load_heif_set_page(heif, heif->page, FALSE))
+		return -1;
+
+	if (vips_foreign_load_heif_get_clli_current(heif, &clli)) {
+		g_info("heifload: setting CLLI from content light level");
+
+		vips_image_set_int(out, "clli-max-content-light-level",
+			clli.max_content_light_level);
+		vips_image_set_int(out, "clli-max-frame-average-light-level",
+			clli.max_frame_average_light_level);
+	}
+
+	return 0;
+}
+#endif /*HAVE_HEIF_CONTENT_LIGHT_LEVEL*/
+
 static int
 vips_foreign_load_heif_set_header(VipsForeignLoadHeif *heif, VipsImage *out)
 {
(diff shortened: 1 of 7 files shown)


gregbenz:
If anyone needs a test image:
[P3 HDR.avif.zip](https://github.com/user-attachments/files/29354238/P3.HDR.avif.zip)

Transcoding this image demonstrates preservation of wide gamut, HDR, and CLLI metadata.

libvips/foreign/heifsave.c
```diff
@@ -557,7 +630,8 @@ vips_foreign_save_heif_build(VipsObject *object)
 #endif
 	gboolean has_alpha;
 
-	if (VIPS_OBJECT_CLASS(vips_foreign_save_heif_parent_class)->build(object))
+	if (vips__foreign_save_build_with_force_keep(object,
+			VIPS_FOREIGN_KEEP_CICP))
```
kleisauke: As noted in https://github.com/WordPress/gutenberg/issues/79559#issuecomment-4808396378, this doesn't make much sense because it would make `VIPS_FOREIGN_KEEP_CICP` redundant.

If we decide that CICP color metadata should always be preserved (I'm not yet convinced that's the right approach), we should remove the `VIPS_FOREIGN_KEEP_CICP` enum value instead. Since this hasn't been released yet, we can still make that change without breaking API/ABI compatibility.

gregbenz: @kleisauke Whether we should always keep CICP is a great question, especially as it has been standard practice to discard ICC profiles in some cases. I'll lay out my rationale here for consideration.

In earlier days of the internet, bandwidth was of course extremely limited and almost no monitors supported a gamut wider than sRGB. Stripping ICC profiles made sense because they were often a significant percentage of the size of downloaded assets, especially small assets like icons, button graphics, etc., and removing them often had no negative impact. Wide-gamut display support arrived much later, and by that point there was already a long precedent for stripping color metadata.

So there is a lot of history behind stripping ICC profiles, but the rationale for it has eroded and the cost of it has grown significantly. Common ICC profiles are still often 500-1000 bytes, but the standard is extremely flexible. Profiles can be much larger, especially with complex LUTs or unusual device profiles. So there is arguably still a reason to consider stripping ICC profiles in some workflows, though removing them can make the image effectively useless in many cases.

CICP is different. The signaling is tiny, roughly 20 bytes in HEIF/AVIF including the colr box that contains it, and fixed-size: color primaries, transfer characteristics, matrix coefficients, and range. This includes:

- Color primaries, such as sRGB/Rec. 709, Display P3, or Rec. 2020. Without this information, a wider-gamut image may be interpreted incorrectly and can look significantly desaturated.
- Transfer characteristics, such as Rec. 709 / Rec. 2020 for SDR, or PQ / HLG for HDR.
- Matrix coefficients, which tell a decoder or color-management system how to interpret the relationship between RGB and luma/chroma representations such as Y'CbCr, ICtCp, or identity RGB.
- Full / limited range signaling.

The loss of any of that information can result in an image showing up with one or more serious errors: HDR shown as SDR, color significantly desaturated, contrast significantly incorrect, etc.

The bottom line is that CICP has effectively no meaningful file-size cost, and there is no privacy consideration in this data, unlike generalized XMP or other descriptive metadata. Removing it can effectively break a large and growing percentage of modern images. I can see no strong argument in favor of removing this compact color signaling when it is needed for proper interpretation of the pixels in the image.

I considered three possible scenarios before I submitted the PR:

1. Developer opts in to keeping color when stripping other metadata, such as a `keep: 'cicp'` flag. Making this new and optional seems likely to cause CICP to be stripped accidentally or without understanding the ramifications.
2. When a developer requests `keep: 'icc'`, interpret this as a request to preserve color and keep CICP as well. This aligns reasonably well with user intent, but the specific language is murky. It also still leaves developers deciding when to deliberately remove information that may be vital to proper color interpretation.
3. Always keep CICP for HEIF/AVIF save, as this PR does. This takes the perspective that CICP is not ancillary descriptive metadata, but compact color signaling needed to interpret the encoded samples. I believe that is the correct model. Without it, the image can be effectively ruined in many cases, and the old file-size rationale for stripping ICC profiles does not apply here.

gregbenz: > As noted in [WordPress/gutenberg#79559 (comment)](https://github.com/WordPress/gutenberg/issues/79559#issuecomment-4808396378), this doesn't make much sense because it would make `VIPS_FOREIGN_KEEP_CICP` redundant.
> 

That makes sense. The helper was intended to keep the change scoped to HEIF/AVIF and avoid changing non-HEIF save behavior, but I agree it creates an awkward semantic where `VIPS_FOREIGN_KEEP_CICP` remains public while `heifsave` forces it internally.

I see two cleaner directions:

1. Keep `VIPS_FOREIGN_KEEP_CICP` as an explicit opt-in and remove the HEIF forced-retention path.
2. Treat CICP as required color signaling rather than strippable metadata, and remove `VIPS_FOREIGN_KEEP_CICP` before release.

My preference is still the second model for CICP, because the NCLX/CICP values are needed to interpret the encoded samples and have negligible size/privacy cost. But I agree the implementation should match that model cleanly rather than making the keep flag redundant for HEIF.

mertalev: Perhaps a useful precedent: FFmpeg does automatically propagate CICP and passing `-map_metadata -1`, which strips all ancillary metadata, does not clear CICP.

I personally think it's okay to always keep CICP signaling, but if we do this then it should probably also apply for the other CICP-compatible savers or it will be very confusing.

gregbenz: @mertalev Are you referring to updates for other file formats like JXL? If so, which formats?

I would not propose to alter how ICC is handled. The size is not trivial like CICP and in some cases can be very significant. Additionally, it is not a small defined set of values. ICC is inherently different in my opinion, given its level of complexity is very high.

My PR is just scoped for AVIF as I wanted to minimize the complexity of the request. I would generally agree that this approach to always preserve CICP/NCLX or similar code point implementations makes sense. Without it, the image will be decoded incorrectly unless it coincidentally is encoded in a way that aligns with the assumptions used when the data is missing (ie, only SDR sRGB images will render properly).

mertalev: I think JXL and PNG are the other two, yes.

gregbenz: @mertalev That aligns with my understanding. I believe the complete list of formats using "CICP" currently supported by libvips include: AVIF (covered here), HEIF (covered here as it uses the same path through libheif), **JXL** (structured JxlColorEncoding - conceptually aligned but not same as NCLX in color box like AVIF, for example includes white point, no full range flag, enums are similar but not exactly the same, etc), and **PNG** (cICP chunk is new and I am not sure how widely supported for decoding at this time). 

I have not evaluated these paths deeply in libvips, so this is my best guess based on a quick review. I believe webP does not support a CICP-like system for enumerating color encoding (similar to JPG - supports ICC or would be implicitly assumed to be sRGB).

It is my understanding that libjxl will always write color values. If you don't provide them, I believe it assumes nonlinear sRGB if the input pixels are integers or linear sRGB for float. So it would explicitly assume the missing values when encoding rather than leave it to the decoder. There is a JXL_COLOR_SPACE_UNKNOWN, though that should only be used if the input is truly unknown, which would leave things to the decoder. A JXL without ICC or JxlColorEncoding would be invalid. So untagged is invalid for JXL - stripping the color info isn't really an option and the output would be more clearly a bug if something like a P3 PQ source is transcoded to sRGB. With a quick review of libvips, it appears that using keep=none will incorrectly mark images as sRGB regardless of actual color - and it appears that CLLI-like data required for signaling peak HDR values for proper tone mapping may be lost (given Safari's clipping of AVIF lacking CLLI, I won't be surprised if this is an issue for JXL, but I have not tested that). 

I think 1-2 follow-on PRs after this one would be nice updates for PNG / JXL pending a more thorough review of gaps. The same logic applies to preserve information critical to proper decoding of color information (though probably not as urgent as the AVIF updates here given limited decoder support with those formats for wide gamut / HDR if my list below is complete).

kleisauke: Good points, thanks. I've just opened PR #5110 to remove the `VIPS_FOREIGN_KEEP_CICP` flag.

gregbenz: Thanks, that direction makes sense to me and aligns with the model I was hoping for: CICP as compact color signaling rather than optional ancillary metadata.

If #5110 lands first, I can rebase this PR and remove the now-unnecessary `VIPS_FOREIGN_KEEP_CICP` / forced-keep plumbing here, while keeping the HEIF/AVIF NCLX and CLLI load/save changes.

Starbix:
Just for completeness: PNG has the same metadata, see [here](https://www.w3.org/TR/png-3/#cLLI-chunk)

gregbenz:
@kleisauke I rebased this after #5110 and narrowed the PR to the HEIF/AVIF load/save path. It no longer adds a public keep flag or generic forced-retention plumbing.

HEIF/AVIF NCLX is color signaling needed to interpret encoded samples. After #5110, CICP is no longer treated as ancillary metadata, so this PR completes the HEIF/AVIF side by loading NCLX into libvips CICP metadata and writing complete CICP metadata back as NCLX on save.

This also preserves CLLI when present and valid. CLLI is only copied when complete CICP/NCLX signaling is present; libvips does not synthesize or recompute it.

The multi-page load behavior is conservative:
- for one selected page, copy that page's CICP/CLLI
- for multiple selected pages, attach CICP/CLLI only when all selected pages agree
- otherwise omit the metadata rather than applying one page's signaling to all pages

This PR also avoids writing invalid identity-matrix CICP with chroma subsampling by forcing 4:4:4 when `cicp-matrix-coefficients` is RGB/identity.

Verification performed locally:
- Meson configure detects libheif 1.23.0 and the CLLI get/set APIs
- Meson compile passes
- `git diff --cached --check` passes
- `test_cicp.py` syntax check passes
- real HDR AVIF resize with `keep=none` strips EXIF/XMP while preserving NCLX and CLLI
- smoke-tested `profile=` ICC plus HDR NCLX
- smoke-tested identity matrix plus requested subsampling, confirming output remains 4:4:4

Broader HDR metadata policy, MDCV/mastering display metadata, and downstream wasm-vips/Gutenberg integration remain follow-up work.

mertalev:
> The multi-page load behavior is conservative:
> * for one selected page, copy that page's CICP/CLLI
> * for multiple selected pages, attach CICP/CLLI only when all selected pages agree
> * otherwise omit the metadata rather than applying one page's signaling to all pages

The behavior for one selected page makes sense, but the check for multiple pages seems a bit non-obvious to me. I'd have guessed that it would use the first selected page. I think this is a similar case to having different ICC profiles for each page, so it'd be good to check how libvips behaves there for multi-page.

gregbenz:
@mertalev  Good point, thanks. I amended this to match the existing HEIF metadata behavior.

The loader now copies CICP/CLLI from the non-thumbnail first selected page, the same page currently used for general metadata/ICC, rather than adding a special all-pages-identical rule for these fields.

I also renamed the related test so it no longer implies the removed all-pages-identical policy.

gregbenz:
Following up now that the PR description has been updated to match the amended implementation. The branch incorporates #5110, CICP/CLLI follows the existing HEIF first-selected-page metadata behavior, and CI is green.

When you have a chance, could you take another look? Please let me know if there are any remaining scope, API, or semantics concerns I should address.

test/test-suite/test_cicp.py
```diff
@@ -27,6 +32,45 @@ def make_cicp_image(r, g, b, primaries=1, transfer=1, mc=0, fmt="uchar"):
...
+    im.set_type(
+        pyvips.GValue.gint_type,
+        "clli-max-frame-average-light-level",
+        max_frame_average_light_level
+    )
+
+    return im
+
+
+def have_libheif_clli():
```
kleisauke: We usually mark the test as `xfail` instead of querying pkg-config, see e.g.:
https://github.com/libvips/libvips/blob/736059b2e8e3ca84012b68a9ffdc3528bdeb1e8f/test/test-suite/test_foreign.py#L346
(in this case you'll probably need `raises=AssertionError` instead of `raises=pyvips.error.Error`)

test/test-suite/test_cicp.py
```diff
@@ -299,6 +356,18 @@ def test_scRGB2CICP_metadata_set(self):
         assert result.get("cicp-matrix-coefficients") == 0
         assert result.get("cicp-full-range-flag") == 1
 
+    @skip_if_no("scRGB2CICP")
+    def test_scRGB2CICP_strips_clli_metadata(self):
+        im = pyvips.Image.black(1, 1, bands=3).copy(interpretation="scrgb") \
+            + [0.5, 0.5, 0.5]
+        add_clli(im)
+        result = im.scRGB2CICP(colour_primaries=PRIMARIES_BT2020,
+                                transfer_characteristics=TRANSFER_PQ)
```
kleisauke: nit: indentation looks incorrect here.
```suggestion
                               transfer_characteristics=TRANSFER_PQ)
```

test/test-suite/test_cicp.py
```diff
@@ -441,8 +510,10 @@ def test_heif_cicp_identity_matrix_saves(self):
         im = make_cicp_image(128, 100, 80,
                              primaries=PRIMARIES_DISPLAY_P3,
                              transfer=TRANSFER_PQ)
-        buf = im.heifsave_buffer(compression="av1")
-        assert len(buf) > 0
+        buf = im.heifsave_buffer(compression="av1",
+                                  subsample_mode="on")
```
kleisauke: Ditto.
```suggestion
                                 subsample_mode="on")
```

test/test-suite/test_cicp.py
```diff
@@ -490,6 +579,20 @@ def test_heif_cicp_matrix_coefficients_preserved(self):
...
         assert out.get("cicp-matrix-coefficients") == 6
 
+    @skip_if_no("heifsave")
+    def test_heif_cicp_matrix_coefficients_preserved_with_subsample(self):
+        im = make_cicp_image(128, 128, 128,
+                             primaries=PRIMARIES_DISPLAY_P3,
+                             transfer=TRANSFER_PQ, mc=6,
+                             width=64, height=64)
+        buf = im.heifsave_buffer(compression="av1",
+                                  subsample_mode="on")
```
kleisauke: Ditto.
```suggestion
                                 subsample_mode="on")
```

test/test-suite/test_cicp.py
```diff
@@ -511,6 +614,99 @@ def test_heif_hdr_nclx_with_icc(self):
...
         assert out.get("cicp-colour-primaries") == PRIMARIES_BT2020
 
+    @skip_if_no("heifsave")
+    def test_heif_hdr_nclx_with_profile_option(self):
+        im = make_cicp_image(128, 128, 128,
+                             primaries=PRIMARIES_BT2020,
+                             transfer=TRANSFER_PQ)
+
+        buf = im.heifsave_buffer(compression="av1",
+                                  profile=SRGB_FILE)
```
kleisauke: Ditto.
```suggestion
                                 profile=SRGB_FILE)
```

meson.build
```diff
@@ -547,6 +547,9 @@ if libheif_dep.found()
     cfg_var.set('HAVE_HEIF_GET_DISABLED_SECURITY_LIMITS', cpp.has_function('heif_get_disabled_security_limits', prefix: '#include <libheif/heif.h>', dependencies: libheif_dep))
     # heif_security_limits.max_total_memory added in 1.20.0
     cfg_var.set('HAVE_HEIF_MAX_TOTAL_MEMORY', cpp.has_member('struct heif_security_limits', 'max_total_memory', prefix: '#include <libheif/heif.h>', dependencies: libheif_dep))
+    cfg_var.set('HAVE_HEIF_CONTENT_LIGHT_LEVEL',
```
kleisauke:
```suggestion
    # heif_image_set_content_light_level added in 1.15.0
    # heif_image_handle_get_content_light_level added in 1.19.0
    cfg_var.set('HAVE_HEIF_CONTENT_LIGHT_LEVEL',
```
See commit https://github.com/strukturag/libheif/commit/80bc6f9631f80d981b5f757c368125cd88e8c5a7 and https://github.com/strukturag/libheif/commit/f45abc8b140c1974a6f85eb88cfda42b7620cc5c for reference.

libvips/foreign/heifsave.c
```diff
@@ -357,6 +411,26 @@ vips_foreign_save_heif_write_page(VipsForeignSaveHeif *heif, int page)
...
+		heif_content_light_level content_light_level;
+
+		if (vips_foreign_save_heif_get_cicp(save->ready,
+				&colour_primaries, &transfer_characteristics,
+				&matrix_coefficients, &full_range_flag) &&
+			vips_foreign_save_heif_get_clli(save->ready,
+				&content_light_level))
+			heif_image_set_content_light_level(heif->img,
+				&content_light_level);
+	}
```
kleisauke: This can be simplified to:
```suggestion
	heif_content_light_level content_light_level;

	if (vips_image_get_typeof(save->ready, "cicp-colour-primaries") &&
		vips_image_get_typeof(save->ready, "cicp-transfer-characteristics") &&
		vips_image_get_typeof(save->ready, "cicp-matrix-coefficients") &&
		vips_image_get_typeof(save->ready, "cicp-full-range-flag") &&
		vips_foreign_save_heif_get_clli(save->ready, &content_light_level))
		heif_image_set_content_light_level(heif->img, &content_light_level);
```

libvips/foreign/heifload.c
```diff
@@ -558,6 +572,107 @@ vips_foreign_load_heif_set_page(VipsForeignLoadHeif *heif,
...
+
+static int
+vips_foreign_load_heif_set_cicp_metadata(VipsForeignLoadHeif *heif,
+	VipsImage *out)
+{
+	VipsHeifCicp cicp;
+
+	if (vips_foreign_load_heif_set_page(heif, heif->page, FALSE))
+		return -1;
+
```
kleisauke: I don't think this call is necessary.
```suggestion
```
Since it's already called here:
https://github.com/libvips/libvips/blob/5375f204ef6a7c8da3aa62ebcb25b27aef1b407d/libvips/foreign/heifload.c#L694-L698

libvips/foreign/heifload.c
```diff
@@ -558,6 +572,107 @@ vips_foreign_load_heif_set_page(VipsForeignLoadHeif *heif,
...
+static int
+vips_foreign_load_heif_set_clli_metadata(VipsForeignLoadHeif *heif,
+	VipsImage *out)
+{
+	VipsHeifClli clli;
+
+	if (vips_foreign_load_heif_set_page(heif, heif->page, FALSE))
+		return -1;
+
```
kleisauke: Ditto (unless I've misunderstood something).
```suggestion
```

libvips/foreign/heifsave.c
kleisauke: Or perhaps just:
```suggestion
	heif_content_light_level content_light_level;

	if (vips_foreign_save_heif_get_clli(save->ready, &content_light_level))
		heif_image_set_content_light_level(heif->img, &content_light_level);
```

(assuming CLLI signaling can be used independently of CICP)

libvips/foreign/heifsave.c
```diff
@@ -326,7 +379,8 @@ vips_foreign_save_heif_write_page(VipsForeignSaveHeif *heif, int page)
 		 * write both colr boxes so the NCLX is preserved. ICC alone
 		 * cannot describe PQ or HLG.
 		 */
-		if (vips_image_get_typeof(save->ready, VIPS_META_ICC_NAME) &&
+		if ((save->profile ||
+				vips_image_get_typeof(save->ready, VIPS_META_ICC_NAME)) &&
```
kleisauke: This change (+ the `test_heif_hdr_nclx_with_profile_option` test) should probably be split into a separate PR to keep this one focused on CLLI preservation only.

kleisauke:
Commit 20443027619160ba4216e1ce0c45d8d1306c9a8a incorporates my feedback. I also adjusted the PR title to remove the "HEIF/AVIF NCLX" part, since that was already preserved and none of the changes in this PR affect it (it was likely a remnant of the earlier changes, before PR #5110 was merged).

Only the `save->profile` change mentioned above should be split out, otherwise, this already looks to be in good shape.

libvips/foreign/heifsave.c
gregbenz: Thanks, agreed. I removed the `save->profile` condition and its corresponding test, leaving this PR focused on CLLI preservation. I also updated the description and added a focused test confirming that CLLI round-trips when the input has no CICP metadata.

I’ll handle the `profile=` NCLX/ICC behavior in a separate follow-up after this PR merges.

kleisauke (approved):
Looks great, thank for doing this work @gregbenz.

I suppose the next step would be to support this in `pngload` and `pngsave` as well, since PNG also supports it via the [`cLLI` chunk](https://www.w3.org/TR/png-3/#cLLI-chunk) (see e.g. the sample image in #5051).

@jcupitt Would you like to review this as well?

gregbenz:
@kleisauke Agreed, sounds good and I'm happy to submit those similar updates for PNG. I'll plan to do that as a separate PR after this is merged, unless you suggest otherwise.

kleisauke:
> I'm happy to submit those similar updates for PNG. I'll plan to do that as a separate PR after this is merged, unless you suggest otherwise.

Sounds good to me. For future PRs, though, please disclose any use of an LLM. It was pretty clear that one was involved here (e.g. `python3 -m py_compile` doesn't make much sense, you should use `python3 -m pytest` instead).
https://github.com/libvips/libvips/blob/master/CONTRIBUTING.md#disclosure

jcupitt:
@kleisauke thanks for pinging me, I had a read and this looks reasonable. Nice work everyone!

I have a more general question though -- is it always correct to remove CICP and CLLI on conversion to and from scRGB?

Looking at other libvips use of `vips_image_remove()` in git master, we have:

```
$ grep vips_image_remove */*.c
colour/CICP2scRGB.c:	vips_image_remove(colour->out, "cicp-colour-primaries");
colour/CICP2scRGB.c:	vips_image_remove(colour->out, "cicp-transfer-characteristics");
colour/CICP2scRGB.c:	vips_image_remove(colour->out, "cicp-matrix-coefficients");
colour/CICP2scRGB.c:	vips_image_remove(colour->out, "cicp-full-range-flag");
colour/profile_load.c:		vips_image_remove(image, VIPS_META_ICC_NAME);
conversion/autorot.c:		(void) vips_image_remove(image, field);
conversion/autorot.c:	(void) vips_image_remove(image, VIPS_META_ORIENTATION);
foreign/foreign.c:		vips_image_remove(image, field);
foreign/foreign.c:			!vips_image_remove(in, VIPS_META_ICC_NAME))
foreign/uhdrsave.c:	vips_image_remove(image, "gainmap");
foreign/uhdrsave.c:	vips_image_remove(image, "gainmap-data");
iofuncs/header.c: * vips_image_remove:
iofuncs/header.c:vips_image_remove(VipsImage *image, const char *name)
iofuncs/header.c:		printf("vips_image_remove: remove \"%s\" on shared image\n", name);
```

`autorot.c` is removal of EXIF orientation to prevent double autorot, which seems fine.

`foreign.c` is the code for `keep` handling.

`uhdrsave.c` is stripping the gainmap before using the base JPEG saver to save the SDR.

This makes CICP handling look like the odd one out. For example, converting an ICC-tagged image to scRGB will not remove the ICC profile.

From memory, we've gone back and forth on this in the past. I think the idea behind the current position (never remove any signalling automatically) is that this is best left to the application layer, since it has much better knowledge of processing intent.

kleisauke:
Great, let's merge!

> I have a more general question though -- is it always correct to remove CICP and CLLI on conversion to and from scRGB?

I think it's probably fine for now. The `CICP2scRGB` operation is only called automatically from `uhdrsave` and `scRGB2CICP` is not used by any loader/saver (at the moment). However, if we start integrating these operations into more conversion paths, we should probably revisit this to avoid running into an issue similar to #4321.

> For example, converting an ICC-tagged image to scRGB will not remove the ICC profile.

I looked into this while reviewing PR #5123. Would it make sense to add logic to `vips_icc_is_compatible_profile()` to handle this? That would allow incompatible ICC profiles to be silently dropped during save.
https://github.com/libvips/libvips/blob/916b968bd993c7f93a4474486bfd5c57995309de/libvips/foreign/foreign.c#L1771-L1785

jcupitt:
> Would it make sense to add logic to `vips_icc_is_compatible_profile()` to handle this?

Oh, maybe. Hmmmm!

Perhaps this should be punted to a follow-up PR.

jcupitt:
I made https://github.com/libvips/libvips/issues/5138 as a reminder to look again at automatic removal of CICP signalling.

gregbenz:
> Is it always correct to remove CICP and CLLI on conversion to and from scRGB?

I believe CICP and CLLI should be considered separately.

CICP describes the color encoding of the pixel values. After conversion to scRGB, the source primaries and transfer characteristics no longer describe the output pixels, so retaining the original CICP tuple unchanged would be incorrect. Internally, the output is identified as scRGB by its libvips interpretation.

When converting from scRGB to an encoded output, CICP should describe the target encoding. Existing values can carry over only when the corresponding primaries, transfer characteristics, matrix and range remain unchanged. The matrix must describe the resulting RGB or YCbCr sample representation; chroma subsampling alone does not necessarily change it.

CLLI is different because it summarizes content luminance for use in tone mapping (a browser could calculate these values and in practice I see that Chrome does well without CLLI, while Safari clips HDR values). A colorimetric CICP-to-scRGB conversion can preserve physical luminance, in which case MaxCLL and MaxFALL may remain valid. Conversely, conversion from scRGB can involve clipping, gamut mapping or other changes that make the original values inaccurate.

So preserving CLLI is sometimes correct and sometimes not (when the content is visibly altered). Ideally, CLLI would be recalculated for the final exported image when luminance has changed. For a straightforward decode and re-encode that preserves the luminance semantics and does not clip or tone map the pixels, carrying it through is reasonable.

AVIF, HEIF and PNG support native CLLI. JPEG XL uses different, analogous tone-mapping fields such as `intensity_target` and `min_nits`, so those would need format-specific handling rather than a direct metadata copy.

Generally, known and valid CLLI should be preserved or written when exporting HDR in a format that supports it. It should not be synthesized when the values are unknown.

PR #5105, opened 2026-06-25
maplut: clamp index in complex LUT path to fix OOB read

sage-mode-hunter:
maplut clamps the lookup index in every code path except the complex-LUT ones for uchar input:
- loopc, loop1c and loop1cm index tlut[p[x] * 2] straight from the pixel with no bound
- the table holds only maplut->clp + 1 entries (taken from the LUT image size), so a LUT shorter than 256 entries plus a pixel value past its length reads off the end of the table
- the matching loops for wider input types (loop1cg etc.) already clamp to clp and bump the overflow counter

Reproduced as an ASAN heap-buffer-overflow read with a 16-entry complex LUT and a uchar value of 200; clamps to the final entry after.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -31,6 +31,7 @@ date-tbd 8.19.0
 - vipsthumbnail: add '@' modifier for size to pixel count
 - dcrawload: add half-size option
 - uhdrsave: expose peak-brightness and max-content-boost parameters [dshatz]
+- maplut: clamp index in complex LUT path
 
 6/6/26 8.18.3
 
--- a/libvips/histogram/maplut.c
+++ b/libvips/histogram/maplut.c
@@ -178,10 +178,15 @@ vips_maplut_start(VipsImage *out, void *a, void *b)
 				OUT *tlut = (OUT *) maplut->table[z]; \
 \
 				for (x = 0; x < ne; x += b) { \
-					int n = p[x] * 2; \
+					unsigned int index = p[x]; \
 \
-					q[0] = tlut[n]; \
-					q[1] = tlut[n + 1]; \
+					if (index > maplut->clp) { \
+						index = maplut->clp; \
+						seq->overflow++; \
+					} \
+\
+					q[0] = tlut[index * 2]; \
+					q[1] = tlut[index * 2 + 1]; \
 					q += b * 2; \
 				} \
 			} \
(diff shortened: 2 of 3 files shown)


kleisauke:
I just opened PR #5117 for this instead, due to the concerns outlined in this comment:
https://github.com/bblanchon/ArduinoJson/pull/2223#issuecomment-4863849515

You'll still be credited in the ChangeLog, but your profile looks very suspicious. I also couldn't find anything related to "CODEOPS47" besides the GitHub profiles for @jmestwa-coder and @metsw24-max, which look the same to me.

kleisauke:
Hmm, I found a few more profiles:
- @naruto-lgtm
- @nvxbug
- @uchiha-bug-hunter
- @netliomax25-code
- @aizu-m
- @rootvector2
- @SABITHSAHEB

They all look very similar to me and may be operated by the same person.

jcupitt:
I agree, let's close this PR in favour of #5117

sage-mode-hunter:
Thanks for the review, and for still crediting the report in the ChangeLog — no problem that you went with #5117.

One thing I'd like to clear up: these accounts aren't a single person or automated. We're a small group of engineers at Digiscrypt Technologies ([digiscrypt.com](https://digiscrypt.com/)). The two names you noticed have a simple explanation — CODEOPS47 is just our internal team/group name, and BugQore is an upcoming project of ours that isn't public yet, so there's not much to find on it externally. That's why our profiles look similar: same team, similar bios, accounts created around the same time. I get how that pattern looked from the outside, and that's on us.

We'll consolidate under a single, verifiable Digiscrypt organization on GitHub going forward so it's clear who we are. Appreciate the work you both put into libvips.

lovell:
This looks exactly like the sort of response an LLM might generate :laughing:

sage-mode-hunter:
Ha, fair enough 😅 English isn't our first language, so we lean on tools to help us sound professional in replies — I guess that's why it comes out a bit robotic. But we're a real small team, just trying to contribute and get better at this. No hard feelings, and thanks again for the fix and the credit 🙏

sage-mode-hunter:
we're happy to hop on a Google Meet whenever suits you, just share a time 🙂 And I'll be honest: yeah, we do use AI/LLMs in our workflow, like a lot of devs do these days. But nothing goes out without a human checking it first — we reproduced and tested this one ourselves before raising it. So please don't judge us too hard 🙏 we're genuinely just trying to do good work.

PR #5109, opened 2026-06-28
enumtypes: drop `G_GNUC_CONST`

kleisauke:
See: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/5223.


lovell (approved):

kleisauke:
Looking at https://gitlab.gnome.org/GNOME/glib/-/merge_requests/5225, I think we should backport this to `8.18` to prevent instances from being unexpectedly optimized away when building with GCC trunk.

/cherry-pick f8be485a187e1bb64902194bed288adfaa22766f

PR #5111, opened 2026-06-29
shrinkv: multiply by 1/vshrink in float FAVG to match shrinkh

dloebl:
Brings `shrinkv`'s float average path in line with the `shrinkh` float path from #5077, which already multiplies by the reciprocal instead of dividing per pixel:

https://github.com/libvips/libvips/blob/f8be485a187e1bb64902194bed288adfaa22766f/libvips/resample/shrinkh.c#L140

`FAVG` now computes `1.0 / vshrink` once and multiplies (applies to FLOAT, DOUBLE, COMPLEX, DPCOMPLEX). Avoids a DIV in the loop.


kleisauke (approved):
Thanks!

PR #5112, opened 2026-06-30
reducev: block the float reduce to auto-vectorise

dloebl:
`reducev_float_tab` summed each output pixel's taps in a separate inner reduction, which can't be auto-vectorised. This blocks 32 contiguous output elements with the tap loop outermost, so the multiply-add auto-vectorises well and ILP (Instruction-level parallelism) is also more effective.
Taps are summed in the same order, so the output is **bit-exact**.

Affects the `FLOAT`/`COMPLEX` reduce path.
~40% faster on a float `resize` (lanczos3), and **~3x faster on the isolated `reducev` pass itself**.

`-march=native -O3` on an Apple M3 Pro (clang 21.0.0)

**Resize**:
```bash
vips gaussnoise float.v 16000 16000   # float image, so resize hits the float reduce

VIPS_CONCURRENCY=1 /usr/bin/time ./build-reducev-baseline/tools/vips resize float.v out_base.v 0.5
> 1.06 real         0.98 user         0.12 sys

VIPS_CONCURRENCY=1 /usr/bin/time ./build-reducev-branch/tools/vips resize float.v out_block.v 0.5
> 0.64 real         0.56 user         0.12 sys

sha256sum out_base.v out_block.v 
> 1d8cbf8bc662bf0ecca5ffa655af0db166d817823080ae3257fa1964649657d7  out_base.v
> 1d8cbf8bc662bf0ecca5ffa655af0db166d817823080ae3257fa1964649657d7  out_block.v
```

**Isolated `reducev`**:
```
VIPS_CONCURRENCY=1 /usr/bin/time vips reducev float.v out.v 3.0

              parent    branch   speedup
FLOAT         0.58s     0.19s    ~3.0x
```

Found with https://github.com/davebcn87/pi-autoresearch running Claude Opus 4.8. Manually verified and tuned.

### DIFF
--- a/libvips/resample/reducev.cpp
+++ b/libvips/resample/reducev.cpp
@@ -458,26 +458,28 @@ static void inline reducev_float_tab(VipsReducev *reducev,
 	const int l1 = lskip / sizeof(T);
 
 	/* Accumulate a block of output elements at once with the tap loop
-	 * innermost. Consecutive output elements are contiguous within each
-	 * input line, so the inner multiply-add over the block auto-vectorises.
-	 * Each element still sums its taps in the same order as the
-	 * reduce_sum() below, so the result is unchanged.
+	 * outermost. Consecutive output elements are contiguous within each
+	 * input line, so the inner multiply-add over the block can auto-vectorise,
+	 * and the per-element accumulators form independent dependency chains
+	 * for better instruction-level parallelism (ILP). Each element still
+	 * sums its taps in the same order as the reduce_sum() below, so the
+	 * result is unchanged.
 	 */
-	constexpr int block = 32;
+	constexpr int block_size = 32; // multiple of common SIMD widths, tuned
 
 	int z = 0;
-	for (; z + block <= ne; z += block) {
+	for (; z + block_size <= ne; z += block_size) {
 		/* One accumulator per output element in the block, zeroed for
 		 * each new block.
 		 */
-		double sum[block] = {};
+		double sum[block_size] = {};
 		for (int i = 0; i < n; i++) {
 			const double c = cy[i];
 			const T *restrict p = in + z + i * l1;
-			for (int k = 0; k < block; k++)
+			for (int k = 0; k < block_size; k++)
 				sum[k] += c * p[k];
 		}
-		for (int k = 0; k < block; k++)
+		for (int k = 0; k < block_size; k++)
 			out[z + k] = sum[k];
 	}
 
--- a/libvips/resample/reducev.cpp
+++ b/libvips/resample/reducev.cpp
@@ -495,8 +495,7 @@ static void inline reducev_float_tab(VipsReducev *reducev,
 
 	/* Tail: fewer than a full block.
 	 */
-	if (z < ne)
-		reducev_block<T>(out + z, in + z, ne - z, lskip, cy, n);
+	reducev_block<T>(out + z, in + z, ne - z, lskip, cy, n);
 }
 
 /* Ultra-high-quality version for double images.
--- a/libvips/resample/reducev.cpp
+++ b/libvips/resample/reducev.cpp
@@ -460,6 +460,8 @@ static void inline reduce_sum_array(T *restrict out, const T *restrict in,
 	int width, int lskip, const CT *restrict cy, int n)
 {
 	const int l1 = lskip / sizeof(T);
+
+	g_assert(width <= REDUCEV_BLOCK);
 	IT sum[REDUCEV_BLOCK] = { 0.0 };
 
 	for (int i = 0; i < n; i++) {
--- a/libvips/resample/reducev.cpp
+++ b/libvips/resample/reducev.cpp
@@ -488,12 +488,12 @@ static void inline reducev_float_tab(VipsReducev *reducev,
 
 	int z;
 
-	/* Complete blocks. Passing the constant block size lets this vectorise.
+	/* do as many complete blocks as we can
 	 */
 	for (z = 0; z + REDUCEV_BLOCK <= ne; z += REDUCEV_BLOCK)
 		reducev_block<T>(out + z, in + z, REDUCEV_BLOCK, lskip, cy, n);
 
-	/* Tail: fewer than a full block.
+	/* do the tail (fewer than a full block)
 	 */
 	reducev_block<T>(out + z, in + z, ne - z, lskip, cy, n);
 }
--- a/libvips/resample/reducev.cpp
+++ b/libvips/resample/reducev.cpp
@@ -445,6 +445,34 @@ static void inline reducev_signed_int_tab(VipsReducev *reducev,
 	}
 }
 
+/* Reduce a block of "width" output elements at once (width <= REDUCEV_BLOCK),
+ * with the tap loop outermost and the element loop innermost. Consecutive
+ * output elements are contiguous within each input line, so the inner
+ * multiply-add vectorises, and the per-element accumulators form independent
+ * dependency chains for better instruction-level parallelism (ILP). Each
+ * element sums its taps in the same order as reduce_sum(), so the result is
+ * identical.
+ */
+static constexpr int REDUCEV_BLOCK = 32; // multiple of common SIMD widths, tuned
+
+template <typename T, typename CT, typename IT = typename LongT<T>::type>
+static void inline reduce_sum_array(T *restrict out, const T *restrict in,
+	int width, int lskip, const CT *restrict cy, int n)
+{
+	const int l1 = lskip / sizeof(T);
+	IT sum[REDUCEV_BLOCK] = { 0.0 };
+
+	for (int i = 0; i < n; i++) {
+		const CT c = cy[i];
+		const T *restrict p = in + i * l1;
+		for (int k = 0; k < width; k++)
+			sum[k] += c * p[k];
+	}
+
+	for (int k = 0; k < width; k++)
+		out[k] = sum[k];
+}
+
 /* Floating-point version.
  */
 template <typename T>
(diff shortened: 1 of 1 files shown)


dloebl:
**For reviewers:** I aim to get feedback on this PR first and, once it's merged, get #5114 ready for review as well - it's the generalized version, which shows very promising results for other band formats too.

jcupitt:
Hi @dloebl, I'm also excited by this PR and keen to merge it! I won't get a chance to review it until weds though.

libvips/resample/reducev.cpp
```diff
@@ -457,7 +457,35 @@ static void inline reducev_float_tab(VipsReducev *reducev,
...
+	 * result is unchanged.
+	 */
+	constexpr int block_size = 32; // multiple of common SIMD widths, tuned
+
+	int z = 0;
+	for (; z + block_size <= ne; z += block_size) {
+		/* One accumulator per output element in the block, zeroed for
+		 * each new block.
+		 */
+		double sum[block_size] = {};
```
jcupitt: This is just the float and complex path, so we could use `float` here rather than `double`. If you change `c` below as well, I see a nice almost x2 speedup.

Master:

```
$ VIPS_CONCURRENCY=1 /usr/bin/time vips reducev x.v x2.v 2
0.89user 0.27system 0:01.17elapsed 99%CPU (0avgtext+0avgdata 1162932maxresident)k
40inputs+1029328outputs (0major+27773minor)pagefaults 0swaps
```

With double:

```
$ VIPS_CONCURRENCY=1 /usr/bin/time vips reducev x.v x2.v 2
0.45user 0.28system 0:00.97elapsed 75%CPU (0avgtext+0avgdata 1163504maxresident)k
48inputs+1029328outputs (0major+27772minor)pagefaults 0swaps
```

With float for sum and c:

```
$ VIPS_CONCURRENCY=1 /usr/bin/time vips reducev x.v x2.v 2
0.26user 0.27system 0:00.77elapsed 70%CPU (0avgtext+0avgdata 1163320maxresident)k
32inputs+1029328outputs (0major+27774minor)pagefaults 0swaps
```

jcupitt:
I found some time heh. This is great and, as you say, it generalises really easily.

Have you had a look at reduceh? That's the harder case.

libvips/resample/reducev.cpp
kleisauke: Unfortunately, I think this needs to stay `double`, see e.g.:
https://github.com/libvips/libvips/blob/f8be485a187e1bb64902194bed288adfaa22766f/libvips/resample/templates.h#L548-L553

libvips/resample/reducev.cpp
```diff
@@ -457,7 +457,35 @@ static void inline reducev_float_tab(VipsReducev *reducev,
...
+				sum[k] += c * p[k];
+		}
+		for (int k = 0; k < block_size; k++)
+			out[z + k] = sum[k];
+	}
+
+	/* Process the remaining tail elements (fewer than a full block).
+	 */
+	for (; z < ne; z++)
 		out[z] = reduce_sum<T>(in + z, l1, cy, n);
```
kleisauke: Note to self: inlining `reduce_sum()` here doesn't seem to make any difference, see e.g.:
https://godbolt.org/z/5EvT6hfzb

kleisauke:
This is great! Here's a Compiler Explorer link that shows how GCC vectorizes it:
https://godbolt.org/z/6n8hczToa

libvips/resample/reducev.cpp
jcupitt: You're right that it'll no longer be bit-identical, but I think the delta should be small and bounded, shouldn't it? This change would sum 32 floats into a float, so I suppose max error would be 5 / 2 bits of mantissa of the 24 (is that right?).

jcupitt:
`reduce_sum()` is no longer a good primitive for this, since it does a complete convolution for a single element. We need something that does a batch of pixels, with the rotated inner loop to help the vectoriser.

How about something like:

```C++
template <typename T, typename CT, typename IT = typename LongT<T>::type>
static void inline reduce_sum_array(T *out, const T *in,
    int block_size, int lskip, const CT *cy, int n)
{
    IT restrict sum[block_size] = { 0.0 };

    for (int i = 0; i < n; i++) {
        const CT c = cy[i];
        const T *restrict p = in + i * (lskip / sizeof(T));

        for (int k = 0; k < block_size; k++)
            sum[k] += c * p[k];
    }

    for (int k = 0; k < block_size; k++)
        out[k] = sum[k];
}
```

(we need a better name for this func!)

So that's doing `block_size` pixels at once in a vector-friendly way. Now you can write `reducev_float_tab()` as:

```C++
template <typename T>
static void inline reducev_float_tab(VipsReducev *reducev,
    VipsPel *pout, const VipsPel *pin,
    const int ne, const int lskip, const double *cy)
{
    T *out = (T *) pout;
    const T *in = (T *) pin;
    const int n = reducev->n_point;

    constexpr int block_size = 32; // multiple of common SIMD widths, tuned

    int z;

    // do as many complete blocks as we can
    for (z = 0; z + block_size <= ne; z += block_size)
        reduce_sum_array<T>(out + z, in + z, block_size, lskip, cy, n);

    // do the tail
    reduce_sum_array<T>(out + z, in + z, ne - z, lskip, cy, n);
}
```

(edit: oops, removed `for` from tail)

Which looks a bit neater and avoids different processing for the body and the tail. Because we're passing a const for `block_size` to the first instantiation, it'll vectorise.

You can also use `T` for the intermediates, eg.:

```C++
template <typename T, typename CT>
static void inline reduce_sum_array(T *restrict out, const T *restrict in,
    int block_size, int lskip, const CT *restrict cy, int n)
{   
    for (int k = 0; k < block_size; k++)
        out[k] = 0.0;
    
    for (int i = 0; i < n; i++) {
        const T c = cy[i];
        const T *restrict p = in + i * (lskip / sizeof(T));
        
        for (int k = 0; k < block_size; k++)
            out[k] += c * p[k];
    }
}
```

Which is 2x faster again, but of course drops precision as noted.

libvips/resample/reducev.cpp
dloebl: Good point. Yes, exactly - I've tried to keep it bit-exact for this PR and the generic version. Maybe we can try the `double` to `float` change in another follow-up PR?

libvips/resample/reducev.cpp
```diff
@@ -445,6 +445,34 @@ static void inline reducev_signed_int_tab(VipsReducev *reducev,
...
+ * identical.
+ */
+static constexpr int REDUCEV_BLOCK = 32; // multiple of common SIMD widths, tuned
+
+template <typename T, typename CT, typename IT = typename LongT<T>::type>
+static void inline reduce_sum_array(T *restrict out, const T *restrict in,
+	int width, int lskip, const CT *restrict cy, int n)
+{
+	const int l1 = lskip / sizeof(T);
+	IT sum[REDUCEV_BLOCK] = { 0.0 };
```
jcupitt, line 465: How about using `width` here? Otherwise callers could (potentially) set `width` to a larger value and write over the end of the array.

libvips/resample/reducev.cpp
```diff
@@ -445,6 +445,34 @@ static void inline reducev_signed_int_tab(VipsReducev *reducev,
...
+ * output elements are contiguous within each input line, so the inner
+ * multiply-add vectorises, and the per-element accumulators form independent
+ * dependency chains for better instruction-level parallelism (ILP). Each
+ * element sums its taps in the same order as reduce_sum(), so the result is
+ * identical.
+ */
+static constexpr int REDUCEV_BLOCK = 32; // multiple of common SIMD widths, tuned
+
+template <typename T, typename CT, typename IT = typename LongT<T>::type>
+static void inline reduce_sum_array(T *restrict out, const T *restrict in,
```
jcupitt: I don't like the name `reduce_sum_array()` :( 

Maybe `reducev_array()`? What do you think?

libvips/resample/reducev.cpp
dloebl: This would change it to a VLA. I see:
`../libvips/resample/reducev.cpp:463:9: error: variable-sized object may not be initialized
  463 |         IT sum[width] = { 0.0 };`

`warning: variable length arrays in C++ are a Clang extension [-Wvla-cxx-extension]`

And I'm also not sure how that would interact with auto-vectorization. How about we add an `assert` instead that's only active in debug builds? 

Something like:
```C
g_assert(width <= REDUCEV_BLOCK);
```

libvips/resample/reducev.cpp
jcupitt: OK, let's stick with double for now.

libvips/resample/reducev.cpp
dloebl: I've added an assert with 762746a730d5aaf614e2cbe7ac8fa707c7ef9b76

jcupitt: Oh, strange, it's working fine here with gcc, and with no warnings.

Sure, an assert is fine too.

libvips/resample/reducev.cpp
dloebl: Maybe `reducev_block` is a good fit? I've renamed it with 1aed00e314f092056278e9c1ecdfc3eb862851e9

jcupitt:
Oh haha try changing this to:

```C++
template <typename T>
static void inline reducev_float_tab(VipsReducev *reducev,
    VipsPel *pout, const VipsPel *pin,
    const int ne, const int lskip, const double *cy)
{
    T *out = (T *) pout; 
    const T *in = (T *) pin;
    const int n = reducev->n_point;

    reducev_arrayf<T>(out, in, ne, lskip, cy, n);
}
```

ie. no stepping in 32-pixel units. It still vectorises, at least with gcc.

edit: though that's obvious, now I think about it. Working in blocks is still a good idea since it avoids a VLA and keeps stack use down. But the comment about a fixed block size helping the vectoriser should probably go.

libvips/resample/reducev.cpp
kleisauke: FWIW, here's a [`ResampleScope`](https://github.com/jsummers/resamplescope) comparison of `double` versus `float` precision.

| `double` | `float` |
| :---: |  :---: |
|  |  |
| https://github.com/kleisauke/resize-comparison/commit/56478f1eb051e4167b06fc5298022622870e45f0 | https://github.com/kleisauke/resize-comparison/commit/328462467a177faf29ff51fcdb77ae196e756806 |

The `double` implementation produces a noticeably cleaner result, while `float` introduces visible precision artifacts. Based on this comparison, I think `double` is the better choice here.

dloebl:
> no stepping in 32-pixel units. It still vectorises, at least with gcc

Good point. It's not a requirement for auto-vectorization. I've updated the comments.
Though comparing the `REDUCEV_BLOCK` approach to processing everything at once, `reducev` is still ~1.5-1.7 times faster for me (Apple M3 Pro) - I guess that's because a small block can fit into registers (block size is known at compile time)?

> Have you had a look at reduceh? That's the harder case.

I've briefly looked at it, but it does seem to be trickier. Maybe something for a follow-up too 😅

libvips/resample/reducev.cpp
jcupitt: Oh, interesting! But what's the vertical scale?

If I reducev two 0-255 value images, one with float and one with double, I see a max difference of 0.000061, or about 1 part in 4m. 

That's 22 bits of mantissa, so we've only lost 2 bits of 24 by using float, roughly what you'd expect from summing 32 floats (32 is 2 ** 5, so max error of 5 / 2 bits).

jcupitt (approved):

jcupitt:
> Though comparing the REDUCEV_BLOCK approach to processing everything at once, reducev is still ~1.5-1.7 times faster for me (Apple M3 Pro) - I guess that's because a small block can fit into registers (block size is known at compile time)?

OK, that's another good reason to keep the block.

The two approaches run at the same speed when compiling with gcc on my AMD x64 CPU.

jcupitt:
Ooop, I forgot to say that a note in the ChanegLog would be useful, since this is such a nice improvement.

I think you are planning a followup which generalises this, so perhaps we can add the note then.

dloebl:
> I forgot to say that a note in the ChanegLog would be useful, since this is such a nice improvement.
I think you are planning a followup which generalises this, so perhaps we can add the note then.

Sounds good! I'll add a ChangeLog entry to the generic version in #5114

jcupitt:
\o/

PR #5115, opened 2026-07-03
project: add "combine" option

### LINKED ISSUE #5113: vips_project: add combine option
Add a `VipsCombine` option to `vips_project()` to allow max / min / sum of rows.

### Discussed in https://github.com/libvips/libvips/discussions/2017

<sup>Originally posted by **zoranpopovic** September  2, 2018</sup>
I'm trying to compute max intensity for each row of a large image file.  right now I do it with a rank full-width rank filter over the whole image which is wasteful. 
```
img = img.rank(img.width, 1, img.width-1)
col_mip = img.extract_area(img.width/2, 0, 1, img.height)
```

Any idea how to do this more efficiently.</div>

jcupitt:
To compute max or min projection, as well as sum.

Closes: https://github.com/libvips/libvips/issues/5113

### DIFF
--- a/libvips/arithmetic/project.c
+++ b/libvips/arithmetic/project.c
@@ -58,7 +58,7 @@ typedef struct {
 	 */
 	void *columns;
 
-	/* Vertical array: conbination of all rows.
+	/* Vertical array: combination of all rows.
 	 */
 	void *rows;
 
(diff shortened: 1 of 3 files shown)


libvips/arithmetic/project.c
```diff
@@ -277,25 +326,25 @@ vips_project_stop(VipsStatistic *statistic, void *seq)
 	int hsz = in->Xsize * in->Bands;
 	int vsz = in->Ysize * in->Bands;
 
+	// I think this is always true
+	g_assert(sub_hist->init);
```
kleisauke: CIFuzz seems to fail on this assertion, but I've not investigated it further.

jcupitt: Ooop yes I was thinking about this in the shower this morning. I think the assumption is wrong, I'll fix it.

kleisauke (approved):
Nice!

jcupitt:
\o/ thanks for the review!

PR #5118, opened 2026-07-05
Add apngsave

jcupitt:
Add APNG save.

- [x] look at compression settings (APNG has some new ones)
- [x] use vips_image_get_loop() everywhere

Derived from #4944


jcupitt:
APNG save will follow in a second PR once this is merged.

Tests will go in once APNG load is in, since they test round-trips.

Test against the libpng 1.8 branch here: https://github.com/pnggroup/libpng/tree/libpng18 (this will become libpng 2.0, probably later this year)

A web browser (eg. firefox) is convenient for testing the output of this PR.

Run eg.:

```
$ vips copy dancing-banana.gif[n=-1] x.png
```

To make:

x

(this ^^^ is a static image if your browser does not have apng support)

kleisauke (approved):
Nice!

PR #5122, opened 2026-07-11
Add apngload

jcupitt:
Add APNG load.

Derived from #4944


jcupitt:
It all seems to work, and all tests pass. Here's another handy test APNG:

numbers

(from https://github.com/SixLabors/ImageSharp/issues/2708)

Test against the libpng 1.8 branch here: https://github.com/pnggroup/libpng/tree/libpng18 (this will become libpng 2.0, probably later this year).

Run with eg.:

```
$ vips copy numbers.png[n=-1] x.v
$ vipsdisp x.v
```

Test two-band (GA) animated PNG:

```
$ vips extract_band numbers.png[n=-1] alpha.v 3
$ vips extract_band numbers.png[n=-1] r.v 0
$ vips bandjoin "r.v alpha.v" x.png
```

Test palette animated PNG:

```
$ vips copy numbers.png[n=-1] x.png[palette]
```

You can also add PNG to the list of animated formats in vipsdisp and use vipsdisp directly with:

```patch
diff --git a/src/tilesource.c b/src/tilesource.c
index 5f7d3fe..f0d513a 100644
--- a/src/tilesource.c
+++ b/src/tilesource.c
@@ -212,6 +212,7 @@ tilesource_open(Tilesource *tilesource, int level)
 
        else if (vips_isprefix("webp", tilesource->loader) ||
                vips_isprefix("jxl", tilesource->loader) ||
+               vips_isprefix("png", tilesource->loader) ||
                vips_isprefix("gif", tilesource->loader)) {
                /* These formats have pages all the same size and support page and n.
                 */
```

This is a hand-rewrite of the LLM-generated load code in #4944 with the following changes:

- rather than ad-hoc manipulation of memory buffers, this PR uses regions and manipulates them with `vips_region_copy()`, `vips_region_paint()` etc. 
- it adds `vips_region_blend_over()` to blend two regions (this function could be usefully SIMDed!)
- when loading the APNG metadata, rather than a lot of call-once functions with mysterious and hard to track side effects, it has a single block of ifdefed code in `png2vips_header()`
- `_apng_generate()` handles any fetch geometry
- some efficiency improvements (eg. load directly to the animation if the blend mode is not OVER)
- various small bugfixes

I kept the nice tests untouched.

We should get rid of `vipspng.c` and fold that code into `pngload.c` and `pngsave.c`. I'll make a PR for that once this PR is merged.

libvips/foreign/vipspng.c
```diff
@@ -935,12 +1126,246 @@ png2vips_generate(VipsRegion *out_region,
...
+			&read->dispose_rect,
+			read->dispose_rect.left,
+			read->dispose_rect.top);
+		break;
+
+	case PNG_fcTL_DISPOSE_OP_NONE:
+		break;
+
+	default:
+		printf("png2vips_apng_read_next_frame: unknown dispose\n");
```
kleisauke:
```suggestion
		g_assert_not_reached();
```

should be (fuzzer-)safe, it's checked here:
https://github.com/pnggroup/libpng/blob/d1d0abeffede1cc898ddc3d0e600839cf026d749/pngset.c#L1486-L1489

kleisauke (approved):
Nice!

libvips/foreign/vipspng.c
jcupitt: Oh, nice! I always forget to do a final pass looking for dbg code.

jcupitt:
And thank you for the review!

PR #5123, opened 2026-07-11
icc: fix black output for float input

### LINKED ISSUE #4674: Issue when applying an icc profile when colourspace is scRGB
Hi,
I have a weird issue that has started with version 8.17.x regarding colouspaces and icc profiles.

I have this sample image
Image
That has an embedded profile `Adobe RGB (1998)`, the thing is that if I convert this file into colorspace `scRGB` and then import a profile then the result is a black image. This was working fine in versions <= 8.16.x

Image

I've made a test case to show you what I mean:

[use-vips-func.c](https://github.com/user-attachments/files/22403056/use-vips-func.c)
```c
#include <stdio.h>
#include <vips/vips.h>

int
main(int argc, char **argv)
{
	VipsImage *in;
	double mean;
	VipsImage *out;
	VipsImage *out2;

	if (VIPS_INIT(argv[0]))
		vips_error_exit(NULL);

	vips_cache_set_max_mem(0);
	vips_cache_set_max(0);

	vips_concurrency_set(1);

	if (argc != 3)
		vips_error_exit("usage: %s infile outfile", argv[0]);

	in = vips_image_new_from_file(argv[1], NULL);
	if (vips_avg(in, &mean, NULL))
		vips_error_exit(NULL);


	/* generate photo nexative - replace with other vips_ funcs */
	if (vips_colourspace(in, &out, VIPS_INTERPRETATION_scRGB, NULL))
		vips_error_exit(NULL);

	g_clear_object(&in);
	size_t length;
	const void *data;
	if (vips_image_get_blob(out, VIPS_META_ICC_NAME, &data, &length))
		vips_error_exit(NULL);

	if (vips_icc_is_compatible_profile(out, data, length)) {
		printf("Compatible profile\n");
	} else {
		printf("Not compatible profile\n");
	}

	if (vips_icc_import(out, &out2, "input_profile", "sRGB_v4_ICC_preference.icc", "embedded", TRUE, NULL))
		vips_error_exit(NULL);

	g_clear_object(&out);

	if (vips_image_write_to_file(out2, argv[2], NULL))
		vips_error_exit(NULL);

	g_clear_object(&out2);

	return 0;
}
```

If I convert it to some other colourspace then is fine.
Do you know if it's normal or did I just stumbled on a bug?
Thank you very much for the help.

lancylot2004:
Hi! 👋

I ran into this pushing 32-bit float TIFFs (Lightroom exports, embedded Linear Rec. 2020) into [Immich](https://immich.app/) - anything that went through an ICC transform came out pure black. It turned out to be a libvips thing, and it looks like the same bug as #4674.

I think the cause is that float input to `icc_import` / `icc_transform` gets cast down to `uchar` before the lcms transform. The fix hands lcms the float encodings (`TYPE_RGB_FLT` and so on) for float input, which is what the Lab/XYZ PCS paths already do. Integer paths are untouched, and I've tried to add a test that a float image transforms the same as the equivalent 8-bit one.

Apologies for jumping straight to a PR without discussing first — I saw @jcupitt had already earmarked scRGB + `icc_transform` for the 8.19 cycle in #4674, so I may well be missing context or stepping on planned work. Very happy to rework the approach, or drop it if rather fold it into that effort.

Closes #4674.

## Before / after

Here's a test image of my own, and the image attached in #4674.

| | Original | Before | After |
|---|---|---|---|
| **32-bit float TIFF** | [tif.tif](https://github.com/user-attachments/files/29931092/tif.tif) | tif-before | tif-after |
| **#4674 sample** | issue | issue-before | issue-after |

(The #4674 "after" still differs a little from the original - should be the scRGB round-trip in the poster's repro being a bit lossy.)

My TIFF:

```
$ vipsheader -a example.tif
L1027162.tif: 5152x2904 float, 3 bands, scrgb, tiffload
  format: float
  bits-per-sample: 32
  interpretation: scrgb
  icc-profile-data: 496 bytes of binary data   # "Linear Rec. 2020"
```

It's a plain 32-bit float export straight out of Lightroom, with an embedded profile.

Generated with:

```bash
# 32-bit float TIFF
ninja -C build && ./build/tools/vips icc_transform in.tif out.jpg srgb --embedded

# #4674 sample — the poster's exact steps
ninja -C build && \
  ./build/tools/vips colourspace in.jpg /tmp/s.v scrgb && \
  ./build/tools/vips icc_import /tmp/s.v /tmp/out.v --embedded
```


mertalev:
The 4674 image looks off beyond lossiness, as if the wrong transform got applied. It'd be good to pin down why it looks different.

lancylot2004:
> The 4674 image looks off beyond lossiness, as if the wrong transform got applied. It'd be good to pin down why it looks different.

I think it's happening because `vips_colourspace ... scrgb` linearises the pixels but keeps the original gamma-sRGB `icc-profile-data` attached, so `icc_import --embedded` applies the profile's tone curve to already-linear values.

Deciding how libvips should handle that stale profile/interpretation mismatch seems like a separate issue with a larger blast radius than the float-encoding fix here. Does that sounds about right?

mertalev:
Yup, that makes sense to me. I'm not a maintainer though, so it's up to the libvips maintainers on how to proceed.

kleisauke:
It looks like these `TYPE_*` macros in LCMS are generated from a small set of bitfield macros:
https://github.com/mm2/Little-CMS/blob/lcms2.19.1/include/lcms2.h#L776

So, I think we only need to store the pixel type (e.g. `PT_RGB`, `PT_CMYK`, etc.) and whether floating-point encoding is supported. See e.g. commit https://github.com/kleisauke/libvips/commit/3efdca343ab28f6ca77516e6142c7946d2d4b1ec.

> > The 4674 image looks off beyond lossiness, as if the wrong transform got applied. It'd be good to pin down why it looks different.
> 
> I think it's happening because `vips_colourspace ... scrgb` linearises the pixels but keeps the original gamma-sRGB `icc-profile-data` attached, so `icc_import --embedded` applies the profile's tone curve to already-linear values.
> 
> Deciding how libvips should handle that stale profile/interpretation mismatch seems like a separate issue with a larger blast radius than the float-encoding fix here. Does that sounds about right?

FWIW, we did try to fix this, but ran into issues, see e.g. #4321.

lancylot2004:
Thanks for the #4321 pointer!

I've folded your [commit](https://github.com/kleisauke/libvips/commit/3efdca343ab28f6ca77516e6142c7946d2d4b1ec) onto this branch: ran it against the existing suite plus the new float test, no regressions.

> FWIW, we did try to fix this, but ran into issues, see e.g. https://github.com/libvips/libvips/issues/4321.

Makes sense. Sounds like that'll be a separate issue then!

How does everything look now?

kleisauke:
Great, I simplified it further with commit 232681ad6ecd7a3fbdc2b228f39442c89b685ed2.

kleisauke (approved):
LGTM, but I'll leave the final review to @jcupitt since my commits in this PR need to be reviewed as well.

jcupitt (approved):
Very neat!

jcupitt:
I tried it out and it seems to work well! The fix is tidy too, nice work.

We'll need to look again at `colourspace` breaking profiles, but as you say that's a separate issue.

lancylot2004:
Thank you @jcupitt and @kleisauke! If all looks good I'll leave you guys to merge :)

I added a ChangeLog entry since that seems like what most PRs are doing - feel free to change

PR #5125, opened 2026-07-13
dcrawload: move other loaders ahead of dcrawload

### LINKED ISSUE #5124: dcrawload takes priority over JPEG for files with wrong extension
## Bug report

Given a JPEG file with .DNG extension, libvips errors with `dcrawload: unable to read: Unsupported file format or not RAW file`.

### Description

Since dcrawload matches based on file extension and has a higher priority than other loaders, an incorrect file extension can cause it to process images that aren't actually RAW.

### Reproducible example

This contains a regular JPEG image renamed to `sample.dng`: [sample.zip](https://github.com/user-attachments/files/29979080/sample.zip)

```
unzip sample.zip && vips jpegsave sample.dng sample_out.jpg
```

### Expected result

It should sniff the file and detect that it's actually a JPEG, as was the behavior in libvips 8.17.3 when magickload was used.

### Actual result

It assumes that it's a DNG and fails to process the image.

### Environment

- OS: Debian Trixie
- libvips version: 8.18.4

### Severity / impact

This affects Immich assets backed up from iOS. If a DNG file is edited, has a portrait effect, etc., the image that's uploaded is actually a JPEG render and not a DNG. The app apparently incorrectly sets the file extension as DNG in this case. While this is itself a bug, it previously worked and it's difficult to retroactively correct these files even if the bug is fixed moving forward.

- [ ] Crash / data loss.
- [x] Incorrect result.
- [ ] Build / install problem.
- [ ] Performance / memory issue.
- [ ] Other (please specify).

### Additional context

I think lowering the priority of the loader would be enough to fix it, still ahead of TIFF but behind loaders like JPEG that wouldn't be confused with a RAW image.

mertalev:
### Description

Since dcrawload matches on filename, there's a chance of false positives when an image has the wrong extension. This PR moves jpegload and heifload ahead of it since they have strict `is_a` implementations that should take priority if they match. tiffload is intentionally still lower since RAW images can look like TIFFs, so the file extension is actually a more reliable disambiguation there.

I was initially going to change `dcrawload`'s priority, but realized tiffload and jpegload share the same priority and heifload had a lower priority than tiffload, so raising jpegload and heifload is what I landed on.

Fixes #5124

### DIFF
--- a/libvips/foreign/dcrawload.c
+++ b/libvips/foreign/dcrawload.c
@@ -468,8 +468,8 @@ vips_foreign_load_dcraw_class_init(VipsForeignLoadDcRawClass *class)
 
 	operation_class->flags |= VIPS_OPERATION_UNTRUSTED;
 
-	/* We need to be ahead of JPEG and TIFF, since many cameras use those
-	 * formats as containers. We are slow to open, but we only test the
+	/* We need to be ahead of TIFF, since many cameras use that
+	 * format as a container. We are slow to open, but we only test the
 	 * filename suffix, so that's fine.
 	 */
 	foreign_class->priority = 100;
--- a/libvips/foreign/heifload.c
+++ b/libvips/foreign/heifload.c
@@ -1105,6 +1105,7 @@ vips_foreign_load_heif_class_init(VipsForeignLoadHeifClass *class)
 {
 	GObjectClass *gobject_class = G_OBJECT_CLASS(class);
 	VipsObjectClass *object_class = (VipsObjectClass *) class;
+	VipsForeignClass *foreign_class = (VipsForeignClass *) class;
 	VipsForeignLoadClass *load_class = (VipsForeignLoadClass *) class;
 
 	vips__heif_init();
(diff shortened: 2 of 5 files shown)


jcupitt (approved):

jcupitt:
See the (short!) discussion on the referring issue for more background to this.

kleisauke:
macOS CI failure is due to:
https://github.com/libvips/libvips/blob/38a4c4fec9653922e7f41bc57eaa4c2031fcd8ac/libvips/foreign/uhdrload.c#L602-L604

jcupitt:
Ah, good spot. @mertalev you'll need to bump uhdrload priority too.

kleisauke:
This might also need to be updated:
https://github.com/libvips/libvips/blob/38a4c4fec9653922e7f41bc57eaa4c2031fcd8ac/libvips/foreign/openslideload.c#L981-L988
and the "JPEG and" should be removed from this comment:
https://github.com/libvips/libvips/blob/38a4c4fec9653922e7f41bc57eaa4c2031fcd8ac/libvips/foreign/dcrawload.c#L471-L475

mertalev:
Good catches! I applied those changes, so here's hoping it all works.

kleisauke (approved):

PR #5126, opened 2026-07-15
colour: add missing `cmsSig4colorData` case

kleisauke:
Noticed while reviewing PR #5123.


jcupitt (approved):
Huh funny! I suppose I thought 4color was the same as cmyk.

PR #5127, opened 2026-07-15
icc: set input interpretation for CMYK profiles

kleisauke:
I'm not sure why we didn't this earlier, but it seems consistent with the LAB and XYZ cases above. I think this would be useful for multiband images, e.g.:
```python
#!/usr/bin/env python3
import pyvips

# One yellow pixel
pixel = bytes([255, 255, 0, 255])
multiband = pyvips.Image.new_from_memory(pixel, 1, 1, 4, 'uchar')
#multiband = multiband.colourspace('cmyk')
srgb = multiband.icc_transform('srgb', input_profile='cmyk')
print(srgb(0, 0)) # before: [25.0, 25.0, 37.0] after: [253.0, 238.0, 0.0, 255.0]
```
(i.e. the generic multiband image would be interpreted as CMYK now)

Use of CMYK profiles with sRGB images are still blocked:
```console
$ vips icc_transform ~/zebra.jpg x.jpg srgb --input-profile=cmyk

(vips:258568): VIPS-WARNING **: 16:00:00.497: profile incompatible with image
```

Noticed this while reviewing PR #5123.

### DIFF
--- a/libvips/colour/icc_transform.c
+++ b/libvips/colour/icc_transform.c
@@ -342,6 +342,7 @@ vips_icc_build(VipsObject *object)
 
 		case cmsSigCmykData:
 		case cmsSig4colorData:
+			code->input_interpretation = VIPS_INTERPRETATION_CMYK;
 		case cmsSig5colorData:
 		case cmsSig6colorData:
 		case cmsSig7colorData:
@@ -352,7 +353,6 @@ vips_icc_build(VipsObject *object)
 		case cmsSig12colorData:
 			/* Treat as forms of CMYK.
 			 */
-			code->input_interpretation = VIPS_INTERPRETATION_CMYK;
 			vips_icc_pick_input_format(code, icc, info);
 			break;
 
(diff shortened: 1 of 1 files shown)


libvips/colour/icc_transform.c
```diff
@@ -332,12 +330,10 @@ vips_icc_build(VipsObject *object)
 		case cmsSig12colorData:
 			/* Treat as forms of CMYK.
 			 */
-			info = vips_icc_info(
-				cmsGetColorSpace(icc->in_profile));
-
```
kleisauke, line 357: This was removed as it is already done here:
https://github.com/libvips/libvips/blob/38a4c4fec9653922e7f41bc57eaa4c2031fcd8ac/libvips/colour/icc_transform.c#L281-L282

jcupitt:
I think we didn't have a good * -> CYMK transform back in the day, so we didn't try to transform the input automatically. You're right, we have one now, so perhaps we should set `input_interpretation` for the cmyk4 case.

Is this right for the 5, 6, 7 etc. colour case? We probably don't want to force to CMYK for them.

kleisauke:
Good point. Commit 69747140f71dda4eb6426cc329f0add601684459 limits this to the `cmsSigCmykData` and `cmsSig4colorData` cases.

jcupitt (approved):

PR #5129, opened 2026-07-16
fuzz: add failing allocator fuzzer

kleisauke:
Add a fuzzer that deterministically injects allocation failures. This is mostly low-hanging fruit and some allocations in dependencies (e.g. GLib) abort the process instead of returning `NULL`, which limits the approach somewhat.

Context: https://github.com/google/oss-fuzz/issues/1675.

### DIFF
--- a/fuzz/failing-alloc.c
+++ b/fuzz/failing-alloc.c
@@ -1,0 +1,45 @@
+/* Originally from HarfBuzz: https://github.com/harfbuzz/harfbuzz/blob/main/src/failing-alloc.c
+ * SPDX-License-Identifier: MIT-Modern-Variant
+ */
+
+#include <stdlib.h>
+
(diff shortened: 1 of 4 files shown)


kleisauke:
I'll abandon this. I let it fuzz for more than a week, but it required building libvips with `-Dtiff=disabled -Dheif=disabled -Dimagequant=disabled` since libtiff, libaom and libimagequant do not always check for allocation failures. Also, as noted in PR https://github.com/mm2/Little-CMS/pull/584, some of these cases would likely never occur in the real world.

Note that libvips itself also does not always check for allocation failures, see e.g. commit https://github.com/kleisauke/libvips/commit/35495785d920a749643dd1a87047dc246a420b61. Though, I'm not sure whether it is worth fixing these cases.

PR #5130, opened 2026-07-16
exif: guard against allocation failures

kleisauke:
Found by #5129.

### DIFF
--- a/libvips/foreign/exif.c
+++ b/libvips/foreign/exif.c
@@ -214,7 +214,7 @@ vips_exif_load_data_without_fix(const void *data, size_t length)
 		/* Ensure "Exif" prefix as loaders may not provide it.
 		 */
 		void *data_with_prefix;
-		data_with_prefix = g_malloc0(length + 6);
+		data_with_prefix = g_malloc(length + 6);
 		memcpy(data_with_prefix, "Exif\0\0", 6);
 		memcpy((char *) data_with_prefix + 6, data, length);
 		exif_data_load_data(ed, data_with_prefix, length + 6);
@@ -874,7 +874,7 @@ tag_is_utf16(ExifTag tag)
 
 /* Set a libexif-formatted string entry.
  */
-static void
+static int
 vips_exif_alloc_string(ExifEntry *entry, unsigned long components)
 {
 	ExifMem *mem;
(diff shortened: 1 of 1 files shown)


lovell (approved):

kleisauke:
Sorry, I found a few more after letting it fuzz a bit longer. I'll leave inline comments on the lines that could use some clarification.

libvips/foreign/exif.c
```diff
@@ -214,7 +214,7 @@ vips_exif_load_data_without_fix(const void *data, size_t length)
 		/* Ensure "Exif" prefix as loaders may not provide it.
 		 */
 		void *data_with_prefix;
-		data_with_prefix = g_malloc0(length + 6);
+		data_with_prefix = g_malloc(length + 6);
```
kleisauke, line 217: Drive-by: switched to `g_malloc()` because the entire buffer is filled immediately after allocation, so `calloc()`'s zero-initialization is unnecessary. Also, a `NULL` check is unnecessary here because `g_malloc()` aborts on allocation failure.

libvips/foreign/exif.c
```diff
@@ -1022,13 +1022,13 @@ vips_exif_set_tag(ExifData *ed, int ifd, ExifTag tag, write_fn fn, void *data)
...
 	else {
-		entry = exif_entry_new();
+		if (!(entry = exif_entry_new()))
+			return -1;
 
 		/* tag must be set before calling exif_content_add_entry.
 		 */
 		entry->tag = tag;
 		exif_content_add_entry(ed->ifd[ifd], entry);
-		exif_entry_unref(entry);
```
kleisauke, line 1031: `exif_entry_unref(entry)` was moved to below because `exif_content_add_entry()` will only increase the reference on the `entry` if `realloc()` succeeds.
https://github.com/libexif/libexif/blob/v0.6.26/libexif/exif-content.c#L155-L161

(otherwise, a use-after-free can occur, as `exif_entry_unref()` will call `free()` when the reference count drops to zero)

lovell (approved):

kleisauke:
Closing in favor of PR #5133, sorry for the back-and-forth on this.

PR #5132, opened 2026-07-18
heifload: extract ISO 21496-1 HDR gain maps from HEIC auxiliary images

cosminpolifronie:
Hi! I was looking into converting my photo library from HEIC HDR photos to JXL HDR photos. In this search, I have come to find out some limitation in libjxl, which mean that HEIC HDR wasn't directly convertable to JXL HDR, at least not using gain maps.

I knew that technically it should be possible to convert the HEIC HDR gain mapped to an actual 10-bit HDR image. One thing led to another, and in the end I have spent 5+ hours prompting GLM-5.2 until the output image matches the input image. I am a Software Developer, and I have been reviewing the code written by AI as I progressed.

**I know there are a lot of implementations of HDR gain maps in HEIC, but luckily, in my case, my Samsung phone is actually using the ISO standard for this. Apple is not using the ISO standard in the latest iOS version, but a somewhat simplified version which is not covered by this patch.**

The text below has been written completely by AI. I have left in a lot of details so that the changes are justified in detail. I am somewhat familiar with image processing pipelines, but I'm nowhere near a knowledgeable person related to HDR pipelines or color spaces.

Please use this as a proof of concept and ask for as many refinements as you wish.

The attached photo has been taken by me, and you have full rights to use it in any way you wish. I don't claim any copyright on it, it can be considered open domain.
[samsung_hdr.zip](https://github.com/user-attachments/files/30158446/samsung_hdr.zip)

___________________________

# heifload: extract ISO 21496-1 HDR gain maps from HEIC auxiliary images

## Overview

This PR adds support for extracting ISO 21496-1 / Adobe HDR gain maps from HEIC files in `heifload`, and fixes several issues in `uhdr2scRGB` that prevented correct HDR rendering when converting gain-map-based HEIC images (e.g. from Samsung S25 Ultra) to native HDR formats like JPEG XL.

## Background

Modern smartphones (Samsung S25 Ultra, Apple iPhone 16, etc.) store HDR photos as HEIC files containing:
1. An 8-bit SDR primary image (the base rendition)
2. A gain map auxiliary image (stored as an HEIC aux item with type `urn:com:samsung:photo:2024:aux:hdrgainmap` or `urn:com:apple:photo:2020:aux:hdrgainmap`)
3. Gain map metadata as an Adobe `hdrgm:` XML namespace (ISO 21496-1)

Previously, `heifload` only loaded the primary SDR image and discarded the gain map entirely. The `uhdr2scRGB` operation (which applies the gain map math to recover HDR) existed but was only usable with Google UltraHDR JPEGs via `uhdrload`.

### 1. `heifload.c` — ISO 21496-1 gain map extraction

Added four functions:

- **`vips_foreign_load_heif_is_gainmap_aux()`** — detects gain map aux images by matching the `"hdrgainmap"` substring in the aux type URN. Covers Samsung, Apple, and any future ISO URN variants.

- **`vips_foreign_load_heif_parse_gainmap_xml()`** — SAX-based parser for the Adobe `hdrgm:` XML namespace using GLib's `g_markup_parse_context`. Extracts the 8 standard attributes (`GainMapMin`, `GainMapMax`, `Gamma`, `OffsetSDR`, `OffsetHDR`, `HDRCapacityMin`, `HDRCapacityMax`, `BaseRenditionIsHDR`).

  **Critical conversion**: ISO 21496-1 `GainMapMin`/`GainMapMax`/`HDRCapacityMin`/`HDRCapacityMax` are **log2 values** (stops), while libultrahdr's `min_content_boost`/`max_content_boost`/`hdr_capacity_min`/`hdr_capacity_max` are **linear ratios** (1.0 = no boost, 4.9 = 4.9× boost). The parser converts with `exp2()`. Without this conversion, `log2(0)` produces `-inf` and the output is all zeros.

- **`vips_foreign_load_heif_decode_aux()`** — decodes the gain map aux image. Uses a format-aware approach:
  1. Decodes as YCbCr 4:2:0
  2. Checks Cb/Cr deviation from neutral (128). If within ±8, the gain map is effectively monochrome (the small deviation is HEVC compression noise) → extract the Y (luma) plane only (1-band)
  3. If Cb/Cr deviate significantly, the gain map is truly multi-channel → re-decode as interleaved RGB (3-band)

  **Why Y-plane extraction**: Gain maps are stored as HEVC 4:2:0, where the Y plane contains the actual gain data at full resolution. The Cb/Cr planes are subsampled (half resolution in each dimension) and contain no useful gain information. Decoding as RGB would upsample the chroma (4:2:0 → 4:4:4), introducing block boundary artifacts that become visible "streaks" when amplified by the exponential gain map boost formula (`exp2(boost)`). By extracting the Y plane directly, we avoid chroma upsampling entirely.

  **Stride padding handling**: The Y plane may have `stride > width` (alignment padding, e.g. width=750 but stride=752). The code copies row-by-row to create tightly packed data for `vips_image_new_from_memory`. Without this, misaligned rows produce random noise in the output image.

- **`vips_foreign_load_heif_load_gainmap()`** — orchestrator that:
  1. Enumerates aux images on the primary handle via `heif_image_handle_get_number_of_auxiliary_images` + `heif_image_handle_get_list_of_auxiliary_image_IDs`
  2. For each gain map aux, decodes the raster (via `vips_foreign_load_heif_decode_aux`) and attaches it as `"gainmap"` VipsImage metadata
  3. Enumerates metadata blocks on the **aux handle** (not the primary) to find the `cdsc`-referenced MIME item with the `hdrgm:` XML. This is critical: the XML is attached to the gain map image item (item 55 in Samsung's layout), not to the primary image item (item 49). The existing metadata loop in `vips_foreign_load_heif_set_header` only enumerates metadata on the primary handle, so it never sees the gain map XML.
  4. Parses the XML and populates the 8 `gainmap-*` metadata fields, matching the format produced by `uhdrload.c`

  The gain map metadata is attached using the same field names as `uhdrload.c` (`gainmap-max-content-boost`, `gainmap-gamma`, etc.), so the existing `uhdr2scRGB` operation works unchanged for the gain map math.

### 2. `uhdr2scRGB.c` — gain map math fixes

- **RGB path gain map linearization**: Changed `vips_v2Y_8[p2[x]]` (sRGB inverse OETF) to `p2[x] / 255.0f` (raw normalization). The sRGB inverse OETF was incorrectly applied to gain map values when `Gamma=1.0`. The gain map's encoding is controlled by the `Gamma` parameter, not sRGB. The mono path already used raw normalization; the RGB path was inconsistent. For Samsung's gain map (max pixel=178/255=0.698), this was a 1.58× brightness error (sRGB decode gives 0.445, linear gives 0.698).

- **Primaries conversion**: Added `vips_uhdr2scRGB_detect_primaries()` which detects the SDR image's colour primaries and applies the appropriate conversion matrix to BT.709 (scRGB). Detection order:
  1. CICP metadata (`cicp-colour-primaries`): codes 1 (BT.709), 9 (BT.2020), 12 (Display P3) → direct matrix lookup
  2. ICC profile: uses lcms2's `cmsGetProfileInfoASCII` to read the profile description, searches for "P3" (Display P3) or "2020" (BT.2020)
  3. Default: identity matrix (assume sRGB/BT.709)

  **Why this matters**: Samsung S25 Ultra's SDR base uses "DCI-P3 D65 Gamut with sRGB Transfer" (confirmed via `exiftool` on the ICC profile). Without primaries conversion, the Display P3 colours are treated as sRGB, causing both colour shifts (less vibrant) and brightness errors. The conversion matrix is applied after the gain map math but before the `VIPS_UHDR_TO_SCRGB` scaling. Negative values from out-of-gamut colours are clamped to 0.

  **lcms2 instead of raw byte search**: The initial implementation searched for "P3" in the raw ICC profile bytes. This failed because ICC profiles store text descriptions in structured tags (ICCM profileInfoType), not as plain ASCII. Using lcms2's `cmsGetProfileInfoASCII` properly extracts the description.

- **Matrix sharing**: Moved all 9 primaries-to-BT.709 matrices and the lookup function from `CICP2scRGB.c` to `pcolour.h` as shared static inline functions (`vips_cicp_get_primaries_matrix()`, `vips_cicp_apply_matrix()`). Both `CICP2scRGB.c` and `uhdr2scRGB.c` now use these shared definitions, eliminating duplication.

### 3. Test samples and tests

Added three HEIC test samples to `test/test-suite/images/`:
- `samsung-heic-gainmap.heic` — Samsung S25 Ultra, ISO 21496-1 `hdrgm:` XML
- `apple-heic-gainmap.heic` — iPhone 16, Apple `HDRGainMap:` XML (gain map image attached, math metadata not parsed — out of scope)
- `oneplus-heic.heic` — OnePlus 13, no gain map (negative test case)

Added 5 tests to `test/test-suite/test_foreign.py`:
- `test_heifload_gainmap_samsung` — verifies all 8 `gainmap-*` fields populated with correct values (GainMapMax=2.3 → max_content_boost≈4.92, etc.)
- `test_heifload_gainmap_apple` — verifies gain map image attached but XML not parsed (Apple format not yet supported)
- `test_heifload_no_gainmap` — verifies SDR HEIC has no gainmap metadata
- `test_heifload_gainmap_uhdr2scrgb` — verifies end-to-end HEIC → `uhdr2scRGB` → scRGB float with HDR values >1.0
- `test_heifload_gainmap_to_jxl_hdr` — verifies full pipeline: HEIC → `uhdr2scRGB` → `scRGB2CICP` (PQ) → `jxlsave` with correct CICP metadata

## Technical issues encountered and resolved

1. **ISO 21496-1 log2 vs. libultrahdr linear**: `GainMapMin=0.0` caused `log2(0)=-inf` → all-zero output. Fixed by converting with `exp2()` before populating metadata fields.

2. **Gain map XML on aux handle, not primary**: The `hdrgm:` XML (item 59) is `cdsc`-attached to the gain map image item (item 55), not the primary (item 49). The existing metadata loop only saw primary-handle metadata. Fixed by enumerating metadata on the aux handle.

3. **Chroma upsampling streaks**: Decoding the gain map as RGB caused HEVC 4:2:0 chroma upsampling artifacts that became visible streaks when amplified by `exp2()`. Fixed by decoding as YCbCr 4:2:0 and extracting the Y plane, with format-aware fallback to RGB for true multi-channel gain maps.

4. **Stride padding noise**: The Y plane had `stride=752` but `width=750` (2 bytes/row padding). `vips_image_new_from_memory` assumed packed data, causing row misalignment and random noise. Fixed by copying row-by-row when `stride != width`.

5. **sRGB inverse OETF on linear gain map**: The RGB gain map path applied `vips_v2Y_8[]` (sRGB→linear) to gain map values, but `Gamma=1.0` means the gain map is already linear. This caused a 1.58× brightness reduction. Fixed by using raw normalization (`p2[x]/255.0f`), matching the mono path.

6. **Display P3 primaries not converted**: `uhdr2scRGB` output scRGB (BT.709) but the SDR base was Display P3. Without conversion, colours were less vibrant and brightness was off. Fixed by detecting primaries from CICP or ICC (via lcms2) and applying the conversion matrix with negative clamping.

7. **ICC text search failure**: Initial implementation searched for "P3" in raw ICC bytes, but ICC profiles store descriptions in structured tags. Fixed by using lcms2's `cmsGetProfileInfoASCII`.

## Out of scope

- Apple's proprietary `HDRGainMap:` XML namespace (iPhone 16+) is not parsed. The gain map image is still attached, but the math metadata fields are not populated. Mapping `HDRGainMapHeadroom` to ISO 21496-1 parameters would require research into Apple's gain map formula.
- Pre-iOS 17 Apple HDR (JPEG+MPF) is handled separately by `uhdrload.c` via libultrahdr.
- The `uhdr2scRGB` primaries conversion uses lcms2's profile description string matching ("P3", "2020") rather than comparing the actual XYZ matrix coefficients. This is a pragmatic heuristic that works for all known real-world ICC profiles but could theoretically fail for custom profiles.

## Example Python API

```python
import pyvips

def heic_hdr_to_jxl(input_path, output_path, distance=1.0):
    """Convert an HEIC image with ISO 21496-1 HDR gain map to a native HDR JXL.

    The output JXL is encoded as 16-bit PQ (SMPTE ST 2084) with BT.2100
    colour primaries, suitable for display on HDR-capable monitors.

    Args:
        input_path: Path to the input HEIC file (e.g. Samsung S25 Ultra HDR photo)
        output_path: Path for the output JXL file
        distance: JXL compression distance (0=lossless, 1.0=visually lossless, higher=more compression)
    """
    # Load the HEIC file with gain map extraction enabled.
    # unlimited=True disables libheif's security limits, which are
    # triggered by the large number of items in Samsung/Apple HEIC files.
    im = pyvips.Image.heifload(input_path, unlimited=True)

    # Apply the gain map math to recover HDR linear light.
    # This reads the "gainmap" VipsImage metadata and gainmap-* fields
    # populated by heifload, and produces scRGB float output where
    # 1.0 = 80 nits (SDR reference white).
    # The SDR image's colour primaries (e.g. Display P3) are automatically
    # detected from CICP or ICC metadata and converted to BT.709 (scRGB).
    scrgb = im.uhdr2scRGB()

    # Encode as 16-bit PQ with BT.2100 primaries.
    # PQ (Perceptual Quantizer, SMPTE ST 2084) maps linear light to
    # 10-12 bit code values optimized for human perception.
    # BT.2100 primaries cover the full HDR colour gamut.
    # matrix_coefficients=0 (identity) because JXL stores RGB natively.
    pq = scrgb.scRGB2CICP(
        colour_primaries=9,           # BT.2100
        transfer_characteristics=16,  # PQ (SMPTE ST 2084)
        matrix_coefficients=0,        # identity (RGB)
        depth=16,                     # 16-bit container for 10-bit PQ codes
    )

    # Save as JXL with visually-lossless compression.
    # The JXL header will signal PQ + BT.2100 + intensity_target=10000 nits
    # (auto-defaulted by libjxl for PQ transfer function).
    pq.jxlsave(output_path, distance=distance)


# Usage:
heic_hdr_to_jxl(
    input_path="/Users/cosminpolifronie/Downloads/heic_samples/samsungs25u.heic",
    output_path="/Users/cosminpolifronie/Downloads/samsung_hdr_output.jxl",
)
```

## Verification

The output JXL has been visually verified to match the original Samsung S25 Ultra HEIC when rendered on a Liquid Retina XDR display (macOS 26, Preview/Safari), and also in Windows on an Alienware AW3423DWF display. The pipeline produces:
- 3000×4000 16-bit PQ JXL, ~1.6 MB
- `cicp-transfer-characteristics: 16` (PQ)
- `cicp-colour-primaries: 9` (BT.2100)
- Peak brightness: ~630 nits (matching the gain map's `HDRCapacityMax=2.3` → `2^2.3 ≈ 4.9× SDR`)
- No visible streaks, artefacts, or colour shifts


jcupitt:
Hi @cosminpolifronie,

Thanks for this. libvips loads UltraHDR images via google's libultrahdr. Right now, this only handles JPEG images, but there's a new version being developed that adds support for gainmaps in other formats, including HEIC. They have the new code in an experimental branch for now, but you can try it out.

My feeling is that waiting for the next libultrahdr is probably the simplest solution for us.

cosminpolifronie:
Hi! I will check it out and compare it with this impleme tation, just out of curiosity.

Sadly I see that this new libultrahdr release has been mentioned for a year now, with no release in sight. The functionality has been implemented in august last year.

I'm not holding my breath for a release there. The library looks almost abandoned.

Changes in this PR are quite minimal and fit very well with the currently existing path, at least for the ISO implementation in HEIC. Even if libultrahdr is used for loading, I think you will still need some similar code to actually read the correct fields from the photo and apply them, no?

mertalev:
libultrahdr is active; in fact, it just got support for Apple's gain maps last month. They just haven't made a release in a while. I feel like waiting for them is the most natural option. Building out something here would just turn into tech debt once upstream support lands.

It might be good to work with the libultrahdr branch and see what changes would be needed on libvips's end with that, since that's the more realistic integration path.

jcupitt:
> Even if libultrahdr is used for loading, I think you will still need some similar code to actually read the correct fields from the photo and apply them, no?

I'm not sure, but I'd expect the ultrahdr metadata to cover this. I suppose we'll see when they release the new version.

In the meantime, I'll close this PR. Please reopen if libultrahdr fails to add HEIC (and AVIF I guess?) support.

kleisauke:
[libultrahdr v2.0.0](https://github.com/google/libultrahdr/releases/tag/v2.0.0) supports AVIF/HEIC gain maps via libheif. See the API changes for details:
https://github.com/google/libultrahdr/blob/v2.0.0/docs/api_changes_v1.5_to_v2.0.md

Some work is needed in libvips to support this, as the `vips_image_get_gainmap()` API and the `uhdrload` / `uhdrsave` operations are currently tied to JPEG.

kleisauke:
Actually, libultrahdr v2.0.0 depends on an unmerged libheif PR, so it can't be easily integrated or tested yet. See PR #5177 for details.

PR #5133, opened 2026-07-20
exif: use GLib allocator for libexif objects

kleisauke:
Create `ExifData` and `ExifEntry` instances with a custom `ExifMem` that delegates to GLib allocation functions. This ensures all libexif-managed memory uses `g_malloc0()`, `g_realloc()` and `g_free()`. GLib allocation failures abort, so no allocation error handling is required.

Supersedes: #5130.


jcupitt (approved):

jcupitt:
A neater solution!

PR #5136, opened 2026-07-23
jp2kload: support palette images

### LINKED ISSUE #5027: jp2kload: add support for palette images
## Bug report

I'm running into some problems when I try to read paletted JP2 (JPEG 2000 Part 1) images with pyvips. Here's an example of a paletted JP2:

[palettedImage.jp2](https://github.com/openpreserve/jpylyzer-test-files/raw/refs/heads/master/files/palettedImage.jp2)

I use this as input for this script:

```python
import pyvips

imageIn = "palettedImage.jp2"

im1 = pyvips.Image.new_from_file(imageIn, access="sequential")
stats = im1.stats().tolist()
```

Result:

```
Traceback (most recent call last):
  File "/home/johan/test/test-palette.py", line 7, in <module>
    stats = im1.stats().tolist()
            ^^^^^^^^^^^
  File "/home/johan/kb/digitalisering/tifftojp2/.venv/lib/python3.12/site-packages/pyvips/vimage.py", line 1353, in call_function
    return pyvips.Operation.call(name, self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/johan/kb/digitalisering/tifftojp2/.venv/lib/python3.12/site-packages/pyvips/voperation.py", line 302, in call
    raise Error(f'unable to call {operation_name}')
pyvips.error.Error: unable to call stats
  jp2kload: Failed to decode all used components

jp2kload: Failed to decode the codestream in the JP2 file
```

For regular (non paletted) JP2s everything works as expected. Paletted TIFFs work fine as well (here's an example: [poppies.tif](https://github.com/tlnagy/exampletiffs/raw/refs/heads/master/poppies.tif))

Since Libvips uses OpenJPEG for JPEG 2000 decoding, I also tried decoding some paletted JP2s directly  with OpenJPEG's opj_decompress tool (from v2.5.0 of libopenjp2-tools):

```
opj_decompress -i palettedImage.jp2 -o palettedImage-opj.tif
```

This works without problems, suggesting that the issue in LibVips is not caused by an inability of OpenJPEG to decode paletted JP2s, but rather by Libvips not interpreting the decoded image data correctly.

**Environment**

 - OS: Linux Mint 22.3
 - Vips: 8.15.1

**Additional context**

I'm using Python 3.12.3, with Pyvips installed in local mode, i.e. installed using:

```
sudo apt install libvips-dev --no-install-recommends
pip install pyvips
```

Socialpranker:
**AI disclosure**: written by Claude (Anthropic, Sonnet 5). The entire patch, root-cause analysis, and verification below were produced by an AI agent; I (the human account owner) reviewed the diff and the verification before submitting.

Fixes #5027.

### What was happening

For a paletted JP2, `opj_read_header()` reports the palette *index* component only (`numcomps == 1`, with `color_space` already set to the post-expansion value, e.g. `OPJ_CLRSPC_SRGB` from the `colr` box). openjpeg then expands the palette itself inside `opj_decode()` / `opj_get_decoded_tile()`, so the decoded image has three components. jp2kload trusted the header, made a one-band image, and the decode-vs-container sanity check then failed with `decoded image does not match container` (on current master; older releases like the 8.15.1 in the issue report failed earlier, inside openjpeg, with "Failed to decode all used components").

openjpeg's public API exposes nothing about `pclr`/`cmap` before decode, so the header-time fix has to be a prediction: **a colour colourspace (sRGB/sYCC/eYCC) with fewer than three components can only mean a palette** — the `pclr`/`cmap` boxes will expand the indexes to full colour during decode. When that signature is seen, declare three output bands. Everything downstream then works unchanged: the decoded image really does have three components, `vips_foreign_load_jp2k_pack()` reads `numcomps` from the *decoded* image, and the existing `is_match` check passes naturally — while still rejecting genuinely malformed files.

### issue412.jp2

An interesting side effect: `test/test-suite/images/issue412.jp2` — the "horrible thing [with] a header that doesn't match the decoded pixels" that the test suite expected to be *rejected* — turns out to be exactly this kind of file (header: 1 component, sRGB; decode: 3 components). With this change it loads and decodes **bit-identically to `opj_decompress`** (max abs diff 0.0), so I updated that test to expect success rather than an exception. Flagging this explicitly since it changes the documented behavior of that fixture.

### Testing

Verified for real against a local meson build on macOS:

- **The reporter's exact script** (`new_from_file(..., access="sequential")` + `.stats()`) on their sample file ([palettedImage.jp2](https://github.com/openpreserve/jpylyzer-test-files/raw/refs/heads/master/files/palettedImage.jp2)): fails on unpatched master with `decoded image does not match container`, succeeds with this patch (3 bands, uchar, srgb).
- **Pixel correctness**: both the reporter's file and `issue412.jp2` decode bit-identically to `opj_decompress` output (`(vips - ref).abs().max() == 0.0`), across random, `access="sequential"`, `oneshot=True`, and `thumbnail` paths. Both `opj_decode()` and `opj_get_decoded_tile()` were probed directly with a small C program to confirm palette expansion behaves the same on the tiled and untiled paths.
- **Non-regression**: `Bretagne2_1.j2k`, `Bretagne2_4.j2k`, `world.jp2` load unchanged; full `test_foreign.py` + `test_conversion.py` + `test_colour.py` run green (88 passed / 19 skipped for missing optional deps / 3 xpassed, same as master).
- **New fixture**: rather than vendoring the reporter's sample (PROBA2 solar image, "public domain (license unknown)" per the jpylyzer-test-files README — unclear provenance for redistribution), I added `test/test-suite/images/palette.jp2`: a 380-byte JP2 assembled by hand (opj_compress lossless codestream of an 8-bit index image, wrapped in `jP  `/`ftyp`/`jp2h(ihdr+colr+pclr+cmap)`/`jp2c` boxes) with a 16-entry RGB palette and known per-pixel values. The new test checks exact expanded pixel values on the random, sequential, and oneshot paths, and fails on unpatched master.

### Known limitation

If a palette maps low-precision indexes to a *wider* LUT (e.g. 8-bit indexes → 16-bit palette entries), the header-predicted band format (from the index component's `prec`) won't match the decoded format, and `is_match` will still reject the file — same failure as today, no regression, but not yet supported. Supporting that would need the palette bit depth, which openjpeg doesn't expose before decode. Happy to note this in the docs if you'd like.


libvips/foreign/jp2kload.c
```diff
@@ -527,6 +533,29 @@ vips_foreign_load_jp2k_get_ycc(opj_image_t *image)
...
+ * palette itself during decode, but the header only shows the palette
+ * index component, so we must spot this case and predict the shape of the
+ * expanded image.
+ *
+ * A colour colourspace with fewer than three components can only mean a
+ * palette: the pclr/cmap boxes will expand the indexes to full colour
+ * during decode.
+ */
+static gboolean
+vips_foreign_load_jp2k_get_palette(opj_image_t *image)
```
jcupitt: Rather than adding a palette member, how about changing this to a function which predicts the bands from a jp2 image?

Eg. maybe:

```C
static int
vips_foreign_load_jp2k_get_bands(opj_image_t *image)
{
	switch (image->color_space) {
	case OPJ_CLRSPC_SRGB:
	case OPJ_CLRSPC_SYCC:
	case OPJ_CLRSPC_EYCC:
        // numcomps can be 1 for palette images
        return 3;

	default:
		return image->numcomps;
	}
}
```

Now just call that for the `vips_image_init_fields()` below.

jcupitt:
Does jp2 support RGBA? That might mess things up. Or palette CMYK!

jcupitt:
Hi again @Socialpranker, thanks for this. It's nice it's such a simple fix.

I think we could probably make it quite a bit simpler still, I added a note.

Socialpranker:
Checked both of those against openjpeg 2.5.4 rather than guessing, by reading the header and then fully decoding, printing `numcomps` and `color_space` at each stage:

```
palette.jp2      HEADER: numcomps=1 color_space=SRGB  ->  DECODED: numcomps=3 color_space=SRGB
issue412.jp2     HEADER: numcomps=1 color_space=SRGB  ->  DECODED: numcomps=3 color_space=SRGB
rgba.jp2         HEADER: numcomps=4 color_space=SRGB  ->  DECODED: numcomps=4 color_space=SRGB
gray_alpha.jp2   HEADER: numcomps=2 color_space=GRAY  ->  DECODED: numcomps=2 color_space=GRAY
gray.jp2         HEADER: numcomps=1 color_space=GRAY  ->  DECODED: numcomps=1 color_space=GRAY
```

So yes, jp2 does RGBA, and it comes through as `numcomps=4` with `SRGB` in the header. That's why the check is `numcomps < 3` rather than an unconditional 3: your sketch returns 3 for every `SRGB`/`SYCC`/`EYCC` image, which would drop the alpha band on the RGBA case above and make `vips_image_init_fields` disagree with what openjpeg actually hands back.

Grey plus alpha is the other one I was worried about, since it's also under three components, but it reports `GRAY`, not `SRGB`, so it lands in the `default` branch and keeps its 2 bands. A colour colourspace with fewer than three components really does only happen for palette images.

Palette CMYK can't reach the palette branch at all: `OPJ_CLRSPC_CMYK` is its own enum value, so it falls to `default` and keeps `numcomps`. I don't have a paletted CMYK jp2 to hand and openjpeg's own CMYK handling is separate from the pclr/cmap path, so I'd rather leave that case behaving exactly as it does today than predict a geometry I can't test.

Happy to fold the function into the shape you described, keeping the `< 3` guard, if you'd like the palette member gone:

```C
static int
vips_foreign_load_jp2k_get_bands(opj_image_t *image)
{
	switch (image->color_space) {
	case OPJ_CLRSPC_SRGB:
	case OPJ_CLRSPC_SYCC:
	case OPJ_CLRSPC_EYCC:
		/* numcomps is 1 for palette images, which openjpeg expands to
		 * three bands during decode.
		 */
		return image->numcomps < 3 ? 3 : image->numcomps;

	default:
		return image->numcomps;
	}
}
```

That drops the `palette` struct member and the separate detect call, which is the simplification you were after. Say the word and I'll push it.

jcupitt:
Sure, that looks reasonable. Thanks!

Socialpranker:
Done in 4907755 — the `palette` member is gone, and `vips_foreign_load_jp2k_get_bands()` is called straight from `vips_image_init_fields()`.

One deviation from your sketch: returning a flat `3` for the colour cases would drop the alpha on RGBA, since those arrive as `numcomps=4 color_space=SRGB` (from the header check I ran earlier). `VIPS_MAX(3, image->numcomps)` keeps 4 there and still lifts a palette's single index component to 3.

Checked against the built lib:

```
palette.jp2      bands=3  srgb
issue412.jp2     bands=3  srgb
world.jp2        bands=3  srgb
rgba.jp2         bands=4  srgb
gray_alpha.jp2   bands=2  b-w
gray.jp2         bands=1  b-w
```

Python suite is 310 passed / 40 skipped.

libvips/foreign/jp2kload.c
```diff
@@ -527,6 +527,29 @@ vips_foreign_load_jp2k_get_ycc(opj_image_t *image)
 	return ycc;
 }
 
+/* Predict the number of bands the decoded image will have.
+ *
+ * openjpeg expands palettes itself during decode, but the header only
```
jcupitt: Remove this paragraph.

ChangeLog
```diff
@@ -1,5 +1,6 @@
 date-tbd 8.19.0
 
+- jp2kload: support palette images [Socialpranker]
```
jcupitt: This should go at the end of the list of changes for 8.19.

libvips/foreign/jp2kload.c
Socialpranker: Removed in f7b0666 — just the one-line comment now.

ChangeLog
Socialpranker: Moved to the end of the 8.19.0 list in f7b0666.

jcupitt:
Great! Thank you for doing this work @Socialpranker

PR #5141, opened 2026-07-25
PNG: add cLLI load and save support

gregbenz:
## What

Add PNG cLLI chunk support when libvips is built against libpng with the fixed-point cLLI APIs.

- Load PNG cLLI values into the existing common CLLI metadata fields.
- Save a cLLI chunk when both common fields are present, valid, and representable by libpng.
- Preserve cLLI when saving with `keep=none`, consistent with CLLI being HDR signaling rather than ancillary metadata.
- Keep cICP and cLLI working together in the same PNG.

PNG stores cLLI values in 0.0001 nit units, while the common libvips fields use whole nits. Loading therefore rounds to the nearest whole nit. This small quantization is trivial and intentionally allowed so PNG can reuse the representation introduced for HEIF/AVIF without adding format-specific metadata fields.

This follows the PNG work discussed and accepted as a separate follow-up in [#5104](https://github.com/libvips/libvips/pull/5104#pullrequestreview-4756619565).

## Tests

- Added byte-level PNG chunk tests for load, save, chunk ordering, rounding, range limits, missing/invalid values, `keep=none`, and cICP coexistence.
- Added a compile-time-gated C round-trip test so capable libpng builds cannot silently skip the behavior.
- Verified the dedicated PNG cLLI test, the CICP/CLLI test module, the full Python test suite, and the Meson build.

## AI disclosure

Disclosure: I used an LLM to assist with implementation and review. I reviewed and tested all changes.


test/test-suite/test_cicp.py
```diff
@@ -1,11 +1,156 @@
...
+
+
+def require_png_clli_read():
+    if not png_clli_read_supported():
+        pytest.skip("libpng build lacks fixed-point cLLI read support")
+
+
+def require_png_clli_write():
+    if not png_clli_write_supported():
+        pytest.skip("libpng build lacks fixed-point cLLI write support")
```
kleisauke: It would probably be neater to have a single `require_png_clli_support()` function that does the same thing as `test_png_clli.c`.

This way, you wouldn't need the `parse_png_chunks()`, `insert_png_clli()`, `get_png_clli_chunks()` and `assert_png_clli()` helper functions above.

test/test-suite/test_cicp.py
```diff
@@ -653,3 +798,130 @@ def test_png_cicp_lossless_roundtrip(self):
...
+    @pytest.mark.parametrize(
+        "max_cll,max_fall,expected_cll,expected_fall",
+        [
+            (7_694_359, 1_825_000, 769, 183),
+            (0x7fffffff, 0, 214_748, 0),
+            (0, 10_000, 0, 1),
+        ],
+        ids=["fractional", "libpng-maximum", "zero-and-nonzero"]
+    )
+    def test_png_clli_load(self, max_cll, max_fall,
```
kleisauke: This test can be removed, since a full round-trip is already tested below.

test/test-suite/test_cicp.py
```diff
@@ -653,3 +798,130 @@ def test_png_cicp_lossless_roundtrip(self):
...
+    @pytest.mark.parametrize(
+        "max_cll,max_fall,save_options",
+        [
+            (1624, 182, {}),
+            (1624, 182, {"keep": "none"}),
+            (0, 214_748, {"palette": True}),
+        ],
+        ids=["default", "keep-none", "range-boundaries"]
+    )
+    def test_png_clli_save(self, max_cll, max_fall, save_options):
```
kleisauke:
```suggestion
    def test_png_clli_roundtrip(self, max_cll, max_fall, save_options):
```

test/test-suite/test_cicp.py
```diff
@@ -653,3 +798,130 @@ def test_png_cicp_lossless_roundtrip(self):
...
+        require_png_clli_read()
+
+        im = make_png_test_image()
+        add_clli(im, max_cll, max_fall)
+
+        encoded = im.pngsave_buffer(**save_options)
+        assert_png_clli(
+            encoded,
+            (max_cll * 10_000, max_fall * 10_000)
+        )
```
kleisauke: This can be removed, as it's already covered by the `.get()` test below (and I'd like to get rid of those helpers).
```suggestion
```

test/test-suite/test_cicp.py
```diff
@@ -653,3 +798,130 @@ def test_png_cicp_lossless_roundtrip(self):
...
+        ids=[
+            "missing-max-fall",
+            "missing-max-cll",
+            "negative-max-cll",
+            "negative-max-fall",
+            "large-max-cll",
+            "large-max-fall",
+        ]
+    )
+    def test_png_clli_invalid_or_incomplete_not_written(
```
kleisauke: I think a round-trip test plus `test_png_cicp_and_clli_coexist` below is sufficient in this case, so this test can be removed as well.

test/test-suite/test_cicp.py
```diff
@@ -653,3 +798,130 @@ def test_png_cicp_lossless_roundtrip(self):
...
+        im = make_cicp_image(
+            37_888, 22_272, 12_544,
+            primaries=PRIMARIES_DISPLAY_P3,
+            transfer=TRANSFER_PQ,
+            fmt="ushort"
+        )
+        add_clli(im)
+
+        encoded = im.pngsave_buffer()
+        assert_png_clli(encoded, (16_240_000, 1_820_000))
```
kleisauke: Ditto.
```suggestion
```

test/test-suite/test_cicp.py
```diff
@@ -653,3 +798,130 @@ def test_png_cicp_lossless_roundtrip(self):
...
+        assert out.get("cicp-colour-primaries") == PRIMARIES_DISPLAY_P3
+        assert out.get("cicp-transfer-characteristics") == TRANSFER_PQ
+        assert out.get("clli-max-content-light-level") == 1624
+        assert out.get("clli-max-frame-average-light-level") == 182
+
+    @skip_if_no("pngsave")
+    def test_png_clli_not_created_when_absent(self):
+        encoded = make_png_test_image().pngsave_buffer()
+
+        assert get_png_clli_chunks(encoded) == []
```
kleisauke: Ditto.
```suggestion
```

test/test_png_clli.c
kleisauke, line 1: This test can be removed. I assume it was added to verify the `PNG_READ_cLLI_SUPPORTED`, `PNG_WRITE_cLLI_SUPPORTED` and `PNG_FIXED_POINT_SUPPORTED` defines, but those can also be tested from Python (by checking wherever the `clli-*` metadata items are round-tripped).

For some context, we only add C tests when it cannot be easily tested from Python or when doing so would slow down the Python test suite (e.g. `test_timeout_webpsave` and `test_timeout_gifsave` were added in C for this reason, since Meson executes tests in parallel while pytest does not).

kleisauke (commented):
LGTM, aside from the test suite comments.

It might also be worth adding an entry to the `ChangeLog` for this change. Don't forget to credit yourself!

gregbenz:
Thanks for the review. I simplified the tests as suggested: replaced the custom PNG chunk helpers with a single runtime support check, consolidated the coverage into round-trip and CICP/cLLI coexistence tests, and removed the separate C test and Meson wiring. I also added the ChangeLog entry.

The focused PNG tests and full `test_cicp.py` suite pass locally.

libvips/foreign/vipspng.c
```diff
@@ -712,6 +712,22 @@ png2vips_header(Read *read, VipsImage *out, gboolean header_only)
 	}
 #endif
 
+	/* Read cLLI chunk and normalize from 0.0001 nit units to whole nits.
+	 */
+#if defined(PNG_READ_cLLI_SUPPORTED) && \
```
kleisauke: `png_get_cLLI_fixed()` is guarded with `PNG_cLLI_SUPPORTED` in libpng.
https://github.com/pnggroup/libpng/blob/d1d0abeffede1cc898ddc3d0e600839cf026d749/png.h#L2019-L2031
```suggestion
#if defined(PNG_cLLI_SUPPORTED) && \
```

libvips/foreign/vipspng.c
```diff
@@ -1299,6 +1315,44 @@ vips_png_add_original_icc(Write *write)
 	return 0;
 }
 
+#if defined(PNG_WRITE_cLLI_SUPPORTED) && \
```
kleisauke: `png_set_cLLI_fixed()` is guarded with `PNG_cLLI_SUPPORTED` in libpng.
https://github.com/pnggroup/libpng/blob/d1d0abeffede1cc898ddc3d0e600839cf026d749/png.h#L2033-L2045

```suggestion
#if defined(PNG_cLLI_SUPPORTED) && \
```

libvips/foreign/vipspng.c
```diff
@@ -1529,6 +1583,16 @@ write_vips(Write *write,
 	if (setjmp(png_jmpbuf(write->pPng)))
 		return -1;
 
+#if defined(PNG_WRITE_cLLI_SUPPORTED) && \
```
kleisauke: Ditto.

```suggestion
#if defined(PNG_cLLI_SUPPORTED) && \
```

kleisauke (approved):
LGTM, thank you!

PR #5143, opened 2026-07-28
IM_FREEF: fix error with glib 2.43+

jcupitt:
the same fix we made for VIPS_FREEF a while ago


kleisauke:
Fixes: https://github.com/libvips/nip2/issues/127.

kleisauke (approved):
The commit message should say glibc 2.43 instead of gcc15, but otherwise this looks good to me.

jcupitt:
Oh yes, I'd forgotten. I had to fix this to build nip2, since I needed the paintbox sigh.

I might try porting the nip2 paintbox to nip4.

PR #5144, opened 2026-07-28
reduceh: fix missing comma in fallback shuffle table

dloebl:
Small correction in the `tbl3_hi` in the Highway < v1.1.0 fallback path of `reduceh_hwy.cpp`:

```c
-1, -1 -1, -1
```

But it should be this instead:
```c
-1, -1, -1, -1
```

The missing comma causes the 15 element to be an implicit `0`. I haven't tested what this caused in the output.


kleisauke (approved):
Nice catch! This appears to be a regression introduced in commit 46dde2e6500a2bb0bfde860cd0ce4b5e1110075b. I'm not entirely sure how much impact it had, but hopefully most users have already upgraded to Highway 1.1.0 or later.

PR #5157, opened 2026-08-01
improve dev checklist

jcupitt:
- separate performance and security items
- put security recommendations first

thanks harry


jcupitt:
I suppose we could add something about putting image handling into a microservice in its own isolated container, but maybe that's obvious.

kleisauke:
Sorry, I missed this. LGTM!

PR #5160, opened 2026-08-01
Add initial Rust operations prototype

dloebl:
With the [recent Rails CVE ](https://discuss.rubyonrails.org/t/cve-2026-66066-possible-arbitrary-file-read-and-remote-code-execution-in-active-storage-variant-processing/91432)and all the memory safety bugs we found with fuzzing this year, I was wondering what it would take to get native Rust operations in libvips.
I've added a simple dummy `bmpload` implemented in Rust (`libvips/foreign/bmpload.rs`) and a simple `vips_ffi.rs` layer that's the interface to GObject/libvips.
The `bmpload` only supports very basic BMPs (uncompressed, 24-bit) and it's intended as a prototype to demonstrate how to implement a vips loader in Rust. 

I'd be happy to receive feedback and hear whether this is a general direction you're interested in pursuing.

```
meson setup build -Drust=enabled
meson compile -C build

./build/tools/vips bmpload test/test-suite/images/sunset.bmp sunset.jpg
```
And that's the output (`sunset.jpg`):
sunset

**ToDo:**
- [x] ABI safety of `vips_ffi.rs` (switched to bindgen at build time)
- [ ] Recover from Rust panics in release build
- [ ] Figure out if mixing Rust and C/C++ source code works well in OSS Fuzz
- [ ] Explore if `no_std` is an option
- [x] Impact on library size (~+300KB, Thanks to LTO that strips most of Rust's stdlib)
- [ ] Check toolchain/platform support
- [ ] Prototyped with Claude Fable 5 with a few manual edits - I still have to refactor most of it


jcupitt:
Hi @dloebl, sure, I like Rust, this sounds interesting.

We should check things like support between various compilers and platform libraries too, I remember C++ linking used to be rather awkward on some of them.

A full Rust rewrite is also tempting, but that'd probably have to be something for libvips 9.

dloebl:
> We should check things like support between various compilers and platform libraries too, I remember C++ linking used to be rather awkward on some of them

Good point, added to the TODO. Linking should be less awkward than C++: The Rust parts are static libs with `rust_abi: 'c'`, so they export a plain C ABI and bundle their own runtime. I guess the larger concern is probably `rustc` availability on niche platforms.

> A full Rust rewrite is also tempting, but that'd probably have to be something for libvips 9.

Makes sense. I wonder if this could be achieved incrementally. Once this becomes a stable feature, we could allow only new loaders implemented in Rust and move libvips's operations to Rust one by one. This would also allow libvips to remain backward compatible. Loaders would benefit the most because they parse untrusted input.

jcupitt:
> I guess the larger concern is probably rustc availability on niche platforms.

True, Kleis would be the person to ask about that I guess.

> I wonder if [libvips in rust] could be achieved incrementally. Once this becomes a stable feature, we could allow only new loaders implemented in Rust and move libvips's operations to Rust one by one. This would also allow libvips to remain backward compatible. Loaders would benefit the most because they parse untrusted input.

Yes, that would be pretty easy. There are some basic architectural things it'd be nice to fix too, such as the error system, which would mean swapping the core to rust as well. 

For vips8 we did a new core, then a compat API, then rewrote the operations one by one. If we keep gobject as the base then doing the same for vips9 might not be too painful.

Something to think about later!

kleisauke:
> * [ ]  Prototyped with Claude Fable 5 with a few manual edits - I still have to refactor most of it

As of today, libvips does not accept AI-generated contributions:

See also comment https://github.com/libvips/libvips/discussions/3331#discussioncomment-4937045 about creating out-of-source dynamically loadable modules, which may be relevant here.

PR #5162, opened 2026-08-02
text: differentiate between missing and invalid fontfile

lovell:
Ensures that the `fontfile`, when specified, exists on the filesystem instead of being silently ignored.

On macOS, [pangocairo defaults to Core Text](https://gitlab.gnome.org/GNOME/pango/-/blob/main/pango/pangocairo-fontmap.c#L75) but needs to be configured to use fontconfig for `fontfile` to work, so also warn when `PANGOCAIRO_BACKEND` is missing or set to another value should `FcConfigAppFontAddFile` fail.

As discussed downstream at https://github.com/lovell/sharp/issues/4577


libvips/create/text.c
```diff
@@ -434,22 +434,38 @@ vips_text_build(VipsObject *object)
...
 	if (text->fontfile &&
 		!g_hash_table_contains(vips_text_fontfiles, text->fontfile)) {
-		/* This can fail if you eg. add the same font from two
-		 * different files. Just warn.
-		 */
-		if (!FcConfigAppFontAddFile(NULL,
-				(const FcChar8 *) text->fontfile))
-			g_warning("unable to load fontfile \"%s\"",
+		unsigned char buf[4];
+		if (vips__get_bytes(text->fontfile, buf, 4) != 4) {
```
kleisauke: I think using `!vips_existsf("%s", text->fontfile)` would be neater, see e.g.:
https://github.com/libvips/libvips/blob/362b4920ce58b9054aee1a60757d45d8b5dbf757/libvips/foreign/foreign.c#L614-L620

libvips/create/text.c
```diff
@@ -434,22 +434,38 @@ vips_text_build(VipsObject *object)
...
-			g_strdup(text->fontfile), NULL);
-
-		/* We need to inform that pango should invalidate its
-		 * fontconfig cache whenever any changes are made.
-		 */
-		if (PANGO_IS_FC_FONT_MAP(vips_text_fontmap))
-			pango_fc_font_map_cache_clear(
-				PANGO_FC_FONT_MAP(vips_text_fontmap));
+			return -1;
+		} else {
```
kleisauke: To avoid nesting, we usually skip else-statements after early returns.

libvips/create/text.c
```diff
@@ -434,22 +434,38 @@ vips_text_build(VipsObject *object)
...
+			if (!FcConfigAppFontAddFile(NULL,
+					(const FcChar8 *) text->fontfile)) {
+				g_warning("unable to load fontfile \"%s\": "
+					"invalid or already registered)",
+					text->fontfile);
+#if defined(__APPLE__)
+				const char *backend = g_getenv("PANGOCAIRO_BACKEND");
+				if (!backend || strcmp(backend, "fontconfig") != 0)
```
kleisauke: I think using `!PANGO_IS_FC_FONT_MAP(vips_text_fontmap)`  would be neater. And perhaps we should also remove the macOS guard (since this warning might be useful on other platforms as well).

libvips/create/text.c
lovell: Thanks, updated, plus added a `vips_isdirf` check too.

libvips/create/text.c
```diff
@@ -434,22 +434,37 @@ vips_text_build(VipsObject *object)
...
+				_("fontfile \"%s\" is a directory"), text->fontfile);
+			g_mutex_unlock(&vips_text_lock);
+			return -1;
+		}
 		if (!FcConfigAppFontAddFile(NULL,
-				(const FcChar8 *) text->fontfile))
-			g_warning("unable to load fontfile \"%s\"",
+				(const FcChar8 *) text->fontfile)) {
+			g_warning("unable to load fontfile \"%s\": "
+					  "invalid or already registered)",
```
kleisauke:
```suggestion
					  "invalid or already registered",
```

kleisauke (approved):
LGTM!

PR #5165, opened 2026-08-03
threadpool: guard progress reporting with `allocate_lock` mutex

kleisauke:
A non-functional change in terms of that progress is still reported single-threaded for each work unit. Previously, this was done on the main thread during the semaphore tick. It now happens on a worker thread, which may affect users who expect `VipsThreadpoolProgressFn` / `vips_image_eval()` callbacks to always run on the main thread (though eval callbacks could already be invoked from worker threads in some cases, see: https://github.com/libvips/libvips/pull/5092#issuecomment-4923822692).

Split out from #5092.


jcupitt (approved):

jcupitt:
I'd still like to report progress back in the calling thread, but I agree that if we're letting workers report too, it should at least be single-threaded.

PR #5166, opened 2026-08-03
colour: fix hue wrap-around in dE00 and CMC2LCh

binggao1230:
Two hue wrap-around bugs in `libvips/colour`, found checking `dE00` against the published CIEDE2000 test data.

**`vips_col_dE00()` gets the 0/360 hue wrap wrong three ways.** The wrapped mean hue is `fabs(h1' + h2' - 360) / 2`, which is `180 - (h1' + h2') / 2` when the sum is below 360 rather than the `180 + (h1' + h2') / 2` the formula asks for; the branch tests `|dh'| < 180` when the boundary belongs to the unwrapped case; and the wrapped `dh'` is `360 - (h1' - h2')`, right magnitude but inverted relative to `dL'` and `dC'`, so the `R_T dC' dH'` term comes out with the wrong sign. Now written as Sharma eq. (10) and (14).

Master matches 30 of the 34 published pairs (Sharma, Wu and Dalal, Color Res. Appl. 30(1) 2005), this branch matches 33. Over 300k random sRGB pairs, 9.0% of master's results are off by more than 0.1 dE00, 3.0% by more than 5, worst case 17.8. Pure red vs pure blue: 58.30 on master, 52.88 correct.

Pair 14 still fails and is unchanged here. `h1'` and `h2'` are exactly 180 apart there, which is where dE00 is discontinuous, and in double precision they land one ulp on the wrapped side; pair 10 is the same case and happens to land the other way. Anything I tried just moved the knife edge, so the test skips 14 and pins the same situation with a pair whose hues come out exact (`a' = 0`, so `h'` is exactly 90 and 270).

**`vips_col_Chcmc2h()` indexes the wrong entry at the top of the hue table.** It interpolates towards `hI[r][(known + 1) % 360]`, but the table has 361 entries and `known` is clipped to 359, so the modulo only ever fires at the end of the range, where it picks up entry 0 (hue ~0) instead of entry 360. Hues in the last degree came back from CMC near zero: `h = 359.5` at `C = 40` round tripped to 161.8. `vips_col_Lcmc2L()` and `vips_col_Ccmc2C()` alongside it index `known + 1` with no modulo.

Tests are the 34 pair table and a CMC hue round trip in `test_colour.py`, both red before and green after. The rest of `test/test-suite` is unchanged (3 failures here on master as well, libexif and libpng related). The published values were cross-checked against an independent implementation of eq. (2)-(24) and against `skimage.color.deltaE_ciede2000`; all three agree on all 34 pairs. No API change, so no ChangeLog entry.

This work was done with AI assistance under my direction; I have verified every change myself and am accountable for it.

### DIFF
--- a/libvips/colour/UCS2LCh.c
+++ b/libvips/colour/UCS2LCh.c
@@ -206,7 +206,7 @@ vips_col_Chcmc2h(float C, float hcmc)
 	known = VIPS_CLIP(0, known, 359);
 
 	return hI[r][known] +
-		(hI[r][(known + 1) % 360] - hI[r][known]) * (hcmc - known);
+		(hI[r][known + 1] - hI[r][known]) * (hcmc - known);
 }
 
 static void *
--- a/libvips/colour/dE00.c
+++ b/libvips/colour/dE00.c
@@ -133,9 +133,16 @@ vips_col_dE00(float L1, float a1, float b1,
 	 */
 	double Ldb = (L1d + L2d) / 2;
 	double Cdb = (C1d + C2d) / 2;
-	double hdb = fabs(h1d - h2d) < 180
-		? (h1d + h2d) / 2
-		: fabs(h1d + h2d - 360) / 2;
+
+	/* Mean hue, see Sharma eq. (14). 180 belongs to the unwrapped case.
+	 */
+	double hdb = h1d + h2d;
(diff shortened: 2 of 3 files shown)


kleisauke:
Please do not use LLMs to generate code or PR descriptions when contributing to libvips:

PR #5168, opened 2026-08-03
test: prefer use of xfail marker

kleisauke:


lovell (approved):
:+1: for removing conditional statements from tests

PR #5172, opened 2026-08-04
Update toolchain requirement to C11

kleisauke:
GLib 2.89.2 updated its toolchain requirement to C11 in https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3574, so let's do the same. This would fix a lot of noisy `-Wtypedef-redefinition` warnings when building with Clang, see e.g.:
https://github.com/lovell/sharp-libvips/actions/runs/30836350475/job/91762528607#step:4:9239
https://github.com/kleisauke/wasm-vips/actions/runs/30895159808/job/91946196579#step:5:8412


jcupitt (approved):

jcupitt:
An exciting new world!

PR #5173, opened 2026-08-05
openexrload: more validation

jcupitt:
it seems libopenexr does very little checking

thanks Hygge Halcyon

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+date tbd 8.18.5
+
+- openexrload: more validation [Hygge Halcyon]
+
 3/8/26 8.18.5
 
 - maplut: fix possible OOB read with complex images [sage-mode-hunter]
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,4 @@
-date tbd 8.18.5
+date tbd 8.18.6
 
 - openexrload: more validation [Hygge Halcyon]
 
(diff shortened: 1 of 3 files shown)


kleisauke:
CI failures can probably(?) be fixed by replacing `return -1;` with `return NULL;` on the lines changed in this PR.

jcupitt:
:facepalm: thanks!

ChangeLog
```diff
@@ -1,3 +1,7 @@
+date tbd 8.18.5
```
```suggestion
date tbd 8.18.6
```

kleisauke (approved):

libvips/foreign/openexr2vips.c
```diff
@@ -182,16 +180,41 @@ read_new(const char *filename, VipsImage *out)
...
 		read->tile_width = ImfTiledInputTileXSize(read->tiles);
 		read->tile_height = ImfTiledInputTileYSize(read->tiles);
+
+		// openexr does little sanity checking
+		if (read->tile_width > 8192 ||
+			read->tile_width <= 0 ||
+			read->tile_height > 8192 ||
+			read->tile_height <= 0) {
+			vips_error("exr2vips", _("bad tile geometry"));
+			return NULL;
```
lovell, line 189: Do we need to call `read_destroy` on the error path when `out` is `NULL` e.g. when called via `vips__openexr_istiled`?

jcupitt: Oh hmm you're right, although there's a close handler, it's not always set.

I've tried to make this simpler and more reliable. Thanks!

lovell (approved):

kleisauke (approved):

PR #5179, opened 2026-08-09
CI: validate `author_association` in `/cherry-pick` commands

kleisauke:
Restrict the `/cherry-pick` command to users with an `author_association` of `OWNER`, `MEMBER` or `COLLABORATOR`. Without this check, the `/cherry-pick` command could succeed for external users on older issues or PRs that have a milestone attached.

Also simplify the event handling by using comment fields first (`github.event.comment`) and falling back to issue fields (`github.event.issue`), without checking if `github.event.action` is `opened`.

Follow up to: #5134.


jcupitt:
Ooops yes that sounds like a good idea!

jcupitt (approved):

PR #5190, opened 2026-08-19
heif: use struct tag for heif_content_light_level (fix build on libheif 1.19.x)

### LINKED ISSUE #5189: Build fails against libheif 1.19.x: heif_content_light_level used as a bare type
## What happens

Building current `master` fails to compile the HEIF loader/saver when the system libheif is a **1.19.x** release (e.g. Debian trixie ships libheif 1.19.8):

```
libvips/foreign/heifload.c:794:2: error: unknown type name 'heif_content_light_level'; use 'struct' keyword to refer to the type
  794 |  heif_content_light_level clli;
      |  ^
```

(and the same for `heifsave.c`).

## Root cause

The CLLI support added in #5104 declares variables and a parameter with the **bare typedef name** `heif_content_light_level`:

- `libvips/foreign/heifload.c:794` — `heif_content_light_level clli;`
- `libvips/foreign/heifsave.c:299` — `heif_content_light_level *content_light_level)`
- `libvips/foreign/heifsave.c:415` — `heif_content_light_level clli;`

But libheif only added that **typedef** in **1.20.0** (`libheif/api/libheif/heif_color.h`: `typedef struct heif_content_light_level {…} heif_content_light_level;`). In the **1.19.x** series the type exists solely as a **struct tag** (`libheif/api/libheif/heif.h`: `struct heif_content_light_level {…}`, no typedef), and libheif's own accessors take `struct heif_content_light_level *`.

`HAVE_HEIF_CONTENT_LIGHT_LEVEL` is set purely from function presence (`meson.build`):

```
cfg_var.set('HAVE_HEIF_CONTENT_LIGHT_LEVEL',
    cpp.has_function('heif_image_handle_get_content_light_level', …) and
    cpp.has_function('heif_image_set_content_light_level', …))
```

`heif_image_handle_get_content_light_level` was added in libheif **1.19.0**, so on 1.19.x the block is compiled in — but the bare type name is not valid in C without the `struct` keyword, so the build fails.

## Why it isn't seen everywhere

- libheif **< 1.19**: the accessor functions are absent, `HAVE_HEIF_CONTENT_LIGHT_LEVEL` is unset, and the block is compiled out. (This is why e.g. Ubuntu Noble builds clean.)
- libheif **≥ 1.20**: the typedef exists, so the bare name compiles.
- libheif **1.19.x**: functions present (flag on) but typedef absent → **build breaks**. Debian trixie (1.19.8) is squarely in this window.

## Affected versions

master / 8.19.0-dev — introduced by #5104 (2026-07-24), present through current HEAD (`923111819e`). Releases predating #5104 are unaffected.

## Fix

Use the struct tag — `struct heif_content_light_level` — in the three declarations. It is valid on the 1.19.x struct tag and the 1.20+ typedef alike, and in both C and C++, so it keeps the feature working on 1.19.x (the accessor functions are present there) rather than disabling it. PR to follow.

## Reproduce

Build `master` in an environment with libheif 1.19.x installed (e.g. `debian:trixie`, which ships libheif 1.19.8):

```
docker run --rm debian:trixie bash -euxc '
  apt-get update -qq
  apt-get install -y -qq --no-install-recommends ca-certificates git meson \
    ninja-build build-essential pkg-config libglib2.0-dev libexpat1-dev libheif-dev
  git clone --depth 1 https://github.com/libvips/libvips.git /src && cd /src
  meson setup build -Dheif=enabled
  ninja -C build libvips/*heif*.so.p/foreign_heifload.c.o   # fails to compile
'
```

Verified with a before/after compile on `debian:trixie` (libheif 1.19.8): master fails as
above; adding the `struct` keyword at the three declaration sites compiles both
`foreign_heifload.c.o` and `foreign_heifsave.c.o` clean, with no other change.

jeremy:
The CLLI support added in #5104 declares variables and a parameter with the bare
type name `heif_content_light_level`. libheif only added that typedef in 1.20.0
(`heif_color.h`); in the 1.19.x series the type is available solely as the struct
tag (`struct heif_content_light_level`, `heif.h`), and libheif's own accessors
take `struct heif_content_light_level *`.

`HAVE_HEIF_CONTENT_LIGHT_LEVEL` is set from function presence
(`heif_image_handle_get_content_light_level`, added in 1.19.0), so on libheif
1.19.x the block is compiled in but fails to build in C — the bare name is not a
type without the `struct` keyword. Builds against libheif < 1.19 compile the
block out, and ≥ 1.20 have the typedef, so only the 1.19.x window breaks (e.g.
Debian trixie ships libheif 1.19.8).

Use the struct tag, which is valid on both 1.19.x and 1.20+ and in C and C++.
This keeps the feature enabled on 1.19.x rather than gating it off, since the
accessor functions are present there.

Fixes #5189.

### DIFF
--- a/libvips/foreign/heifload.c
+++ b/libvips/foreign/heifload.c
@@ -791,7 +791,7 @@ vips_foreign_load_heif_set_header(VipsForeignLoadHeif *heif, VipsImage *out)
 	}
 
 #ifdef HAVE_HEIF_CONTENT_LIGHT_LEVEL
-	heif_content_light_level clli;
+	struct heif_content_light_level clli;
 
 	if (heif_image_handle_get_content_light_level(heif->handle, &clli)) {
 		g_info("heifload: setting CLLI from content light level");
--- a/libvips/foreign/heifsave.c
+++ b/libvips/foreign/heifsave.c
@@ -296,7 +296,7 @@ vips_foreign_save_heif_get_cicp(VipsImage *image,
 #ifdef HAVE_HEIF_CONTENT_LIGHT_LEVEL
 static gboolean
 vips_foreign_save_heif_get_clli(VipsImage *image,
-	heif_content_light_level *content_light_level)
+	struct heif_content_light_level *content_light_level)
 {
 	int max_content_light_level;
 	int max_frame_average_light_level;
@@ -412,7 +412,7 @@ vips_foreign_save_heif_write_page(VipsForeignSaveHeif *heif, int page)
 #endif
 
 #ifdef HAVE_HEIF_CONTENT_LIGHT_LEVEL
-	heif_content_light_level clli;
+	struct heif_content_light_level clli;
 
 	if (vips_foreign_save_heif_get_clli(save->ready, &clli))
 		heif_image_set_content_light_level(heif->img, &clli);


kleisauke (approved):
Thanks!

PR #5191, opened 2026-08-22
draw_line: add draw-point

jcupitt:
This PR adds a `draw-point` option to the line draw operation. 

You can use this to override the default "plot a point" action with your own code, perhaps drawing a disc instead. The nip4 paintbox uses this to draw wide lines.

Also add an autoptr for VipsRect (nip4 uses this too).


kleisauke (approved):

jcupitt:
Thanks for the review!

Issue #4931, opened 2026-03-05
Remaining UBSan reproducers for int format images


kleisauke:
This issue tracks the remaining UBSan reproducers for images tagged as `VIPS_FORMAT_INT`, similar to the one fixed in PR #4922 (and related).

<details>
  <summary>Reproducer - <code>merge</code></summary>

```console
$ printf 'P5\n2 2\n65536\n' > max_int_2x2.ppm
$ printf '\x7F\xFF\xFF\xFF%.0s' {1..4} >> max_int_2x2.ppm
$ vips cast max_int_2x2.ppm max_int_2x2.v int
$ vips merge max_int_2x2.v max_int_2x2.v x.v horizontal 0 0
../libvips/mosaicing/lrmerge.c:645:4: runtime error: signed integer overflow: 4096 * 2147483647 cannot be represented in type 'int'
    #0 0x7fe9deda997b in lr_blend /home/kleisauke/libvips/build/../libvips/mosaicing/lrmerge.c:645:4
    #1 0x7fe9ded7c75f in vips__merge_gen /home/kleisauke/libvips/build/../libvips/mosaicing/lrmerge.c:1035:8
    #2 0x7fe9deb67fe2 in vips_region_generate /home/kleisauke/libvips/build/../libvips/iofuncs/region.c:1614:6
    #3 0x7fe9deb49ea3 in vips_region_fill /home/kleisauke/libvips/build/../libvips/iofuncs/region.c:867:7
    #4 0x7fe9deb6779c in vips_region_prepare /home/kleisauke/libvips/build/../libvips/iofuncs/region.c:1682:7
    #5 0x7fe9de580767 in vips_copy_gen /home/kleisauke/libvips/build/../libvips/conversion/copy.c:140:6
    #6 0x7fe9deb67fe2 in vips_region_generate /home/kleisauke/libvips/build/../libvips/iofuncs/region.c:1614:6
    #7 0x7fe9deb49ea3 in vips_region_fill /home/kleisauke/libvips/build/../libvips/iofuncs/region.c:867:7
    #8 0x7fe9deb6779c in vips_region_prepare /home/kleisauke/libvips/build/../libvips/iofuncs/region.c:1682:7
    #9 0x7fe9deab89e7 in vips_image_write_gen /home/kleisauke/libvips/build/../libvips/iofuncs/image.c:2600:6
[...]

PR #5192, opened 2026-08-24
convi: sum using 128 bit integer where available

lovell:
Uses the `__int128` extension type provided by gcc and clang on 64-bit platforms.

https://stackoverflow.com/questions/63029428/why-is-int128-t-faster-than-long-long-on-x86-64-gcc suggests this approach might be slightly faster too.

Fixes https://issues.oss-fuzz.com/issues/550629907

### DIFF
--- a/libvips/convolution/convi.c
+++ b/libvips/convolution/convi.c
@@ -129,6 +129,12 @@
 
 #include "pconvolution.h"
 
+#ifdef __SIZEOF_INT128__
+#define VIPS_CONVI_INT128 __int128
+#else
+#define VIPS_CONVI_INT128 int64_t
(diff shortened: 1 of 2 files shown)


libvips/convolution/convi.c
```diff
@@ -129,6 +129,12 @@
 
 #include "pconvolution.h"
 
+#if defined(__SIZEOF_INT128__)
```
kleisauke: nit:
```suggestion
#ifdef __SIZEOF_INT128__
```

kleisauke (approved):
Great! This will also fix the `convi` reproducer in #4931.

PR #5193, opened 2026-08-25
heifload: set untrusted for libheif before 1.23.2

jcupitt:
also bump version to 8.18.6

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+25/8/26 8.18.6
+
+- heifload: tag as UNTRUSTED for libheif before 1.23.2
+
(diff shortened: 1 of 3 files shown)


kleisauke (approved):

kleisauke:
TIL about the `LIBHEIF_HAVE_VERSION` macro (which has been available in all libheif releases since commit https://github.com/strukturag/libheif/commit/b57c3e8615896fc716dfad83119d1285561e19fd). It's much simpler than performing these checks in Meson. :+1:

jcupitt:
Yes, maybe we should revise `meson.build`? A low priority!

PR #5199, opened 2026-08-31
vips_target_finish: drop the two params it no longer takes

darkdi:
The block still describes the old signature, the function has taken only the target since it became a wrapper around `vips_target_end()`.


jcupitt:
Oooop! Nice find, thank you @darkdi

PR #5200, opened 2026-09-02
add vips_image_copy_draw(), vips_check_draw()

jcupitt:
A threadsafe thing to get an image you can draw on.

Deprecate vips_image_inplace().

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -48,6 +48,7 @@ date-tbd 8.19.0
 - draw_line: add draw-point optional parameter [jcupitt]
 - VipsRect: add g_autoptr cleanup
 - convi: sum using 128 bit integer, where available, before clipping [lovell]
+- add vips_check_draw(), vips_image_copy_draw() [jcupitt]
 
 25/8/26 8.18.6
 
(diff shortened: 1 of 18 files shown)


jcupitt:
`vips_image_inplace()` is not threadsafe, so this is supposed to be better. Draft since I'm still testing it.

jcupitt:
There's a matching nip4 branch here: https://github.com/libvips/nip4/tree/add-copy-draw

It seems to work OK, I just need to improve progress reporting.

jcupitt:
Right now, this no longer works:

```
$ vips copy k2.jpg x.v
$ vips draw_flood x.v 0 100 100
draw_flood: image cannot be drawn on
```

I think I can fix it.

kleisauke:
Here's a minimal reproducer for the CIFuzz failures:
```console
$ vips labelregions x.jpg x.v
```

jcupitt:
Yes, there are some internal uses of the direct draw functions which need updating. I'll have a go.

jcupitt:
I think this is all done now. The test suites and examples/ for nip4, pyvips and ruby-vips all work unmodified, and the paintbox seems a lot better.

`vips_image_inplace()` is not deprecated (that would break too many things), but the docs do point you to the new threadsafe alternative.

kleisauke (approved):
Great!

Issue #4628, opened 2025-08-12
Wrong dimensions when using SVG units such as "mm"


urish:
## Bug report

**Describe the bug**

The following SVG uses the "mm" units, and renders incorrectly:

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="100" height="100" viewBox="0 0 100 100">
  <rect x="0" y="0" width="13.22mm" height="13.22mm" fill="red" />
  <rect x="50" y="50" width="13.22mm" height="13.22mm" fill="blue" />
</svg>
```

This is what it looks like with pyvips (3.0.0):

<img width="100" height="100" alt="Image" src="https://github.com/user-attachments/assets/7e2a81ce-6cbe-4793-82c7-8d81282620a8" />

This is what it looks like when converted with rsvg-convert (v2.58.0) and in both Chrome / Firefox:

<img width="100" height="100" alt="Image" src="https://github.com/user-attachments/assets/dd34c0a1-5e53-438c-b60d-3151b946b9b2" />

From what I can tell, it seems like the units are scaled down by a factor of 1.33.

I also tried playing around with the `dpi` and `scale` parameters of the `svgload` function, but they only scale the resulting image, without affecting the content within.

**To Reproduce**

Save the provided svg to "test-units.svg", and then run the following python script:

```python
import pyvips
image = pyvips.Image.svgload('test-units.svg')
image.write_to_file('test-units-py.png')
```

**Environment**
(please complete the following information)
- OS: Windows 11
- pyvips: 3.0.0
- pyvips-binary: 8.17.1

kleisauke:
libvips uses a density of 72 DPI by default, whereas `rsvg-convert` uses 96 DPI.

So, the following two commands produce identical images:
```console
$ rsvg-convert --dpi-x=72 --dpi-y=72 test-units.svg -o x.png
$ vips svgload test-units.svg x2.png
```

Though, you might expect these two commands to also produce identical images, as you said:
```console
$ rsvg-convert test-units.svg -o x.png
$ vips svgload test-units.svg x2.png --dpi=96
```

But they don't. Applying this patch seems to fix that:
```patch
--- a/libvips/foreign/svgload.c
+++ b/libvips/foreign/svgload.c
@@ -546,16 +546,13 @@ vips_foreign_load_svg_get_scaled_size(VipsForeignLoadSvg *svg,
 	double width;
 	double height;
 
-	/* Get dimensions with the default dpi.
+	/* Get dimensions with the target dpi.
[...]

PR #5205, opened 2026-09-03
svgload: improve dpi= and scale= behaviour

jcupitt:
We were calling rsvg_handle_set_dpi() and also scaling rendering. This works fine for SVG files with pixel dimensions in the header, but fails for real-world dimensions (eg. `width="1000mm"`), where librsvg will *also* scale. You get dpi being applied twice, and image size starts to scale with the square of dpi.

This PR moves svgload closer to the behaviour of librsvg itself. It makes the dpi= parameter only scale DPI, and the scale= parameter only scale rendering.

See https://github.com/libvips/pyvips/issues/1424 and https://github.com/libvips/pyvips/issues/1424

Thanks Czoczaj

### DIFF
--- a/libvips/foreign/svgload.c
+++ b/libvips/foreign/svgload.c
@@ -179,11 +179,9 @@ vips_utf8_strcasestr(const char *haystack_start, const char *needle_start,
 			 * might end half-way through a utf-8 character, so we
 			 * need to be careful not to run off the end.
 			 */
-			gunichar a =
-				g_utf8_get_char_validated(haystack_char,
-					haystack_start + len_bytes - haystack);
-			gunichar b =
-				g_utf8_get_char_validated(needle_char, -1);
+			gunichar a = g_utf8_get_char_validated(haystack_char,
+				haystack_start + len_bytes - haystack);
+			gunichar b = g_utf8_get_char_validated(needle_char, -1);
 
 			/* Invalid utf8?
 			 *
@@ -211,9 +209,8 @@ vips_utf8_strcasestr(const char *haystack_start, const char *needle_start,
 			if (g_unichar_tolower(a) != g_unichar_tolower(b))
 				break;
 
-			haystack_char =
-				g_utf8_find_next_char(haystack_char,
-					haystack_start + len_bytes);
+			haystack_char = g_utf8_find_next_char(haystack_char,
+				haystack_start + len_bytes);
 
 			/* End of haystack. There can't be a complete needle
 			 * anywhere.
@@ -223,8 +220,7 @@ vips_utf8_strcasestr(const char *haystack_start, const char *needle_start,
 
 			/* needle_char will never be NULL.
 			 */
-			needle_char =
-				g_utf8_find_next_char(needle_char, NULL);
+			needle_char = g_utf8_find_next_char(needle_char, NULL);
 		}
 
 		if (i == needle_len)
@@ -355,8 +351,7 @@ vips_foreign_load_svg_build(VipsObject *object)
 	}
 #endif /*HAVE_CAIRO_FORMAT_RGBA128F*/
 
-	return VIPS_OBJECT_CLASS(vips_foreign_load_svg_parent_class)
-		->build(object);
+	return VIPS_OBJECT_CLASS(vips_foreign_load_svg_parent_class)->build(object);
 }
 
 static VipsForeignFlags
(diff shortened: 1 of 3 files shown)


jcupitt:
Marked as a draft since this is likely to break things and needs testing.

jcupitt:
Sample image:

```svg
<?xml version="1.0" encoding="utf-8" ?>
<svg baseProfile="full" 
        version="1.1"
        width="1000mm" 
        height="1000mm"    
  xmlns="http://www.w3.org/2000/svg" 
  xmlns:ev="http://www.w3.org/2001/xml-events" 
  xmlns:xlink="http://www.w3.org/1999/xlink">
  <defs></defs> 
  <rect fill="#ffffff" height="1000mm" width="1000mm" x="0" y="0" />
</svg>
```

This has a physical dimension (mm) in the header and will scale with DPI. For example:

```
$ vips copy temp.svg x.png
$ vipsheader -a x.png
x.png: 2835x2835 uchar, 4 bands, srgb, pngload
width: 2835
height: 2835
bands: 4
format: uchar
coding: none
interpretation: srgb
xoffset: 0
yoffset: 0
xres: 2.83465
yres: 2.83465
```

2835 pixels printed at 72 DPI will be 1000mm across and have a resolution of 2.835 pixels/mm (the xres value above).

```
$ vips copy temp.svg[dpi=25.4] x.png
$ vipsheader -a x.png
x.png: 1000x1000 uchar, 4 bands, srgb, pngload
width: 1000
height: 1000
bands: 4
format: uchar
coding: none
interpretation: srgb
xoffset: 0
yoffset: 0
xres: 1
yres: 1
```

1000 pixels printed at 25.4 DPI will be 1000mm across and have a res of 1 pixel/mm.

```
$ vips copy temp.svg[dpi=25.4,scale=2] x.png
$ vipsheader -a x.png
x.png: 2000x2000 uchar, 4 bands, srgb, pngload
width: 2000
height: 2000
bands: 4
format: uchar
coding: none
interpretation: srgb
xoffset: 0
yoffset: 0
xres: 1
yres: 1
```

Applying a scale changes the number of pixels but does not change the DPI (it'll print 2000mm across now).

jcupitt:
Sample image:

```svg
<svg id="レイヤー_1のコピー" viewBox="0 0 100 100">
  <circle cx="50" cy="50" r="30" fill="#000" />
</svg>
```

That SVG has a size in pixels, so it no longer scales with DPI:

```
$ vips copy circle.svg x.png
$ vipsheader -a x.png
x.png: 100x100 uchar, 4 bands, srgb, pngload
width: 100
height: 100
bands: 4
format: uchar
coding: none
interpretation: srgb
xoffset: 0
yoffset: 0
xres: 2.83465
yres: 2.83465
```

Though xres/yres do change:

```
$ vips copy circle.svg[dpi=144] x.png
$ vipsheader -a x.png
x.png: 100x100 uchar, 4 bands, srgb, pngload
width: 100
height: 100
bands: 4
format: uchar
coding: none
interpretation: srgb
xoffset: 0
yoffset: 0
xres: 5.66929
yres: 5.66929
```

You must now use `scale=`:

```
$ vips copy circle.svg[dpi=144,scale=2] x.png
$ vipsheader -a x.png
x.png: 200x200 uchar, 4 bands, srgb, pngload
width: 200
height: 200
bands: 4
format: uchar
coding: none
interpretation: srgb
xoffset: 0
yoffset: 0
xres: 5.66929
yres: 5.66929
```

jcupitt:
And here's the bug in master. Using:

```svg
<?xml version="1.0" encoding="utf-8" ?>
<svg baseProfile="full" 
        version="1.1"
        width="1000mm" 
        height="1000mm"    
  xmlns="http://www.w3.org/2000/svg" 
  xmlns:ev="http://www.w3.org/2001/xml-events" 
  xmlns:xlink="http://www.w3.org/1999/xlink">
  <defs></defs> 
  <rect fill="#ffffff" height="1000mm" width="1000mm" x="0" y="0" />
</svg>
```

(same as above) you see:

```
$ vips copy temp.svg[dpi=25.4] x.png
$ vipsheader -a x.png
x.png: 353x353 uchar, 4 bands, srgb, pngload
width: 353
height: 353
bands: 4
format: uchar
coding: none
interpretation: srgb
xoffset: 0
yoffset: 0
xres: 1
yres: 1
```

You can see the pixel dimensions are wrong (should be 1000, instead 353). The SVG has been accidentally scaled twice.

353 comes from:

```
2835 pixels (the natural size of the image) / (72 / 25.4) ^ 2 == 353
```

ie. it's been shrunk by the square of the DPI.

kleisauke:
Ah, interesting. I looks like `pdfload` might need a similar fix (e.g. grep for `total_scale`).

jcupitt:
Oh, maybe, I've not tested `pdfload`. 

The horrible problem here is that calling `rsvg_handle_set_dpi()` only changes the rendering scale if the SVG header is sized in real-world units (!!). So unless you parse the header yourself, you don't know if the image is being scaled or not (!!!!?!?!##?@!!), and if you need to scale again yourself.

I think the only sane fix is to split the functionality of `dpi=` and `scale=` into two different controls, but maybe I've missed another solution?

Hopefully poppler does not have this problem! Or maybe it does? Probably not since (I think?) PDF always uses real-world dimensions.

libvips/foreign/svgload.c
kleisauke: Should we remove "If you set both, they combine." here?

(I synced that with `pdfload` in commit b27f6edbb49cd0643e42341c85a0a5100c457fb3)
https://github.com/libvips/libvips/blob/2558dffb50dd3ed024a6f2673eeef5b9a557ce9c/libvips/foreign/foreign.c#L2763-L2764

kleisauke (approved):
Great! I also tested this with the reproducer from https://github.com/libvips/libvips/issues/4628#issuecomment-3196518706 and it now works without `--scale=.75`.

libvips/foreign/svgload.c
jcupitt: Oh, good point, I missed that. I'll look at the docs again.

kleisauke (approved):

PR #5206, opened 2026-09-03
buildlut: limit the output lut size

jcupitt:
Plus some minor reformatting.

thanks Ada Logics

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,7 @@
 tbd 8.18.7
 
 - tiffload: copy colormaps on page load [Ada Logics]
+- buildlut: limit output lut size [Ada Logics]
 
 25/8/26 8.18.6
 
--- a/libvips/create/buildlut.c
+++ b/libvips/create/buildlut.c
@@ -142,15 +142,14 @@ vips_buildlut_build_init(VipsBuildlut *lut)
 			xhigh = v;
 	}
 
-	if (xlow < INT_MIN || xhigh > INT_MAX ||
-		xhigh - xlow + 1 > VIPS_MAX_COORD) {
+	if (xlow < INT_MIN ||
+		xhigh > INT_MAX)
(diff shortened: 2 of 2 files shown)


ChangeLog
```diff
@@ -2,6 +2,7 @@
 
 - openexrload: more validation [Hygge Halcyon]
 - heifload: tag as UNTRUSTED for libheif before 1.23.2
+- buildlut: limit output lut size [Ada Logics]
```
kleisauke: 8.18.6 was already released, so this needs to be moved to a new 8.18.7 entry (+ version bump in Meson).

kleisauke (approved):

ChangeLog
jcupitt: Oops! I'll fix it.

libvips/create/buildlut.c
kleisauke: Perhaps we should remove this check as well (since we're checking the same thing below against a lower value)?

jcupitt: good idea, done.

libvips/create/buildlut.c
```diff
@@ -143,23 +142,24 @@ vips_buildlut_build_init(VipsBuildlut *lut)
 			xhigh = v;
 	}
 
-	if (xlow < INT_MIN || xhigh > INT_MAX ||
-		xhigh - xlow + 1 > VIPS_MAX_COORD) {
+	if (xlow < INT_MIN ||
+		xhigh > INT_MAX)
```
kleisauke: Should fix CI:
```suggestion
		xhigh > INT_MAX) {
```

jcupitt: :facepalm:

kleisauke (approved):

jcupitt:
Thanks!

PR #5207, opened 2026-09-04
pdfiumload: check for buffer too small

jcupitt:
FPDF_GetMetaText() does nothing if the passed-in buffer is too small, so you must check the returned length against the maximum buffer size.

Thanks Ada Logics

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,7 @@
 
 - openexrload: more validation [Hygge Halcyon]
 - heifload: tag as UNTRUSTED for libheif before 1.23.2
+- pdfiumload: check for buffer too small [Ada Logics]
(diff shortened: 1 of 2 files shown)


ChangeLog
```diff
@@ -2,6 +2,7 @@
 
 - openexrload: more validation [Hygge Halcyon]
 - heifload: tag as UNTRUSTED for libheif before 1.23.2
+- pdfiumload: check for buffer too small [Ada Logics]
```
kleisauke: This also needs to be moved to 8.18.7.

jcupitt: Done.

kleisauke (approved):

PR #5209, opened 2026-09-04
hough_line: add a featurespace bounds check

jcupitt:
it was just an assert before

Thanks Ada Logics


kleisauke (approved):
LGTM. Do we need a changelog entry for this as well?

jcupitt:
Done. Thank you!

PR #5212, opened 2026-09-10
cache: ensure entry invalidation uses mutex

lovell:
As reported by Albert Ebb.

~~I can't get ASAN to complain about the current behaviour so this is unlikely to be a security problem.~~ I can get ASAN to complain when using internal-ish API to force the issue.

It's worth making this change though as all updates to the operation cache should be using the mutex.

In addition, because we now invalidate entries within the mutex, we have to separate the removal of a cache entry from the freeing of that entry (in this case with judicious use of the "steal" functions) to prevent a deadlock.

### DIFF
--- a/libvips/iofuncs/cache.c
+++ b/libvips/iofuncs/cache.c
@@ -696,11 +696,17 @@ vips_cache_operation_get(VipsOperation *operation)
 }
 
 /* Remove an operation from the cache.
+ * Caller must free the returned entry.
  */
-static void
+static VipsOperationCacheEntry *
 vips_cache_remove(VipsOperation *operation)
 {
-	g_hash_table_remove(vips_cache_table, operation);
+	VipsOperationCacheEntry *entry =
+		g_hash_table_lookup(vips_cache_table, operation);
+	if (entry)
+		g_hash_table_steal(vips_cache_table, operation);
+
+	return entry;
 }
 
 static void *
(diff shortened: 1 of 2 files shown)


libvips/iofuncs/cache.c
```diff
@@ -696,11 +696,17 @@ vips_cache_operation_get(VipsOperation *operation)
 }
 
 /* Remove an operation from the cache.
+ * Caller must free the returned entry.
  */
-static void
+static VipsOperationCacheEntry *
 vips_cache_remove(VipsOperation *operation)
```
jcupitt, line 702: This is a bit confusing to me -- the hash table is built with a thing to auto-free entries on removal, so you wouldn't expect a remove method to not free the entry.

How about renaming as `vips_cache_steal()`?

jcupitt: In `_buildp()` we have:

```C
    /* We need to remove the existing cache entry if it's been tagged
     * as invalid, if it's been blocked, or someone has requested
     * revalidation. 
     */
    if (hit) {
        if (hit->invalid ||
            (flags & VIPS_OPERATION_BLOCKED) ||
            (flags & VIPS_OPERATION_REVALIDATE)) {
            vips_cache_remove(hit->operation);
            hit = NULL;
        }
    }    
```

I suppose that should have a free now.

libvips/iofuncs/cache.c
```diff
@@ -806,15 +812,21 @@ vips_entry_ref(VipsOperationCacheEntry *entry)
 }
 
 static void
-vips_cache_invalidate_cb(VipsOperation *operation,
-	VipsOperationCacheEntry *entry)
+vips_cache_invalidate_cb(VipsOperation *operation, void *user_data)
```
jcupitt, line 815: How about splitting this up? Perhaps:

```C
static void
vips_cache_invalidate_cb(VipsOperation *operation)
{
#ifdef DEBUG
    printf("vips_cache_invalidate: ");
    vips_object_print_summary(VIPS_OBJECT(operation));
#endif /*DEBUG*/
        
    g_mutex_lock(&vips_cache_lock);

    VipsOperationCacheEntry *entry = vips_cache_operation_get(operation);
    if (entry)
        entry->invalid = TRUE;
    
    g_mutex_unlock(&vips_cache_lock);
}  

static void
vips_cache_invalidate_cb(VipsOperation *operation, void *user_data)
{
    vips_cache_invalidate(operation);
}  
```

So the `_cb()` version is only used as a callback.

libvips/iofuncs/cache.c
jcupitt: `vips__cache_once_init()` sets `vips_cache_free_cb()` as the free callback, but I don't think that is ever used now.

Maybe set the free callback to NULL and rename `vips_cache_free_cb()` as `vips_cache_free()`?

jcupitt:
Ah I think I see. I agree this needs fixing.

I made some comments to clean up the naming a little, though maybe that should happen in master rather than here. I think I saw one leak too.

lovell:
Thanks for the review John, I've applied your suggested updates and removed the possible leak. Is this OK for 8.18 or should I update the destination branch for inclusion in 8.19?

